PR #1, opened 2016-03-26
Project Setup

cquiroz:
This PR setups the basis of the OCS3/Seqexec project. Intentionally no code is included, so we can discuss the project structure.
At this stage this project should be able to build the seqexec though it is still open how it will be deployed
The same LICENSE usen in OCS is included


modules/edu.gemini.seqexec.server/build.sbt
```diff
@@ -0,0 +1,20 @@
...
+  "io.argonaut"        %% "argonaut"                  % "6.1",
+  "commons-httpclient" % "commons-httpclient"         % "2.0",
+  // OCS bundles
+  SpModelCore,
+  POT,
+  EpicsACM
+)
+
+unmanagedJars in Compile ++= Seq(
+  new File((baseDirectory in ThisBuild).value, "lib/squants_2.11-0.6.1.jar")
```
cquiroz: Do we want to put these in the repository instead of having the binary?

project/Settings.scala
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Application settings and dependencies
+ */
+object Settings {
+  object Definitions {
+    /** The name of the application */
+    val name = "ocs3"
+
+    /** Project version */
+    val version = "2016001.1.1"
```
cquiroz: Should me use `2016.1.1.1` as version? What's the role of the `00` there?

cquiroz: Should we mark the version as SNAPSHOT?

modules/edu.gemini.seqexec.server/build.sbt
swalker2m: I don't have a strong preference on this but my inclination is to leave it in the lib directory until squants has been updated and it can go into the managed dependencies.  Otherwise we have another thing that we have to depend on being up and running and reachable in order to build.  I mean, in general in the future.  For now obviously you'll be pulling in OCS dependencies from artifactory anyway.  What do you think?

tpolecat: Sorry I forgot what the issue is with Squants. Is it not available as a managed dependency?

cquiroz: It is available as 0.5.3 but not 0.6.1. We are building the new version, so we have the binary. There is no ETA for 0.6.1 reaching maven central
I don't have a very strong opinion, but if we go with artifactory it worries me that it will try to reload it very often if it is a SNAPSHOT

cquiroz: For OCS dependencies I like going through artifactory as it will easily give you access to the sources.

tpolecat: Could we publish it as `0.6.1-GEM01` or something so it's not a `SNAPSHOT`?

Like Shane I don't really care, but I think in general this is what people do.

project/Settings.scala
swalker2m: Hmm, I hadn't seen this version alternative.  I wonder if we should abandon the semester in the version number altogether and just count up like normal software products.

tpolecat: We should put the version in its own SBT file (as done [here](https://github.com/tpolecat/atto) for instance) and yes, mark it as a snapshot. I'd like to get set up with sbt-release and all that eventually.

README.md
```diff
@@ -1,2 +1,65 @@
...
+
+Some build-level tasks are still pending
+
+- [ ] Support deployable artifacts
+- [ ] Include modules cross-compiled to both JVM and JS backends
+- [ ] Enable travis
+- [ ] Unify JVM/JS scalaz versions
+
+## Structure
+The project contains a single `modules` subtree. Inside that module each of the bundles is located. Web applications are further devided internally to have the server side, client side and shared code compiled to their respective backends
```
swalker2m: Web applications and cross-built libraries as well.

project/Settings.scala
cquiroz: What's the benefit of `version.sbt` over putting it in the top level `build.sbt`?

tpolecat: @cquiroz if you do that the `sbt-release` plugin can roll the version for you (by rewriting that file).

tpolecat: @swalker2m yeah I have similar thoughts on the version thing. I suggest we start with `0.0.1-SNAPSHOT` like normal software projects and see how that goes. The fewer weird things we have from the outset the better off we will be I think.

cquiroz: @tpolecat Sounds good. Were you thinking about `sbt-assembly` for deployment?

modules/edu.gemini.seqexec.server/build.sbt
```diff
@@ -0,0 +1,20 @@
...
+libraryDependencies ++= Seq(
+  ScalaZCore,
+  ScalaZConcurrent,
+  "io.argonaut"        %% "argonaut"                  % "6.1",
+  "commons-httpclient" % "commons-httpclient"         % "2.0",
+  // OCS bundles
+  SpModelCore,
+  POT,
+  EpicsACM
+)
```
swalker2m: I like these dependency variables like `ScalaZCore`.  Is there a reason to treat `argonaut` and `httpclient` differently?

project/Settings.scala
tpolecat: @cquiroz I had not gotten that far. [`sbt-native-packager`](https://github.com/sbt/sbt-native-packager) is worth a look ... it knows how to do startup scripts and that kind of thing.

swalker2m:
Looks fantastic to me.

modules/edu.gemini.seqexec.server/build.sbt
tpolecat: I kind of like the simplicity of just writing them all out, except when there are a bunch of modules with the same version in which case just that segment can be factored out. I don't feel strongly about it though.

cquiroz: Not really, the only "reason" is that they are not in use any other place. We could unify libraries making all of them as variables

LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 1998-2015 Association of Universities for Research in Astronomy, Inc. (AURA)
```
arturog8m: Time to add 2016 to this....

modules/edu.gemini.seqexec.server/build.sbt
arturog8m: I like the idea of using Artifactory for binary stuff that it isn't ours in general; that way, we prevent us from putting something there that we shouldn't.  I haven't checked in details all the stuff in `lib`, probably it's all open source and there is no concern at all.

swalker2m: I think only the latest version is cross built for scala.js, for what it's worth.  We depend on squants in core so if we want to cross build it we'll need the latest version.  I'm also no sure whether Florian's additions have been officially incorporated into squants.

cquiroz: There is no PR for florian's changes yet

tpolecat: Yes, all of Floripan's PRs have been merged.

tpolecat: @cquiroz was  there anything beyond [these](https://github.com/garyKeorkunian/squants/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+author%3Afnussber)?

cquiroz: Yes, there is one with `SpectralIrradiance` that is not in PR form yet. We may decide to post it ourselves

cquiroz:
I updated the project with the following changes:

- All dependencies are expressed as variables
- Version set to 0.0.1-SNAPSHOT and set in version.sbt
- Store squants-gemini in our artifactory server and use it as a managed dependency
- Fixed year in the license

Let me know if more changes are needed before merging

PR #4, opened 2016-03-28
Copy legacy code from seqexec-server in OCS

cquiroz:
This PR brings the modules `edu.gemini.seqexec.server` and `edu.gemini.seqexec.shared` into OCS3.


cquiroz:
This code has been already reviewed on the OCS
do you know if there is a way to keep the history? it is probably hard given the layout is different

tpolecat:
It is possible (and desirable) to maintain history so maybe see if [this](http://blog.mattsch.com/2015/06/19/move-directory-from-one-repository-to-another-preserving-history/) works?

cquiroz:
Thanks for the tip. The procedure works with some adjustments. I also removed with extra commits the OSGi activators. The history seems to be all preserved

swalker2m:
Will the shell commands be replaced eventually?

cquiroz:
That is a good question. Since there is no shell currently available I'd say no. OTOH there is no reason we couldn't make those commands into REST endpoints or even a web-based shell

swalker2m:
Right, that makes sense.  These particular shell commands might not be worth much but in general we'll want a way to do this type of thing.

:+1:

cquiroz:
maybe we should create a ticket to remind us about this. GH or JIRA?

jluhrs:
I used the shell commands a lot, for testing. My vote if for trying to keep that functionality.

cquiroz:
Added issue #5 and ready to merge

Issue #8, opened 2016-03-29
Seqexec Queue State


cquiroz:
Seqexec Queue state must be read from the backend and displayed on the UI

PR #12, opened 2016-03-31
Seqexec Queue load

cquiroz:
This PR builds on the Web UI by loading the contents of the Seqexec Queue (Canned data ATM), Implements #8. At startup it will request the current queue using an ajax request and display the contents of the queue on the queue table. Some notes:

* Both backends can be used to serve the application
* This PR starts using [Diode](http://ochrons.github.io/diode/index.html), which is the scala.js version of the [Flux Architecture](http://facebook.github.io/flux/) for React applications
* The code contains logic to display an indicator while loading and to display proper error in case of a failure

Below there is a small video (á la @tpolecat) with the happy path and an error case. In both cases the server has an artificial 3 seconds delay, otherwise you cannot see the loading process while serving locally

[HappyCase](https://www.dropbox.com/sc/oydtgm5be9tf9f1/AAAk2LStHPPjCyifyjS9aGMPa)
[ErrorCase](https://www.dropbox.com/sc/iywl4mo4tfnnt8x/AABvI8HNBbKPVSfUmKKaMmL9a)

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
@@ -33,14 +33,14 @@ object QueueTableBody {
       <.tbody(
         // Render after data arrives
         p.queue().render( q =>
-          q.queue.map(Some.apply).padTo(minRows, None).collect {
-            case Some(s) =>
+          q.queue.map(Some.apply).padTo(minRows, None).zipWithIndex.collect {
+            case (Some(s), i) =>
               <.tr(
                 ^.classSet(
                   "positive" -> (s.state == SequenceState.Running),
                   "negative" -> (s.state == SequenceState.Error)
                 ),
-                ^.key := s"item.queue.${s.id}",
+                ^.key := s"item.queue.$i",
                 <.td(
                   ^.cls := "collapsing",
                   s.id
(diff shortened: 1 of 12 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
```diff
@@ -1,17 +1,86 @@
...
+                ),
+                <.td(s.state.toString),
+                <.td(s.instrument),
+                <.td(
+                  SeqexecStyles.notInMobile,
+                  s.error.map(_ => <.p(Icon("attention"), " Error")).getOrElse(<.p("-"))
+                )
+              )
+            case _ =>
+              emptyRow(s"time.queue." + (1000*math.random).toInt)
```
swalker2m: Would you explain a bit about the keys?  Normal rows are `item.queue.` followed by something and empty rows `time.queue.` followed by a random number which may not be unique?  What is the key being used for?

cquiroz: Now that I look at this, the `random` part should be improved. Regardless, React requires that components in an array have a unique key so that it can efficiently render changes to the virtual dom. Interestingly they don't need to be globally unique, just in an array or sequence. in this case, I just took a shortcut but will improve it with a `zipWithIndex` call

swalker2m:
This looks good to me. :+1:

PR #17, opened 2016-04-04
Seqexec commands

cquiroz:
This PR restores to some extent the capability of sending commands to the seqexec using a command line interface. These commands used to be available in the OSGi felix console but with that not present in the current setup, a REST api interface is made available. The `Commands` class was taken from the old OSGi commands and adapted to conform to the new setup, it doesn't need to parse all the args and can return richer responses

A backend for both http4s and play is provided. Additionally, for the play backend, the routes were split to into different objects like done for http4s

The API is documented in the README including examples of **curl** commands equivalent to the old code


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Commands.scala
```diff
@@ -0,0 +1,172 @@
...
+  def showCount(obsId: String): CommandError \/ CommandResponse
+  def showStatic(obsId: String): CommandError \/ CommandResponse
+  def showStatic(obsId: String, system: String): CommandError \/ CommandResponse
+  def showDynamic(obsId: String, step: String): CommandError \/ CommandResponse
+  def showDynamic(obsId: String, step: String, system: String): CommandError \/ CommandResponse
+  def run(obsId: String): CommandError \/ CommandResponse
+  def stop(obsId: String): CommandError \/ CommandResponse
+  def continue(obsId: String): CommandError \/ CommandResponse
+  def state(obsId: String): CommandError \/ CommandResponse
+}
```
swalker2m: I think I would be tempted to give `CommandError \/ CommandResponse` a type alias like

````scala
type CommandResult[A] = CommandError \/ A
````

but that's just a personal preference.  The more explicit form could be considered clearer.

swalker2m: Oh, I guess the response isn't an arbitrary type.

````scala
type CommandResult = CommandError \/ CommandResponse
````

maybe then.

swalker2m:
Looks good to me. :+1:

PR #19, opened 2016-04-05
sbt-buildinfo into the build

cquiroz:
I'm raising this PR for discussion as I don't need this right away. I want, at some moment, display the version of the software on the UI. I'd think the best way to do this is by bringing the information from sbt avoiding repetition, and that's exactly what the [sbt-buildinfo](https://github.com/sbt/sbt-buildinfo) plugin does.

This PR integrates buildinfo into the build and will generate a managed class with the content as

```scala
package edu.gemini.seqexec.web.client

import scala.Predef._

/** This object was generated by sbt-buildinfo. */
case object BuildInfo {
  /** The value is "edu_gemini_seqexec_web_client". */
  val name: String = "edu_gemini_seqexec_web_client"
  /** The value is "0.1-SNAPSHOT". */
  val version: String = "0.1-SNAPSHOT"
  override val toString: String = {
    "name: %s, version: %s" format (
      name, version
    )
  }
}
```

I found one caveat though. You need to generate the class before opening the project on IDEA. If you open it before it won't compile as the `BuildInfo` class won't be on idea's classpath. It works fine if you do an initial compile and then open it with IDEA. I believe the same happens for any generated code, see [bug](https://youtrack.jetbrains.com/issue/SCL-9301)

What do you think, is this the right approach?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,8 @@ Note that IDEA doesn't support running tests on scala.js. See this [issue](https
 
 ## IDEA
 
+Before opening the project in IDEA, run a top-level compile at least once to generate code that will be picked up by IDEA
+
 IDEA can open directly the project's sbt. Go to 
 
 ```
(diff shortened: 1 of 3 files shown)


tpolecat:
`sbt-buildinfo` is fine, but note that it's easy enough to [roll your own](https://github.com/tpolecat/doobie/blob/series/0.2.x/build.sbt#L135-L153) if we want more information or a different format. The IDEA thing doesn't bother me. 

:+1: 

PS - You can do `managedSources` at the `sbt>` prompt to force all the code generators to run.

PR #20, opened 2016-04-08
Search by obsid

cquiroz:
This PR adds a UI to search an obsid from the ODB (Via the Seqexec server) and display it on a list. From here you can click on the + button and add the sequence to the queue. Both backends are supported

Note that we are still not saving the queue to the backend, this PR is mostly about the UI

screenshot 2016-04-07 20 58 06

Some things todo:

[ ] Check that the search result scrolling is done correctly
[ ] Some events are miss-firing on Firefox when adding a sequence to the queue
[ ] Expand the search area with animation
[ ] Add a button to close the search area
[ ] Seqexec server should indicate the state and instrument of a sequence


swalker2m:
It's a bit difficult to follow UI code in general, but we looked through this in the morning and it seems good to me.  :+1:

PR #28, opened 2016-04-11
Button component fixes firefox bug

cquiroz:
this PR brings a React button component and fixes a bug on the Search results button that wouldn't work on firefox. This version works correctly on firefox and chrome

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/SequenceSearch.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/SequenceSearch.scala
@@ -7,6 +7,7 @@ import edu.gemini.seqexec.web.client.model._
 import edu.gemini.seqexec.web.client.semanticui.SemanticUI._
 import edu.gemini.seqexec.web.client.semanticui.elements.button.Button
 import edu.gemini.seqexec.web.client.semanticui.elements.icon.Icon
+import edu.gemini.seqexec.web.client.semanticui.elements.icon.Icon.{IconPlus, IconSearch}
 import edu.gemini.seqexec.web.common.Sequence
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.prefix_<^._
@@ -59,9 +60,7 @@ object SequenceSearchResultsBody {
               <.td(u.instrument),
               <.td(
                 ^.cls := "collapsing",
-                Button(Button.Props(icon = true, circular = true, onClick = onAdding(p.searchResults, u)),
-                  Icon(Icon.Props("plus"))
-                )
+                Button(Button.Props(icon = Some(IconPlus), circular = true, onClick = onAdding(p.searchResults, u)))
               )
             )
           }
@@ -151,7 +150,7 @@ object SequenceSearch {
             ^.onChange ==> onChange,
             ^.value := s.searchText
           ),
-          Icon(Icon.Props("search", link = true, onClick = search))
+          IconSearch.copy(Icon.Props(id = IconSearch.p.id, link = true, onClick = search))
         ),
         <.div(
           ^.cls := "menu"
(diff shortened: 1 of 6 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/button/Button.scala
```diff
@@ -0,0 +1,70 @@
...
+      "basic"     -> p.basic,
+      "inverted"  -> p.inverted,
+      "circular"  -> p.circular
+    )
+  def component = ReactComponentB[Props]("Button")
+    .renderPC((_, p, c) =>
+      if (p.animated == NotAnimated)
+        <.button(
+          ^.cls := "ui button",
+          p.color.map(u => ^.cls := u),
```
swalker2m: Does this change the value of `.cls`?  If there is a color does that mean it isn't a "ui button"?

Maybe it would be nice if the colors were defined in a trait/case object enumeration somewhere.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/button/Button.scala
```diff
@@ -0,0 +1,70 @@
...
+          ^.tabIndex := p.tabIndex,
```
swalker2m: What is a `tabIndex`?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/SequenceSearch.scala
```diff
@@ -58,9 +59,8 @@ object SequenceSearchResultsBody {
               <.td(u.instrument),
               <.td(
                 ^.cls := "collapsing",
-                <.button(
-                  ^.cls := "circular ui icon button",
-                  Icon(Icon.Props("plus", onClick = onAdding(p.searchResults, u)))
+                Button(Button.Props(icon = true, circular = true, onClick = onAdding(p.searchResults, u)),
+                  Icon(Icon.Props("plus"))
```
swalker2m: Is there a way to just specify the `Icon` and not have to also set the property `icon`?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/button/Button.scala
cquiroz: It will add the class, not replace it, so you'd get e.g. "ui button red". Now that you mention it, I should try to integrate scalacss color support (https://japgolly.github.io/scalacss/book/features/typed-values.html) into these components

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/button/Button.scala
cquiroz: It is similar to the tab index of a swing component. It is required in this case because otherwise you cannot click into a button with just the keyboard.

swalker2m:
I have some questions about how things work (see comments) but this looks good.  👍

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/SequenceSearch.scala
cquiroz: Actually yes, you can do just `Icon("plus")`. We could also enumerate the possible icons, it is a long list but it would make it typesafer

cquiroz: I didn't totally get your comment, you were talking about having the Icon twice. The `Button` class could be improved to take an `Icon` in its props and assume the `icon` flag if that's the case

cquiroz:
Thanks for the comments, I improved the definitions to pass an `Icon` directly to a `Button` and added many Icon definitions to be used directly in the code rather than resorting to String ids

swalker2m:
Cool! 👍

PR #34, opened 2016-04-14
Sequence display

### LINKED ISSUE #21: Display a sequence when selected on the queue
On the queue area, when a sequence is selected, a detailed view of the sequence should be displayed on the lower section of the page

cquiroz:
This PR lets the user click on a sequence on the queue area and display its contents on the lower part of the screen. Sorry for the size, it is a lot of markup but there are some important model changes too

This fixes #21 even though there is still a lot of functionality that should be added. To keep the PR size reasonable, I've added new tasks for the missing bits

Here's a screenshot

screenshot 2016-04-14 12 54 05


andrewwstephens:
In the "Headers" field could you change "SOS" to "Operator"?

andrewwstephens:
We should also show the observation title in the "Queue" box at the top left.

andrewwstephens:
Could we display the file name for each executed and ongoing step?

andrewwstephens:
I suggest pulling the "Headers" box out of the "Running Sequences" box since they are independent of the instrument / sequence.

cquiroz:
Thanks @andrewwstephens for the feedback. I'll change the "SOS" label and make new issues for the other suggestions

cquiroz:
@andrewwstephens I'm not clear about having the title in the queue box, given that we could have multiple sequences running, Or do you mean have it on the table, per sequence?

andrewwstephens:
@cquiroz - Yes, I meant it would be useful to display the title of each sequence in the Queue table.  Maybe something like: ObsID, Instrument, Title, State.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/resources/seqexec-semantic-search.html
```diff
@@ -210,155 +210,268 @@
...
+                      </div>
+                      <div class="field">
+                        <label>Image Quality</label>
+                        <div class="ui fluid selection dropdown">
+                          <div class="default text">Select</div>
+                          <i class="dropdown icon"></i>
+                          <input type="hidden" name="iq">
+                          <div class="menu">
+                            <div class="item" data-value="male">IQ20</div>
+                            <div class="item" data-value="female">IQ70</div>
```
swalker2m: Hmm, pardon my ignorance but what is the difference between male and female values?

swalker2m:
I'm afraid I don't have much input to give for the client code. 👍

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/resources/seqexec-semantic-search.html
cquiroz: Sorry, copy paste, Those should be unique ids per item, likely something like `iq20`, `iq70` etc

PR #36, opened 2016-04-16
DhsClient simulator

### LINKED ISSUE #35: mock DHS Client
It would be good to have the ability to replace the current DHS client with a mock version for development

cquiroz:
This PR gives me an opportunity to dig into the seqexec server and fix a specific problem I'm having. I want to test with a mock dhs client. The current implementation only lets you use the real DHS client making it impossible harder to work without the VPN and it maybe risky to do calls to the real DHS

It is still an open question how'll configure the Seqexec, right now it is set to simulated instruments, simulated tcs and the new mock dhs client but we need to find out a way to properly configure it for different scenarios

This PR fixes #35

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
@@ -1,10 +1,8 @@
 package edu.gemini.seqexec.server
 
-import java.util.logging.Logger
-
 import argonaut._
 import Argonaut._
-import org.apache.commons.httpclient.{HttpMethod, HttpClient}
+import org.apache.commons.httpclient.HttpClient
 import org.apache.commons.httpclient.methods.{EntityEnclosingMethod, PutMethod, PostMethod}
 
 import scala.io.Source
@@ -12,9 +10,6 @@ import scala.io.Source
 import scalaz.concurrent.Task
 import scalaz.EitherT
 
-/**
- * Created by jluhrs on 11/5/15.
- */
 object DhsClient {
 
   val baseURI = "http://cpodhsxx:9090/axis2/services/dhs/images"
@@ -38,7 +33,7 @@ object DhsClient {
   }
 
   implicit def errorDecode: DecodeJson[Error] = DecodeJson[Error]( c => for {
-      t <- (c --\ "type").as[ErrorType]
+      t   <- (c --\ "type").as[ErrorType]
       msg <- (c --\ "message").as[String]
     } yield Error(t, msg)
   )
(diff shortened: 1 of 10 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
```diff
@@ -128,7 +96,61 @@ object DhsClient {
         internalKeywordConvert(k4), internalKeywordConvert(k5), internalKeywordConvert(k6)))
   }
 
-  implicit def keywordEncode: EncodeJson[InternalKeyword] = EncodeJson[InternalKeyword]( k =>
+}
+
+/**
+  * Implementation of DhsClient that interfaces with the real DHS oven the http interface
```
swalker2m: The "DHS oven" sounds intimidating!  :smile:

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
```diff
@@ -151,14 +173,32 @@ object DhsClient {
...
       "Unable to write keywords for image " + id)
+}
+
+/**
+  * Implementation of the Dhs client that simulates a dhs without external dependencies
+  */
+class DhsClientSim(date: LocalDate) extends DhsClient {
+  private val counter = new AtomicInteger(0)
+
+  val format = DateTimeFormatter.ofPattern("yyyyMMdd")
```
swalker2m: Cool, we should be using this everywhere instead of `SimpleDateFormat`!

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
cquiroz: Maybe I was thinking about DHS hell 😄

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
cquiroz: Yes! this formatter is immutable and thread-safe

swalker2m:
Looks good to me. 👍  Cool that you were able to use `Reader` for the client.  Seems like the kind of thing it was created for.

cquiroz:
I'm using `Reader` for Depenency Injection though we may need to change that as we have more dependencies, e.g. choosing between Tcs real and simulator, instruments, etc

tpolecat:
It would probably be cleaner to write `Reader[DhsClient, SeqAction[A]]` as `Kleisli[SeqAction, DhsClient, A]` and use some kind of abstract `Config` rather than `DhsClient` to allow you to add more information to it as the project evolves.

cquiroz:
Right, we are not just going to have the `DhsClient` as the only dependency for the server so we'll have to expand this. I'll add the suggested type on the next PR

tpolecat:
It would probably make sense to redesign this stuff in terms of `Free` because the type is getting complicated. Hopefully I'll have some time to work on this before the sun burns out.

Issue #43, opened 2016-04-21
Sequence run events


cquiroz:
We need to provide events while a sequence is running, e.g. when it starts, when a step completes, etc

PR #44, opened 2016-04-21
Step events

cquiroz:
This PR is the basis to fix issue #43. It uses a `scalaz-stream` process to publish events whenever a step completes. Those events are then taken on the backend and published as websocket frames.

The code is probably not very pure but it is working and avoids getting too deep into the complexity of the execution engine code.

Several items are pending but they could be postponed or worked on parallel while the UI is implemented

* We should support publishing on the play backend, e.g. use [streamz](https://github.com/krasserm/streamz) 
* Verify persistence of web sockets connection. It seems http4s is disconnecting them if they are idle for a while.
* Expose other events like pause, step error, completion
* Move event classes to a common module compiled to js and jvm


cquiroz:
Any takers on this PR? Note that several bits are just comments to guide me on what the functions do, let me know if you need clarifications on how it works

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
```diff
@@ -3,66 +3,99 @@ package edu.gemini.seqexec.server
...
-    stateRef.modify(_ + (id -> s))
+  def stepEvent(id: SPObservationID, s: ExecState): SeqexecEvent = s.completed.reverse match {
+    case Nil        => SequenceStartEvent(id.toString)
+    case Ok(r) :: _ => StepExecutedEvent(id.toString, s.completed.size, s.remaining.size, r.observeResult.dataId)
+    case l          => NullEvent //Fallback
+  }
+
+  private def recordState(id: SPObservationID)(s: ExecState): Task[ExecState] =
+    // This marks the current state for observation id
+    eventsQueue.publishOne(stepEvent(id, s)) >> stateRef.modify(_ + (id -> s)) >> Task.delay(s)
```
cquiroz: This is the key change on this PR. The function `recordState` is called on each step with  a snapshot of the current execution state. the call `eventsQueue.publishOne` pushes a `SeqexecEvent` into a scalaz-stream `Process` that in turn is sent to the web clients

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
```diff
@@ -80,6 +113,10 @@ class ExecutorImpl private (cancelRef: TaskRef[Set[SPObservationID]], stateRef:
       })
   }
 
+  // Important, make it a def so each client gets a new one
+  def sequenceEvents: Process[Task, SeqexecEvent] = Process.emit(SeqexecConnectionOpenEvent) ++ eventsQueue.subscribe
```
cquiroz: This the rest of the world listen for events generated inside the execution engine

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -17,18 +17,22 @@ import Scalaz._
...
         s     <- ExecutorImpl.read(obsId)
       } yield (obsId, s)
 
       r match {
         case \/-((i, s)) => Ok(write(List(Sequence(i.stringValue(), SequenceState.NotRunning, "Flamingos2", s.toSequenceSteps, None))))
         case -\/(e)      => NotFound(SeqexecFailure.explain(e))
       }
+    case GET -> Root / "seqexec" / "events" =>
+      // Stream seqexec events to clients
+      Ok(ExecutorImpl.sequenceEvents.map(write(_)))
```
cquiroz: http4s can take the `sequenceEvents` and stream them to web clients

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -14,13 +14,14 @@ import Scalaz._
...
 object Step {
+  val dhsClient = DhsClientSim(LocalDate.now())
+
   type Step = EitherT[Task, NonEmptyList[SeqexecFailure], StepResult]
 
-  def parConfig(config: List[SeqAction[ConfigResult]]):
-    EitherT[Task, NonEmptyList[SeqexecFailure], List[ConfigResult]] =
-      EitherT(Nondeterminism[Task].gather(config.map(_.run)).map(_.map(_.validationNel).sequenceU.disjunction))
+  private def parConfig(config: List[SeqAction[ConfigResult]]): EitherT[Task, NonEmptyList[SeqexecFailure], List[ConfigResult]] =
+    EitherT(Nondeterminism[Task].gather(config.map(_.run)).map(_.map(_.validationNel).sequenceU.disjunction))
```
swalker2m: Could this be (slightly) simplified to

````scala
  EitherT(Nondeterminism[Task].gather(config.map(_.run)).map(_.traverseU(_.validationNel).disjunction))
````

File: modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala (hunk 1 above)
Comment (swalker2m) [inline modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala]: I've never worked with scalaz.stream, but it seems okay to me. 👍  @tpolecat might have more constructive comments.  I'm curious why the input state is returned in the resulting task.

File: modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala (hunk 4 above)
Comment (cquiroz) [inline modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala]: To be honest, I can't parse either version 😄

File: modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala (hunk 1 above)
Comment (cquiroz) [inline modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala]: Right, I attempted another variant to extract the `ExecState` where I'd use the return type in the Task instead of `Task[Unit]`. I left it there but it is not needed for this use case

**Comment by tpolecat** [comment]
Hi sorry, I have not been watching OCS3 closely. I will look at this today.

File: modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala (hunk 4 above)
Comment (tpolecat) [inline modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala]: Yes, `_.map(_.validationNel).sequenceU` can be written `_.traverseU(_.validationNel)`.

File: modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala [hunk 5]
```diff
@@ -126,7 +129,7 @@ object Executor { self =>
       (liftA(liftT(go)) |@| liftA(gets(_.nextStep.nonEmpty)))(_ && _)
 
     // Run to completion, or to error, as long as not cancelled or complete
-    def runT(go: Task[Boolean], saveState: ExecState => Task[Unit]): ExecActionF[Unit] =
+    def runT[A](go: Task[Boolean], saveState: ExecState => Task[A]): ExecActionF[Unit] =
```
tpolecat: What is the intent of this parameterization? You're introducing a type variable that's never used.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
tpolecat: This seems Good. Stylistically you should prefer `*>` over `>>` because it is less constrained (it only requires `Apply` but `>>` requires `Bind` ... there is really never any reason to use `>>`).

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
cquiroz: Right, I was trying to obtain the `ExecState` out of this call instead of returning `Task[Unit]` but it may not be necessary now that I took another route

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
```diff
@@ -3,66 +3,99 @@ package edu.gemini.seqexec.server
...
 
 /** An executor that maintains state in a pair of `TaskRef`s. */
 class ExecutorImpl private (cancelRef: TaskRef[Set[SPObservationID]], stateRef: TaskRef[Map[SPObservationID, Executor.ExecState]]) {
+  type ExecutionFailures = NonEmptyList[SeqexecFailure] \/ Unit
+  type ExecutionResult = (ExecState, ExecutionFailures)
 
   private var loc = new Peer("localhost", 8443, null)
 
+  // Queue to keep events generated when executing a sequence
+  private val eventsQueue = async.topic[SeqexecEvent]()
```
tpolecat: Constructing this is a side-effect so I would prefer that it be passed to the constructor.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -14,13 +14,14 @@ import Scalaz._
 import scalaz.concurrent.Task
 
 object Step {
+  val dhsClient = DhsClientSim(LocalDate.now())
```
tpolecat: This is a side-effect.

tpolecat:
This seems fine for now. I think the whole design probably needs review as the requirements of the execution model have changed a lot as features have been added.

👍

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
cquiroz: This needs to be constructed somewhere and being Step an object there are no good places. I guess it could be parameter of the `step` method and be passed up the stack

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
cquiroz: If it were a param on the constructor, wouldn't be essentially the same? Why in that case it wouldn't be a side-effect?

tpolecat: It pushes the effect out to whoever is constructing the `ExecutorImpl`. It's not a huge deal since this whole thing needs review.

cquiroz: so, the trouble is with constructing the topic there instead of being a param, I guess the exact equivalence would be a constructor parameter with a default value?

tpolecat: Yes, but a side-effecting default value is probably even worse ;-)

cquiroz:
Incorporated some of the suggestions from the PR, in particular moving the initialization of the dhsclient and the events topic up the stack to the executor

PR #45, opened 2016-04-21
Use common edu.gemini.seqexec.odb bundle

cquiroz:
This PR is the counterpart of the ocs PR removing the `edu.gemini.seqexec.shared` bundle. It removes the code from ocs3 using the artifact from the repo instead


swalker2m:
Makes sense to me. 👍

Issue #41, opened 2016-04-20
Console for websocket messages


cquiroz:
We should have a development-only console displaying the web socket messages sent by the seqexec

PR #46, opened 2016-04-22
WebSockets console and new common module

cquiroz:
This PR brings fixes for #41 and a new module containing code shared between `edu.gemini.seqexec.server` and the web client.

The new module contains some classes, in particular `SeqexecEvent` (Added in PR #44) and moves them to a common `edu.gemini.seqexec.model` module, cross-compiled to jvm and js. This lets the web client to listen to them over a websocket connection. Note that we already have a cross compiled `edu.gemini.seqexec.web.shared` module but I couldn't make the dependencies work right to use it from the server module. Eventually the web shared module will be merged onto the top level `edu.gemini.seqexec.model` module

the UI was updated to include a Console displaying websocket messages as sent from the client. The code is designed so, that the console is only visible in development as this is of no use to regular users

Here's a screenshot of the console in firefox

screenshot 2016-04-22 13 17 16


modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/dhs.scala
```diff
@@ -0,0 +1,8 @@
+package edu.gemini.seqexec.model
+
+object dhs {
+
+  // Common data types
+  type ObsId = String
```
swalker2m: Is this a placeholder?  The observation ID should eventually be ported to Scala, placed in `core` and cross compiled once squants is ready.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
```diff
@@ -49,14 +50,28 @@ class SearchHandler[M](modelRW: ModelRW[M, Pot[SeqexecCircuit.SearchResults]]) e
...
+  }
+}
+
+/**
+  * Handles actions related to the development console
+  */
+class DevConsoleHandler[M](modelRW: ModelRW[M, SectionVisibilityState]) extends ActionHandler(modelRW) {
+  implicit val runner = new RunAfterJS
+
+  override def handle = {
```
swalker2m: I think it would help me in general if we explicitly type what methods produce.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/model.scala
```diff
@@ -54,11 +67,26 @@ case class SequencesOnDisplay(instrumentSequences: Zipper[SequenceTab]) {
   }
 }
 
+case class WebSocketsLog(log: List[SeqexecEvent]) {
+  // Upper bound of accepted events or we may run out of memory
+  val maxLength = 100
+  def append(s: String):WebSocketsLog = copy((log :+ read[SeqexecEvent](s)).take(maxLength))
```
swalker2m: Should be `maxLength - 1` ?

swalker2m:
I'm not really a good candidate to review this, but for what it's worth it looks fine to me. 👍

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/dhs.scala
cquiroz: Yes, it is a placeholder. It was on the original seqexec server and I moved it here to make it shared between the web server and seqexec server
It should be a `SPObservationID` but that's a Java class

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
cquiroz: Ok, I'll add types to them. In this particular case handle is `PartialFunction[AnyRef, ActionResult[M]]` where `M` is essentially our root model

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/model.scala
cquiroz: 👍

cquiroz:
Thanks for this. Ideally I'd like to separate non-GUI PRs from GUI-only PRs (which maybe quicker to check) to facilitate review, but it is difficult to separate them, especially with the current pace of development

PR #47, opened 2016-04-27
Run tracking

### LINKED ISSUE #33: Completion message
A message should be played when a sequence completes

### LINKED ISSUE #37: Show file name per step
Once a step completes we should display the filename on the step's table

cquiroz:
This PR is rather large and contains quite a few changes to both the model and the UI. The code covers what was shown on the Apr 26th demo. It allows to run sequences and follow the execution on the web ui with events sent over the WebSockets channel

Given that I cannot fully extract the state of execution from the engine some of the complexities in this PR come from simulating those events at the higher level. In particular, aborting a sequence is very much emulated

Here's a screenshot of a sequence being executed

screenshot 2016-04-27 15 55 20

Fixes #37 
Fixes #33


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
```diff
@@ -28,15 +28,16 @@ class ExecutorImpl private (cancelRef: TaskRef[Set[SPObservationID]], stateRef:
...
+    case Nil                               => List(SequenceStartEvent(id.toString))
+    case Ok(r) :: _ if s.remaining.isEmpty => List(StepExecutedEvent(id.toString, s.completed.size, s.remaining.size, r.observeResult.dataId), SequenceCompletedEvent(id.toString))
+    case Ok(r) :: _                        => List(StepExecutedEvent(id.toString, s.completed.size, s.remaining.size, r.observeResult.dataId))
+    case l                                 => List(NullEvent) //Fallback
   }
 
   private def recordState(id: SPObservationID)(s: ExecState): Task[Unit] =
     // This marks the current state for observation id
-    eventsQueue.publishOne(stepEvent(id, s)) *> stateRef.modify(_ + (id -> s))
+    stepEvent(id, s).map(eventsQueue.publishOne).sequenceU *> stateRef.modify(_ + (id -> s))
```
swalker2m: `stepEvent(id, s).traverseU(eventsQueue.publishOne)` ...

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/model.scala
```diff
@@ -7,20 +7,164 @@ import Scalaz._
...
+  case object NotDone extends StepState
+  case object Running extends StepState
+  case object Done    extends StepState
+  case object Error   extends StepState
+  case object Abort   extends StepState
+  case object Paused  extends StepState
+
+  implicit val show = Show.shows[StepState] {
+    case Running    => "Running"
+    case NotDone    => "Pending"
```
swalker2m: Would `Pending` be a better name for the case object?

swalker2m:
Probably someone less ignorant about the UI code should evaluate these, but in lieu of that 👍

cquiroz:
Thanks, I have some code there using lenses and state that I was unsure about,if you can check it

swalker2m:
@cquiroz I saw that, it all looked great to me.  The live demo the other day was excellent too.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
cquiroz: Besides one option being shorter, is there other difference on `traverseU` insteado of `map(...).sequenceU`?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/model.scala
cquiroz: Right, I expect these objects to be properly defined on spmodel at some moment, for now they are mostly UI model classes

PR #48, opened 2016-05-04
Terminal ui

cquiroz:
This PR is mostly for fun. I wanted to make an old-fashion terminal like the one we had on felix but on the web. This PR takes the [jquery.terminal plugin](http://terminal.jcubic.pl/), adds a scala.js facade and integrates with the web based command terminal. The plugin has support for history (even across browser restarts), shortcuts, tab completion

It is also a good opportunity to check how to make multiple webapps under the same umbrella server

Note that this is not a generic terminal API, this is meant only for the seqexec, in the future it could be expanded to let other apps include commands

Here's a screenshot

screenshot 2016-05-04 15 30 08


sraaphorst:
I'm not familiar at all with scala.js, but this was pretty cool and I think I picked up a few things 👍

cquiroz:
In general you shouldn't need to be very aware that you are in scala.js, It should be essentially scala. This particular PR is a bit different since it implements a facade for jquery-terminal, though it is very small at that

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client.cli/src/main/scala/edu/gemini/seqexec/web/client/cli/SeqexecTerminal.scala
```diff
@@ -0,0 +1,191 @@
...
+      .greetings(banner + s"\nVersion: ${OcsBuildInfo.version}\n")
+      .completion((t: Terminal, c: String, callback: CompletionCallback) => callback(cmdStrings.toJSArray)))
+  }
+
+  val banner = """  ___
+                 | / __| ___ __ _ _____ _____ __
+                 | \__ \/ -_) _` / -_) \ / -_) _|
+                 | |___/\___\__, \___/_\_\___\__|
+                 |             |_|               """.stripMargin
+}
```
jluhrs: The 'x' looks weird.

jluhrs:
It looks OK, at least the little I understood. It has my :+1:

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client.cli/src/main/scala/edu/gemini/seqexec/web/client/cli/SeqexecTerminal.scala
arturog8m: http://patorjk.com/software/taag/#p=display&v=1&f=Star%20Wars&t=Seqexec

PR #56, opened 2016-05-12
Play Websockets

### LINKED ISSUE #55: Support websockets channel on the play backend

cquiroz:
The Websocket channel was made using `scalaz-stream`. The http4s backend was able to use the stream directly, but the play backend was left pending as it is not compatible with `scalaz-stream` preferring `akka-streams` instead.

This PR uses the `streamz` library to convert from one type of stream to the other, letting the play backend use the same data source

Fixes #55


jluhrs:
:+1:

cquiroz:
Rebased onto develop and updated to the latest streamz

PR #58, opened 2016-05-20
Authentication part I

cquiroz:
This PR sets the basis to support authentication on the Seqexec webapp.

It can do authentication against the LDAP/AD server, with support for a test in-memory auth database for testing. This is done with the [UboundId SDK](unboundid.com/products/ldapsdk/) library. Besides authentication we can read the display name

Once auth is established it will construct a [JSON Web Token](https://jwt.io/) and return it to the client as a cookie. See the rationale behind using a [cookie for storing](https://stormpath.com/blog/where-to-store-your-jwts-cookies-vs-html5-web-storage) the token. Tokens are built using the [JWT Scala](https://github.com/pauldijou/jwt-scala) library using HMAC with SHA-512 and a single secret key

Note that this approach is stateless, we don't need to store the cookies as they can be decoded and authenticated

Two routes (login/logout) are added to the backends to let the UI interact with the authentication

There are several steps pending:
* Make the UI to login/logout
* Protect the routes with authentication
* Support authentication on the web socket channel
* Ensure we are XSRF resistant
* Ensure cookies expire properly
* Improve configuration, the secret key cannot be on the clear as in this PR
* Check that the user is allowed to use the seqexec (LDAP Groups?)

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/authentication.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/authentication.scala
@@ -20,12 +20,26 @@ trait AuthenticationService {
   def authenticateUser(username: String, password: String): AuthResult
 }
 
-object AuthenticationService {
-  type AuthResult = AuthenticationFailure \/ UserDetails
-
+object AuthenticationConfig {
   // TODO externalize
   val sessionTimeout = 8 * 3600
   val onSSL = false
+  val key = "secretkey"
+
+  val testMode = true
+  val ldapHost = "gs-dc6.gemini.edu"
+  val ldapPort = 3268
+
+  val ldapService = new LDAPService(ldapHost, ldapPort)
+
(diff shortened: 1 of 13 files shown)


jluhrs:
:+1:

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/authentication.scala
```diff
@@ -0,0 +1,87 @@
...
+    def authenticateUser(username: String, password: String): AuthResult = {
+      @tailrec
+      def go(l: List[AuthenticationService]): AuthResult = l match {
+        case Nil      => -\/(NoAuthenticator)
+        case x :: Nil => x.authenticateUser(username, password)
+        case x :: xs  => x.authenticateUser(username, password) match {
+            case u @ \/-(_) => u
+            case -\/(e)     => go(xs)
+          }
+      }
```
swalker2m: The failure will be whatever is reported by the last authenticator, but I imagine that's fine.

swalker2m:
I don't know much about this stuff or the libraries involved but it all looks quite reasonable to me. 👍

PR #60, opened 2016-06-07
Experiment to test some Seqexec Executor concepts.

jluhrs:
This pull is not intended for merging with the central ocs3 repository. Its purpose is to expose some of the ideas I have for the Sequence Executor engine. See the README.md for more details.

### DIFF
--- a/SimpleGame/README.md
+++ b/SimpleGame/README.md
@@ -1,0 +1,18 @@
+I wrote this program as an exercise to explore some concepts for the Sequence Executor engine. The program is an incomplete implementation of a simple game, where a ball moves inside a box at constant speed, changing direction in response to user inputs. The ball position update is driven by a simulated periodic timer. The periodic timer and the user inputs are modeled as streams of data.
+The concepts I wanted to test are presented below. 
+# Model the basic execution unit as a state machine that produces status updates.
+The execution of a sequence can be modeled as a state machine that produces a stream of status updates (I’m using the term status update to refer to a snapshot of the sequence execution state, with the information relevant to the clients). The output stream will ultimately feed the web server to send updates to the clients. The state machine is constructed as a chain of smaller state machines. Each one takes care of an action in the sequence and produces a status update.
+A second related idea is to compose the state that is passed around with the sequence execution state and the sequence commands (just the ones that affect a running sequence: `stop`, `setBreakpoint`). The action state machine will then be composed of two stages. The first one gets the input commands from the real world and updates the commands part of the state. The second one executes the action according to the previous state and the commands. For example, if the commands section has `stop==true`, then the action part should end the execution and close the output stream. That structure isolates the interaction with the real world from the rest of the processing. Part of this idea is implemented in _Game.scala_, where the state machine created by `updateSpeed` takes a player input and updates the speed direction in the state. The updated state is then passed to the state machine created by the function `tick`, which provides a new state with an updated position.
+# Time control everywhere
+Another concept I want to use is to treat time as an input (an idea I first read about in John Carmack’s .plan). To achieve this I need to:
+Have a single source of time (and forbid having calls to system timers everywhere)
+Timestamp every external event.
+Use timestamps to determine time order and time concurrency.
+# What do we gain ?
+Introducing these concepts offers several interesting advantages:
+- Modeling inputs and outputs as streams, combined with the time control, makes it easier to write unit tests. 
+- It also makes possible to capture input events from the real world, and reproduce them later for debugging purposes. Because of time control, reproduction can be done at different speeds.
+- It makes relatively easy to add new commands without affecting the code structure.
+# What is next ?
+As fun as basic streams are, they are not the solution. My plan is to use scalaz-streams. I‘ll start by introducing them to this little game, as an exercise.
+There are a lot of details that I still haven’t figure out for the Sequence Executor engine. For example, this game is driven by the periodic timer: every time the timer ticks, the game machine produces an output with the new player position. At those times, the player inputs are read by polling. But, what should drive the Executor engine? The responses from the EPICS systems? What if such response takes too long? Wouldn’t that make the engine unresponsive to user commands?
--- a/SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/ConsoleKeys.scala
+++ b/SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/ConsoleKeys.scala
@@ -1,0 +1,41 @@
+package edu.gemini.experimental.jluhrs
+
+import edu.gemini.experimental.jluhrs.EventStream._
+import edu.gemini.experimental.jluhrs.Game._
+
+import scalaz.concurrent.Task
+import scalaz.stream.{Process, async}
+import scala.concurrent.duration._
+import sys.process._
+
+/**
+  * Created by jluhrs on 6/15/16.
+  */
+object ConsoleKeys {
+  // Build a process that produces player inputs from keystrokes
+  private val inputQ = async.boundedQueue[Event[PlayerInput]](10)
+
+  val inputP: Process[Task, Event[PlayerInput]] = inputQ.dequeue
+
+  private val keySource: Task[Unit] = Task {
+
+    // Some munbo jumbo to get keystrokes unbuffered.
+    (Seq("sh", "-c", "stty -icanon min 1 < /dev/tty") !)
+    (Seq("sh", "-c", "stty -echo < /dev/tty") !)
+
+    while (true) {
+      val c = Console.in.read
+      c match {
+        case 119 => inputQ.enqueueOne((System.nanoTime nanoseconds, Up)).unsafePerformSync
+        case  97 => inputQ.enqueueOne((System.nanoTime nanoseconds, Left)).unsafePerformSync
+        case 115 => inputQ.enqueueOne((System.nanoTime nanoseconds, Down)).unsafePerformSync
+        case 100 => inputQ.enqueueOne((System.nanoTime nanoseconds, Right)).unsafePerformSync
+        case _ => ()
+      }
+    }
+    ()
+  }
+
+  def start = keySource.unsafePerformAsync(_ => ())
+
+}
--- a/SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/ConsoleKeys.scala
+++ b/SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/ConsoleKeys.scala
@@ -1,5 +1,7 @@
 package edu.gemini.experimental.jluhrs
 
+import java.util.concurrent.atomic.AtomicBoolean
+
 import edu.gemini.experimental.jluhrs.EventStream._
 import edu.gemini.experimental.jluhrs.Game._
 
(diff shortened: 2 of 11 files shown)


cquiroz:
Can you extract the state of the game at any given time?

jluhrs:
Kinda. The output is a stream of `Position`s, which are extracted from the game state at each cycle in `runGameCycle`. I could have put the whole game state in the output. But I guess you are thinking of the use case in Seqexec when a client requests the state of a sequence for the first time. I didn't contemplate that case. From what I have read so far of scalaz-stream, we could put the output in a `Topic`, and create a `Process` that listen to that `Topic` and to client requests for sequence state, to produce the reply with the latest sequence state.

SimpleGame/README.md
```diff
@@ -0,0 +1,18 @@
...
+A second related idea is to compose the state that is passed around with the sequence execution state and the sequence commands (just the ones that affect a running sequence: `stop`, `setBreakpoint`). The action state machine will then be composed of two stages. The first one gets the input commands from the real world and updates the commands part of the state. The second one executes the action according to the previous state and the commands. For example, if the commands section has `stop==true`, then the action part should end the execution and close the output stream. That structure isolates the interaction with the real world from the rest of the processing. Part of this idea is implemented in _Game.scala_, where the state machine created by `updateSpeed` takes a player input and updates the speed direction in the state. The updated state is then passed to the state machine created by the function `tick`, which provides a new state with an updated position.
+# Time control everywhere
+Another concept I want to use is to treat time as an input (an idea I first read about in John Carmack’s .plan). To achieve this I need to:
```
swalker2m: Maybe you could provide a reference to this?

SimpleGame/README.md
```diff
@@ -0,0 +1,18 @@
...
+Have a single source of time (and forbid having calls to system timers everywhere)
+Timestamp every external event.
+Use timestamps to determine time order and time concurrency.
+# What do we gain ?
+Introducing these concepts offers several interesting advantages:
+- Modeling inputs and outputs as streams, combined with the time control, makes it easier to write unit tests. 
+- It also makes possible to capture input events from the real world, and reproduce them later for debugging purposes. Because of time control, reproduction can be done at different speeds.
+- It makes relatively easy to add new commands without affecting the code structure.
+# What is next ?
+As fun as basic streams are, they are not the solution. My plan is to use scalaz-streams. I‘ll start by introducing them to this little game, as an exercise.
```
swalker2m: You're most certainly right but it might be good to explicitly state what you hope to gain by using scalaz-streams.

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/EventStream.scala
```diff
@@ -0,0 +1,26 @@
...
+  * Created by jluhrs on 6/2/16.
+  */
+
+// My attempt to describe external inputs as a stream of values with a timestamp.
+
+object EventStream {
+  import Timer._
+
+  type Event[A] = (Time, A)
+  type EventStream[A] = Stream[Option[Event[A]]]
```
swalker2m: Why `Option[Event[A]]`?  Is there a difference between a `None` entry and just not having an entry?

SimpleGame/README.md
cquiroz: One place where scalaz-stream are useful is to be used as source for websocket events

SimpleGame/README.md
```diff
@@ -0,0 +1,18 @@
...
+# Time control everywhere
+Another concept I want to use is to treat time as an input (an idea I first read about in John Carmack’s .plan). To achieve this I need to:
+Have a single source of time (and forbid having calls to system timers everywhere)
+Timestamp every external event.
+Use timestamps to determine time order and time concurrency.
+# What do we gain ?
+Introducing these concepts offers several interesting advantages:
+- Modeling inputs and outputs as streams, combined with the time control, makes it easier to write unit tests. 
+- It also makes possible to capture input events from the real world, and reproduce them later for debugging purposes. Because of time control, reproduction can be done at different speeds.
```
cquiroz: In the particular case of the seqexec we probably want to get the instrument output (Or input to the seqexec) too. How possible would it be to, e.g. simulate instrument failures using this scheme?

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/EventStream.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  // Returns all the events in the stream that happened before a given time.
+  def pastEvents[A](es: EventStream[A], t: Time): (List[Event[A]], EventStream[A]) = {
+    def f (s: EventStream[A], acc: List[Event[A]]): (List[Event[A]], EventStream[A]) = s match {
+      case Some((t1, a)) #:: rest =>  if (t1 > t) (acc, s)
+                                      else f (rest, (t1, a) :: acc)
```
cquiroz: I think for a very long stream this will blow the stack, possible you may re-arrange it to make it tail recursive

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/EventStream.scala
```diff
@@ -0,0 +1,26 @@
...
+      case _                      => (acc, s)
+    }
+    f(es, List()) match {
+      case (l, s) => (l.reverse, s)
+    }
+  }
```
swalker2m: I'm probably missing something but it almost seems this could be accomplished with `es.takeWhile { case (t0, a) => t0 < t }`?

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Main.scala
```diff
@@ -0,0 +1,39 @@
...
+    System.out.print(s)
+    System.out.flush();
+
+    //This is just to animate the frames. The real timing must come from the input timing stream.
+    Thread.sleep(500)
+  }
+
+  def main(args: Array[String]): Unit =
+    (runGame(width, height, periodicTimer(0.0, 1.0),
+      Stream[Option[Event[PlayerInput]]](Some(5.0, Right), Some(15.0, Up), Some(25.0, Right))
```
cquiroz: So, in the seqexec case, this would be a stream of, e.g. user actions and instrument inputs?

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Game.scala
```diff
@@ -0,0 +1,92 @@
...
+  */
+
+/* The game logic is here.
+ * It's just an object that moves at constant speed, and refresh its position at given times.
+ * The direction of movement is affected by some input events.
+ */
+object Game {
+
+  type Position = (Double, Double)
+  type SpeedDir = (Double, Double)
```
swalker2m: I know this is just a toy exercise but `case class`es to give these two concepts distinct types and make clear what they contain would be useful.

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Game.scala
```diff
@@ -0,0 +1,92 @@
...
+  sealed trait PlayerInput
+
+  sealed trait DirectionalInput extends PlayerInput
+
+  case object Up extends DirectionalInput
+  case object Down extends DirectionalInput
+  case object Right extends DirectionalInput
+  case object Left extends DirectionalInput
+
+  implicit def decodeSpeedDir(i: DirectionalInput): SpeedDir = i match {
```
swalker2m: This is to allow you to use a `DirectionalInput` where a `SpeedDir` is expected?  This kind of implicit conversion is generally discouraged because it can be hard to follow.  You can make the transition explicit by adding a method `toSpeedDir` in `DirectionalInput` like so:

````scala
sealed trait DirectionalInput extends PlayerInput {
  def toSpeedDir: SpeedDir = this match {
    // conversion logic here
  }
}
````

or I guess just remove `implicit` and call this directly when you need it.

File: SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Timer.scala [hunk 10]
```diff
@@ -0,0 +1,17 @@
+package edu.gemini.experimental.jluhrs
+
+/**
+  * Created by jluhrs on 5/19/16.
+  */
+
+// The source of time (can we really control time ?)
+object Timer {
+  // For the sake of simplicity, Time is defined as Double
+  type Time = Double
```
swalker2m: Is time just an ever increasing counter used for ordering events or is it a timestamp corresponding to a real-world time?  Either way, why a `Double`?

SimpleGame/README.md
jluhrs: Yes. In a more generic sense, all external stimulus should be considered an input event.

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/EventStream.scala
jluhrs: You are right. I'm not sure what I was thinking. It was about how to deal with an empty input stream, but I should have just used Stream.Empty.

SimpleGame/README.md
jluhrs: Someone put all those .plan files in Github. Here is the one about time: https://raw.githubusercontent.com/ESWAT/john-carmack-plan-archive/master/by_day/johnc_plan_19981014.txt

SimpleGame/README.md
jluhrs: It occurred to me that one way to model the Sequence Executor was to view it as a machine that processes inputs and produce outputs. Inputs and outputs deal with the real world, while the processing can be kept functionally pure. Input and output cookies, with a creamy funcional filling, if you like. The tutorial I read about scalaz-stream starts with the same idea, that is why I thought it could be a good match. But I'll not really know until I get my hands dirty.

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/EventStream.scala
jluhrs: Yes, you are right. Proof that I'm still not fluent in Scala.

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Main.scala
jluhrs: That's correct! Here is a candy for you :candy:

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Timer.scala
jluhrs: No real reason.

cquiroz:
I think this PR seems fine, though I still miss somewhat the connection to the seqexec engine. Would the next version be updated to consider the uses cases required for a seqexec, even if applied to the game?

I find the notion of time as in input very interesting. It would be great if we could save a `seqexec session` and replay it for testing at full speed

swalker2m:
I agree with Carlos.  All of my comments were on technicalities of the implementation.  I wan't sure what you meant by "the state machine is constructed as a chain of smaller state machines", but the ideas sound promising in general to me.

jluhrs:
New commit! Now SimpleGame uses scalaz-stream to trigger periodic processing and handling user inputs (use the key W, A, S, D to direct the O)
One obvious flaw is that I didn't figure out out to finish the game gracefully.
I didn't try to implement the idea of having only one source for time.

jluhrs:
New commit. The main change is that now it is possible to terminate the program pressing Esc.

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/ConsoleKeys.scala
```diff
@@ -0,0 +1,41 @@
...
+import sys.process._
+
+/**
+  * Created by jluhrs on 6/15/16.
+  */
+object ConsoleKeys {
+  // Build a process that produces player inputs from keystrokes
+  private val inputQ = async.boundedQueue[Event[PlayerInput]](10)
+
+  val inputP: Process[Task, Event[PlayerInput]] = inputQ.dequeue
```
cquiroz: Note that this will let you have a single listener on the queue, that maybe fine depending on your use case. For the seqexec I used a topic instead since we have many clients

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Game.scala
```diff
@@ -29,63 +37,42 @@ object Game {
...
+    State[GameState, Unit] { s => {
+      (s.copy(t=time, position=modPos(move(time-s.t, s.position, s.speedDir, s.velocity), s.boardWidth, s.boardHeight ) ), ()) }
+    }
   }
 
-
-  def runGameCycle(s0: GameState, ss: => Stream[State[GameState, Unit]]): Stream[Position] = {
-    s0.position #:: runGameCycle(ss.head(s0)._1, ss.tail)
+  // State machine to run one game cycle.
+  def gameCycle(in: Event[Input]): State[GameState, \/[GameState, GameState]] = in match {
```
cquiroz: `\/[GameState, GameState]`? doesn't seem too informative

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/Main.scala
```diff
@@ -22,18 +29,17 @@ object Main {
...
 
-  def main(args: Array[String]): Unit =
-    (runGame(width, height, periodicTimer(0.0, 1.0),
-      Stream[Option[Event[PlayerInput]]](Some(5.0, Right), Some(15.0, Up), Some(25.0, Right))
-    ).take(50).toList map (frame(width, height, border, _))).map(displayFrame)
+  def main(args: Array[String]): Unit = {
+
+    ConsoleKeys.start
+
+    (runGame(width, height, (awakeEvery(100 milliseconds).map((_, TimeTick)) merge ConsoleKeys.inputP))
```
cquiroz: So this merges the time tick with the console input, nice.
How would it work with the real time?

SimpleGame/src/main/scala-2.11/edu/gemini/experimental/jluhrs/ConsoleKeys.scala
```diff
@@ -17,25 +19,34 @@ object ConsoleKeys {
...
 
-    while (true) {
-      val c = Console.in.read
-      c match {
-        case 119 => inputQ.enqueueOne((System.nanoTime nanoseconds, Up)).unsafePerformSync
-        case  97 => inputQ.enqueueOne((System.nanoTime nanoseconds, Left)).unsafePerformSync
-        case 115 => inputQ.enqueueOne((System.nanoTime nanoseconds, Down)).unsafePerformSync
-        case 100 => inputQ.enqueueOne((System.nanoTime nanoseconds, Right)).unsafePerformSync
-        case _ => ()
+    while (continue.get) {
```
cquiroz: This will likely work but I wonder if there is some other way, sort of sending a poison pill to the stream and it would terminate itself

PR #64, opened 2016-06-10
LDAP Free

cquiroz:
This PR contains  a rewrite of the ldap query classes in terms of free monads. It is probably overkill for this case since we essentially have only one type of operation/interpreter but it is a good learning exercise. 

The PR is mostly for learning about Free monads and it doesn't add much though I included the ability to read a users' groups and thumbnail photo from AD

Many thanks to @tpolecat to get this done


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
```diff
@@ -0,0 +1,119 @@
...
+  */
+object FreeLDAPAuthenticationService {
+  import UserDetails._
+  import LdapConnectionOps._
+
+  sealed trait LdapOp[A]
+  // Operations on ldap
+  object LdapOp {
+    // Operation to authenticate a user, It returns true if it works
+    case class AuthenticateOp(username: String, password: String) extends LdapOp[UID]
```
swalker2m: Returns a `UID`, not a `Boolean`, if it works?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
```diff
@@ -0,0 +1,119 @@
...
+    // Read the user display name
+    case class UserDisplayNameOp(uid: UID)                        extends LdapOp[DisplayName]
+    // Reads the name, groups and thumbnail
+    case class DisplayNameGrpThumbOp(uid: UID)                    extends LdapOp[(DisplayName, Groups, Option[Thumbnail])]
+  }
```
swalker2m: I'd be inclined to add a `ConnectOp` to the algebra, but maybe that's stupid?  Did you consider it or should it obviously not be included for some reason?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
cquiroz: Yes, returns the user id from LDAP. If it fails it throws an exception
in Free monads you can either go this way just letting the operations break through an exception or handle the exceptions in each case. From @tpolecat I got this is the regular/easiest way to handle this but there are cases when you want to catch the exception

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
cquiroz: A `ConnectOp` would mean the algebra kind of knows that we are connecting to something. Note that the interpreter could just use an in-memory database, like in the test case. I think we don't need a `ConnectOp` but I can certainly be wrong

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
swalker2m: I was just wondering why the comment read "It returns true if it works" and yet it is an `LdapOp[UID]`.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
```diff
@@ -0,0 +1,119 @@
...
+  import UserDetails._
+  import LdapConnectionOps._
+
+  sealed trait LdapOp[A]
+  // Operations on ldap
+  object LdapOp {
+    // Operation to authenticate a user, It returns true if it works
+    case class AuthenticateOp(username: String, password: String) extends LdapOp[UID]
+    // Read the user display name
+    case class UserDisplayNameOp(uid: UID)                        extends LdapOp[DisplayName]
```
cquiroz: This is maybe for @tpolecat. To call `UserDisplayNameOp` and `DisplayNameGrpThumbOp` I need to previously call `AuthenticateOp`. Can I enforce that restriction somehow?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
cquiroz: Oh, initially it was a boolean, the comment is stale. I thought about just doing auth true/false but you need the uid for the next operations

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
swalker2m: That makes sense, and yet the `IO` and `Task` transformations certainly know and cannot be done without a connection.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/test/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationServiceSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+  case class MockAuthDB(users: Map[UID, (String, DisplayName)], acceptEmptyPwd: Boolean) {
+    def authenticate(u: String, p: String): UID = {
+      // This checks if the username and password but lets it bypass it
+      if (users.contains(u) && ((u == p && p.nonEmpty) || acceptEmptyPwd)) u else throw new RuntimeException()
+    }
+    def displayName(uid: UID): DisplayName = ~users.get(uid).map(_._2)
+  }
+
+  // Natural transformation to Id with a mock auth db
+  def toMockDB(db: MockAuthDB): LdapOp ~> Id =
```
cquiroz: @swalker2m  In the case of an in memory database you don't need to know about the connection

swalker2m:
Looks great.  👍

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
tpolecat: You will get the `UID` is via `AuthenticateOp` so this seems ok to me as-is. Treating a foreign `UID` as an error condition seems ok. There are ways to enforce this at compile-time but I don't think it's worth the complexity here.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
```diff
@@ -0,0 +1,119 @@
...
+          case LdapOp.DisplayNameGrpThumbOp(uid) => IO(c.nameGroupsThumb(uid))
+      }
+    }
+
+  // Natural transformation to Task
+  def toTask(c: LDAPConnection): LdapOp ~> Task =
+    new (LdapOp ~> Task) {
+      def apply[A](fa: LdapOp[A]) =
+        fa match {
+          case LdapOp.AuthenticateOp(u, p)       => Task(c.authenticate(u, p))
```
tpolecat: You should use `Task.delay` here; `Task.apply` will cause the operation to run on an implicit thread pool which probably isn't what you want. In fact the whole interpreter could be `Task.delay(toIO(c).unsafePerformIO)` if you're feeling lazy.

tpolecat:
Agree, this is great. I hope we see more things done this way. 👍

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
cquiroz: `UID` is just an alias for `String`, so you can easily supply one. But it could be a proper type and then set the constructors in such a way that it could only be build inside `AuthenticateOp`

But as you said, it would be overkill in this case

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/FreeLDAPAuthenticationService.scala
cquiroz: What would be the difference between using `Task.now` instead of `Task.delay`? It isn't that clear from the docs

tpolecat: `Task.now` computes a constant value when the `Task` is constructed. `Task.delay` captures a side-effect and evaluates it each time the computation is run. The latter is what you want if the argument is a side-effect, as it is here.

cquiroz: Right, it makes sense

cquiroz:
Ready to merge with the latest suggestions and rebased on develop

PR #65, opened 2016-06-15
Client logging

### LINKED ISSUE #14: Client logging
We should support logging with support for sending the logs to the server for remote storage. See as an example [SPA Tutorial logger](https://github.com/ochrons/scalajs-spa-tutorial/tree/master/client/src/main/scala/spatutorial/client/logger)

cquiroz:
This PR integrates onto the client the scala.js version of  java util logging. This lets us use the same api on the frontend and backend. Additionally it will post the log messages to the backend so that we can have a record of the client activity

Fixes #14

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/LogMessage.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/LogMessage.scala
@@ -1,0 +1,45 @@
+package edu.gemini.seqexec.web.common
+
+import java.util.logging.{Level, LogRecord}
+
+import upickle.Js
+
+import scala.collection.breakOut
+
+/** Minimal class to transfer log messages from client to server
+  * We only care about level and message
+  */
+case class LogMessage(level: Level, msg: String)
+
+object LogMessage {
+  def fromLogRecord(record: LogRecord): LogMessage = {
+    // We may want to encode time, but the timezone on the client cannot be
+    // easily recorded. Discard exceptions too
+    LogMessage(record.getLevel, record.getMessage)
+  }
+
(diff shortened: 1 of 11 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/LogMessage.scala
```diff
@@ -0,0 +1,45 @@
...
+  // Map levels to numbers to reduce space
+  private val Int2LevelMapping: Map[Int, Level] = Map(0 -> Level.OFF,
+    1 -> Level.SEVERE,
+    3 -> Level.WARNING,
+    4 -> Level.INFO,
+    5 -> Level.CONFIG,
+    6 -> Level.FINE,
+    7 -> Level.FINER,
+    8 -> Level.FINEST,
+    9 -> Level.ALL)
```
swalker2m: You skipped `2`?  I suppose you could also make a `List` in the order you want, zip with index, and turn it into a `Map`.  Is it worth worrying about saving space?  The text of the message itself is presumably much larger anyway?

cquiroz: Right silly mistake. You're right, it is a small space saving compared to the message. I will remove this when we go to binary anyway and compression should take care of this

swalker2m:
Seems cool that you can send log messages back to the server.  Presumably we'll want to be judicious about using it though. 👍

cquiroz:
I think it is very powerful, then we know, e.g. when somebody opened the client and from where in a central place. We are not going to have access to the client logs. As you correctly point out it can generate a lot of traffic, thus only INFO and higher messages are sent and we should be moderate. Some of the actions we can log in the server regardless, e.g. when a user runs a sequence

Issue #67, opened 2016-06-20
Improve launch script to run as daemon


cquiroz:

cquiroz:
Done

Issue #62, opened 2016-06-07
Define a mechanism to configure applications


cquiroz:
system properties, configuration file

Other options:
[PureConfig](https://github.com/melrief/pureconfig)
[Ficus](https://github.com/iheartradio/ficus)
[Knobs](https://github.com/oncue/knobs)

tpolecat:
My vote is likely to be for `Knobs` because it provides a pure interface and has a mechanism for notifying applications when the the config changes, which could be helpful.

PR #68, opened 2016-06-20
Seqexec distribution

### LINKED ISSUE #10: Package seqexec js application with FullOptJS
Ideally the packaged javascript should be compressed and optimized with FullOptJs rather than just FastOptJS

### LINKED ISSUE #42: SLF4J configure
Both http4s and play use slf4j but at the moment it is not configured. Add a default useful configuration sending the logs to file and/or console

### LINKED ISSUE #53: Make a standalone distribution
Ideally on RPM and tar.gz format with launch scripts

cquiroz:
This is a rather large change on the build system of ocs3, so I should be checked by @gemini-hlsw/core  as much as possible. It lets us deploy the seqexec server either locally or to a separate server, at least on linux 64.

The structure of the sbt build had to be changed quite a bit as I had many troubles having references between projects. Unfortunately `ProjectRef` doesn't seem to work for cross-compiled projects so I adopetd a flat structure where all the projects are inside `Build.scala`. Note that we'll need to phase this out when adopting `sbt-1.0`. If you know a better way to refer to projects not located at the top let me know

How to build the distributions has been documented [here](https://github.com/gemini-hlsw/ocs3/wiki/Deployment)

A live running seqexec with this method is available [here](http://sbfgpidev1.cl.gemini.edu:9090/), This instance doesn't do much as is not connected to the odb but at least you can login 😄 

The build system now will:

* Support making a local distribution for testing
* Support making a test distrbition in tar ball format
* Generate different logging configuration for each configuration
* Embed and use a specific jre
* Optimize the javascript for production

Note that travis will build the distributions. This may slow down the build quite a bit so we could enable them only for non-PR builds

There are tasks still pending but this is an initial step

- [ ] Build with RPM #66
- [ ] Support configuration #62 (This must include hiding secret data for deployment)
- [ ] Run as daemon #67

This PR also

Fixes #42 Configuring logging for different distributions
Fixes #53 Lets the building of standalone distributions
Fixes #10 Packages the javascript file in optimized format


app/seqexec-server-test-l64/src/templates/bash-template
```diff
@@ -0,0 +1,343 @@
+#!/usr/bin/env bash
```
cquiroz: There is no need to review this script. It is taken from sbt-native-packager and slightly modified to support running on a custom jre. Hopefully sbt-native-packager will take a PR with this change

app/seqexec-server-test-l64/src/templates/bash-template
```diff
@@ -0,0 +1,343 @@
...
+      -d|-debug) debug=1 && shift ;;
+
+    -no-version-check) no_version_check=1 && shift ;;
+
+           -mem) echo "!! WARNING !! -mem option is ignored. Please use -J-Xmx and -J-Xms" && shift 2 ;;
+     -jvm-debug) require_arg port "$1" "$2" && addDebugger $2 && shift 2 ;;
+
+          -main) custom_mainclass="$2" && shift 2 ;;
+
+     -java-home) require_arg path "$1" "$2" && jre=`eval echo $2` && java_cmd="${jre}/bin/java" && shift 2 ;;
```
cquiroz: This line is the only one changed compared to the original launcher

cquiroz:
It would be important to merge this PR before danny adds his new module. Could @swalker2m and @tpolecat review this?

cquiroz:
The PR has been updated after discussion with @swalker2m and the large common build split into several traits for easier handling

Note that this may not be compatible with sbt 1.0 but we'll wait for 1.0 to be released and mature before attempting to port it

cquiroz:
Will merge after discussion with Shane

PR #69, opened 2016-06-22
Binary encoding

### LINKED ISSUE #18: Compare upickle and boopickle for serialization
For large objects (e.g. a sequence) the gain of binary serialization could be important, compare both approaches to check if it is worth it. We should compare having gzip compression on the backends

cquiroz:
This PR uses [boopickle](https://github.com/ochrons/boopickle) to convert the messages between client and server from json to binary. This produces an important reduction on the payload size especially due to the repetition of values in sequences

The method used here doesn't forbid to have also a json representation if needed for non-scala clients though it wasn't implemented

GZip is also enabled to gain and even larger more reduction

Given the possible incompatibilities of binary websockets and gzip compression this was tested on

* Chrome/OSX
* Firefox 38/OSX
* Safari/OSX
* Safari/iOS
* Firefox 38/Linux

Fixes #18


cquiroz:
Rebased onto develop and updated boopickle to the latest version 1.2.4

jdnavarro:
I'd like to test manually that the messages go back and forth through the websocket without issues. Are you using any tool to mock the server side of the websocket? is there something like socat or netcat for websockets?

cquiroz:
I test it against the real server. there is a tool called wscat that can be useful but with binaries it won't be very useful

jdnavarro:
I still haven't tested manually but a quick glance everything looks fine with me. The only thing is, although it may not be relevant for this concrete PR, that it looks like it would be possible to extract some helper functions to DRY a bit the code. Is there a reason not to do so?

cquiroz:
Can you point some examples of repeated code to removed? Sometimes when they are very small things I tend to just leave them repeated

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/services/SeqexecWebClient.scala
```diff
@@ -72,6 +85,7 @@ object SeqexecWebClient {
   def log(record: LogRecord): Future[Unit] =
     Ajax.post(
       url = s"$baseUrl/log",
-      data = default.write(LogMessage.fromLogRecord(record))
+      responseType = "arraybuffer",
+      data = Pickle.intoBytes(LogMessage.fromLogRecord(record))
     ).map(_.responseText)
 }
```
jdnavarro: There seems to be some repeated code for setting up the `url`, the `responseType` and then pickling and unpickling. Perhaps a helper function which takes the `url` and whatever it needs to pickle and unpickle would make the code more concise. However, on a second thought, it may not be worth extracting, so please don't take this comment too seriously :)

cquiroz: I guess that's possible, I think it is very small in this case especially because they are repeated parameters rather than logic and in some cases we don't have data, e.g. on the `stop` command

jdnavarro:
Additionally I've tested it with Chrome 51 and Firefox 47 in Android 6.0.1 and works flawlessly. 👍

cquiroz:
Excellent, I'm happy somebody else has been able to test it, Now let's see how I get travis to comply

PR #70, opened 2016-06-28
Port to diode 1.0.0

cquiroz:
This is a small PR updating the diode library to the version 1.0.0. It can be considered pure UI code

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/NavBar.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/NavBar.scala
@@ -36,7 +36,7 @@ object NavBar {
               IconTerminal.copy(p = IconTerminal.p.copy(link = true, circular = true, onClick = Callback {SeqexecCircuit.dispatch(ToggleDevConsole)}))
             )
           ),
-          userConnect(x => TopMenu(x))
+          userConnect(TopMenu.apply)
         )
       )
     )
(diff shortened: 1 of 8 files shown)


jdnavarro:
Is `ActionType` import necessary now? Everything seems to be based on `Action` now. Am I understanding it correctly?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/NavBar.scala
```diff
@@ -35,7 +36,7 @@ object NavBar {
               IconTerminal.copy(p = IconTerminal.p.copy(link = true, circular = true, onClick = Callback {SeqexecCircuit.dispatch(ToggleDevConsole)}))
             )
           ),
-          SeqexecCircuit.connect(_.user)(TopMenu(_))
+          userConnect(x => TopMenu(x))
```
jdnavarro: Perhaps it's the pointfree style bias coming from Haskell but isn't eta reduction a good practice in these cases? Not sure how the syntax would be, would `(TopMenu(_))` work?

jdnavarro:
I'm not sure I understand how diode uses the `ReactConnector` and what are the advantages of separating the connection in 2 steps. Apparently it should save some unnecessary mounting/unmounting but I can't see how it benefits in this case.

jdnavarro:
Overall my comments were mostly nit-picking. With my limited understanding I don't see anything wrong with this PR.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/NavBar.scala
cquiroz: `(TopMenu(_))` may work, it depends a bit if the compiler can infer the type of x, I'll try

cquiroz:
About `ActionType`. It brings a typeclass into the context to let the calls to `dispatch` accept `Action` making it typesafer

cquiroz:
About `ReactConnector` it is a change on the latest diode, it improves performance by not rebuilding the connector each time the component is re-rendered. Interestingly you can't even compile the old style

For changes check [CHANGELOG](https://github.com/ochrons/diode/blob/master/CHANGES.md)

PR #71, opened 2016-06-29
Reconnecting WebSocket

cquiroz:
This PR brings an implementation of WebSocket reconnections. The application tries to open the connection at startup and if it fails, or the connection breaks at a later point, it will try to reconnect after some delay. This is implemented as a state machine using diode to handle state transitions

The delay is growing doubling every time a reconnect is attempted, up to a max of 1 min.

Note that this PR doesn't contain UI, that will come later

Another missing bit is re-syncing the data. Once we reconnect we need to query the current state of the server to update the UI

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
@@ -192,10 +192,8 @@ class WebSocketHandler[M](modelRW: ModelRW[M, Option[WebSocket]]) extends Action
 
     val host = document.location.host
 
-    def onOpen(e: Event): Unit = {
-      println("Open")
+    def onOpen(e: Event): Unit =
       SeqexecCircuit.dispatch(Connected)
-    }
 
     def onMessage(e: MessageEvent): Unit = {
       \/.fromTryCatchNonFatal(read[SeqexecEvent](e.data.toString)) match {
@@ -204,14 +202,12 @@ class WebSocketHandler[M](modelRW: ModelRW[M, Option[WebSocket]]) extends Action
       }
     }
 
-    def onError(e: ErrorEvent): Unit = {
-      println("Error " + e)
-      //SeqexecCircuit.dispatch(ConnectionError(e.message))
-    }
+    def onError(e: ErrorEvent): Unit =
+      SeqexecCircuit.dispatch(ConnectionError(e.message))
 
     def onClose(e: CloseEvent): Unit = {
       println("Close ")
-      //SeqexecCircuit.dispatch(ConnectionClosed)
+      SeqexecCircuit.dispatch(ConnectionClosed)
     }
 
     val ws = new WebSocket(s"ws://$host/api/seqexec/events")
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
```diff
@@ -180,14 +201,85 @@ class GlobalLogHandler[M](modelRW: ModelRW[M, GlobalLog]) extends ActionHandler(
...
+
+    val ws = new WebSocket(url)
+    ws.onopen = onOpen _
+    ws.onmessage = onMessage _
+    ws.onerror = onError _
+    ws.onclose = onClose _
+    Connecting(ws)
+  }.recover {
+    case e: Throwable => NoAction
+  }
```
jdnavarro: This looks like perfectly respectable callback style code, the typical case to deal with websockets in JS. This is not large enough to become a significant issue, but I was wondering if it'd be too much trouble, or possible at all, to use a more monadic API, like with `Task.async`. It also would set the tone for a more monadic style in future commits minimizing the risk of ending up in callback hell.

jdnavarro:
Would there be any performance issues if we render the entire state of the UI from scratch whenever a resync is needed? If that's acceptable, implementing the synchronization would be much easier.

If that's not the case it may be worth looking into some libraries that take care of optimizations and edge cases when resyncing specific pieces of the UI state.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
cquiroz: Scalaz `Task` is not available in Scala.js though Monix has one https://monix.io/docs/2x/eval/task.html

cquiroz:
A good portion of React is the ability to detect partial changes to the DOM and only re-render that part making it much faster. I'm not too worried about keeping the UI in sync, diode ensure data can only flow one way (hence diode)

Do you mean the data sync after a reconnect?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
cquiroz: In any case this is one of the few places where callbacks are used, and it has all to do with interacting with javascript, in this case the `WebSocket` API. As we see you need to inject listeners to the API to make it work, I don't see a particular benefit of a monadic style in this particular case.

cquiroz:
With respect to data sync on reconnect, that could be potentially complex, I guess we need to link the existence of the WS connection to anything that can change the model. After all if there is no backend you cannot do anything.

Then it would be a matter of replacing the stale model with a fresh one from the backend. The evil will be on the details 😄

jdnavarro:
Data sync on reconnect is definitely a case but a more broad case is whenever the UI depends on some data coming from the server to keep the UI state consistent. React is great dealing with the DOM when everything needed is in the browser itself, but dealing with data dependencies from the server is still left to the programmer. It can get really hairy, specially failures scenarios are considered. I don't know if this a concern, if replacing the stale model with a fresh one works in this case it would simplify things a lot. We can continue discussing off-thread, this is not really related to this PR anymore 👍

cquiroz:
The seqexec ui is relatively dumb, it essentially just display what is sent from the server. How I envision it is that the seqexec-server has the current global state and we just get the state on startup and the same would be done for a reconnect

Issue #40, opened 2016-04-20
Reconnecting websocket


cquiroz:
We need to support websockets that can reconnect. See 
https://github.com/rleibman/scalajs-reconnecting-websocket/blob/master/src/main/scala/net/leibman/reconnecting/ReconnectingWebSocket.scala

PR #72, opened 2016-07-02
Reconnecting WS ui

cquiroz:
This PR completes the WebSockets reconnection task with UI updates. It will display a message on the UI and disable buttons when the connection is lost

Here's a screenshot:

screenshot 2016-07-01 17 46 02

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
@@ -318,6 +318,14 @@ object PotEq {
 }
 
 /**
+  * Utility class to let components more easily switch parts of the UI depending on the context
+  */
+case class ClientStatus(u: Option[UserDetails], w: WebSocketConnection) {
+  def isLogged = u.isDefined
+  def isConnected = w.ws.isReady
+}
+
+/**
   * Contains the model for Diode
   */
 object SeqexecCircuit extends Circuit[SeqexecAppRootModel] with ReactConnector[SeqexecAppRootModel] {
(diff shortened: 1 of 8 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
```diff
@@ -259,19 +261,21 @@ class WebSocketHandler[M](modelRW: ModelRW[M, Option[WebSocket]]) extends Action
...
     case ConnectionClosed(d) =>
-      logger.fine("Retry connecting in "+ d)
-      val effect = Effect(Future(WSConnect(d)))
-      updated(None, effect)
+      val next = math.min(60000, math.max(250, value.nextAttempt * 2))
+      logger.fine("Retry connecting in "+ next)
+      val effect = Effect(Future(WSConnect(next)))
+      updated(WebSocketConnection(Pending(), next), effect)
   }
 }
```
jdnavarro: I'm not sure how browsers handle nowadays the sudden lost of a TCP connection. If I unplug the network cable I still need to refresh to get the connection retry (OSX/Firefox). If there were messages unable go through the websocket from the webclient, do you know if the browser closes the connection by itself? Not sure how to test it.

cquiroz: At application level you have essentially no control on how the browser handles the connection. From what I've seen the connection won't drop right away but it will take a while until the TCP stack realizes that the other endpoint is not responding.

In this app the requests to the browser are not really complete until we get a confirmation over the WebSocket channel so I don't think it is a problem

jdnavarro:
Looks fine to me 👍

cquiroz:
This PR fixes #40

PR #74, opened 2016-07-04
Additional gitignores

jdnavarro:
This commit is a good chance to include more ignores. Let me know if you want to list more.


cquiroz:
Looks good to me :+1:

PR #75, opened 2016-07-04
Test improvements

cquiroz:
This PR fixes some compilation errors on the tests. These tests are disabled at the build level as there is a **yet-unsolved** conflict between the available runtime environments (Rhino/node.js/phantom) and jquery

At least this PR will not let check in with non-compilable tests

Additionally this PR advances scala.js version to 0.6.10


jdnavarro:
It works on my machine now :+1:

PR #76, opened 2016-07-11
Add scalastyle configuration file

jdnavarro:
This was taken from the `cats` project whose style seems to align better with
the current one than the default configuration from scalastyle.

This could also be used to enforce proper styling by having travis fail if scalastyle throws warnings. However, speaking with @cquiroz and @tpolecat, we think it may not be a good idea to do so until we get used to it first.

From scalastyle website:

> You can enable scalastyle in Intellij by selecting *Settings->Editor->Inspections*, then searching for Scala style inspections.

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -1,0 +1,94 @@
+<scalastyle>
+ <name>Scalastyle standard configuration</name>
+ <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"></check>
+ <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
+  <parameters>
+   <parameter name="maxFileLength"><![CDATA[800]]></parameter>
+  </parameters>
+ </check>
+ <check level="error" class="org.scalastyle.file.WhitespaceEndOfLineChecker" enabled="true"></check>
+ <check level="error" class="org.scalastyle.scalariform.ClassNamesChecker" enabled="true">
+  <parameters>
+   <parameter name="regex"><![CDATA[[A-Z][A-Za-z]*]]></parameter>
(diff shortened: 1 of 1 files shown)


cquiroz:
Is this exactly as the cats style? Is there human-readable documentation on the style guidelines?
I'd be ok on this, since it is not enforced nobody will go mad about this 😄 

Does it work for you on ensime?

tpolecat:
You can run it as part of the CI build and make it mandatory, which probably makes sense. The things I want to check for end up being correctness issues so it's basically a lightweight linter.

jdnavarro:
@cquiroz I'm using it with ensime. I haven't inspected the xml file to check the exact differences between the default configuration file and the one from cats, but I'm noticing the default one complains about several issues which aren't an issue with the cats configuration, like package names and implicit imports. OTOH, cats style seems more picky on spaces for vertical alignment, which seems something acceptable in ocs3 code.

cquiroz:
@tpolecat Do you want to add something like `WartRemover`?

tpolecat:
@cquiroz yes, it would be worth looking at WartRemover too. I would like to get as much FP-oriented linting as possible. The problem is that you sometimes have to break the rules when you're interacting with Java APIs, so I often have to turn these things off. But it's worth a try.

tpolecat:
You can also get a lot of useful stuff from [compiler flags](http://tpolecat.github.io/2014/04/11/scalac-flags.html) … you may have already done this.

jdnavarro:
The style as it is in this in this PR is not that useful yet. I'm using a global scalastyle configuration that I still think needs more tweaking. Once I feel it's stable enough I'll submit a new PR.

PR #78, opened 2016-07-12
AD Server name changes

cquiroz:
The AD server names were changing rendering the logging functionality of ocs3 inoperable
This PR fixes the server names and adds support for failover. The server will attempt first cpo and if it fails it will attempt sbf before giving up


jdnavarro:
I logged in with my login credentials and the dummy user and both work properly. I'm not sure about which LDAP server is being used when I'm logging in or how to test the login with a user in a different LDAP server, but as far as I can test this works  👍

cquiroz:
As per @jdnavarro suggestion this was tested on failure mode and a bug found and fixed

PR #79, opened 2016-07-21
Remove play-based backend

cquiroz:
After some time developing ocs3 with 2 backends, `play` and `http4s` it is a good time to re-evaluate. It has been possible to develop both backends side-by-side (with some effort) demonstrating that both are capable of full-filling our needs, however it would be good to maintain only one.

**http4s pros**: Very much compatible with the rest of the ocs3 backend, native support for `Task` and `scalaz-streams`
**http4s cons**: Non-existing documentation, some features half-baked like multi-part, security cannot be assessed given how little it is being used

**play pros**: Well documented and support for many types of technologies, stronger security record
**play cons**: There is an impedance mismatch when using `scalaz-stream` and `Task` based APIs. Even though play supports  many front-end oriented technologies, we are not really using them

Authentication via JWT isn't supported by either backend so adapters were created for both backends.

In summary, `http4s` seems less mature and lacks documentation compared to `play`, however, given that our code base is biased to use `typelevel` and `scalaz` constructs, and we can make a seqexec running on `http4s`, it seems that `http4s` is a better fit

This PR removes the play backend and dependencies based on the previous discussion


swalker2m:
Agree. 👍

PR #80, opened 2016-07-25
Application configuration

### LINKED ISSUE #62: Define a mechanism to configure applications

cquiroz:
This PR uses the suggested [Knobs](http://oncue.github.io/knobs/) library to configure the seqexec. The configuration for the web server/authentication and odb to use are now in configuration files that are included on deployment

This PR is fairly involved with many changes to the launcher and to objects whose configuration was hardcoded. Probably the most interesting changes are on the `WebServerLauncher` class using `Kleisli` to configure and execute the application start

Note that with this change it is possible to make a test release pointing to a test odb

The PR updates http4s to 0.14 to use the improved launcher classes on the new release

Fixes #62


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/resources/app.conf
```diff
@@ -0,0 +1,34 @@
...
+mode = "dev"
+
+# Authentication related settings
+authentication {
+    # Indicates how long a session is valid in hrs
+    sessionLifeHrs = 8
+    # Name of the cookie to store the session
+    cookieName = "SeqexecToken"
+    # Secret key for JWT tokens
+    secretKey = "Ofm@uYN5zcRL0qwNlMR75%m70HLG@Rz9INv1Bhv"
```
jdnavarro: I guess it's too early to worry about the real secret keys, but does `Knobs` provide an easy way to store secret data? I'm just curious, nothing wrong with dummy secret keys in source code.

cquiroz: That is a very good concern. In this case we don't care since it is a development/test instance, I have already code to get the "secret" configuration for **production** stored in svn. I didn't want to make this PR more complex than it is

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -1,38 +1,111 @@
...
+  val executorConf: Task[SeqexecConfiguration] =
+    config.map { cfg =>
+      val host = cfg.require[String]("seqexec-engine.odb")
+      SeqexecConfiguration(host)
     }
+
+  /**
+    * Configures the Seqexec executor
+    */
+  def seqexecExecutor: Kleisli[Task, SeqexecConfiguration, Unit] = Kleisli { conf =>
```
jdnavarro: In Scalaz I've seen Kleislis are used `ReaderT` monad. Is this the reason to wrap the executor in a Kleisli with its configuration (as opposed to just pass the configuration as a function argument) or is it something related to how `http4s` works?

cquiroz: It is unrelated to `http4s`. This is my first attempt at using `Kliesli` and I'm following `knobs` recommendations:

> Where possible, we recommend designing your applications as a free monad or use a reader monad transformer like Kleisli[Task,Config,A] to "inject" your configuration to where it's needed



On the `server` function I read the configuration for each "subsystem" and run them in the main for loop

jdnavarro:
Maybe this concern is not really about this PR, but I noticed that the logging side-effect is unsafe now. I'm not saying it's wrong but since logging is such a common use case for webapps is this normal `http4s` apps? Or in other words, doesn't `http4s` provide some guidelines about how to do logging safely?

cquiroz:
There is no guidelines about `logging`. In general we have been doing it this way, just drop a log call wherever we need it without worrying about making it pure.

Perhaps @swalker2m and @tpolecat can comment on that

I personally don't think it is worth to complicate the code for that but it is worth to discuss it

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -1,38 +1,111 @@
...
+    * Configuration for the seqexec engine
+    */
+  case class SeqexecConfiguration(odbHost: String)
+
+  // Attempt to get the file or throw an exception if not possible
+  val configurationFile: File = new File(new File(baseDir, "conf"), "app.conf")
+
+  // Read the config, first attempt the file or default to the classpath file
+  val config: Task[Config] = knobs.loadImmutable(
+    Required(FileResource(configurationFile) or ClassPathResource("app.conf")) :: Nil)
```
jdnavarro: If I'm not mistaken, the configuration will read just once when starting the server, right? Not a problem for this PR, but will there be a need in the future to reload the configuration without restarting the server?

cquiroz: Knobs has support for checking config file changes and support reloading, but I didn't want to go there. I think it is a valid point, I can imagine cases where you want to reload the config. On the other hand it worries me that somebody will edit the configuration and do a restart it right away by mistake

In general we have tended to have configs set in stone, even inside the jar files to avoid accidental editing but I can see how it could be useful in some cases.

tpolecat:
If you're working in `Task` or `IO` then logging usually needs to be as well, otherwise you are logging the construction of the computation rather than its execution. This is a subtle point. I think most of our logging is kind of useless so we do need to figure out some guidelines but we don't need to do so in this PR.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/common/LogInitialization.scala
```diff
@@ -8,7 +8,26 @@ import scalaz._
 import Scalaz._
 import scalaz.effect.IO
 
-trait LogInitialization {
+trait AppBaseDir {
+  /**
+    * Calculates the base dir of the application based on the location of "this" class jar file
+    * It will throw an exception if unable to find the base dir
+    */
+  def baseDir: File = {
```
tpolecat: Why not `IO[File]` here?

jdnavarro:
@cquiroz, the worst scenario I see for unsafe logging, mainly when laziness is involved, is to have log messages come when the thing you are logging is not really happening at that moment. I'm not completely sure, but there may be also alteration in the order of logging messages, but that may be a extreme edge case.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
tpolecat: If you're not composing kleisli arrows then there's no reason to use the wrapper, but if big hunks of your application are parameterized by `SeqexecConfiguration` then it's a more convenient type to use. Seems reasonable to me.

jdnavarro:
@tpolecat, wrapping in `IO` or `Task` would make logging safe, but do we want to shove every effect we come across into `IO` or `Task`? I agree that's a discussion for another time though.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/authentication.scala
```diff
@@ -72,19 +106,10 @@ object AuthenticationService {
...
-    * From the user details it creates a JSON Web Token
-    */
-  def buildToken(u: UserDetails): String =
-    // Given that only this server will need the key we can just use HMAC. 512-bit is the max key size allowed
-    Jwt.encode(JwtClaim(write(u)).issuedNow.expiresIn(3600), AuthenticationConfig.key, JwtAlgorithm.HmacSHA256)
+  // Launcher for the authentication service taking a configuration
+  val authServices: Kleisli[Task, AuthenticationConfig, AuthenticationService] = Kleisli(c =>
+    if (c.devMode) Task.now(AuthenticationService(c))
+    else Task.now(AuthenticationService(c))
+  )
```
tpolecat: What is the intent here? The check for `c.devMode` doesn't do anything.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/authentication.scala
```diff
@@ -17,45 +19,77 @@ case class GenericFailure(msg: String) extends AuthenticationFailure
...
+    */
+  def decodeToken(t: String): AuthResult =
+    (for {
+      claim <- Jwt.decode(t, config.secretKey, Seq(JwtAlgorithm.HmacSHA256)).toDisjunction
+      token <- \/.fromTryCatchNonFatal(read[JwtUserClaim](claim))
+    } yield token.toUserDetails).leftMap(m => DecodingFailure(m.getMessage))
+
+  val sessionTimeout: Long = config.sessionLifeHrs * 3600
+
+  override def authenticateUser(username: String, password: String): AuthResult = authServices.authenticateUser(username, password)
```
tpolecat: Small thing, but please put a newline after the `=` here.

tpolecat:
@jdnavarro if we're using `IO` or `Task` (and really it should be one or the other in a given program) it's important to be consistent about capturing effects otherwise it makes things even harder to reason about.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/authentication.scala
cquiroz: You're right. `devMode` is used inside the constructor of `AuthenticationService`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/common/LogInitialization.scala
cquiroz: This is called from inside `IO` but I guess you want to make it explicit that is run inside `IO`?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
tpolecat: It has been very useful in the past to be able to change logging levels at runtime, so this kind of facility is nice to have.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/common/LogInitialization.scala
tpolecat: It's public, so yes. `IO` actions should be small so we can compose them. Putting a big block of side-effecty stuff inside of `IO { ... }` doesn't buy us much.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
jdnavarro: If it's not absolutely necessary I'd also rather not support hot reloading of configuration. Would restarting the server manually after a configuration change be safe enough to prevent accidental restarts?

If hot reloading is not a priority then, I'm trying to understand what do we gain from `Knobs`. Wouldn't parsing a JSON file with the configuration when starting the server be good enough?

cquiroz: `Knobs` won`t help on that, as `java.util.logging` has its own style of configuration. While other logging frameworks allow changing config at runtime `j.u.l.i` has no provisions for that.

I guess it is possible to use knobs file watching utility with a custom code that changes logging config. Possibly in a separate PR

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
jdnavarro: Using a `Kleisli` as a ReaderT monad so that the seqexec can access the configuration at any time is perfectly reasonable. The problem is that the seqexec will have its own StateT monad to control the state of execution and I already experienced first hand how painful combining monad transformers of multiple layers is in Scala (the Kleisli and the StateT would have to be stacked).

In any case I'm working now on replacing the StateT monad with a Free Applicative, so we don't need to be concerned with this `Kleisli` now for this PR.

But my point is that we need to at some point to discuss what do we do with multiple effects, do we ditch monad transformers entirely? Do we allow transformers of just one layer only? If we allow them then, which layer takes precedence in this particular case (Kleisli or StateT)?

Sorry for hijacking the thread. We can continue discussing this elsewhere.

cquiroz: Pelase note that this code works with the current `ExecutorImpl` which is not using any type of `Monad` wrapper. I'd rather wait to see the final form of the future executor before worrying about the integration

cquiroz:
@jdnavarro `Knobs` provides more than just reading the file (Which is not json but more like HOCON), it can for example substitute environment variables, or reference internal values. I guess a worry about `Knobs` is adding yet another axis of dependencies on the `scalaz-*` family

cquiroz:
@tpolecat The Log initialization was done on a previous PR using `IO` while the app configuration uses `Task` as it fits both `knobs` and `http4s`.

I explicitly do them separately since I want logging to be configured first than anything, as we do on the ocs. In the ocs case, the situation is different though, since we don't have control on the order OSGi bundles are initialized.

Do you think it makes sense to do logging config in `Task` to be consistent?

tpolecat:
It probably makes sense to settle on either `Task` or `IO` within a given app. If you need async support use `Task`, otherwise `IO`.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/common/LogInitialization.scala
```diff
@@ -6,73 +6,77 @@ import java.util.logging.LogManager
...
+trait AppBaseDir {
+  /**
+    * Calculates the base dir of the application based on the location of "this" class jar file
+    * It will throw an exception if unable to find the base dir
+    */
+  def baseDir: Task[File] = Task.delay {
+    val clazz = this.getClass
+    val fileName = clazz.getResource(s"/${clazz.getName.replace(".", System.getProperty("file.separator"))}.class").getFile
 
-  type LogConf[A] = EitherT[IO, LogConfError, A]
```
jdnavarro: From a FP stand point it may be easy to disagree with this change, but for the record, I just wanted to say that I think getting getting rid of this type for this case is a great idea. :+1:

jdnavarro:
I can't find anything to point to now 😄 . Fine with me 👍

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/common/LogInitialization.scala
cquiroz: What's the FP point of view to disagree?

app/seqexec-server-test-l64/src/main/resources/app.conf
```diff
@@ -0,0 +1,18 @@
+#
+# Seqexec server configuration for the testing server
+#
+
+# mode for the test server is production but we point to the test odb
+mode = "production"
```
cquiroz: Now the application conf only needs what changes from the default configuration

cquiroz:
The PR was updated with the following changes:

* Use `Task` instead of `IO` in all places
* Simplify error management by raising an exception if, e.g., the logging configuration is missing
* Support config overriding, changing the way knobs reads the config it is possible to only supply at the app level the differences between the app config and the default config
* Separate the initialization of `AuthenticationServices` from the Auth logic itself

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/common/LogInitialization.scala
jdnavarro: It's very nice that in FP you can deal with errors statically, but the thing here is, that since you already have to handle exceptions raised from `IO`, `Task`s in this case, you don't gain much by having 2 ways of dealing with errors. Dealing statically with errors in the `EitherT` gives you the false sense of security that you are already dealing with exceptions but it's not true, since the IO errors can be raised at any moment and you still have to deal with them.

But this is a controversial topic that can generate many heated debates. I'm of the camp that `EitherT Error IO a` is usually a smell of:
 
1. You are dealing with genuine static errors where you shouldn't. The code dealing with `Either Error a` needs to be factored out from the `IO` code.

2. You are shoehorning exceptions that are meant to be handled in the traditional out-of-band way with `try`/`catch` blocks, like most IO errors.

Maybe I'm guilty of extrapolating my knowledge of Tasks too literally from Haskell `IO`, so I'm still open to change my opinion on this matter if it turns out my assumptions are wrong. For example, is it possible to properly handle asynchronous exceptions (like an interrupt anywhere in the code) statically within the `Task`s API? Can `Task`s guarantee there won't be any raised exceptions *leaking* through?

tpolecat: Exceptions in `Task` are analogous to those in `IO`. You can make exceptions observable via `.attempt` which takes `Task[A]` to `Task[Throwable \/ A]` or via combinators based on that operation. If the task is run via an `.unsafeXXX` method that doesn't trap exceptions then any such exception will be thrown on the calling thread. Exceptions on other thread pools are routed back to the caller. There is no notion of "detached" asynchronous execution; every executing task has a logical "caller".

jdnavarro: @tpolecat so you mean to say that `Task`s are always structured as a tree of threads, and when a `Task` throws an exception it can only be handled at the top level `Task`, when doing `unsafeXXX`, even for *async* exceptions (like a user interruption). Not being able to run *detached* threads rules out some concurrency patterns, but it also rules out many sources of complexity, specially when dealing with exceptions.

At a first glance, it seems that `.attempt` catches all exceptions indiscriminately which looks a bit [dangerous](https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Exception.html#g:4). But if it's only used at the top level and you make sure any unhandled exceptions are re-raised then I think it's quite reasonable to use it.

But I definitely have to read more about `Task`s because it's quite different to what I'm used to.

tpolecat: `Task` is just like `IO` but the `bind` effect allows the computation to be suspended and rescheduled on another thread, if the task has affinity for a thread pool (this is determined on construction). This is not directly observable, but it allows parallel execution of the applicative-like and traverse-like operations defined on `Nondeterminism`.

`.attempt` allows exceptions to be observed (and, as you note, ignored if you're not careful) at any point in the computation (not just at the top level), just as you can do with Haskell's `IO`. But it is possible to build up fine-grained combinators based on this primitive operation, as [here](https://github.com/tpolecat/doobie/blob/series/0.3.x/yax/core/src/main/scala/doobie/util/catchable.scala#L49-L68).

cquiroz:
I think logging is very valuable. If we refer to things like AGS that are very verbose I agree that most of it is useless but in the case of Seqexec we need a trail of actions for fault reporting.

This is more often than not the only way to find whether a fault happened at the seqexec level, instrument level, etc

One of the plus sides of having authentication is that we can have a record of not only when but who did it again making fault resolving simpler

jdnavarro:
@cquiroz Of course, I understand the need to log as much as possible. My only objection to logging using `Task`s is that we'll have to use `Task`s everywhere throughout the code whenever we need to log something. This would make the code less *pure* than it could be. Ideally, if we keep the `Task` layer of the code as thin as possible, we'd have an easier time dealing with simpler side effects in the core logic of the application. After all, within a `Task` any kind of side effect in the real world is possible, like launching missiles or moving a huge telescope 😄.

For functions that only need to just log what's being done (which I guess there'll be plenty of those), the logging effect could be managed with just a writer monad or a DSL with free monads or some other mechanism to manage the logging effect without bringing in all kind of side effects, including those from the real world.

In any case I'm pointing this potential issue in order to start the conversation about managing side effects. I agree that at this stage is irrelevant, but I do think it may be significant later when the codebase becomes much larger.

cquiroz:
@jdnavarro I don't expect to create a `Task` just to log but would certainly add logging inside the Task, let's say right before configuring an instrument. In my opinion add another layer of monads for that case is quite overkill

jdnavarro:
@cquiroz if logging is only needed when there is a side effect in the real world, using `Task` for logging is fine by me. But are you sure the need to log won't happen in places where nothing in the real world needs to happen?

tpolecat:
Pure `Writer` is not a good solution for logging in the presence of exceptions because you lose your state when you unwind the stack. I am fine if we consider logging unobservable, but we need to be clear about what we're logging. Normally we want to log the execution of a task, not its construction. And if you're not careful about things you can get these confused.

We should also think about something like [funnel](https://github.com/oncue/funnel) for metrics, which in some cases is what we're trying to achieve with noisy logging. I'm going to include it in the prototype back end I'm working on now so we'll see how it looks.

cquiroz:
Certainly logging should be inside execution, it is quite easy to make the mistake of doing it in construction, unfortunately

The main scenario I'm concerned, is that we often need to correlate what happened in the seqexec and link it to the errors on the instruments/telescope/etc. Most instruments are very noisy in terms of logging and the only way to make sense of it is by relating events on one side and the other. Often times, the most important item is the time of logging rather than the message itself

I'm quite fine with having metrics but one way or another we should support the above scenario in operations

tpolecat:
@cquiroz have we thought about logging to GEA? I have no idea how that works. If it's important to see how this stuff is interleaved maybe we could be more direct about it.

swalker2m:
I think GEA is only for EPICS-based systems.

cquiroz:
GEA is not really for that. In an ideal world all the systems would log into a central site and you could do some reasonable correlation across subsystems. You could see somebody editing the ODB, executing on Seqexec and how the instruments work

It would be quite a task to retrofit systems to do that. In the case I know the best GPI (a single instrument) has at least 3 different ways of logging from 5 machines 😢 

When a GPI fault happens 90% of the work is log navigation

tpolecat:
Well we can certainly talk to EPICS … I'm just trying to push on this to figure out what we're really trying to accomplish. A firehose of non-machine-readable text is a very primitive mechanism.

tpolecat:
I want to have persistent structured logging (i.e., database table(s)) in the new ODB so maybe we can improve the situation a bit. Anyway I didn't intend to hijack this issue. But I'd like to keep the question of logging open for now.

cquiroz:
It is very primitive and time consuming indeed, and at least at the high level we should be able to do better.

The issue comes with the legacy real-time systems. If I remember the grand plan William had for real-time included centralized logging and monitoring but I'm not sure what happened to that. And again it would be quite an effort to retrofit it to old systems

jdnavarro:
Alright now I got that you are referring to the logging of the instruments, not the logging generated in the application itself. For that I see why a pure `Writer` won't work.

cquiroz:
It is application level logging. What we need is to know, e.g., user started sequence Y at time X and record in each step when you the system to configure and the result, especially failures.

As mentioned we need to log the time actions happen, not when they are constructed

jdnavarro:
@cquiroz Maybe I'm again translating too literally the semantics of Scalaz `Writer` and `Task`  to their equivalents in Haskell, but I fail to see why is there a problem between construction vs execution when using `Writer`. If `Writer` is properly implemented why it's not just doing the right thing? Can you point to me to any example or blog post or some documentation where this gotchas with `Writer` are explained?

cquiroz:
I guess my points can be summarized as:

* Logging is really important in many cases and we have to record it at certain key locations. It is especially important in the seqexec that communicates with hardware distributed in many nodes
* I think just having a call log.info("xxx") in the code is not a big deal. Doesn't seem to me to warrant an extra layer of complexity

Regardless of the discussion, do you think it would be ok to merge this PR?

cquiroz:
I'm going to merge this task to move on. I guess is worthwhile at some moment to review the options for metrics and the usability of the selected configuration system

PR #81, opened 2016-08-04
Track sequence running step

### LINKED ISSUE #50: Focus on the running step
When a sequence is running the table should scroll to the currently running element

cquiroz:
This PR changes the Ui to always have the running step visible as required on the seqexec specs document. Basically if you have more steps than visible on the steps table, the ui should autoscroll to make the running step always visible

Changing the scroll position on a page can only be done on plain javascript and that collides with React way of handling the DOM. This was solved by creating  a React `Ref` to the scroll pane and hook into React lifecycle events to calculate the new scroll position, store it in the component's state and then make it effective

There are still some open issues with this like

* If the user scrolls manually, should the position be reset?
* Pausing is not supported in this schema
* For fully executed sequences, should we focus on the first or the last row?

I'll expect user feedback to continue polishing this feature

Fixes #50

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
@@ -12,6 +12,7 @@ import edu.gemini.seqexec.web.common.{SeqexecQueue, Sequence, SequenceState}
 import japgolly.scalajs.react.vdom.prefix_<^._
 import japgolly.scalajs.react._
 
+import scala.scalajs.js
 import scalacss.ScalaCssReact._
 import scalaz.syntax.show._
 
@@ -141,7 +142,7 @@ object QueueTableLoading {
         ^.cls := "ui header item",
         p.queue.renderPending(_ => <.span(IconCircleNotched.copyIcon(loading = true), "Loading..."))
       )
-    ).build
+    ).build.withKey("key.queue.loading")
 
   def apply(p: ModelProxy[Pot[SeqexecQueue]]) = component(Props(p()))
 }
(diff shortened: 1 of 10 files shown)


jdnavarro:
@cquiroz is there any way to load a dummy sequence to test this manually?

cquiroz:
Not really, you have to connect it to some odb. But now that you mention it perhaps we should include that as a test feature. Having some test sequences available to load under some names could be useful

jdnavarro:
Without being able to watch the functionality it's a bit hard to suggest UI changes or improvements, so I'll look into creating some mock sequences.

Regarding the code of this PR, it looks alright to me but I must acknowledge, that since I'm still not that familiar with `diode` or using `react` with  scala.js, you shouldn't take my review too seriously :wink:.

👍

cquiroz:
Here's a video of the process, the auto scroll happens at step 9 where it goes out of view

https://dl.dropboxusercontent.com/u/1834070/SeqexecScrolling.mov

Note that this has been tested in Chrome OSX/Firefox 38 OSX/Safari OSX/Firefox Linux/iOS

cquiroz:
Reading the code gives you very little about what it does, I guess that's the nature of  UI development. In any case, there are no `diode` related changes on the code, all is `react` and plain old Javascript/CSS/DOM dark magic

jdnavarro:
Very nice video. I don't know what a real user would expect, but would make it sense keeping the *current* running sequence always focus on the first row? That would make it easy to match the expected scrolling position when loading a sequence URL from scratch.

Also, if the running sequence fixed were fixed to the first row always, the perception that is advancing may be lost, so I don't know how much trouble would be to insert a subtle scroll animation in order to fix it. Does semantic-ui include some kind of CSS transition out of the box for such case?

cquiroz:
You're right, it is difficult to predic what the user may want, for example if the user scrolls manually should you anyway reset it. I tried putting the runnig row first but you loose sense that things are progressing, but that may be just my taste

`jquery` offers some animation but I'm trying to avoid `jquery` if possible. Semantic does animation with pure css too but in general for this case I think animation is not needed.

I'm happy with the current state and would go back to update after user testing (And admittedly don't want to get more into DOM manipulation 😄  for a while)

andrewwstephens:
I think this looks really good!  Ideally the executing step should be near the middle of the visible table so that the user can see the resulting file numbers as well as get a sense of how many steps remain.

Issue #57, opened 2016-05-12
Support assets fingerprinting


cquiroz:
We should add a unique id to static assets and then we could cache them forever on the client

PR #83, opened 2016-08-16
Improved caching 

### LINKED ISSUE #61: Generate html during the build
This would let us remove duplicated html files and support the work for #57 
It likely requires an sbt task

cquiroz:
This PR adds caching headers to static assets in order to improve performance. With this code the static resources should be cached on the browser for a year. This is enabled only in production.

Assets' name contain the build timestamp, this way it is possible to easily do upgrades and have a strong cache policy at the same time. To support this we need to generate the `index` file "dynamically" (In practice is only generated once when the server start)

An additional improvement on this PR is serving fonts used by SemanticUI locally instead of remotely from google, supporting disconnected operation

Fixes #61 and #57


jdnavarro:
Will the http4s web server be directly exposed without a reverse HTTP proxy in front? I'm used to dealing with static files and cache busting mostly through NGINX configuration which makes it very easy. I was wondering whether maintaining the configuration for static assets may be easier with NGINX than with http4s. But if there are no plans to introduce a reverse proxy in the first place I don't think it's worth considering introducing it just for that option though.

cquiroz:
There are no plans for that. nginx would certainly be faster but it makes he setup more complex as you need to copy the assets, probably during deploy. At least I don't have plans for that on the seqexec but perhaps we do that in the future if we expose it to the outside

jdnavarro:
Looks fine to me 👍

PR #84, opened 2016-09-07
CLI authentication

### LINKED ISSUE #63: Support authentication on the terminal client

cquiroz:
This PR adds authentication when using the web command line terminal. A side effect benefit is that we can now protect all the command routes, like `run`, `pause`, `stop` from being executed without proper authentication

The credentials are the same as on the regular web ui

screenshot 2016-09-07 13 29 31

Fixes #63


jdnavarro:
This is really cool 👍 . Do you expect to add extra information (perhaps more verbose logging) when using the CLI?

cquiroz:
The terminal client is very simple you can send commands and it will display the response, which is normally not very verbose. If you are thinking about displaying how a sequence is being executed that would be harder as there is no websocket channel is this application

The idea of the CLI is to replace, to an extent, the felix console. The felix console was easily more verbose as it would run on the server and show whatever went into `println`. This web-based console doesn't do that

I guess we could setup a listener to `System.out` and `loggers` and send them to the client if it is useful

PR #85, opened 2016-09-21
Seqexec execution with a Zipper

jdnavarro:
With the *Zipper* there is now a deeper data hierarchy with a top level *Queue*, *Sequence*s, *Step*s, *Execution*s and *Action*s. Previously there were only the equivalent of *Execution*s and *Action*s. Also with this structure it comes support for marking `Execution`s as completed instead of remove them.

The executions for each test can be seen printed to the console. The tests themselves only check that the executions finish.

Don't bother too much reviewing this PR since the changes are quite messy. It can be considered as a starting point for the Zipper based execution.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
@@ -10,7 +10,10 @@ import Scalaz._
 case class QState(pending: Queue[Action],
                   current: Current,
                   done: Queue[Result],
-                  status: Status)
+                  status: Status) {
+
+  def isEmpty: Boolean = this.pending.sequences.isEmpty && this.current.isEmpty
+}
 
 object QState {
 
@@ -38,10 +41,13 @@ object QState {
     * Given an index of a current `Action` it replaces such `Action` with the
     * `Result` and returns the new modified `State`.
     *
+    * If after marking the `Action`, all elements in `Current` are `Result`, it
+    * empties `Current` and moves the `Result` to the completed `Queue`
+    *
     * If the index doesn't exist, the new `State` is returned unmodified.
     */
-  def mark(i: Int)(r: Result)(st: QState): QState =
-    current.mod(Current.mark(i)(r)(_), st)
+  def mark(i: Int)(r: Result)(qs: QState): QState =
+    current.mod(Current.mark(i)(r)(_), qs)
 
   /**
     * Returns a new `State` where the next pending `Execution` has been promoted
@@ -52,26 +58,37 @@ object QState {
     * If the `Current` doesn't have all actions completed or there are no more
     * pending `Execution`s it returns None.
     */
-  def next(st: QState): Option[(Execution[Action], QState)] = for {
-    exe3done <- uncurrentify(st.current)
-    qd = Queue.cons(exe3done)(st.done)
-    (exe3pending, qp) <- Queue.uncons(st.pending)
-    (actions, c) = currentify(exe3pending)
-  } yield (actions, QState(qp, c, qd, st.status))
+  def next(qs: QState): Option[QState] = cleanup(qs) >>= (prime(_))
 
+  // None: current not empty
+  //       pending queue empty
+  def prime(qs: QState): Option[QState] =
+    if (qs.current.isEmpty)
+      Queue.uncons(qs.pending).map(
+        t => {
+         val (exe3, q) = t
+         QState(q, currentify(exe3), qs.done, qs.status)
+        }
+      )
+    else None
+
+  def cleanup(qs: QState): Option[QState] =
+    uncurrentify(qs.current).map(exe3 =>
+      QState(qs.pending, Current.empty, Queue.cons(exe3)(qs.done), qs.status)
+    )
   /**
     * Transform an *unconsed* pending `Execution` into `Current` in addition to
     * returning the unwrapped `Execution`.
     */
   // TODO: Use same structure for `Current` and `Queue.Execution3`?
-  private def currentify(exe3: Queue.Execution3[Action]): (Execution[Action], Current) = {
+  private def currentify(exe3: Queue.Execution3[Action]): (Current) = {
 
     def vec(exe: Execution[Action]): Vector[Action \/ Result] = exe.map(_.left).toVector
 
     exe3 match {
-      case -\/(-\/((actions, seqid, stepid))) => (actions, Current(vec(actions), Some((seqid, stepid).left)))
-      case -\/(\/-((actions, stepid))) => (actions, Current(vec(actions), Some(stepid.right)))
-      case \/-(actions) => (actions, Current(vec(actions), None))
+      case -\/(-\/((actions, seqid, stepid))) => Current(vec(actions), Some((seqid, stepid).left))
+      case -\/(\/-((actions, stepid))) => Current(vec(actions), Some(stepid.right))
+      case \/-(actions) => Current(vec(actions), None)
     }
   }
 
@@ -82,14 +99,11 @@ object QState {
   // TODO: Use same structure for `Current` and `Queue.Execution3`?
   private def uncurrentify(current: Current): Option[Queue.Execution3[Result]] = {
 
-    // not available in scalaz?
-    def rights[L, R](xs: List[L \/ R]): List[R] = for { \/-(r) <- xs } yield r
-
-    def unvec(v: Vector[Action \/ Result]): Option[Execution[Result]] =
-      if (v.all(_.isRight)) rights(v.toList).toNel
+    val unvec: Option[Execution[Result]] =
+      if (current.ars.all(_.isRight)) current.results.toNel
       else None
 
-    unvec(current.actions).map(
+    unvec.map(
       exe => current.ctxt match {
         case None => exe.right
         case Some(-\/((seqid, stepid))) => (exe, seqid, stepid).left.left
@@ -117,12 +131,12 @@ case class Queue[A](sequences: List[Sequence[A]]) {
   def isEmpty: Boolean = this.sequences.isEmpty
 }
 
-
 object Queue {
 
   type Execution3[A] = (Execution[A], String, Int) \/ (Execution[A], Int) \/ Execution[A]
 
   def empty[A]: Queue[A] = Queue(List())
+
   def sequences[A]: Queue[A] @> List[Sequence[A]] =
     Lens.lensu((q, ss) => q.copy(sequences = ss), _.sequences)
 
@@ -176,7 +190,7 @@ object Queue {
     *  - x \/ E \/ x: When current `Step` has been completed.
     *  - E \/ x \/ x: When current `Sequence` has been completed.
     *
-    * `uncons`ing on an empty `Queue` returns an empty `Queue`.
+    * `uncons`ing on an empty `Queue` returns `None`.
     */
   def uncons[A](q: Queue[A]): Option[(Execution3[A], Queue[A])] =
     // Queue empty?
@@ -187,7 +201,8 @@ object Queue {
           mseq match {
             // No more Steps in current Sequence, remove Sequence.
             // TODO: listTailPLens?
-            case None => ((exe, seq0.id, stepid).left.left, Queue(q.sequences.tailOption.getOrElse(List())))
+            case None => ((exe, seq0.id, stepid).left.left,
+                          Queue(q.sequences.tailOption.getOrElse(List())))
             // More Steps left in current Sequence, remove `Step` from Sequence.
             case Some(seq) => ((exe, stepid).right.left, head.set(q, seq).getOrElse(q))
           }
@@ -289,10 +304,28 @@ object Step {
   * the `Queue` for proper insertion into the completed `Queue` when all the
   * `Execution`s are done.
   */
-case class Current(actions: Vector[Action \/ Result],
+case class Current(ars: Vector[Action \/ Result],
                    // TODO: The following tuples should be replaced by either
                    // Sequence or Step parameters
-                   ctxt: Option[(String, Int) \/ Int])
+                   ctxt: Option[(String, Int) \/ Int]) {
+
+  def isEmpty: Boolean = this.ars.empty
+
+  def actions: List[Action] = {
+
+    // not available in scalaz?
+    def lefts[L, R](xs: List[L \/ R]): List[L] = for { -\/(r) <- xs } yield r
+
+    lefts(this.ars.toList)
+  }
+
+  def results: List[Result] = {
+    // not available in scalaz?
+    def rights[L, R](xs: List[L \/ R]): List[R] = for { \/-(r) <- xs } yield r
+
+    rights(this.ars.toList)
+  }
+}
 
 object Current {
   val empty: Current = Current(Vector.empty, None)
@@ -302,13 +335,12 @@ object Current {
     * If the index doesn't exist, `Current` is returned unmodified.
     */
   def mark(i: Int)(r: Result)(c: Current): Current =
-    Current(PLens.vectorNthPLens(i).setOr(c.actions, r.right, c.actions), c.ctxt)
+    Current(PLens.vectorNthPLens(i).setOr(c.ars, r.right, c.ars), c.ctxt)
 }
 
 /**
   * The result of an `Action`.
   */
-
 sealed trait Result
 
 object Result {
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
@@ -12,7 +12,7 @@ case class QState(pending: Queue[Action],
                   done: Queue[Result],
                   status: Status) {
 
-  def isEmpty: Boolean = this.pending.sequences.isEmpty && this.current.isEmpty
+  def isEmpty: Boolean = pending.sequences.isEmpty && current.isEmpty
 }
 
 object QState {
@@ -58,18 +58,15 @@ object QState {
     * If the `Current` doesn't have all actions completed or there are no more
     * pending `Execution`s it returns None.
     */
-  def next(qs: QState): Option[QState] = cleanup(qs) >>= (prime(_))
+  def next(qs: QState): Option[QState] = cleanup(qs).flatMap(prime)
 
   // None: current not empty
   //       pending queue empty
   def prime(qs: QState): Option[QState] =
     if (qs.current.isEmpty)
-      Queue.uncons(qs.pending).map(
-        t => {
-         val (exe3, q) = t
-         QState(q, currentify(exe3), qs.done, qs.status)
-        }
-      )
+      Queue.uncons(qs.pending).map {
+        case (exe3, q) => QState(q, currentify(exe3), qs.done, qs.status)
+      }
     else None
 
   def cleanup(qs: QState): Option[QState] =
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -289,10 +304,28 @@ object Step {
...
                    // Sequence or Step parameters
-                   ctxt: Option[(String, Int) \/ Int])
+                   ctxt: Option[(String, Int) \/ Int]) {
+
+  def isEmpty: Boolean = this.ars.empty
+
+  def actions: List[Action] = {
+
+    // not available in scalaz?
+    def lefts[L, R](xs: List[L \/ R]): List[L] = for { -\/(r) <- xs } yield r
```
cquiroz: `leftMap` perhaps?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -52,26 +58,37 @@ object QState {
...
+    uncurrentify(qs.current).map(exe3 =>
+      QState(qs.pending, Current.empty, Queue.cons(exe3)(qs.done), qs.status)
+    )
   /**
     * Transform an *unconsed* pending `Execution` into `Current` in addition to
     * returning the unwrapped `Execution`.
     */
   // TODO: Use same structure for `Current` and `Queue.Execution3`?
-  private def currentify(exe3: Queue.Execution3[Action]): (Execution[Action], Current) = {
+  private def currentify(exe3: Queue.Execution3[Action]): (Current) = {
```
cquiroz: What do you mean by `cons` and `uncons`?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: I was thinking about the [Haskell `lefts`](https://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Either.html#v:lefts). `leftMap` maps over every `left` but I just want to *collect* every left in a list of `\/`. I'm pretty sure there'll be something in scalaz to do that but can't find how it's named.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: To be honest I didn't really know what [`cons`](https://en.wikipedia.org/wiki/Cons) stands for until I looked it up but it's often used when adding an element to the front of something that looks like a single linked list. I guess the term is more widespread in Haskell/Lisp than anywhere else.

I could rename it if there is another term for that in Scala. How is `::` pronounced in Scala?

In any case these functions should be internal to just that module.

jdnavarro:
After writing more tests I realized the completed `Execution`s are not added properly. I'll include the fix in this PR.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
cquiroz: Right `::` is called `cons` in a list. I just wondered what was the use here. I haven't seen `uncons` though

cquiroz: Why's the return type in parentheses?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -52,26 +58,37 @@ object QState {
...
 
     def vec(exe: Execution[Action]): Vector[Action \/ Result] = exe.map(_.left).toVector
 
     exe3 match {
-      case -\/(-\/((actions, seqid, stepid))) => (actions, Current(vec(actions), Some((seqid, stepid).left)))
-      case -\/(\/-((actions, stepid))) => (actions, Current(vec(actions), Some(stepid.right)))
-      case \/-(actions) => (actions, Current(vec(actions), None))
+      case -\/(-\/((actions, seqid, stepid))) => Current(vec(actions), Some((seqid, stepid).left))
```
cquiroz: it is a bit strange to see two `\/` nested. what do they represent?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -187,7 +201,8 @@ object Queue {
           mseq match {
             // No more Steps in current Sequence, remove Sequence.
             // TODO: listTailPLens?
-            case None => ((exe, seq0.id, stepid).left.left, Queue(q.sequences.tailOption.getOrElse(List())))
+            case None => ((exe, seq0.id, stepid).left.left,
+                          Queue(q.sequences.tailOption.getOrElse(List())))
```
cquiroz: Often `List()` is represented as `Nil`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
cquiroz: ok, you can do `collect` on the list. something like

```scala
xs.collect {
   case -\/(e) => e
}
```
or perhaps
```scala
xs.collect {
   case e @ -\/(_) => e
}
```

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -44,29 +44,67 @@ package object engine {
...
+      }
+    }
+
+    status >>= {
+      case Status.Waiting => unit
+      case Status.Running => (
+        gets(_.current.actions) >>= (
+          actions => Nondeterminism[Task].gatherUnordered(
+            actions.zipWithIndex.map(act(_))
+          ).liftM[EngineStateT]
```
cquiroz: Can you remind me what `liftM` does here?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: It's not clear to me how the opposite of `cons` should be named, I've seen it as `uncons` and `snoc`. Monocle uses `snoc` so if we end up using Monocle, `snoc` would be a better name.

I forgot to remove the parentheses, good catch.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: Each sum type of the `\/` represents if either a `Sequence` or a `Step` or a `Execution` in the current `Step` was completed. It's explained in the documentation for that function but I should put a small comment for each case pattern to make the code easier to follow.

`\/` with more than 2 types was strange for me too because in Haskell you can't do that. One of the nice features of `\/` is that it supports sum types of arbitrary number of elements, however in practice it turns out it's not so great because as you can see, pattern matching is not so straightforward.

The solution for this is to create a custom sum data type. This will be necessary, in any case, when adding more complex Steps so I'm writing this down as an improvement to do later.

In any case this *weird* sum types are internal to this object, they are not meant to be used anywhere else.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: Good to know. I think I have more instances of `List()` spread over the code, so I'll change them all.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: This is just for lifting a `Task` into `Engine`: `Task[A] ~> Engine[A]`. `Engine` is a monad transformer of just 1 layer, the StateT with the Seqexec State over `Task` as the base monad.

What I got from the discussion we had a while back, with @swalker2m and @tpolecat too, is that 1 layer is manageable but more than 1 layer become too cumbersome in Scala. It seems that when the need for more transformer layers arise all the solutions lean towards the use of Free Monads for cleaner and more composable effects.

I wanted to experiment first with a simple toy example combining 2 effects using Free monads and coproducts before trying to do it in the Seqexec code, but so far I could get away with just 1 transformer layer without too much trouble.

However, I don't think the code will be able to scale in complexity properly with just 1 layer. For example, another layer I wish we had is a Reader monad (a.k.a Kleisli) with the *scalaz-stream* queue so that we don't have to be passing the queue explicitly as a parameter all over the code. Also, a *Reader* will be inevitable when/if we need to access a global configuration object from the engine.

So at some point all those `liftM` should go away.

jdnavarro:
After writing more tests I realized the completed `Execution`s are not added properly. I'll include the fix in this PR as well.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
cquiroz: In think it's fine this way, unless you think a custom type will add value

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
cquiroz: right, the config system uses `Kliesli` for configuration. About the scalaz-stream `queue` the current implementation has a single `queue` in an object becoming global. Would that work instead of passing it as a param?

jdnavarro: A global mutable variable (in order to create a `queue` we would have to do IO at some point) brings me painful memories from my Python times :). In the concrete case of the `queue` as a parameter it looks like it would be OK to use an unsafe global: it's only created once and it's only read everywhere and not meant to be modified.

But since it seems will have to include the `Kleisli` for configuration anyway, we could piggyback the `queue` in that `Kleisli`.

cquiroz: Actually in ocs3 the queue is passed as a variable, but initialized on the object construction

https://github.com/gemini-hlsw/ocs3/blob/8c859ddb61d920a8faf8d6d5e5f00737dc3812e2/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala#L127

So passing it on the `Kliesli` sounds equivalent

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -10,7 +10,10 @@ import Scalaz._
 case class QState(pending: Queue[Action],
                   current: Current,
                   done: Queue[Result],
-                  status: Status)
+                  status: Status) {
+
+  def isEmpty: Boolean = this.pending.sequences.isEmpty && this.current.isEmpty
```
tpolecat: Minor stylistic thing; it's not idiomatic to use `this.` for members unless there is a need to disambiguate.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -52,26 +58,37 @@ object QState {
     * If the `Current` doesn't have all actions completed or there are no more
     * pending `Execution`s it returns None.
     */
-  def next(st: QState): Option[(Execution[Action], QState)] = for {
-    exe3done <- uncurrentify(st.current)
-    qd = Queue.cons(exe3done)(st.done)
-    (exe3pending, qp) <- Queue.uncons(st.pending)
-    (actions, c) = currentify(exe3pending)
-  } yield (actions, QState(qp, c, qd, st.status))
+  def next(qs: QState): Option[QState] = cleanup(qs) >>= (prime(_))
```
tpolecat: You should be able to write this as `cleanup(qs) >>= prime`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -52,26 +58,37 @@ object QState {
...
+
+  // None: current not empty
+  //       pending queue empty
+  def prime(qs: QState): Option[QState] =
+    if (qs.current.isEmpty)
+      Queue.uncons(qs.pending).map(
+        t => {
+         val (exe3, q) = t
+         QState(q, currentify(exe3), qs.done, qs.status)
+        }
```
tpolecat: A function literal can be expressed in cases.

```scala
Queue.uncons(qs.pending).map {
  case (exe3, q) => QState(q, currentify(exe3), qs.done, qs.status)
}
```

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
tpolecat: `snoc` typically is right-append. I think `uncons` is more typical.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
tpolecat: Shapeless provides an open coproduct type that lets you express `A \/ B \/ C` as `A :+: B :+: C :+: CNil`, and then you can fold by applying a polytypic function. It's much nicer in general but probably not introducing for isolated cases like this.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
tpolecat: `.collect` is probably the most efficient solution but you could also say `.separate._1`. Scalaz would take a PR if you wanted to add this to `MonadPlus`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -44,29 +44,67 @@ package object engine {
...
+  private def execute(q: EventQueue): Engine[Unit] = {
+
+    // Send the expected event when action is executed
+    def act(t: (Action, Int)): Task[Unit] = {
+      val (action, i) = t
+      action >>= {
+        case Result.OK => q.enqueueOne(completed(i))
+        case Result.Error => q.enqueueOne(failed(i))
+      }
+    }
```
tpolecat: Prefer `match` over destructuring assignment because you'll get exhaustiveness checking. Also in general use `flatMap` unless `>>=` is clearly better because it's more efficient.

```scala
... = t match {
  case (action, i) =>
    action.flatMap {
      case ...
    }
}
```

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -44,29 +44,67 @@ package object engine {
...
+
+    status >>= {
+      case Status.Waiting => unit
+      case Status.Running => (
+        gets(_.current.actions) >>= (
+          actions => Nondeterminism[Task].gatherUnordered(
+            actions.zipWithIndex.map(act(_))
```
tpolecat: You can probably say `.map(act)` here.

tpolecat:
I don't really understand the new code review feature, sorry. Some of my comments are out of context because I added them to running conversations. So maybe look at the code tab.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: I try to avoid `collect` if there is another alternative with total functions but this looks like an ideal case.

I didn't know about `.separate._1`. I'll also look into the laws for `MonadPlus` in Scalaz. In Haskell there is no consensus about which laws should a `MonadPlus` obey and I'm curious to see which ones Scalaz sticks to.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: The way `\/` handles more than 2 types with those constructors is a bit hacky. For now it's fine, but I'll keep and eye on Shapeless if I ever need more robust sum types.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: Oh, now I see what you mean, I didn't consider that pattern. I'm reading now about `TaskRef` to understand how they work. I also realize that the `DhsClient` will have to follow a similar pattern to the `scalaz-stream` queue creating, so I think it's worth exploring the possibilities.

In the end, I think it'll be worth having something similar to an `Environment` object embedded in a `Kleisli` with everything that needs to be read-only throughout the Seqexec code. This object can be created only once, when initialized.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: I was reluctant to use `this` but wanted to make `isEmpty` a method of `Queue` instead of a standalone function over `Queue` because it seems more natural for a publicly exposed API.

Is there a more functional way to implement this kind of methods? The Syntax Ops packages seem a bit overkill but perhaps they are worth writing for these cases anyway.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: Oh, I thought the destructuring assignment were syntactic sugar for `_ match {...`. Also, TIL `flatMap` and `>>=` are not the same! I thought they were aliases of each other. Good to know.

tpolecat: No, destructing assignments are not checked carefully so you can get a match error at runtime. `>>=` is equationally identical to `flatMap` but has higher overhead because there is an intervening implicit conversion. In practice it rarely matters but it's good to remember.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: Never mind my previous comment. I just realized the parameters of a case class are in scope in its case class block.

jdnavarro:
The completion of Executions is still broken but since I'm realizing it will take some time to fix it, will it be OK to merge this PR as it is?

PR #86, opened 2016-09-21
feature/OCSADV-523

jluhrs:
A function that takes a sequence Config builds the initial sequence state for the execution engine.
It does not take into account the step state read from the ODB, so for now it will try to run steps already executed.
This commit as some conflicts with existing code. Task SEQNG-5 will clean those issues.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -1,0 +1,54 @@
+package edu.gemini.seqexec.server
+
+import edu.gemini.seqexec.engine.{Action, Result, Sequence, Step}
+import edu.gemini.seqexec.server.SeqexecFailure.{Execution, UnrecognizedInstrument}
+import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
+import edu.gemini.spModel.obscomp.InstConstants._
+import edu.gemini.spModel.seqcomp.SeqConfigNames._
+
+import scalaz.\/
+import scalaz.NonEmptyList._
+import scalaz._
+import Scalaz._
+import scalaz.syntax.TraverseOps
+
+/**
+  * Created by jluhrs on 9/14/16.
+  */
+object SeqTranslate {
+
+  implicit def toAction[A](x: SeqAction[A]): Action = x.run map {
+    case -\/(e) => Result.Error(e)
+    case \/-(r) => Result.OK(r)
+  }
+
+  def step(dhsClient: DhsClient)(i: Int, config: Config): SeqexecFailure \/ (Step[Action]) = {
+    val instName = config.getItemValue(new ItemKey(INSTRUMENT_KEY, INSTRUMENT_NAME_PROP))
+    val instrument = instName match {
+      case GmosSouth.name => Some(GmosSouth)
+      case Flamingos2.name => Some(Flamingos2(Flamingos2ControllerSim))
+      case _ => None
+    }
+
+    instrument.map { a =>
+      val systems = nels(Tcs(TcsControllerEpics), a)
+      // TODO Find a proper way to inject the subsystems
+      Step[Action](i, nels(systems.map(x => toAction(x.configure(config))), nels(toAction(a.observe(config)(dhsClient))))).right
+    }.getOrElse(UnrecognizedInstrument(instName.toString).left[Step[Action]])
+  }
+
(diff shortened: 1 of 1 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -0,0 +1,54 @@
...
+  * Created by jluhrs on 9/14/16.
+  */
+object SeqTranslate {
+
+  implicit def toAction[A](x: SeqAction[A]): Action = x.run map {
+    case -\/(e) => Result.Error(e)
+    case \/-(r) => Result.OK(r)
+  }
+
+  def step(dhsClient: DhsClient)(i: Int, config: Config): SeqexecFailure \/ (Step[Action]) = {
```
cquiroz: Why the parenthesis on `Step[Action]`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -0,0 +1,54 @@
...
+    }.getOrElse(UnrecognizedInstrument(instName.toString).left[Step[Action]])
+  }
+
+  def sequence(dhsClient: DhsClient)(obsId: String, sequenceConfig: ConfigSequence): SeqexecFailure \/ Sequence[Action] =
+  {
+    def nelOfConfigs(l : List[Config]): SeqexecFailure \/ NonEmptyList[Config] =
+      l.toNel.map(_.right).getOrElse(Execution("No steps found in sequence.").left)
+
+    def steps(l : NonEmptyList[Config]): SeqexecFailure \/ NonEmptyList[Step[Action]] =
+      l.zipWithIndex.traverse[({ type λ[β] = SeqexecFailure \/ β })#λ, Step[Action]] { case (c, i) => step(dhsClient)(i, c) }
```
jdnavarro: I guess this won't work without the type lambda, would `traverseU` do the trick?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -0,0 +1,54 @@
...
+  */
+object SeqTranslate {
+
+  implicit def toAction[A](x: SeqAction[A]): Action = x.run map {
+    case -\/(e) => Result.Error(e)
+    case \/-(r) => Result.OK(r)
+  }
+
+  def step(dhsClient: DhsClient)(i: Int, config: Config): SeqexecFailure \/ (Step[Action]) = {
+    val instName = config.getItemValue(new ItemKey(INSTRUMENT_KEY, INSTRUMENT_NAME_PROP))
```
cquiroz: I'd thing `config.getItemValue` can return a `null`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -0,0 +1,54 @@
...
+    val instrument = instName match {
+      case GmosSouth.name => Some(GmosSouth)
+      case Flamingos2.name => Some(Flamingos2(Flamingos2ControllerSim))
+      case _ => None
+    }
+
+    instrument.map { a =>
+      val systems = nels(Tcs(TcsControllerEpics), a)
+      // TODO Find a proper way to inject the subsystems
```
cquiroz: I'd like this to be compatible with the web-side configuration system. we can discuss that further down the line

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
cquiroz: Could/should we use the kind-projector plugin to remove the lambdas?

jluhrs: I'm not willing to use a projector that I don't know, no matter how kind it claims to be.

cquiroz: kind-projector is a scala compiler plugin that is being used on the `gem` project already so it is likely it will endup on ocs3 too

https://github.com/non/kind-projector

It should let you use type lambdas more clearly. I don't know if it worth in this case but perhaps it is useful

cquiroz:
There are compilation errors on travis with this PR. Can you try doing a full compile

jluhrs:
This pull request was included in pull request #87

PR #87, opened 2016-09-26
Feature/SEQNG-5

jluhrs:
Includes Jira tasks OCSADV-523 and SEQNG-5.
I just cleaned the code to make it compile. I removed some old code that is made obsolete by the new execution engine. Some parts that depended on the old code have to be updated, but for now I left them unimplemented.


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -79,12 +79,12 @@ class SeqexecUIApiRoutes(auth: AuthenticationService) extends BooPicklers {
   def userInRequest(req: Request) = req.attributes.get(JwtAuthentication.authenticatedUser).flatten
 
   val protectedServices: HttpService = tokenAuthService { GZip { HttpService {
-      case req @ GET -> Root / "seqexec" / "events"         =>
-        // Stream seqexec events to clients and a ping
-        val user = userInRequest(req)
-
-        WS(Exchange(pingProcess merge (Process.emit(Binary(trimmedArray(SeqexecConnectionOpenEvent(user)))) ++
-          ExecutorImpl.sequenceEvents.map(v => Binary(trimmedArray(v)))), scalaz.stream.Process.empty))
+//      case req @ GET -> Root / "seqexec" / "events"         =>
```
cquiroz: This would certainly break the web UI

cquiroz:
Some of this code looks similar to PR #86 , does this PR replaces it?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
```diff
@@ -28,13 +28,13 @@ object Event {
     * Events generated internally by the Engine.
     */
   sealed trait SystemEvent
-  case class Completed(i: Int) extends SystemEvent
-  case class Failed(i: Int) extends SystemEvent
+  case class Completed[R](i: Int, r: R) extends SystemEvent
+  case class Failed[E](i: Int, e: E) extends SystemEvent
   case object Executed extends SystemEvent
```
jdnavarro: Do `R` and `E` need to be polymorphic? I'm not saying they don't have to be, but if there is a manageable number of types of `Error`s and `Result`s it may be easier to work with concrete types.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -0,0 +1,54 @@
...
+    }.getOrElse(UnrecognizedInstrument(instName.toString).left[Step[Action]])
+  }
+
+  def sequence(dhsClient: DhsClient)(obsId: String, sequenceConfig: ConfigSequence): SeqexecFailure \/ Sequence[Action] =
+  {
+    def nelOfConfigs(l : List[Config]): SeqexecFailure \/ NonEmptyList[Config] =
+      l.toNel.map(_.right).getOrElse(Execution("No steps found in sequence.").left)
+
+    def steps(l : NonEmptyList[Config]): SeqexecFailure \/ NonEmptyList[Step[Action]] =
+      l.zipWithIndex.traverse[({ type λ[β] = SeqexecFailure \/ β })#λ, Step[Action]] { case (c, i) => step(dhsClient)(i, c) }
```
jdnavarro: Would traverseU infer the correct type without the type lambda?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+  * Created by jluhrs on 9/14/16.
+  */
+object SeqTranslate {
+
+  implicit def toAction[A](x: SeqAction[A]): Action = x.run map {
+    case -\/(e) => Result.Error(e)
+    case \/-(r) => Result.OK(r)
+  }
```
jdnavarro: What's the intention of having `toAction` as an implicit function? Where's is it being used as such? I still don't get implicits quite yet.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -0,0 +1,54 @@
...
+      val systems = nels(Tcs(TcsControllerEpics), a)
+      // TODO Find a proper way to inject the subsystems
+      Step[Action](i, nels(systems.map(x => toAction(x.configure(config))), nels(toAction(a.observe(config)(dhsClient))))).right
+    }.getOrElse(UnrecognizedInstrument(instName.toString).left[Step[Action]])
+  }
+
+  def sequence(dhsClient: DhsClient)(obsId: String, sequenceConfig: ConfigSequence): SeqexecFailure \/ Sequence[Action] =
+  {
+    def nelOfConfigs(l : List[Config]): SeqexecFailure \/ NonEmptyList[Config] =
+      l.toNel.map(_.right).getOrElse(Execution("No steps found in sequence.").left)
```
jdnavarro: Maybe it's worth here using the fully qualified import in the form of`SeqexecFailure.Execution`, or maybe rename `Execution` to `ExecFailure` or something similar.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
jluhrs: This has to be rewritten anyway, to connect it to the new execution engine. That will be another task.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
jluhrs: It's just that I know something must go there, but I am not sure at this point what it is.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: That was just to not rewrite the functions that produce the TCS and instruments configuration actions. They produce SeqActions instead of actions, but they can be converted. Eventually the original functions have to be modified to produce Actions.

jdnavarro: Sure, I get that `SeqAction`s have to be converted to `Action`s, and that `toAction` is necessary.

I'm just trying to understand why `toAction` is an implicit function instead of an ordinary one. It's not that I'm against making `toAction` `implicit`, it's just me trying to understand Scala basics 😉 .

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: Well, I didn't use it because Idea was complaining, but I just tried compiling in sbt, and it works.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: I wanted to use implicit conversion. But then I left the explicit conversion in line 36 :(

PR #88, opened 2016-09-26
Upgrades

cquiroz:
This PR updates several libraries used in the seqexec-web plus some minor ui fixes:

scalaJS 0.6.10 -> 0.6.12
scalaDOM 0.9.0 -> 0.9.1
scalaJS-React 0.11.1 -> 0.11.2
scalaCSS 0.4.0 -> 0.5.0
http4s 0.14.1a -> 0.14.7a
jwt 0.7.1 -> 0.8.1
scalatest 0.3.0-RC1 -> 0.3.0

and on the js side
react 15.0.1 -> 15.3.2
semantic ui 2.2 -> 2.2.2

Very minor code changes are required. The most interesting update is scalajs that should let improve testing. The scalacss update also helps to reduce the js output in about 40k

This was tested on Chrome, Safari and Firefox 38 ESR in OSX and Safari in iOS

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
@@ -166,12 +166,12 @@ object QueueAreaTitle {
           implicit val eq = PotEq.seqexecQueueEq
           queueConnect(QueueTableLoading.apply)
         },
-        p.user().map { u =>
+        p.user().fold(<.div()) { _ =>
           <.div(
             ^.cls := "right menu",
             statusAndSearchResultsConnect(SequenceSearch.apply)
-          ): ReactNode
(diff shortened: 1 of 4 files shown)


cquiroz:
Hmm, that is interesting, I hadn't notice but will verify. If so it is certainly a bug

jdnavarro:
Here I'm attaching a screenshot of the debugging console
screen shot 2016-09-26 at 4 54 05 pm

cquiroz:
I cannot reproduce what you're seeing in neither Chrome nor firefox, I tried both with `localhost` and my normal IP and I don't see it

jdnavarro:
Strange... I can reproduce it in a Linux machine too. But this only happens in Firefox, Chrome is fine.

My Firefox version both in Linux and OS X is 49.0.1.

cquiroz:
I can't reproduce it in Linux, but I have Firefox ESR version 38
In OSX Firefox 49.0.1 doesn't happen either

jdnavarro:
It turns out that if I disable all my Firefox addons the error doesn't show up. I don't think my addons are representative of an average Firefox profile 😃 so I think it's safe to ignore my specific case.

👍

PR #89, opened 2016-09-27
Feature/SEQNG-14

jdnavarro:
This introduces *unrooted* Sequences and Steps into the `Queue` data
structure. These are used for keeping track of `Sequences` or `Steps`
that are in an ongoing execution.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
@@ -130,18 +130,32 @@ object Status {
   * A list of Sequences. The `Queue` could be empty of Sequences when waiting
   * for the addition of new ones.
   */
-case class Queue[A](sequences: List[Sequence[A]]) {
+case class Queue[A](
+  sequences: List[Sequence[A]],
+  // Unrooted Steps (without a Queue)
+  steps: List[Step[A]],
+  // Unrooted Executions (without a Step)
+  executions: List[Execution[A]]
+) {
+
   def isEmpty: Boolean = sequences.isEmpty
+
 }
 
 object Queue {
 
   type Execution3[A] = (Execution[A], String, Int) \/ (Execution[A], Int) \/ Execution[A]
 
-  def empty[A]: Queue[A] = Queue(Nil)
+  def empty[A]: Queue[A] = Queue(Nil, Nil, Nil)
 
   def sequences[A]: Queue[A] @> List[Sequence[A]] =
-    Lens.lensu((q, ss) => q.copy(sequences = ss), _.sequences)
+    Lens.lensu((q, s) => q.copy(sequences = s), _.sequences)
+
+  def steps[A]: Queue[A] @> List[Step[A]] =
+    Lens.lensu((q, s) => q.copy(steps = s), _.steps)
+
+  def executions[A]: Queue[A] @> List[Execution[A]] =
+    Lens.lensu((q, e) => q.copy(executions = e), _.executions)
 
   // Next `Sequence`
   def head[A]: Queue[A] @?> Sequence[A] = sequences.partial >=> PLens.listHeadPLens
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -130,18 +130,32 @@ object Status {
   * A list of Sequences. The `Queue` could be empty of Sequences when waiting
   * for the addition of new ones.
   */
-case class Queue[A](sequences: List[Sequence[A]]) {
+case class Queue[A](
+  sequences: List[Sequence[A]],
+  // Unrooted Steps (without a Queue)
```
cquiroz: Somehow the "unrooted" state is very unclear to me. Why would there be steps not in the queue? And the same for executions

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -222,7 +257,12 @@ object Queue {
 /**
   * A list of Steps grouped by target and instrument.
   */
-case class Sequence[A](id: String, steps: NonEmptyList[Step[A]])
+case class Sequence[A](
+  id: String,
+  steps: NonEmptyList[Step[A]],
+  // Unrooted executions
```
cquiroz: Same here, what does it mean that a sequence has executions that don't belong to a step?

jdnavarro: This is the same case as before where an *unrooted* `Execution` represents a completed `Execution` in a yet to be completed `Step`. Once the whole `Step` is completed the *unrooted* `Execution`s are added to the `Step`.

cquiroz: Ok, I guess in my mind an uncompleted execution would have a just a `completed` flag and belong to a `Step`. I guess there are just 2 ways to express the same thing

jdnavarro: Yes, those structures are equivalent.

The thing is that with in the Engine we are all the time working with the *current* execution. This structure with 2 queues and a cursor resembles a [`Zipper`](https://en.wikipedia.org/wiki/Zipper_(data_structure)) which makes the focused element (the current execution) very accessible and facilitates moving back and forth.

Without a `Zipper` we would have to do the work of finding where we are each time we want to change the State.

In any case, a `Zipper` is expected to be *zipped* and *unzipped* from and to its original structure which in this case will be a whole single tree with each element having its own status flag.

cquiroz: Are you using `Zipper` from scalaz? So we'd have `focus`, `lefts` and `rights`. I somehow thought that focus would be in either `lefts` or `rights` but I guess being separated is possible too.

What would be the type in the zipper? at the level of `Step`?

jdnavarro: The `Zipper` structure only exists at the very top level, at the `Execution Queue` level. We could make extra zippers at each level but I don't think it'd be worth it since, AFAIK, when executing an *Execution Queue* (which looks like a tree), we need to precisely locate where is the actual execution in reference to the top level structure.

While executing we need to know in which `Sequence` a `Step` is located and in which position in the `Queue` is that `Sequence`. An `Execution` must always belong to a `Step`, a `Step` must always belong to a `Sequence` and so forth. So you can only create the Zipper when all elements in the tree are properly connected.

The `Zipper` in scalaz works for plain `List`s, but it's the same idea for our custom data type.

Actually, [`TreeLoc`](https://oss.sonatype.org/service/local/repositories/releases/archive/org/scalaz/scalaz_2.11/7.2.5/scalaz_2.11-7.2.5-javadoc.jar/!/index.html#scalaz.TreeLoc) which is a `Zipper` for rose trees resembles much closely the zipper for the *Execution Queue*. The crucial difference is that we don't need to move in 4 directions, only forwards and in a limited way backwards when pausing.

Another difference is that our tree is not recursive, we don't need to have an arbitrary number of deepness levels. In our case having a fixed number of levels deep (just 4) enforced by the type system it's beneficial even if a bit cumbersome to encode.

This guarantees that we can't create tree nodes where they don't belong. Having different types at each level guarantees that we can't treat accidentally a `Step` as if it were a `Sequence`, for example.

jdnavarro: BTW, the concept of `Execution` shouldn't show up outside of the Engine. It internally represents a set of `Action`s that need to be executed in parallel.

cquiroz: If `Execution` shouldn't leak perhaps it should be package protected

cquiroz:
Thanks @jdnavarro on the explanations plus the chat to understand what's going on. I don't want to keep this PR from being merged, so go ahead (at least from my side) as soon as you need to

PR #90, opened 2016-09-27
Tests with jsdom

cquiroz:
This PR fixes one of the trouble running test that interact with the DOM. There has been no reliable way to run these on any of the Scala.js runtime environments, since rhino is to weak and slow to execute `jQuery` and `node` doesn't have a DOM

On this PR we use [jsdom](https://github.com/tmpvar/jsdom) that runs in `node` but provides a DOM basic environment. Note that this requires you installing node and jsdom but that is going to become the default for scala.js anyway

With this PR we can finally run the client-side tests on travis and we can build more meaningful tests in the future

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # OCS3
 
+[![Build Status](https://travis-ci.org/gemini-hlsw/ocs3.svg?branch=develop)](https://travis-ci.org/gemini-hlsw/ocs3)
+
 This project aims to build the next generation Observatory Control System (OCS). At the time of this writing, only the Seqexec has been integrated. This setup has the following goals
 
 - Adhere to the simplest possible sbt structure.
@@ -44,6 +46,10 @@ To tests the javascript side you need to install node and jsdom. In OSX using `b
 brew install node
 npm install jsdom
 ```
+
+`node` should be at least on version `0.6.x`, check it with `node --version` once installed
+
+Don't install jsdom at the global level (with the -g option) as it doesn't work properly with scala.js
 
 ## IDEA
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -30,40 +23,41 @@ ocs3/
...
+`ScalaTest` has been chosen as the test framework since it works in both scala.js and jvm. `ScalaCheck` is also supported running with `ScalaTest`.
+Sample tests using `scalatest` and `scalacheck` are included in some modules. You can run `test` at the root level or on a module basis.
 
 Note that IDEA doesn't support running tests on scala.js. See this [issue](https://github.com/scalatest/scalatest/issues/743)
 
+To tests the javascript side you need to install node and jsdom. In OSX using `brew` you can do:
+
+```
+brew install node
+npm install jsdom
```
cquiroz: Thanks for checking this. the global install doesn't seem to work, you need to do it locally

Unfortunately `jsDependencies` won't help in this case. You use `jsDependencies`  to add js libraries to your program while `node` is the runtime and `jsdom` provides a `DOM` mockup. In the browser both are provided while we use `node` only to run the tests

I hope that when `Rhino` is phased away by scala.js they provide more foolproof instructions on installing `node`

I'll update the README a bit to clarify some of these concepts

cquiroz:
As an extra note, there are absolutely no plans to use node modules in the application and I'd be very skeptical if we ever try to go that way. However in this case there is not really a workaround, the other option would be to use `phantomjs` as a runtime but that brings its own set of problems. 

Ideally we'd run the tests on the browser too. I plan to check the state of https://github.com/scala-js/scala-js-env-selenium

jdnavarro:
I haven't used Selenium in a while, but some years ago I remember it was a pain to configure it properly for each machine where we wanted to test. There are nowadays services that make browser testing much easier, so maybe it's worth looking into those.

cquiroz:
I agree about selenium setup issues. My idea is not to force people to use it but make it instead an optional sbt task that you can run if you want. Ideally travis would run it too, but that may require some extra effort on the travis setup

Issue #82, opened 2016-08-05
Add mock sequences for testing


cquiroz:
We should be able to build mock sequences for testing without connecting to the ODB

PR #92, opened 2016-09-29
Mock Sequences

jdnavarro:
I still plan to add more `Sequences`, but I'm already sending the PR in case @cquiroz detects I'm already doing something wrong.

The filename and location in the `ocs3` directory tree may not be ideal. If you have a better alternative, please, let me know.

Notice this PR is against `develop`, not `feature/engine`.

See #82.


cquiroz:
I think this is fine though I was thinking more on the line of a function that could generate a sequence given some parameters, e.g. step counts.

I think the steps should have a configuration though, that's important to run meaningful simulations

jdnavarro:
I can try to refactor what I've done manually into a function to create the different states, but first I'd like to make sure I'm understanding how the state should change from step to step. Do mock sequences as they are now represent the expected flow of execution?

The configuration `Step`s are key value pairs. Do you know off the top of your head, some dummy parameters that resemble what occurs in reality?

jdnavarro:
Right now this is too half-baked to be worth merging. Since this this is not too urgent I'm keeping them locally in case we need to finish it in the future.

PR #93, opened 2016-10-03
SEQNG-7: Added list of basic tests.

jluhrs:
It is just the list of tests, with no implementation yet.


jdnavarro:
👍 It's useful to know now how the tests should look even if they are not implemented yet. It seems travis is failing in an edge case, I'm repeating the build so that we can merge it.

PR #94, opened 2016-10-06
New Engine state based in a triple Zipper

jdnavarro:
This is the real implementation of the experiment we planned for SEQNG-17.

The external API for the Engine state, which is used for the actual execution, hasn't change much. The internal state has been completed rewritten so that the different levels of the engine state need to know much lesser of each other. This should make the state much more modular and testable.

There are some commits from SEQNG-20 (the function for: `engine state -> web server state`), that I'm still working on, and that now it should be easier to complete, but I think those commits won't get on the way of a review for this PR that much.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
@@ -43,19 +43,19 @@ case class Current(ars: List[Action \/ Result]) {
     if (ars.all(_.isRight)) Some(results)
     else None
 
-}
-
-object Current {
-
-  val empty: Current = Current(Nil)
-
   /**
     * Set the `Result` for the given `Action` index in `Current`.
     *
     * If the index doesn't exist, `Current` is returned unmodified.
     */
-  def mark(i: Int)(r: Result)(c: Current): Current =
-    Current(PLens.listNthPLens(i).setOr(c.ars, r.right, c.ars))
+  def mark(i: Int)(r: Result): Current =
+    Current(PLens.listNthPLens(i).setOr(ars, r.right, ars))
+
+}
+
+object Current {
+
+  val empty: Current = Current(Nil)
 
   def currentify(exe: Execution[Action]): Option[Current] =
     if (!exe.isEmpty) Some(Current(exe.map(_.left)))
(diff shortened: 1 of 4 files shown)


jdnavarro:
There are some build errors when loading of Sequences from ODB, I think. I can fix them once I get the OK for the changes for the engine state in this PR.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -4,388 +4,275 @@ import scalaz._
...
-  def mark(i: Int)(r: Result)(qs: QState): QState =
-    current.mod(Current.mark(i)(r)(_), qs)
+    def rights[L, R](xs: List[L \/ R]): List[R] = xs.collect { case \/-(r) => r }
+
+    rights(ars.toList)
+
+  }
+
+  val uncurrentify: Option[Execution[Result]] =
+    if (ars.all(_.isRight)) Some(results)
```
cquiroz: you can do `ars.all(_.isRight) option results`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -4,388 +4,275 @@ import scalaz._
...
 
-  def isEmpty: Boolean = ars.empty
+  val next: Option[QState] = zipper.next match {
+    // Last execution
+    case None => zipper.uncurrentify.map(QStateF(_, status))
+    case Some(x) => Some(QStateZ(x, status))
+  }
 
-  def actions: List[Action] = {
+  val current: Current = zipper.focus.focus.focus
```
cquiroz: `focus.focus.focus`? Perhaps it would be clearer with a clearer name. is it `currentExecution`?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
```diff
@@ -4,388 +4,275 @@ import scalaz._
...
   /**
-    * Set the `Result` for the given `Action` index in `Current`.
-    *
-    * If the index doesn't exist, `Current` is returned unmodified.
+    * Initialize a `QStateZ` passing a `Queue` of `Action`s. This also takes care
+    * of making the first pending `Execution` `Current`.
     */
-  def mark(i: Int)(r: Result)(c: Current): Current =
-    Current(PLens.vectorNthPLens(i).setOr(c.ars, r.right, c.ars), c.ctxt)
+  // TODO: Make this function `apply`?
```
cquiroz: yes, probably makes sense as apply

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -45,31 +45,19 @@ package object engine {
     */
   def switch(q: EventQueue)(st: Status): Engine[QState] =
     modify(QState.status.set(_, st)) *>
-    whenM (st == Status.Running) (prime *> execute(q)) *>
+    whenM (st == Status.Running) (execute(q)) *>
```
cquiroz: Where is `whenM` defined?

cquiroz:
Why is travis failing?

jdnavarro:
I'm explaining the travis error [above](https://github.com/gemini-hlsw/ocs3/pull/94#issuecomment-252082187).

The comment thread is a bit unclear after the new review feature. I would have liked to have that comment just below the travis fail, but I don't think it's possible right now.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/model.scala
jdnavarro: `current` is the current Execution. I'll put a comment about which each focus level refers to. Outside of this module you should never need to mess up with `focus`, only `current`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: [`whenM`](https://oss.sonatype.org/service/local/repositories/releases/archive/org/scalaz/scalaz_2.11/7.2.5/scalaz_2.11-7.2.5-javadoc.jar/!/index.html#scalaz.Applicative@whenM[A](cond:Boolean)(f:=>F[A]):F[Unit]) is a `Scalaz` function.

I could use explicit imports for `scalaz`, as a code reader I acknowledge is way better than wildcard imports. I think it wouldn't take me so much time now that I'm more familiar with the library. Do you think it's worth it?

cquiroz: I think it is fine as it is. This is mostly my lack of knowledge about Scalaz

jdnavarro:
I'll submit a new PR that, in addition to taking care of @cquiroz comments and fixing the build, will include SEQNG-20. I don't think it's worth merging this until we make sure SEQNG-20 is completed.

PR #95, opened 2016-10-07
Git ignore `buildinfo.properties`

jdnavarro:
I'm not sure if for this commit it's worth making a PR, but I'm doing so just in case.


jdnavarro:
My local `buildinfo.properties` is at `modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/buildinfo.properties` and it gets ignored.

PR #97, opened 2016-10-07
SEQNG-19: Coded the types described in the Interface Definition doc.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  case class SequenceView (
+    status: SequenceState,
+    steps: List[Step],
+    willStopIn: Option[Int]
+  )
+
+  // Log message types
+  type Time = java.time.Instant
```
cquiroz: I wonder if we need a timezone, However, let's leave it as it is since there is no tz support on scala.js yet

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -0,0 +1,98 @@
...
+    object Skipped extends StepState
+    case class Error(msg: String) extends StepState
+    object Running extends StepState
+  }
+
+  trait ActionStatus
+  object ActionStatus {
+    object Pending
+    object Completed
+    case class Running(progress: Double)
```
jluhrs: This type is for the Step internal actions. It is a Double that indicates level of completion, such that 0 <= x <= 1.0

cquiroz: So would it be composed of e.g. TCS setup, instrument setup, etc? How often is the completion updated?

cquiroz:
Looks fine as a starting point, but I'd rather not have `???` in merged code

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: Instant is supposed to be absolute.

jluhrs:
It will take at least a couple of sprints to implement all those functions. So either we merge the code with ??? now and fill them in following PRs, or wait 2 or three weeks and then do a big merge. I support the former.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: Currently the configuration will be 0.0/1.0, but it would be nice to have some updates in between. Observation progress is updated every second.

jdnavarro:
I'd be fine with merging  this with `gemini-hlsw/feature/engine`, but I'd make sure the `???` are gone before merging with `develop`.

cquiroz:
I support @jdnavarro idea, we don't merge `???` into develop, in the interim is fine. We had a production bug  a while ago because we had a `???` in the code

jdnavarro:
We could make the travis build fail if there is any `???` in the code. I'm trying to find a compiler flag to error if there are `???` but can't find anything similar.

A quick and dirty solution would be to add a script to the travis build grepping the code for any `???` left.

PR #98, opened 2016-10-13
SEQNG-17 SEQNG-20 New internal Engine state and state output

jdnavarro:
This includes the new triple Zipper structure for the Engine state and the generation of a minimal output from the engine state needed by the web server.

Previous PR #94 included a review for the Zipper structure. Since then, there has been some changes, including feedback from the review, but the fundamentals are the same.


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/execution.scala
```diff
@@ -0,0 +1,78 @@
...
+case class Current(execution: Execution[Action \/ Result]) {
+
+  val isEmpty: Boolean = execution.isEmpty
+
+  val actions: List[Action] = {
+
+    def lefts[L, R](xs: List[L \/ R]): List[L] = xs.collect { case -\/(l) => l }
+
+    lefts(execution.toList)
+
```
cquiroz: This is just silly me, but the last blank line looks weird

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/execution.scala
```diff
@@ -0,0 +1,78 @@
...
+  val empty: Current = Current(Nil)
+
+  /**
+    * Make an `Execution` `Current` only if all the `Action`s in the execution
+    * are pending.
+    *
+    */
+  def currentify(exe: Execution[Action]): Option[Current] =
+    (!exe.isEmpty).option(Current(exe.map(_.left)))
```
cquiroz: `exe.nonEmpty`?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -26,7 +26,7 @@ package object engine {
     * without interruption. A *sequential* `Execution` can be represented with
     * an `Execution` with a single `Action`.
     */
-  type Execution[A] = NonEmptyList[A]
+  type Execution[A] = List[A]
```
cquiroz: Is this change because we may have empty executions?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -41,58 +41,48 @@ package object engine {
...
     *
     * It also takes care of initiating the execution when transitioning to
-    * `Running` status.
+    * `Running` `Status`.
     */
   def switch(q: EventQueue)(st: Status): Engine[QState] =
     modify(QState.status.set(_, st)) *>
-    whenM (st == Status.Running) (prime *> execute(q)) *>
+    // TODO: Make Status an Equal instance
+    whenM (st == Status.Running) (execute(q)) *>
```
cquiroz: I think this maybe clearer if the `whenM ... *> get` would be on  the same line

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/queue.scala
```diff
@@ -0,0 +1,106 @@
+package edu.gemini.seqexec.engine
+
+import scalaz._
+import Scalaz._
+
+/**
+  * A list of `Sequence`s.
+  */
+case class Queue[+A](sequences: List[Sequence[A]])
```
cquiroz: Why's this covariant?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/state.scala
```diff
@@ -0,0 +1,155 @@
+package edu.gemini.seqexec.engine
+
+import scalaz._
+
+/**
+ * Flag to indicate whether the global execution is `Running` or `Waiting`.
+ */
+sealed trait Status
+
+object Status {
```
cquiroz: Is this status only for for the global state or will it be used at other levels like sequence?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/state.scala
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+  * This is the `State` in Zipper mode, which means is under execution.
+  *
+  */
+case class QStateZ(zipper: QueueZ, status: Status) extends QState { self =>
+
+  val next: Option[QState] = zipper.next match {
+    // Last execution
+    case None => zipper.uncurrentify.map(QStateF(_, status))
```
cquiroz: Often we align the `=>` for pattern matching choices

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/state.scala
```diff
@@ -0,0 +1,155 @@
...
+    */
+  val current: Current =
+    // Sequence
+    zipper
+      // Seqeunce
+      .focus
+      // Step
+      .focus
+      // Execution
+      .focus
```
cquiroz: Nice commenting 👍

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/execution.scala
jdnavarro: I'm still hesitating about whether to leave the blank line or not at the end of the function, I guess your opinion sorts it out :)

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: The reason for replacing `nonEmptyList`s by `List`s all over the code is that we should be able to carry out an actual execution independently at each level of the state hierarchy (Queue -> Sequence -> Step), so the empty case represents the finalization.

However in the case of `Execution` it should still make sense to keep it as `NonEmptyList` since it only can move to the next execution when all actions are completed, not when it's empty. I'll try to change it back.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: Actually, I wanted to get rid of all those `get`s in all the functions of that file. They all end in `*> get` so there should be a way to factor that out. I'll try to take of care of it now for this review.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/queue.scala
jdnavarro: I spent one morning pulling my hair about why I couldn't make `Queue`/`Sequence`/`Step` *Functor* instances when all types seem fine.  I still don't understand how Scala subtyping exactly works but it turns out, that in order to make it type check, I had to make that type parameter *covariant*.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/state.scala
jdnavarro: For now, I could get away with a single status type at every level but I think we would soon have the need for different types of status at each level.

jdnavarro:
It continues in #99

PR #99, opened 2016-10-14
SEQNG-21 SEQNG-17 SEQNG-20 New internal Engine state and Process of `SequenceView`s

jdnavarro:
This includes the new triple Zipper structure for the Engine state and the generation of a minimal output from the engine state needed by the web server.

Previous PR #94 included a review for the Zipper structure. Since then, there has been some changes, including feedback from the review, but the fundamentals are the same.

It also includes the `Process` generates the output states needed by the web server.

~~The commits from #98 are here because SEQNG-21 depends on them. Please, review #98 before this one, so that I can merge it and only the specific commits for SEQNG-21 show up in this PR.~~


cquiroz:
I think this should be much easier to review after you merge #98, Otherwise could you indicate the relevant commits?

jdnavarro:
It's this [one](https://github.com/gemini-hlsw/ocs3/pull/99/commits/887cb16b501083d0af7545ad8406efb3a2effe07) only. But you could wait for the other PR to be merged if you prefer.

cquiroz:
ok, I'll wait for the rebase

jdnavarro:
I pushed the commits taking care of the comments of #98 here.

@cquiroz, sorry for the mess with the PRs, I should have combined the 3 tickets into one 1 PR, they don't make much sense in separate PRs. I'll try to make it easier to review next time.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Handler.scala
```diff
@@ -12,28 +12,45 @@ object Handler {
...
     }
 
-    receive(q) >>= (
+    receive(q).flatMap(
       ev => Process.eval(
-        ev match {
-          case EventUser(ue) => handleUserEvent(ue)
-          case EventSystem(se) => handleSystemEvent(se)
-        }
+        (ev match {
```
cquiroz: Is this splitting incoming requests to see if they are user o system? Could it make sense to have different streams instead?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -41,58 +43,47 @@ package object engine {
     * Changes the `Status` and returns the new `QState`.
```
cquiroz: I guess this doesn't return `Qstate` anymore

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -82,6 +83,59 @@ object SeqexecEngine {
     willStopIn: Option[Int]
   )
 
+  object SequenceView {
+
+    import edu.gemini.seqexec.engine
+
+    // TODO: Better name and move it to `engine`
+    type QueueAR = engine.Queue[engine.Action \/ engine.Result]
```
cquiroz: What `AR` stands for?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -82,6 +83,59 @@ object SeqexecEngine {
...
+      def statusStep(step: StepAR): StepState =
+        engine.Step.status(step) match {
+          case engine.Status.Waiting   => StepState.Pending
+          case engine.Status.Completed => StepState.Completed
+          case engine.Status.Running   => StepState.Running
+        }
+
+      def viewStep(step: StepAR): Step =
+        StandardStep(
+          // TODO: Add configuration parameter to Engine Step
```
cquiroz: What's inside this configuration?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/execution.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import scalaz._
+import Scalaz._
+
+/**
+  * This structure holds the current `Execution` under execution. It carries
+  * information about which `Action`s have been completed.
+  *
+  */
+case class Current(execution: Execution[Action \/ Result]) {
```
cquiroz: We discussed this a bit on the chat. The type `Action \/ Result` seems to me like `Error \/ Ok` while in fact errors will occur in `Result`. Probably doesn't matter but maybe worth some discussion

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Handler.scala
jdnavarro: From what I understood from @jluhrs both types of events will come from different `Process`es but at some point they will have to be merged into one.

The key to achieve consistent concurrency in spite of the asynchronicity of its actions is that in the end the `Engine` only needs to handle a single stream of input events.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: Yeah, the name `QState` has to go away soon.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: It's a poor mnemonic for `Action/Result` but we have to rename it. I already opened a ticket to do a general cohesive renaming because this is just one case of several poor naming choices.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/execution.scala
jdnavarro: I acknowledge it's fair to expect the left part to be some kind of error or failure. I was piggybacking on some functions the `\/` API provides but it's probably worth making this type a distinct one.

cquiroz:
Looks good to me. Have you tried running all the tests. I wonder if travis may complain about `feature/engine` but I guess it is still fairly independent

jdnavarro:
All the tests pass on my machine 💻 . `feature/engine` is already rebased onto `develop` so it should have the latest versions of all the tests overall.

cquiroz:
👍

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/execution.scala
jdnavarro: I'll try take care of this when doing the ticket for the renames.

PR #100, opened 2016-10-18
SEGNG-34 Attach the event with each generated Engine state output

jdnavarro:
With these changes there is now a Process of `SeqexecEvent`s that in theory could be used directly by the client to display the state of execution.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -21,13 +21,40 @@ object SeqexecEngine {
 
   sealed trait SeqexecEvent
   object SeqexecEvent {
-    case class SequenceStart(view: SequenceView) extends SeqexecEvent
-    case class StepExecuted(view: SequenceView) extends SeqexecEvent
-    case class SequenceCompleted(view: SequenceView) extends SeqexecEvent
-    case class StepBreakpointChanged(view: SequenceView) extends SeqexecEvent
-    case class StepSkipMarkChanged(view: SequenceView) extends SeqexecEvent
-    case class SequencePauseRequested(view: SequenceView) extends SeqexecEvent
-    case class NewLogMessage(msg: LogMsg) extends SeqexecEvent
+
+    import edu.gemini.seqexec.engine
+    import edu.gemini.seqexec.engine.Event
+
+    case class SequenceStart(view: List[SequenceView]) extends SeqexecEvent
+    case class StepExecuted(view: List[SequenceView]) extends SeqexecEvent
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -21,13 +21,40 @@ object SeqexecEngine {
...
   object SeqexecEvent {
-    case class SequenceStart(view: SequenceView) extends SeqexecEvent
-    case class StepExecuted(view: SequenceView) extends SeqexecEvent
-    case class SequenceCompleted(view: SequenceView) extends SeqexecEvent
-    case class StepBreakpointChanged(view: SequenceView) extends SeqexecEvent
-    case class StepSkipMarkChanged(view: SequenceView) extends SeqexecEvent
-    case class SequencePauseRequested(view: SequenceView) extends SeqexecEvent
-    case class NewLogMessage(msg: LogMsg) extends SeqexecEvent
+
+    import edu.gemini.seqexec.engine
```
cquiroz: I guess I often put all the imports on the top unless there is some naming conflicts. it's a minor thing

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -21,13 +21,40 @@ object SeqexecEngine {
...
+    import edu.gemini.seqexec.engine.Event
+
+    case class SequenceStart(view: List[SequenceView]) extends SeqexecEvent
```
cquiroz: Perhaps we should put `List[SequenceView]` on its own type, we seems to be using it a lot. WDYT

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: The initial idea was to not pollute the `SeqexecEngine` namespace with similar names coming from the `Engine`. But it turns out that I ended up using qualified imports everywhere so it makes sense to import it from the top. I'll change it.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: This is *very* provisional. The idea here is to have something that can be used directly on the client as it is, so that we can show a quick demo. Isn't a `List[SequenceView]` what the client expects as state from the server?

The equivalent to `List[SequenceView]` has a distinct type named just `Queue`. But I was thinking that we could take care of this in the next iteration after watching the demo.

PR #102, opened 2016-10-19
Quickfixes to run simple demo

jdnavarro:


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -57,5 +59,25 @@ class SeqexecCommandRoutes(auth: AuthenticationService) extends BooPicklers {
...
     case req @ GET  -> Root  / obsId / "state" =>
       Ok(toSequenceConfig("state", commands.state(obsId)))
+
+    // New SeqexecEngine
+
+    // TODO: Add obsId parameter
+    case req @ POST -> Root / "start" =>
+      // TODO: Get rid of `.toString` How do we want to represent input results
+      // now?
+      Ok(SeqexecEngine.start(q).map(_.toString))
```
cquiroz: Does this return a process or a `Task`? In case we want to return a non-String we probably need to define an encoder compatible with http4s

jdnavarro: It returns a `Task[SeqexecFailure \/ Unit]`, so there needs to be a serializer for `SeqexecFailure` which I believe it was used in the previous `executor` implementation, so I'm not sure if we still have to modify `SeqexecFailure` for the client or if we need something else to represent errors in the client.

jdnavarro:
Merging after agreeing in today's sprint meeting, we should have these changes in `feature/engine`. We are thinking about merging into `develop` next week but @cquiroz may want to weight in before doing it.

PR #103, opened 2016-11-08
SEQNG-46 Initial engine unit tests

jdnavarro:
These are the initial unit tests intended to catch the bug causing extra executions to be reported. These tests still don't reveal the bug I was chasing, but since it's not critical, I'm moving it back to the backlog and will revisit it later.

Obviously there is plenty of room for improvement of these tests but I'm sending this PR now because I don't want to get bogged down writing tests when there are other tasks with more priority.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -67,8 +67,10 @@ package object engine {
     */
   def next(q: EventQueue): Engine[Unit] =
     gets(_.next).flatMap {
-      // No more Executions left
+      // Empty state
       case None     => send(q)(finished)
+      // Final State
+      case Some(qs@QStateF(_, _)) => put(qs) *> send(q)(finished)
       // Execution completed, execute next actions
       case Some(qs) => put(qs) *> execute(q)
     }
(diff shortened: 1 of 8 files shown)


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/CurrentSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+    assert(Current.currentify(List(action, action)).nonEmpty)
+    assert(Current.currentify(Nil).isEmpty)
+  }
+
+  "uncurrentify" should "be None when not all actions are completed" in {
+    assert(Current(Nil).uncurrentify.isEmpty)
+    assert(curr.uncurrentify.isEmpty)
+  }
+
+  "marking an index out of bonds" should "not modify the execution" in {
```
cquiroz: bounds?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -57,8 +57,10 @@ package object engine {
     */
   def next(q: EventQueue): Engine[Unit] =
     gets(_.next).flatMap {
-      // No more Executions left
+      // Empty state
       case None     => send(q)(finished)
+      // Final State
+      case Some(qs@QStateF(_, _)) => put(qs) *> send(q)(finished)
```
jdnavarro: I was fiddling with the syntax but couldn't find a way to do what you just said! Thanks for the tip.

PR #104, opened 2016-11-09
SEQNG-49 Load Sequence from from UI

jdnavarro:
Please make sure the load sequence function stub has the correct type signature. If that's the type signature loading the sequence from the UI is trivial.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -18,51 +18,12 @@ import edu.gemini.seqexec.model.SharedModel.SeqexecEvent._
   */
 object SeqexecEngine {
 
-  /**
-    * Emulates TCS configuration in the real world.
-    *
-    */
-  val configureTcs: Action  = for {
-    _ <- Task(println("System: Start TCS configuration"))
-    _ <- Task(Thread.sleep(2000))
-    _ <- Task(println ("System: Complete TCS configuration"))
-  } yield Result.OK(())
-
-  /**
-    * Emulates Instrument configuration in the real world.
-    *
-    */
-  val configureInst: Action  = for {
-    _ <- Task(println("System: Start Instrument configuration"))
-    _ <- Task(Thread.sleep(2000))
-    _ <- Task(println("System: Complete Instrument configuration"))
-  } yield Result.OK(())
-
-  /**
-    * Emulates an observation in the real world.
-    *
-    */
-  val observe: Action  = for {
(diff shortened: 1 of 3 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -103,7 +103,8 @@ class SeqexecUIApiRoutes(auth: AuthenticationService, q: engine.EventQueue) exte
           } yield (obsId, s)
 
           r match {
-            case \/-((i, s)) => Ok(List(Sequence(i.stringValue(), SequenceState.NotRunning, "F2", s.toSequenceSteps, None)))
+            case \/-((i, s)) => SeqexecEngine.load(i) *>
```
cquiroz: I'm not sure what the `load` call does. I'd think you want the output to be used in the response, though it is a `Unit`

jdnavarro: @jluhrs can correct me, but from what I understood, the idea for `load` is to set the initial state in the engine which will send the first output through the websocket.

cquiroz: I think the idea is correct but somewhat I'd expect `load` to do both. read the sequence from the odb and build the internal sequence, so we get away of `ExecutorImpl` altogether

cquiroz: The original design han  the load response only going through plain HTTP. Why would we want it on the socket too?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -78,6 +78,9 @@ object SeqexecEngine {
     )
   )
 
+  // Stub for Sequence loading.
+  def load(oid: SPObservationID): Task[SeqexecFailure \/ Unit] = ???
```
jluhrs: It should be
```
def load(q: engine.EventQueue, seqId: SPObservationID): Task[SeqexecFailure \/ Unit]
```

PR #107, opened 2016-11-15
Keep engine running after finalizing

jdnavarro:
This was necessary in order to reload extra sequences after a sequence was finished. We want to keep the Engine running along with the web server.

I don't know if it's worth sending a PR for this kind of *microcommits*, I could directly push it to `feature/engine` if you prefer.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/HandlerSpec.scala
+++ b/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/HandlerSpec.scala
@@ -103,7 +103,11 @@ class HandlerSpec extends FlatSpec {
     val q = async.boundedQueue[Event](10)
     intercept[Cause.Terminated](
       Nondeterminism[Task].both(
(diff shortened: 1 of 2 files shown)


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/HandlerSpec.scala
```diff
@@ -103,7 +103,11 @@ class HandlerSpec extends FlatSpec {
     val q = async.boundedQueue[Event](10)
     intercept[Cause.Terminated](
       Nondeterminism[Task].both(
-        q.enqueueOne(start),
+        List(
+          q.enqueueOne(start),
+          Task(Thread.sleep(5000)),
+          q.enqueueOne(exit)
```
cquiroz: What's exit? Is an event produced at the end of a queue?

cquiroz:
I'd prefer to have PRs for micro commits, at least for develop. But since you are commit to `feature/engine` it make be ok. What does @jluhrs thinks?

PR #108, opened 2016-11-16
SEQNG-52: Implementation of engine startup options

jluhrs:
Read engine startup settings from configuration file app.cfg. The settings are then applied to the Seqexec engine. The implementation of the settings themselves is a bit dirty (uses mutability), but I think it is OK, has they are not supposed to change in run-time.
I also moved the code that configures the Seqexec engine inside ```SeqexecEngine```

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -1,5 +1,7 @@
 package edu.gemini.seqexec.server
 
+import java.time.LocalDate
+
 import edu.gemini.seqexec.engine.{Action, Result, Sequence, Step}
 import edu.gemini.seqexec.server.SeqexecFailure.UnrecognizedInstrument
 import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
@@ -8,11 +10,21 @@ import edu.gemini.spModel.seqcomp.SeqConfigNames._
 
 import scalaz.Scalaz._
 import scalaz._
+import scalaz.concurrent.Task
 
 /**
   * Created by jluhrs on 9/14/16.
   */
 object SeqTranslate {
+
+  case class Settings(dhsSim: Boolean,
+                    tcsSim: Boolean,
+                    instSim: Boolean,
+                    gcalSim: Boolean)
+
+  private var settings = Settings(false, false, false, false)
+  private val dhsSimulator = DhsClientSim(LocalDate.now)
+  def setConfig(c: Settings): Unit = { settings = c }
 
   implicit def toAction[A](x: SeqAction[A]): Action = x.run map {
     case -\/(e) => Result.Error(e)
@@ -24,12 +36,12 @@ object SeqTranslate {
     val instName = config.getItemValue(new ItemKey(INSTRUMENT_KEY, INSTRUMENT_NAME_PROP))
     val instrument = instName match {
       case GmosSouth.name => Some(GmosSouth)
-      case Flamingos2.name => Some(Flamingos2(Flamingos2ControllerSim))
+      case Flamingos2.name => Some(Flamingos2(if(settings.instSim) Flamingos2ControllerSim else Flamingos2ControllerEpics))
       case _ => None
     }
 
     instrument.map { a =>
-      val systems = List(Tcs(TcsControllerSim), a)
+      val systems = List(Tcs(if(settings.tcsSim) TcsControllerSim else TcsControllerEpics), a)
       // TODO Find a proper way to inject the subsystems
       Step[Action](
         i,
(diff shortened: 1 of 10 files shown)


cquiroz:
Could we make them immutable? In practice it is mostly converting the `object` to `class` and pass the settings as constructor parameters. You can see that on the configuration of the web app classes

jdnavarro:
In general, what sort of code should go into the engine module vs the server module (not the web server module)? I still can't see a clearcut criterion.

What you mean by immutable? Is it doing unsafe IO?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -120,8 +102,8 @@ object WebServerLauncher extends ServerApp with LogInitialization {
       _  <- configLog
       ac <- authConf
       wc <- serverConf
-      sc <- executorConf
-      _  <- seqexecExecutor.run(sc)
+      c <- config
+      _  <- SeqexecEngine.seqexecConfigure.run(c)
```
jdnavarro: Why not follow the existing pattern of configuration? Previously there was not need to do in-place mutations and wrap them into a `Task`. When running this you are just unwrapping the `Task` you just introduced in `seqexecConfigure`, but under the hood you already did an unsafe mutation outside of a `Task`.

I'm not saying this is outright wrong because maybe there is no way to around it, but I'm trying to understand what was the problem with the previous configuration pattern.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -116,4 +118,22 @@ object SeqexecEngine {
...
+
+  def seqexecConfigure: Kleisli[Task, Config, Unit] = Kleisli { cfg: Config => {
+      val odbHost = cfg.require[String]("seqexec-engine.odb")
+      val dhsServer = cfg.require[String]("seqexec-engine.dhsServer")
+      val dhsSim = cfg.require[Boolean]("seqexec-engine.dhsSim")
+      val tcsSim = cfg.require[Boolean]("seqexec-engine.tcsSim")
+      val instSim = cfg.require[Boolean]("seqexec-engine.instSim")
+      val gcalSim = cfg.require[Boolean]("seqexec-engine.gcalSim")
+
+      Task.delay(ODBProxy.host(new Peer(odbHost, 8443, null))) *>
```
cquiroz: I guess this could be something like new `SeqTranslate(setting, dhsClient, odbProxy)` and return it on the `Kleisli`

jluhrs:
About the Execution vs Server separation, I have being trying to keep the logic about running a sequence in Engine, while keeping the interaction with external systems (like the ODB) in Server

jluhrs:
I think I addressed all the concerns with the last update. Instead of using objects with vars for singletons that needed initialization parameters, I use classes and pass an instance around.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -8,47 +10,56 @@ import edu.gemini.spModel.seqcomp.SeqConfigNames._
...
   */
 object SeqTranslate {
 
+  case class Systems(
+                    dhs: DhsClient,
+                    tcs: TcsController,
+                    flamingos2: Flamingos2Controller
+                    )
+
+  private val dhsSimulator = DhsClientSim(LocalDate.now)
```
cquiroz: @jdnavarro I guess this also falls into a side effect?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -120,11 +102,11 @@ object WebServerLauncher extends ServerApp with LogInitialization {
       _  <- configLog
       ac <- authConf
       wc <- serverConf
-      sc <- executorConf
-      _  <- seqexecExecutor.run(sc)
+      c <- config
```
cquiroz: Just a nitpick, could we vertically align the for element?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jdnavarro: Yes, but it might not be worth introducing a `Task` with this PR. With a `TODO` comment to not forget about it in the future would be good enough.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -10,13 +10,23 @@ import scalaz._
...
+import knobs._
 import edu.gemini.seqexec.model.SharedModel._
 import edu.gemini.seqexec.model.SharedModel.SeqexecEvent._
+import edu.gemini.spModel.core.Peer
 
 /**
   * Created by jluhrs on 10/7/16.
   */
-object SeqexecEngine {
+class SeqexecEngine(settings: SeqexecEngine.Settings) {
```
jdnavarro: 👍

PR #110, opened 2016-11-25
Move DhsClientHttp inner classes to the DhsClientHttp object

cquiroz:
This is a small suggestion for a change. It seems strange to have inner classes defined inside another class. Normally you'd put them in the companion object.

I got a compiler warning due to the original code:

```
[warn] /Users/cquiroz/Projects/ocs3/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala:116: The outer reference in this type test cannot be checked at run time.
[warn]   final case class Error(t: ErrorType, msg: String) {
[warn]                    ^
[warn] one warning found
```

What do you think?


swalker2m:
Agree 👍

jdnavarro:
👍

PR #113, opened 2016-11-28
SEGNG-58A: Replace Sequence in the client with SequenceView

cquiroz:
This is the first PR towards SEQNG-58A. It skips the direct response when loading the sequence instead relying on the event through the websocket

It also adds a metadata element to the sequence that should contain extra data in the future

This PR is mostly to get comments about the approach taken to further integrate the new engine into the UI

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
@@ -1,16 +1,16 @@
 package edu.gemini.seqexec.engine
 
+import edu.gemini.seqexec.model.SharedModel.SequenceMetadata
+
 import scalaz._
 import Scalaz._
 
 /**
   * A list of `Step`s grouped by target and instrument.
   */
-case class Sequence[+A](id: String, steps: List[Step[A]])
+case class Sequence[+A](id: String, metadata: SequenceMetadata, steps: List[Step[A]])
 
 object Sequence {
-
-  def empty[A](id: String): Sequence[A] = Sequence(id, Nil)
 
   /**
     * Calculate the `Sequence` `Status` based on the underlying `Action`s.
@@ -23,7 +23,7 @@ object Sequence {
 
   implicit val SequenceFunctor = new Functor[Sequence] {
     def map[A, B](fa: Sequence[A])(f: A => B): Sequence[B] =
-      Sequence(fa.id, fa.steps.map(_.map(f)))
+      Sequence(fa.id, fa.metadata, fa.steps.map(_.map(f)))
   }
 
   // TODO: Proof Foldable laws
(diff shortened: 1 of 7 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -1,17 +1,17 @@
...
+import edu.gemini.seqexec.model.SharedModel.SequenceMetadata
+
 import scalaz._
 import Scalaz._
 
 /**
   * A list of `Step`s grouped by target and instrument.
   */
-case class Sequence[+A](id: String, steps: List[Step[A]])
+case class Sequence[+A](id: String, metadata: SequenceMetadata, steps: List[Step[A]])
```
cquiroz: Right, I also toyed with the idea of making metadata just a `B` type, but that seems to cause trouble on the functor definition

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -97,14 +97,13 @@ class SeqexecUIApiRoutes(auth: AuthenticationService, q: engine.EventQueue, se:
         val user = userInRequest(req)
         user.fold(Unauthorized(Challenge("jwt", "seqexec"))) { _ =>
           val r = for {
-            obsId <- \/.fromTryCatchNonFatal(new SPObservationID(oid)).leftMap((t:Throwable) => Unexpected(t.getMessage))
-            s     <- se.odbProxy.read(obsId)
+            obsId <- EitherT[Task, SeqexecFailure, SPObservationID](Task.delay(\/.fromTryCatchNonFatal(new SPObservationID(oid)).leftMap((t:Throwable) => Unexpected(t.getMessage))))
```
cquiroz: I'm not that happy with the `EitherT` but there is no good way around. perhaps going `Free` would be an option

jdnavarro: In this case the `EitherT` is shortlived so I don't think it will create any problems now, but if we ever we need to combine it with another monad transformer it can get hairy.

For now I think this is fine.

cquiroz: I think it should be possible with a bit of transformations use directly `Task`. I'll look how to do it before merging

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
jluhrs: I have a reason to keep it. Commands are specific to sequences, identified by its unique id. We need to associate ids and sequences in a way that is accessible to the execution engine to find the state of the sequence we are trying to start/stop (i.e. it cannot be hidden inside metadata).

PR #114, opened 2016-11-29
SEQNG-60 Load first Sequence correctly #resolve

jdnavarro:
Following @cquiroz hints about this bug I found out there was a problem with operator preference. Now it should load the first *Sequence* fine, I tried 3 times in a row and worked on my machine 💻.

The `Process` execution is controlled implicitly by `http4s` and I'm pretty sure it creates a different `Process` per websocket. This didn't seem to affect the bug in this PR, but it will be an issue when different users run the Seqexec at the same time.

We will have to look into some kind of broadcast mechanism with `scalaz-stream` (*topics*?) in order to make sure all users run only one engine `Process`. So I'm filing a separate ticket for that case.

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
@@ -80,35 +80,50 @@ class SeqexecUIApiRoutes(auth: AuthenticationService, q: engine.EventQueue, se: 
 
   def userInRequest(req: Request) = req.attributes.get(JwtAuthentication.authenticatedUser).flatten
 
-  val protectedServices: HttpService = tokenAuthService { GZip { HttpService {
-      case req @ GET -> Root / "seqexec" / "events"         =>
-        // Stream seqexec events to clients and a ping
-        val user = userInRequest(req)
-        WS(Exchange(pingProcess ++ Process.emit(Binary(trimmedArray(SeqexecConnectionOpenEvent(user)))) merge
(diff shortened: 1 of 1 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -80,36 +80,44 @@ class SeqexecUIApiRoutes(auth: AuthenticationService, q: engine.EventQueue, se:
...
+  val protectedServices: HttpService =
+    tokenAuthService {
+      GZip {
+        HttpService {
+          case req @ GET -> Root / "seqexec" / "events"         =>
+            // Stream seqexec events to clients and a ping
+            val user = userInRequest(req)
+            WS(
+              Exchange(
+                Process.emit(Binary(trimmedArray(SeqexecConnectionOpenEvent(user)))) ++
```
cquiroz: Is the only change to do the `merge` later between `pingProcess` and `se.eventProcess`?

jdnavarro: Yes, and some extra parentheses to get the associations of the operators right. The rest is just indentation, I should have mentioned it when submitting the PR.

jdnavarro:
I did the rebase after the latest PR from @cquiroz. Since there was a merge conflict, please, watch out for anything that I may reverted back accidentally.

Now when loading a Sequence there is an error in the UI reporting `Got an error during search`, but is this expected? By looking at the web console it looks like it's working correctly.

cquiroz:
I think it is an expected error. It cannot understand the response yet

PR #115, opened 2016-12-02
Version bump

cquiroz:
This PR updates the version of several of our libraries. In particular, scala.js 0.6.13 is going to be important moving forward

Another small update is a change on travis to only build the distribution packages for non-PR builds

The changes reduce time from app 13m to app 8m

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,5 @@ before_script:
   - sudo apt-get install -y rpm
 script:
   - sbt ++$TRAVIS_SCALA_VERSION test:compile test
-  - sbt ++$TRAVIS_SCALA_VERSION seqexec_server/universal:packageZipTarball
-  - sbt ++$TRAVIS_SCALA_VERSION seqexec_server_test_l64/rpm:packageBin
+  - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then sbt ++$TRAVIS_SCALA_VERSION seqexec_server/universal:packageZipTarball; fi'
+  - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then sbt ++$TRAVIS_SCALA_VERSION seqexec_server_test_l64/rpm:packageBin; fi'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,15 @@ scala:
   - 2.11.8
 jdk:
   - oraclejdk8
+# These directories are cached to S3 at the end of the build
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot/
+before_cache:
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -3,6 +3,15 @@ scala:
   - 2.11.8
 jdk:
   - oraclejdk8
+# These directories are cached to S3 at the end of the build
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot/
```
jdnavarro: I did some benchmarks doing precisely that some months ago but didn't notice any difference in build time (I don't remember if it was for OCS or OCS3 though). In any case this is very useful when there are download timeouts 👍 .

project/Settings.scala
```diff
@@ -29,14 +29,14 @@ object Settings {
...
     val uPickle      = "0.4.0"
-    val booPickle    = "1.2.4"
-    val diode        = "1.0.0"
-    val javaTimeJS   = "0.1.0"
+    val booPickle    = "1.2.5"
+    val diode        = "1.1.0"
+    val javaTimeJS   = "0.2.0"
     val javaLogJS    = "0.1.0"
-    val scalaJQuery  = "1.0-RC6"
+    val scalaJQuery  = "1.0"
```
jdnavarro: In case I want to version bump myself in the future, how do you pick the versions? Just look at the latest ones and check that they build correctly together?

project/Settings.scala
```diff
@@ -50,7 +50,7 @@ object Settings {
     val unboundId    = "3.1.1"
     val jwt          = "0.8.1"
     val slf4j        = "1.7.21"
-    val knobs        = "3.8.1a"
+    val knobs        = "3.12.27a"
```
jdnavarro: Did you try `http4s-0.15.0`? It was released yesterday but I was wondering whether the build would break with the that version.

project/SeqexecEngineModules.scala
```diff
@@ -29,10 +29,6 @@ trait SeqexecEngineModules {
     .settings(
       libraryDependencies ++= Seq(BooPickle.value) ++ TestLibs.value
     )
-    .jsSettings(
-      scalaJSUseRhino := false,
```
jdnavarro: I thought Rhino was dead long time ago 😄, but TIL it's still quite active. Is scalajs using node by default now?

project/Settings.scala
cquiroz: That is going to be a bigger update, there are only client side updates. http4s update probably deserve its own PR

project/Settings.scala
cquiroz: Pretty much 😄

PR #116, opened 2016-12-02
SEQNG-58B: Replace Sequence in the client with SequenceView

cquiroz:
This is the next PR towards SEQNG-58. It confirms that events and the model can be performed and adds tests using `scalacheck-shapeless`

It also restores the ability to decode the user sent over the `websocket` channel that essentially makes the login to persist across browser reloads (Authentication still has a time to live of 8h)


modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/SharedModel.scala
```diff
@@ -47,7 +48,7 @@ object SharedModel {
     case class Running(progress: Double) extends ActionStatus
   }
 
-  trait Step {
+  sealed trait Step {
```
cquiroz: `sealed` is need for `scalacheck-shapeless`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -89,7 +89,7 @@ class SeqexecUIApiRoutes(auth: AuthenticationService, q: engine.EventQueue, se:
             val user = userInRequest(req)
             WS(
               Exchange(
-                Process.emit(Binary(trimmedArray(SeqexecConnectionOpenEvent(user)))) ++
+                Process.emit(Binary(newTrimmedArray(ConnectionOpenEvent(user)))) ++
```
cquiroz: This is the key change that will send proper authentication to the client upon connection

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -104,25 +104,16 @@ class SeqexecUIApiRoutes(auth: AuthenticationService, q: engine.EventQueue, se:
...
-                  )
-                )
-                s     <- se.load(q, obsId)
-              } yield (obsId, s)
-
-              r.run >>= {
-                case -\/(e)      => NotFound(SeqexecFailure.explain(e))
-                case \/-((i, _)) => Ok(s"Loaded sequence $i")
-              }
+              for {
```
cquiroz: Note that `EitherT` was removed and the code in general is much simpler now

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/SharedModel.scala
jdnavarro: Are open traits ever used as when doing idiomatic FP in Scala?

It seems to me that the idiomatic way to use sealed traits is when there are a finite number of case classes like here. Otherwise a *type class* pattern is used (like in `simulacrum`). Is my assumption correct?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
```diff
@@ -232,7 +233,7 @@ class WebSocketHandler[M](modelRW: ModelRW[M, WebSocketConnection]) extends Acti
     def onMessage(e: MessageEvent): Unit = {
       val byteBuffer = TypedArrayBuffer.wrap(e.data.asInstanceOf[ArrayBuffer])
       \/.fromTryCatchNonFatal(Unpickle[SeqexecEvent].fromBytes(byteBuffer)) match {
-        case \/-(event) => println(s"Decoding event $event") // SeqexecCircuit.dispatch(NewSeqexecEvent(event))
+        case \/-(event) => SeqexecCircuit.dispatch(NewSeqexecEvent(event))
```
jdnavarro: :facepalm: I completely forgot about that.

cquiroz: The `sealed` comes as scalacheck-shapeless cannot derive an arbitrary if it doesn't know all the possible implementations. You can use normal traits in many cases, e.g. when defining a simple interface

PR #117, opened 2016-12-05
SEQNG-66 Broadcast engine output to each connected websocket

jdnavarro:
Now the engine is launchedonly once when starting the server and uses `scalaz-streams` *topic*s to broadcast the output to every client. I implemented this similar to how was in a previous version of the executor.

I tested it with Firefox and Chrome at the same time.

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
@@ -233,7 +233,7 @@ class WebSocketHandler[M](modelRW: ModelRW[M, WebSocketConnection]) extends Acti
     def onMessage(e: MessageEvent): Unit = {
       val byteBuffer = TypedArrayBuffer.wrap(e.data.asInstanceOf[ArrayBuffer])
       \/.fromTryCatchNonFatal(Unpickle[SeqexecEvent].fromBytes(byteBuffer)) match {
-        case \/-(event) => SeqexecCircuit.dispatch(NewSeqexecEvent(event))
+        case \/-(event) => println(s"Decoding event: $event"); SeqexecCircuit.dispatch(NewSeqexecEvent(event))
         case -\/(t)     => println(s"Error decoding event ${t.getMessage}")
       }
     }
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
@@ -1,6 +1,5 @@
 package edu.gemini.seqexec.web.server.http4s
 
-import edu.gemini.seqexec.model.SharedModel.SeqexecEvent
 import edu.gemini.seqexec.server.Commands
 import edu.gemini.seqexec.server.SeqexecEngine
 import edu.gemini.seqexec.engine
@@ -10,18 +9,15 @@ import edu.gemini.seqexec.web.server.security.AuthenticationService
 import org.http4s._
 import org.http4s.dsl._
 import org.http4s.server.middleware.GZip
-import scalaz.stream.async.mutable.Topic
 
 /**
   * Rest Endpoints under the /api route
   */
-class SeqexecCommandRoutes(auth: AuthenticationService, events: (engine.EventQueue, Topic[SeqexecEvent]), se: SeqexecEngine) extends BooEncoders {
+class SeqexecCommandRoutes(auth: AuthenticationService, inputQueue: engine.EventQueue, se: SeqexecEngine) extends BooEncoders {
 
   val tokenAuthService = JwtAuthentication(auth)
 
   val commands = Commands(se.odbProxy)
-
-  val inputQueue = events._1
 
   val service = tokenAuthService { GZip { HttpService {
     case req @ GET  -> Root  / "host" =>
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
@@ -1,5 +1,6 @@
 package edu.gemini.seqexec.web.server.http4s
 
+import edu.gemini.seqexec.model.SharedModel.SeqexecEvent
 import edu.gemini.seqexec.server.Commands
 import edu.gemini.seqexec.server.SeqexecEngine
 import edu.gemini.seqexec.engine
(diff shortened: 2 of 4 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
```diff
@@ -233,7 +233,7 @@ class WebSocketHandler[M](modelRW: ModelRW[M, WebSocketConnection]) extends Acti
     def onMessage(e: MessageEvent): Unit = {
       val byteBuffer = TypedArrayBuffer.wrap(e.data.asInstanceOf[ArrayBuffer])
       \/.fromTryCatchNonFatal(Unpickle[SeqexecEvent].fromBytes(byteBuffer)) match {
-        case \/-(event) => SeqexecCircuit.dispatch(NewSeqexecEvent(event))
+        case \/-(event) => println(s"Decoding event: $event"); SeqexecCircuit.dispatch(NewSeqexecEvent(event))
```
jdnavarro: I had to restore this in order to watch the events in the web console, I can delete it after this review.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -81,31 +83,44 @@ object WebServerLauncher extends ServerApp with LogInitialization {
...
-      q    = async.boundedQueue[engine.Event](10)
-      ws   <- webServer(as, q, SeqexecEngine(seqc)).run(wc)
-    } yield ws
+    } yield SeqexecEngine(seqc)
+
+    val inq  = async.boundedQueue[engine.Event](10)
+    val out  = async.topic[SeqexecEvent]()
+
+    engineTask flatMap (
+      se => Nondeterminism[Task].both(
```
jdnavarro: I needed to do this in order to force 2 separate threads between the web server and the engine. I couldn't find out how to make the engine run in children thread without blocking the web server. I tried with `Task.force` and `Task.async`.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -9,16 +10,19 @@ import edu.gemini.seqexec.web.server.security.AuthenticationService
...
   * Rest Endpoints under the /api route
   */
-class SeqexecCommandRoutes(auth: AuthenticationService, q: engine.EventQueue, se: SeqexecEngine) extends BooEncoders {
+class SeqexecCommandRoutes(auth: AuthenticationService, events: (engine.EventQueue, Topic[SeqexecEvent]), se: SeqexecEngine) extends BooEncoders {
 
   val tokenAuthService = JwtAuthentication(auth)
 
   val commands = Commands(se.odbProxy)
 
+  val inq = events._1
```
cquiroz: inputQueue?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
cquiroz: Did you try `Task.fork`?

jdnavarro: Sorry, I meant `Task.fork` not `Task.force`.

cquiroz: That maybe related to the `ExecutorService` being used. It maybe that you've exhausted the thread counts so it cannot `fork` another one.

We've discovered that in general is not a good idea to rely on the default one as it uses 2*N (N= processors count) and thus gets exhausted sometimes

jdnavarro: That makes sense, you know if it's possible to configure the `ExecutorService` for http4s?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -9,16 +10,19 @@ import edu.gemini.seqexec.web.server.security.AuthenticationService
 import org.http4s._
 import org.http4s.dsl._
 import org.http4s.server.middleware.GZip
+import scalaz.stream.async.mutable.Topic
 
 /**
   * Rest Endpoints under the /api route
   */
-class SeqexecCommandRoutes(auth: AuthenticationService, q: engine.EventQueue, se: SeqexecEngine) extends BooEncoders {
+class SeqexecCommandRoutes(auth: AuthenticationService, events: (engine.EventQueue, Topic[SeqexecEvent]), se: SeqexecEngine) extends BooEncoders {
```
jluhrs: `events._2` is not used. Did you include it to keep the signature of `SeqexecCommandRoutes` the same as `SeqexecUIApiRoutes` ?

jdnavarro: Well, at some point I think all those extra parameters in the routes should be included in the *http4s* service, so in order to make refactoring easier in the future I was trying to keep adding the same parameters in both endpoints.

But a TODO comment as a reminder should work too, so I'll remove the parameter.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
cquiroz: I don't think so. You probably can create one on the WebServerLauncher and use it there

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
cquiroz: Should the event processes belong to the `SeqexecEngine` parameter?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
jdnavarro: I need to look into how the `ExecutorService` works. I'll leave the changes for another PR though because it'll take some time.

PR #118, opened 2016-12-05
SEQNG-62: Multiple sequences in engine.

jluhrs:
Implemented multiple sequences in engine. EngineState is now a map of sequences. I got rid of Queue, but the files are still there. We need to think how to put Queues back at some point in the future.


jdnavarro:
Could you rebase onto the current `develop`?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -12,6 +13,10 @@ case class Sequence[+A](id: String, metadata: SequenceMetadata, steps: List[Step
 
 object Sequence {
 
+  type Id = String
```
cquiroz: Is this the same as `ObsId`? If so perhaps they should be unified

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -40,69 +44,80 @@ package object engine {
...
-  def load(seq: Sequence[Action]): Handle[Unit] = status.flatMap {
-    case Status.Running => unit
-    case _ => put(Queue.State.init(engine.Queue(List(seq))))
-  }
+  def load(id: Sequence.Id, seq: Sequence[Action]): Handle[Unit] =
+    StateT[Task, EngineState, Unit] { s =>
+      Task {
+        s.get(id).map(t => t.status match {
+          case Status.Running => (s, ())
+          case _ => (s.updated(id, Sequence.State.init(seq)), ())
```
cquiroz: `=>` alignment?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -40,69 +44,80 @@ package object engine {
...
-          ).liftM[HandleStateT]
-        )
-      ) *> send(q)(executed)
-    }
+    getS(id).flatMap(_.map { seq =>
+        seq.status match {
+          case Status.Waiting => unit
+          case Status.Completed => unit
+          case Status.Running => {
+            val a = Nondeterminism[Task].gatherUnordered(seq.current.actions.zipWithIndex.map(act))
```
cquiroz: As discussed in another PR, probably we should be explicit about the `ExecutorService` when doing `gatherUnordered`. We could leave it to another PR though, but let's add a `TODO` reminder

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -28,58 +28,51 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
   }
 
   private def toSeqexecEvent(ev: engine.Event)(svs: List[SequenceView]): SeqexecEvent = ev match {
     case engine.EventUser(ue) => ue match {
-      case engine.Start   => SequenceStart(svs)
-      case engine.Pause   => SequencePauseRequested(svs)
-      case engine.Load(_) => SequenceLoaded(svs)
+      case engine.Start(_)   => SequenceStart(svs)
+      case engine.Pause(_)   => SequencePauseRequested(svs)
+      case engine.Load(_, _) => SequenceLoaded(svs)
```
cquiroz: `=>` alignment?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
jluhrs: Most likely, but I wanted to introduce a type alias to keep the engine isolated from a possible change in the type used to identify sequences (which I think is something planned for the ODB).

cquiroz: I mas thinking about the `ObsId` type on `DhsClient`. Most likely though, that definition is a misnomer, and it should be more like `ImageId` right?

jluhrs: Yes. I'll put a task to change the name in DhsClient.

cquiroz:
@jdnavarro Could we discuss about your doubts? As I mentioned in my comment I don't know this code deep enough to have a strong opinion.

jdnavarro:
The gist of the other approach I discussed with @jluhrs was to just have another Engine event to add an extra Sequence to the front of the Zipper without replacing the current one. This was already done for `Execution`s instead of `Sequence`s but the code was removed. I think that would have been easier to implement and would have required less changes on the code which depends on the engine.

But since most of the work has been already done with this PR, I'm not against merging it. In the long term I agree we'd need something similar to this approach for Sequence parallelization but for a simple demo I wouldn't have worried yet.

cquiroz:
Ok, thanks for the explanation
BTW, we should really focus not on making  a simple demo but on the final product already. we are way behind schedule 😄

arturog8m:
We are indeed!

jdnavarro:
More than a simple demo I was thinking of shipping something like an [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product). I think trying to include the Sequence parallelization in the first release is going to delay us even more, but if we need to ship the final product upfront, I won't complain anymore :)

arturog8m:
MVP is what we need - but we are still way far from reaching that. We need
to be able to run multiple sequences in parallel, this is a core feature of
this system, so although I agree it's complex, I want to address that
problem as soon as possible as it will be best to get as much experience
working with that as we can. Fail quick, recover fast, that is.

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -94,7 +94,7 @@ class packageSpec extends FlatSpec {
       q.enqueueOne(start(seqId)) *>
         // 6 Actions + 4 Executions + 1 start + 1 finished => take(12)
         processE(q).take(12).runLast.eval(qs1)).unsafePerformSync.get._2
-    assert(qs.get(seqId).get.done.length == 1)
+    assert(qs.get(seqId).get.done.length == 2)
```
cquiroz: Does this mean that at the end of the run you've executed 2 sequences?

jluhrs: No, you have 2 executed steps

PR #119, opened 2016-12-07
Feature/error-status

jdnavarro:
Now the `Sequence.status` and `Step.status` methods return the `Status`es from the `SharedModel` including the error status.

`_.status` wasn't used anywhere in the engine so it was easy to switch it directly to the `Status`es from the `SharedModel`. Internally the engine keeps using the keeps using its previous `Status` structure.


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -14,10 +16,18 @@ object Step {
...
     *
     */
-  def status(step: Step[Action \/ Result]): Status =
-    if (step.executions.isEmpty || step.all(_.isLeft)) Status.Waiting
-    else if (step.all(_.isRight)) Status.Completed
-    else Status.Running
+  def status(step: Step[Action \/ Result]): StepState = {
+
+    // At least an Action in this Step errored.
+    if (step.executions.isEmpty || step.any(Execution.errored)) StepState.Error("An action errored")
```
cquiroz: Is it really an error that `executions` is empty?

cquiroz:
I think it would be good to have tests showing the state when there are errors

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
cquiroz: 👍

jdnavarro:
I still have to add the tests for the Sequence Status before merging.

PR #121, opened 2016-12-07
Production packaging

cquiroz:
This PR adds support for making  a package (tarball and rpm) for production. Given that the configuration is secret, the config file is downloaded from the svn repo on demand

This PR also upgrades `sbt-native-packager` to the latest version

As all build updates this contains a fair amount of magic


cquiroz:
RPM generation is basic at the moment. It builds an rpm but I haven't tested them properly. The rpms also don't fully conform to the Gemini standards and I'd like to make sure that init scripts are installed and setup properly

But yes, you can test building the RPMs and installing them. You can even build them in OSX installing `rpmbuild`

jdnavarro:
I'm getting this error when building on my machine:

```
`rpmVendor in Rpm` is empty.  Please provide a valid vendor for the rpm SPEC.
```

Is that expected?

cquiroz:
How are you building the rpm?

jdnavarro:
`sbt rpm:packageBin`

cquiroz:
Did you have any troubles with the svn auth part?

jdnavarro:
Not at all, but I didn't try to install it, only build it. Is it using my svn credentials in the build process?

cquiroz:
Yes, it uses them to fetch the configuration from the protected repo

sraaphorst:
I keep hearing that `rpmbuild` on OS X is broken or has serious limitations. Do you know if this is inaccurate?

cquiroz:
I cannot say for sure because there are many variants of the tools and target linux variants. Historically I have done rpms on OSX and they install fine on linux, but in other occasions, updates to the rpm tools in OSX have broken them.

In general, I'd stick to doing the rpms in Linux in the same target OS, but the OSX option is good for testing

jdnavarro:
I tried the build with `rpmbuild` installed from homebrew. It worked, but next time I think I'll just build it directly in a CentOS VM.

PR #122, opened 2016-12-12
SEQNG-58D: Replace Sequence in the client with SequenceView

cquiroz:
This PR removes the old model and events and updates most of the code to use the new model. 
note also that the ocs version has been updated to match the latest release.

There are some things missing still:

* Improve the tests for the client model handlers
* Fix a few ui glitches
* Restore support for commands on the cli client (this is commented out at the moment)

Sorry for the size of the PR but this is needed to get rid of the legacy model and didn't want to make it even larger adding the missing bits


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -1,6 +1,6 @@
 package edu.gemini.seqexec.engine
 
-import edu.gemini.seqexec.model.SharedModel.{SequenceMetadata, SequenceState}
+import edu.gemini.seqexec.model.Model.{SequenceMetadata, SequenceState}
```
jdnavarro: Nice 👏

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client.cli/src/main/scala/edu/gemini/seqexec/web/client/cli/SeqexecTerminal.scala
```diff
@@ -102,17 +103,17 @@ object SeqexecTerminal extends js.JSApp {
   }
 
   object ShowHandler extends CommandHandler {
-    def width(ks: List[StepConfig]): Int = ks match {
-      case Nil => 0
-      case _   => ks.map(_.key.length).max
+    def width(ks: StepConfig): Int = ks match {
+      case m if m.isEmpty => 0
+      case _              => ks.map(_._2.length).max
     }
```
jdnavarro: I don't have any opinion about this, but is it preferable to an `if ... else ...` statement? I frequently find myself hesitating about what to pick.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client.cli/src/main/scala/edu/gemini/seqexec/web/client/cli/SeqexecTerminal.scala
```diff
@@ -102,17 +103,17 @@ object SeqexecTerminal extends js.JSApp {
...
 
     def staticResponse(c: CliCommand) = c match {
       case SequenceConfig(_, _, _, k) =>
         val pad = width(k)
-        k.sortBy(_.key).map(s => {
-          val paddedKey = s"%-${pad}s".format(s.key)
-          s"$paddedKey -> ${s.value}"}).mkString("\n")
+        k.toList.sortBy(_._1).map(s => {
```
jdnavarro: Just curious, when converting a `Map` to a `List` isn't it sorted by key alphanumerically by default?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/services/SeqexecWebClient.scala
```diff
@@ -25,26 +26,20 @@ object SeqexecWebClient {
     Unpickle[A].fromBytes(ab)
   }
 
-  def read(id: String): Future[List[Sequence]] =
+  def read(id: String): Future[SequencesQueue[SequenceId]] =
```
jdnavarro: `Task`s don't work in the client, right?

cquiroz:
You should be able now to load a sequence and see it displayed on the UI. However you need to have a working spdb set to the latest public release and some observation to load

PR #126, opened 2016-12-16
SEQNG-74 Bugfixes to make it work with EPICS systems.

jluhrs:
The biggest change is the inclusion of initialization of the EPICS connections. Not sure if it is in the right place. The rest are small bug fixes or improvements to simulations.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2ControllerSim.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2ControllerSim.scala
@@ -15,6 +15,7 @@ object Flamingos2ControllerSim extends Flamingos2Controller {
 
   override def observe(obsid: ObsId): SeqAction[ObsId] = EitherT( Task {
     Log.info("Taking Flamingos-2 observation with label " + obsid)
+    Thread.sleep(5000)
     TrySeq(obsid)
   } )
 
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -2,6 +2,7 @@ package edu.gemini.seqexec.server
 
 import java.time.LocalDate
 
+import edu.gemini.epics.acm.CaService
 import edu.gemini.pot.sp.SPObservationID
 import edu.gemini.seqexec.engine
 import edu.gemini.seqexec.engine.Event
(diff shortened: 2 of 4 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -133,9 +134,29 @@ object SeqexecEngine {
...
+      val tcsInit = if(tcsSim) Task(()) else initEpicsSystem(TcsEpics)
+      // More instruments to be added to the list here
+      val instInit = if(instSim) Task(())
+        else Nondeterminism[Task].gatherUnordered(List(Flamingos2Epics).map(initEpicsSystem(_)))
+
+      tcsInit *>
+        instInit *>
+        (for {
+          now <- Task(LocalDate.now)
+        } yield Settings(odbHost, now, dhsServer, dhsSim, tcsSim, instSim, gcalSim) )
```
jdnavarro: Maybe the comprehension could be replaced with `Task(LocalDate.now).map(now => Settings(....`

For comprehensions in Scala use the `Monad` instance, but working with the `Functor`, when possible, is usually simpler.

For this particular case I don't really care though, it's fine with me to leave the comprehension.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -133,9 +134,29 @@ object SeqexecEngine {
...
-      for {
-        now <- Task(LocalDate.now)
-      } yield Settings(odbHost, now, dhsServer, dhsSim, tcsSim, instSim, gcalSim)
+
+    // TODO: Review initialization of EPICS systems
+    def initEpicsSystem(sys: EpicsSystem): Task[Unit] = Task(Option(CaService.getInstance()) match {
+          case None => throw new Exception("Unable to start EPICS service.")
+          case Some(s) => {
+            (sys.init(s)).leftMap {
+                case SeqexecFailure.SeqexecException(ex) => throw ex
```
jdnavarro: Why raise an exception here? I guess we want to make this error crash the whole process, right?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: I added the TCS and instruments initialization and composed them with the task that was already there. But you are right, now it could be written more concisely.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: Yes, that is the reason. That error means that something went wrong loading the EPICS support libraries, which is a fatal error.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2ControllerSim.scala
```diff
@@ -15,6 +15,7 @@ object Flamingos2ControllerSim extends Flamingos2Controller {
 
   override def observe(obsid: ObsId): SeqAction[ObsId] = EitherT( Task {
     Log.info("Taking Flamingos-2 observation with label " + obsid)
+    Thread.sleep(5000)
```
jluhrs: It is just that without the delay the sequence would run too fast.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -133,9 +134,29 @@ object SeqexecEngine {
...
+                case c: SeqexecFailure => throw new Exception(SeqexecFailure.explain(c))
+            }
+          }
+        }
+      )
+
+      val tcsInit = if(tcsSim) Task(()) else initEpicsSystem(TcsEpics)
+      // More instruments to be added to the list here
+      val instInit = if(instSim) Task(())
+        else Nondeterminism[Task].gatherUnordered(List(Flamingos2Epics).map(initEpicsSystem(_)))
```
cquiroz: It seems that when we'll have many instruments you'd call `initEpicsSystem` once per each instrument. Is that necessary given that `initEpicsSystem` just calls `CaService.getInstance` or will be there in the future instrument-specific initalization code?

PR #128, opened 2016-12-19
Feature/seq-status

jdnavarro:
This PR makes sure a Sequence uses a unique status and emits a `SequenceCompleted` event at the of each execution.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
+++ b/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
@@ -1,6 +1,6 @@
 package edu.gemini.seqexec.engine
 
-import edu.gemini.seqexec.model.Model.{SequenceMetadata, SequenceState, StepConfig}
+import edu.gemini.seqexec.model.Model.{SequenceMetadata, StepConfig}
 
 import scalaz._
 import Scalaz._
@@ -77,11 +77,6 @@ class SequenceSpec extends FlatSpec {
     assert(seqzr2.next.isEmpty)
     assert(seqzar0.next.isEmpty)
     assert(seqzar1.next.nonEmpty)
-  }
-
-  "status" should "be Error when empty" in {
-    assert(Sequence.status(seqz0.toSequence) ===
-             SequenceState.Error("An action errored"))
   }
 
 }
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
```diff
@@ -79,9 +79,4 @@ class SequenceSpec extends FlatSpec {
     assert(seqzar1.next.nonEmpty)
   }
 
-  "status" should "be Error when empty" in {
```
cquiroz: Why is this test gone?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -45,11 +45,27 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
 
   def requestRefresh(q: engine.EventQueue): Task[Unit] = q.enqueueOne(Event.poll)
 
-  def eventProcess(q: engine.EventQueue): Process[Task, SeqexecEvent] =
+  def eventProcess(q: engine.EventQueue): Process[Task, SeqexecEvent] = {
+
+    implicit def toSequenceState(st: engine.Status): SequenceState = st match {
+      case engine.Status.Waiting   => SequenceState.Idle
+      case engine.Status.Completed => SequenceState.Completed
+      case engine.Status.Running   => SequenceState.Running
```
cquiroz: don't we have an `Error` status too?

jdnavarro: Previously it was calculated when any `Action` under the `Step`s of a `Sequence` errored but now we are using the engine `status` for each `Sequence` which is just `Waiting` when there is an `error` or a normal `pause`, from the Engine perspective it has to do the same in both cases.

So we have to either check whether there is an error when *printing* the `status` or create another `Engine` `Error` status which would have to set explicitly and handled like `Waiting` by the `engine`.

In either case I think that would need a separate PR.

cquiroz:
I tested this PR and I see the following

* The `SequenceStart` events bring the correct "Running" State, however the first step remains idle (Is this by design?)
* The `StepExecuted` works as before correctly
* I get the final `SequenceCompleted` event

It seems a big improvement

jdnavarro:
Regarding the `Step` status, there is the initial case when no `Action` of any of the `Execution`s under a `Step` has been completed yet, such `Step` is still reported as `Pending`, but I agree it should be reported as `Running`.

The tricky part for this case is that it's that by looking just at the `Step` alone is not possible to know its `status`, we need to know the `Status` of that info from the `Sequence`, too. And by design, the deeper levels in the state must not know anything about the levels above.

This problem will be also show up when checking the `status` of an `Execution`.

Let me try using a helper function that takes the top level status as a parameter when unzipping the state.

jdnavarro:
@cquiroz The delay in the `SequenceStart` is going to take some time, so I already filed  a ticket for it. Aside of it, is this good enough for merging?

cquiroz:
Yes, let's merge it

PR #131, opened 2016-12-21
SEQNG-79 Introduce `Executing` and `Next` system events #resolve

jdnavarro:
In addition to having more granular transition states, this also ensures that a
separate `Task` is launched when the state is updated solving the issue of having delays in between in some state transitions.

When manually testing, the first `Step` is still not shown as `Running` but that's a separate issue that I'll fix in another PR. For this concrete bugfix, please, check that that the `Start` event shows up right after clicking *Start*.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
@@ -24,6 +24,8 @@ sealed trait SystemEvent
 case class Completed[R](id: Sequence.Id, i: Int, r: R) extends SystemEvent
 case class Failed[E](id: Sequence.Id, i: Int, e: E) extends SystemEvent
 case class Executed(id: Sequence.Id) extends SystemEvent
+case class Executing(id: Sequence.Id) extends SystemEvent
+case class Next(id: Sequence.Id) extends SystemEvent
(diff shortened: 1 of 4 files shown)


cquiroz:
I don't see any difference on the client. should I?

jdnavarro:
Before this PR the *Started* event was delayed and came together with the first StepExecuted event, now the *Started* event comes right away, as it should. From the UI this can't be seen properly due to the interference with the other bug related to the Step status, but with the web console you should be able to check that the *Started* event is not delayed as before.

cquiroz:
Right now I see it

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
```diff
@@ -24,6 +24,8 @@ sealed trait SystemEvent
 case class Completed[R](id: Sequence.Id, i: Int, r: R) extends SystemEvent
 case class Failed[E](id: Sequence.Id, i: Int, e: E) extends SystemEvent
 case class Executed(id: Sequence.Id) extends SystemEvent
+case class Executing(id: Sequence.Id) extends SystemEvent
```
cquiroz: Would some of these events go to the client too?

jdnavarro: Now they are going just as log events but we can send the state or something else if that useful for the client.

PR #132, opened 2016-12-22
P1 Backend project skeleton

cquiroz:
This PR adds several bare bones modules for the future phase1 backend server/client that @sraaphorst will work on

It adopts essentially the same tools and structure of the seqexec app. Some classes are moved to a common module to be shared across the applications. This is done only for server classes but most likely client classes will be split in the near future too

### DIFF
--- a/modules/edu.gemini.p1backend/edu.gemini.p1backend.server/src/main/scala/edu/gemini/p1backend/server/http4s/WebServerLauncher.scala
+++ b/modules/edu.gemini.p1backend/edu.gemini.p1backend.server/src/main/scala/edu/gemini/p1backend/server/http4s/WebServerLauncher.scala
@@ -1,19 +1,17 @@
-package edu.gemini.p1backend.server
+package edu.gemini.p1backend.server.http4s
 
 import java.io.File
 import java.util.logging.Logger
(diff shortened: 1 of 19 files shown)


modules/edu.gemini.p1backend/edu.gemini.p1backend.server/src/main/scala/edu/gemini/p1backend/server/http4s/WebServerLauncher.scala
```diff
@@ -0,0 +1,74 @@
...
+      .start
+  }
+
+  /**
+    * Reads the configuration and launches the web server
+    */
+  override def server(args: List[String]): Task[Server] = {
+    // Launch web server
+    for {
+      _  <- configLog
```
sraaphorst: What is this for, out of curiosity, if it isn't used?

cquiroz: That is a Task[Unit] that does the initial configuration of the logging. totally side effecting

PR #133, opened 2016-12-22
Feature/step-status

jdnavarro:
This sets a `Step` as `Running` right away after starting.

With this fix there shouldn't be any more obvious bugs visible from the UI when running a Flamingos2 sequence.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -57,8 +57,6 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
         )
     }
 
-
-
   def load(q: engine.EventQueue, seqId: SPObservationID): Task[SeqexecFailure \/ Unit] = {
     val t = EitherT( for {
         odbSeq <- Task(odbProxy.read(seqId))
@@ -91,7 +89,6 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
     type QueueAR = engine.Queue[engine.Action \/ engine.Result]
     type SequenceAR = engine.Sequence[engine.Action \/ engine.Result]
     type StepAR = engine.Step[engine.Action \/ engine.Result]
-
 
     def viewSequence(seq: SequenceAR, st: SequenceState): SequenceView =
       // TODO: Implement willStopIn
@@ -133,7 +130,6 @@ object SeqexecEngine {
 
   def apply(settings: Settings) = new SeqexecEngine(settings)
 
-
   def seqexecConfiguration: Kleisli[Task, Config, Settings] = Kleisli { cfg: Config => {
       val odbHost = cfg.require[String]("seqexec-engine.odb")
       val dhsServer = cfg.require[String]("seqexec-engine.dhsServer")
@@ -141,7 +137,6 @@ object SeqexecEngine {
       val tcsSim = cfg.require[Boolean]("seqexec-engine.tcsSim")
       val instSim = cfg.require[Boolean]("seqexec-engine.instSim")
       val gcalSim = cfg.require[Boolean]("seqexec-engine.gcalSim")
-
 
     // TODO: Review initialization of EPICS systems
     def initEpicsSystem(sys: EpicsSystem): Task[Unit] = Task(Option(CaService.getInstance()) match {
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -90,34 +90,51 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
     type SequenceAR = engine.Sequence[engine.Action \/ engine.Result]
     type StepAR = engine.Step[engine.Action \/ engine.Result]
 
-    def viewSequence(seq: SequenceAR, st: SequenceState): SequenceView =
+    def viewSequence(seq: SequenceAR, st: SequenceState): SequenceView = {
+
+      def engineSteps(seq: SequenceAR): List[Step] = {
+
+        def viewStep(step: StepAR): Step =
+          StandardStep(
+            step.config,
+            engine.Step.status(step),
+            // TODO: Implement breakpoints at Engine level
+            breakpoint = false,
+            // TODO: Implement skipping at Engine level
+            skip = false,
+            configStatus = Map.empty,
+            // TODO: Implement standard step at Engine level
+            observeStatus = ActionStatus.Pending
+          )
+
+        // Couldn't find this on Scalaz
+        def splitWhere[A](l: List[A])(p: (A => Boolean)): (List[A], List[A]) =
+          l.splitAt(l.indexWhere(p))
+
+        seq.steps.map(viewStep) match {
+          // Find first Pending Step when no Step is Running and mark it as Running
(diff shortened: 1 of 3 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -92,35 +90,51 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
+
+        def viewStep(step: StepAR): Step =
+          StandardStep(
+            step.config,
+            engine.Step.status(step),
+            // TODO: Implement breakpoints at Engine level
+            breakpoint = false,
+            // TODO: Implement skipping at Engine level
+            skip = false,
+            configStatus = Map.empty,
```
cquiroz: How's this different from the first config map?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -92,35 +90,51 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
+            // TODO: Implement standard step at Engine level
+            observeStatus = ActionStatus.Pending
+          )
 
-    def viewSequence(seq: SequenceAR, st: SequenceState): SequenceView =
+        // Couldn't find this on Scalaz
+        def splitWhere[A](l: List[A])(p: (A => Boolean)): (List[A], List[A]) =
```
cquiroz: Is this general? It seems to be called only once

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -92,35 +90,51 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
+          l.splitAt(l.indexWhere(p))
+
+        seq.steps.map(viewStep) match {
```
cquiroz: I guess this updates the first step as running once the sequence is started. Is this real in terms of what happens on the engine or are we simulating them for the clients?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -92,35 +90,51 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
+          // Find first Pending Step when no Step is Running and mark it as Running
+          case steps if st === SequenceState.Running && steps.toList.all(_.status =/= StepState.Running)=>
+            val (xs, (y :: ys)) = splitWhere(steps)(_.status === StepState.Pending)
+            xs ++ (
+              // TODO: Why it doesn't have `.copy`?
```
cquiroz: It mush have a `copy` it is a case class

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: I don't know, I'm just keeping the existing code assuming it was there for some reason. @jluhrs should know.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: This is only when *viewing* the Step. The engine doesn't use the Step status.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: I don't understand what you mean.

cquiroz: I just thought it could have been inlined since is so simple, bun don't worry

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
cquiroz: Do you mean it is only on the engine output?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: I get what you mean. Actually I wrote it as a separate function because I'm pretty sure there should be something very close in Scalaz that I couldn't find it. This way it's easier to replace it later when I stumple upon it in the future.

That's usually what I do when I suspect there is a similar function already in Scalaz.

cquiroz: Sounds good, doesn't `partition` do this?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: Yes, the engine doesn't need, so far, the concept of Step status. The status is only set at the top level, now the Sequence, and flows downstream from there. Aside of not having to keep coordinating separate statuses, it's also impossible to make them inconsistent (i.e. A Sequence Pending but with a Step Running).

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: I already tried it, but it's not exactly what we need here. I found several functions that were almost there :)

cquiroz:
This looks very good, what made the delay disappear?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: I was very surprised, but it's not there. If you find the issue let me know because it doesn't make sense to me either.

cquiroz: I see, it is because steps is `List[Step]` which is a trait and not the `SequenceStep` class

cquiroz: https://gist.github.com/cquiroz/e32cd97cd2133b1c3ddd1a0f7b56389f

jdnavarro:
The delay was taken care in #131. The root of the problem in that case is that the engine logging wasn't accurate. All the messages were logged with a separate thread but for some outputs it wasn't forking when it should.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: Does that gist compile for you? It doesn't for me.

PR #135, opened 2016-12-27
SEQNG-84: Dataset labels are not showing up in web client

cquiroz:
This PR does not fix the issue but identifies where changes are needed. This requires some updates on the engine side that are pending

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -67,6 +67,7 @@ object Model {
     val breakpoint: Boolean
     val skip: Boolean
     val configStatus: Map[SystemName, ActionStatus]
+    val fileId: Option[dhs.ObsId]
     val observeStatus: ActionStatus
   }
 
@@ -76,6 +77,7 @@ object Model {
     override val breakpoint: Boolean,
     override val skip: Boolean,
     override val configStatus: Map[SystemName, ActionStatus],
+    override val fileId: Option[dhs.ObsId],
     override val observeStatus: ActionStatus
   ) extends Step
   // Other kinds of Steps to be defined.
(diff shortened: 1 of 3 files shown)


modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -67,6 +67,7 @@ object Model {
     val breakpoint: Boolean
     val skip: Boolean
     val configStatus: Map[SystemName, ActionStatus]
+    val fileId: Option[dhs.ObsId]
```
cquiroz: I'm not sure this is the right place, Perhaps it should be part of the `observeStatus`. WDYT?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -103,6 +103,8 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
             // TODO: Implement skipping at Engine level
             skip = false,
             configStatus = Map.empty,
+            // Where do I get the file id from?
+            None,
```
cquiroz: This is a key part of the engine update. We need to find how to populate this field from the DHS generated value

jdnavarro:
I actually wanted to take care of this myself because this is also related to the task of creating distinct type for `Step[Action] \/ Step[Result]` and already have a local branch with some work done, but I misunderstood in a previous meeting that this was already taken care of. The idea we discussed at that time was to have something like:

```
sealed trait Step
case class StepPending(...)
case class StepDone(...)
```

And in the `StepDone` case class have the `fileId` parameter but not in the `StepPending` (or perhaps have it there too but wrapped in `Task`). This will get more intricate when we add more types of `Step`s we add but it's a good starting point for now.

Obtaining the actual `fileId` from the DHS should be part of an actual execution.

cquiroz:
Sounds good 👍  I can do the frontend once your PR is merged. The change there is very minimal

PR #136, opened 2016-12-27
SEQNG-85 Enable Stop from Web UI

jdnavarro:
Concretely:

- Wire up UI button with stop route.
- Remove `Pause` button.
- Fix regression where the sequence didn't fully stop after more events were
  introduced.
- Initial attempt of labeling the Step status when in Paused state.

The way the current Step status is displayed in the UI now when paused is strange. This is because there are different cases involving the underlying `Execution`s which are not shown to the client. In the next meeting I can explain with a demo what is going on under the hood and why displaying the Paused status of a Step is trickier than it initially seems.

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -50,7 +50,7 @@ object Model {
     case object Skipped extends StepState
     case class Error(msg: String) extends StepState
     case object Running extends StepState
-    case object Paused extends StepState
+    case object Stopped extends StepState
 
     implicit val equal: Equal[StepState] = Equal.equalA[StepState]
   }
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -50,6 +50,7 @@ object Model {
     case object Skipped extends StepState
     case class Error(msg: String) extends StepState
     case object Running extends StepState
+    case object Paused extends StepState
 
     implicit val equal: Equal[StepState] = Equal.equalA[StepState]
   }
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/SequenceArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/SequenceArea.scala
@@ -33,13 +33,15 @@ object SequenceStepsTableContainer {
 
   class Backend($: BackendScope[Props, State]) {
 
-    def requestRun(s: SequenceView): Callback = $.modState(_.copy(runRequested = true)) >> Callback {
-      SeqexecCircuit.dispatch(RequestRun(s))
-    }
+    def requestRun(s: SequenceView): Callback =
+      $.modState(_.copy(runRequested = true, stopRequested = false)) >> Callback {
+        SeqexecCircuit.dispatch(RequestRun(s))
+      }
 
-    def requestStop(s: SequenceView): Callback = $.modState(_.copy(stopRequested = true)) >> Callback {
-      SeqexecCircuit.dispatch(RequestStop(s))
-    }
+    def requestStop(s: SequenceView): Callback =
+      $.modState(_.copy(stopRequested = true, runRequested = false)) >> Callback {
+        SeqexecCircuit.dispatch(RequestStop(s))
+      }
 
     def render(p: Props, s: State) = {
       <.div(
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/SequenceArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/SequenceArea.scala
@@ -27,7 +27,7 @@ import org.scalajs.dom.document
   * Container for a table with the steps
   */
 object SequenceStepsTableContainer {
-  case class State(runRequested: Boolean, nextScrollPos: Double, autoScrolled: Boolean)
+  case class State(runRequested: Boolean, stopRequested: Boolean, nextScrollPos: Double, autoScrolled: Boolean)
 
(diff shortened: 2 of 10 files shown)


modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -50,6 +50,7 @@ object Model {
     case object Skipped extends StepState
     case class Error(msg: String) extends StepState
     case object Running extends StepState
+    case object Paused extends StepState
```
cquiroz: Maybe to be consistent this should be `Stopped`?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/services/SeqexecWebClient.scala
```diff
@@ -61,7 +61,7 @@ object SeqexecWebClient extends ModelBooPicklers {
     */
   def stop(s: SequenceView): Future[RegularCommand] = {
     Ajax.post(
-      url = s"$baseUrl/commands/${s.id}/stop",
+      url = s"$baseUrl/commands/${s.id}/pause",
```
cquiroz: I thought we were unifying on `stop` for this action

jdnavarro: I used `pause` in this case because it's clashing with an existing route but I can comment out the conflicting one.

arturog8m: What does the other route do? I agree this is confusing.

cquiroz: It is for the old seqexec engine. Feel free to remove those

jdnavarro: The other route it's from the previous implementation and didn't remove it because I don't know if we'll need it later. I'll comment it out.

jdnavarro:
As I mentioned above, it's not obvious to me how to represent in detail the different cases that can happen when stopping. For example, what's the status of a `Step` where all `Execution`s have been completed after stopping? Are we sure we want to make it the same the status of a Step that has more `Execution`s left? There are more edge cases that right now are represented poorly, but I'd like to explain them in the next meeting with the demo in front.

I don't want to go crazy trying to make it all polished taking only my own opinion on these hairy UI situations. I'd rather to consult it with you all in the next meeting before going ahead. Right now, I was trying to get done the essentials of what it has to happen when stopping.

cquiroz:
As a minimum the Stop button should be disabled while stop is being requested. I think having the step where Stop was issued in another color is also something we'll do regardless

Then we can focus on the edge cases

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/services/SeqexecWebClient.scala
jdnavarro: Oh, @cquiroz didn't see your last comment before posting my previous comment. I'll remove those routes then.

jdnavarro:
I took care of disabling the button and a different color for the stopped step. It turns out it wasn't that hard 😌, but there are still rough edges.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/SequenceArea.scala
```diff
@@ -142,9 +142,11 @@ object SequenceStepsTableContainer {
                 p.s.steps.zipWithIndex.map {
                   case (step, i) =>
                     <.tr(
+                      // Available row states: http://semantic-ui.com/collections/table.html#positive--negative
```
cquiroz: 👍

cquiroz:
Tested the last version and it looks much better. We may want to add another message to indicate that we are stoping later

We should do a test with multiple clients too. I think the steps will work fine but I'm not sure about the button's state

What happens if I do a `stop` to a sequence already stopping?

jdnavarro:
By adding another stop message you mean a different button label while stopping, right?

For the case of 2 clients pressing `stop`, I haven't checked it but it should stop when the first client presses it, for the second client it would look as if it stopped earlier than usual.

In any case I think we can find several weird cases when combining `stop`/`start` with multiple clients when there are UI state elements that is not shared among clients, like the disabled state of a button.

I also see several issues with the buttons that we have to iron out. Another one, for example, is having the Run and Stop button in the same position can lead to accidentally clicking the wrong button when multiple clients are using it.

cquiroz:
Perhaps we need to find a way to share this information among clients, e.g. when somebody requests a stop, regardless of the engine

The observation about Run/Stop on the same place is a good one. Perhaps they should all be visible but just disabled. And coloring them could help too

jdnavarro:
I agree that we should think this through.

In the case of `Stop`, aside of the events `Stop` and `Stopped`, an extra `Stopping` event in the engine could fix the inconsistency issues among clients when stopping/running. However, from the underlying systems there is nothing being changed with that extra event, it's just a notification needed by the UI.

So maybe in the end there is no other way around it, but I think we should try to have the engine know nothing about UI and only make it responsible of communicating with the telescope resources.

Maybe we could introduce another intermediate state in the web server with just the UI state that needs to be shared. This state won't need to check the engine every time there is a new user event and could give what the client needs on its own if it can.

In any case I think it's worth planning how to solve this start/stop issue carefully because it should set the trend for how to sync UI state that doesn't touch anything in the underlying systems.

jdnavarro:
With the latest commit, I think this is ready for merging knowing there are still several limitations. These are the tickets I'm going to file for now:

- Sometimes, when stopping, the first Step of a Sequence shows up as `Stopped` when it should be displayed as `Done`.
- Make Stop and Run buttons different colors.
- Don't make the Stop and Run buttons go away, just disable them depending on the status.

There are more glitches I have in my notes where its' no so obvious to know what to do about them so we need to discuss them with a demo in front. This also includes how to show the row of a `Step` that has been completed after being stopped, there are some inconvenient implications for a second client watching.

PR #138, opened 2016-12-30
SEQNG-92 Rename `Stop` -> `Pause

jdnavarro:


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -44,7 +44,7 @@ class SeqexecCommandRoutes(auth: AuthenticationService, inputQueue: engine.Event
       } yield resp
 
     // TODO: Add obsId parameter
```
cquiroz: I guess the todo can be removed

PR #139, opened 2016-12-30
SEQNG-90: Web sockets unit test

cquiroz:
For a while I've been trying to do more tests on the backend. Tests on regular routes is fairly straightforward but for websockets sending data in binary frames it turns out more tricky

The tests below simulate a connection and verify that we get a `ConnectionOpenEvent(u)` on the websocket pipe. The first test is for anonymous users and the second is for a logged in user


cquiroz:
It is complex and you can argue whether we gain much with it. I guess time will tell.
But yes, since it is already there why not use it

PR #141, opened 2016-12-30
SEQNG-89 Add colors Start/Pause buttons

jdnavarro:
`Start` is *Green* and `Pause` is *Red*.


cquiroz:
Do you have a Screenshot?

jdnavarro:
Here you go:


pause
run

cquiroz:
In general we try to use the strong Red only to indicate errors. Also the green seems strong to me. Could we try a subtler shade. I dont know if you can go beyond the list of set colors in semantic

http://semantic-ui.com/elements/button.html#colored

andrewwstephens:
I agree that it should be a bit more subtle; perhaps a red outline and the stronger when the user hovers over it?

jdnavarro:
Following @cquiroz, I changed the colors to *blue* and *teal* for Start and pause respectively. These colors come by default of Semantic UI but if we want to create our own colors, I think it'll be better to create a cohesive color palette covering our needs once we know most of them. If we create new colors *adhoc* we can end up with a great chromatic aberration 😄 .

cquiroz:
I'd rather not have custom colors unless we have a very good reason. Could you post new screenshots?

jdnavarro:
blue
teal

cquiroz:
The blue looks fine though teal is a bit strange. But I say let's merge this anyway and  we can rectify later if needed

PR #142, opened 2017-01-06
SEQNG-93: UI elements for exposure control

cquiroz:
This is an prototype UI to include buttons for exposure control at the row level and a progress bar for the observation process. There is no functionality attached just the following changes:

* Made the running step taller
* Add buttons for pause/stop/abort at the row level
* Added a progress bar that should display how long to complete the observation

The code doesn't tell much so here is a screenshot

screenshot 2017-01-06 14 26 34

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
@@ -119,22 +119,6 @@ object LoadingErrorMsg {
   def apply(p: ModelProxy[SeqexecAppRootModel.LoadedSequences]) = component(Props(p))
 }
 
-object QueueTableLoading {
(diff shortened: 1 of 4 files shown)


swalker2m:
I think the screen shot looks nice.  What is the difference between pausing at the observation level vs pausing the step?  If you stop or abort a step can you pick up at the next step?  I ask because I thought you could stop or abort an observation but not individual steps.

cquiroz:
That's a very good question for @andrewwstephens

andrewwstephens:
@swalker2m - The idea is that the top row of buttons will correspond to the "Sequence" controls from the current Seqexec to pause the sequence at the end of the exposure (the same as putting a break-point) and continue after a sequence has been paused.  The buttons on the active step will correspond to the "Observe" controls in the current Seqexec, which are used to Hold/Stop/Abort the current exposure.

arturog8m:
This looks very good - names of buttons might change in the future as we refine the UI, we were interested at this level on figuring out where the different elements should go. We are missing a few things for sequence control, and maybe adding a bit more information (via tool tips or something) to the observation control. All in all, this is looking good! 👍

cquiroz:
@andrewwstephens brought to my attention that F2 only has Stop/Abort. In the next PR I'll make that an instrument dependent option

PR #143, opened 2017-01-09
SEQNG-97: Add and update observe control buttons based on existing seqexec

cquiroz:
This is a small change to the observation control buttons adding tooltips and making them instrument and state dependent

Here's a screenshot

screenshot 2017-01-09 10 03 44

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -110,11 +110,12 @@ object Model {
      * Returns the observation operations allowed
      * TODO Convert to an Instrument-level typeclass
      */
-    def allowedObservationOperations: List[ObservationOperations] =
+    def allowedObservationOperations(stepStatus: StepState): List[ObservationOperations] =
       metadata.instrument match {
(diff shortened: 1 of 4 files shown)


cquiroz:
After some discussion with science I have changed the code to show for F2 all the buttons as it would be for GMOS-Hammamtsu:

Here's a screenshot with the 4 buttons

screenshot 2017-01-10 11 28 19

And here is a button with a paused step

screenshot 2017-01-10 12 00 52

jluhrs:
I am a bit late to the discussion, but let me say I like how the commands that operate over the currently running exposure are clearly differentiated from the sequence commands.

arturog8m:
I like this very much. Perhaps the two pause buttons and the two stop buttons should have a slightly different icon (perhaps an icon decoration of some sort to one of them) rather than just the color difference. But we can think about that later when more people get to see this. In the meantime, 👍

jluhrs:
One detail: the sequence pause button in the current Seqexec is labeled "Pause Gracefully". It may create confusion if the exposure pause button has the word "gracefully" in the tooltip.

andrewwstephens:
Could we add an "Abort" button for good measure?

andrewwstephens:
When an exposure is paused we should also show the options to "Stop" or "Abort"

cquiroz:
@jluhrs Pause gracefully would be shown only for GMOS Nod&Shuffle

PR #145, opened 2017-01-11
SEQNG-99: Add missing buttons for sequence control

cquiroz:
This PR is another mockup of the UI updating the sequence control buttons. After some discussions this has been distilled to 2 buttons: `Run sequence from step N` and `Pause`

At the beginning we should start from step 1

screenshot 2017-01-11 17 40 41

However the user can select starting from a later step clicking the correct row:

screenshot 2017-01-11 17 40 47

Once we run only Pause is visible

screenshot 2017-01-11 09 43 44

And if we pause we return to the `Run from step N` where N is the next step to be executed and it can be again overriden manually:

screenshot 2017-01-11 17 39 53

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -88,6 +88,7 @@ object Model {
     case object Completed         extends SequenceState
     case object Running           extends SequenceState
     case object Idle              extends SequenceState
+    case object Paused            extends SequenceState
     case class Error(msg: String) extends SequenceState
 
     implicit val equal: Equal[SequenceState] = Equal.equalA[SequenceState]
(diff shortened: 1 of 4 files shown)


andrewwstephens:
Nice!  The "Run from step N" button avoids the need to have both the sequence "Run" and "Continue" buttons, and makes it very clear what will happen when you click it.

If a sequence has some executed steps, we may want to label the button "Continue from step N", where N defaults to the first unexecuted step.

andrewwstephens:
It would be nice to make it more obvious when a step is paused mid-exposure versus when the sequence is paused between steps.

cquiroz:
I added several small fixes to details discovered during review:

* Change the text from Run to Continue if some steps have been executed
* Correctly select the step if a step is paused but the exposure completed regardless
* Override the start step with a double click and ignore clicks on the config button
* Fixes a bug where the step to run is miscalculated if the page is reloaded

There are still several iterations to complete this but this can probably close the mockup stage

PR #146, opened 2017-01-13
SEQNG-11 Implemented step rollback, added tests.

jluhrs:
The rollback implementation stores some data in the Step.Zipper to remember how to go back to the original Step.Zipper (the one with no Execution run). The rollback is called on the sequence, and propagated to the current Step. It is done in response to a start (if the Sequence was not paused, the rollback has no effect)
This pull request includes tests for pause and resume. I feel the implementation of the tests is a bit too intrusive. Maybe we could build some helper functions to abstract a bit the access to the guts of the engine?


cquiroz:
Can you explain a little bit what does it mean a step rollback? Is it in response to a sequence pause?

PR #149, opened 2017-01-19
SEQNG-98: Find a place for F2 sequence execution errors

cquiroz:
As in other cases this PR is to share the UI design. This is how we could display errors on a given step. I'm not clear if we need the ack system. Do we?

If so, is this something that others should see, meaning if one user presses ack, the others don't need to see it? We could discuss this a bit, it may complicate the backend model

Here is a screenshot in the error state:

screenshot 2017-01-19 15 07 01

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -114,38 +114,7 @@ object Model {
     status: SequenceState,
     steps: List[Step],
     willStopIn: Option[Int]
-  ) {
-    /**
-     * Returns the observation operations allowed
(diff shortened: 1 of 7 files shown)


cquiroz:
This is still a work in progress, pending PR #148 and the comments from our users, So yes, the Ack button will be removed

cquiroz:
Updated according to the last discussion. The main change is to enable the Sequence level controls over the Observation level controls:

screenshot 2017-01-24 18 00 43

arturog8m:
This is looking really nice. In the last screenshot, it says "Press Continue to re-try". Is Continue a button/link you can press or is it referring to different button I'm failing to see in the UI?

cquiroz:
It should match the name of the blue button above the table. That's the idea but yes there is a small bug present

cquiroz:
Fixed and here's a new screenshot

screenshot 2017-01-25 00 14 32

PR #150, opened 2017-01-24
Set FileId for completed Step

jdnavarro:
The `fileId` is set in the Step when an action returns successfully an `Observed` `Result`.

fileid

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
@@ -77,6 +77,10 @@ object Execution {
 sealed trait Result
 
 object Result {
-  case class OK[R](r: R) extends Result
+  case class OK(response: Response) extends Result
   case class Error[E](e: E) extends Result
+
+  sealed trait Response
+  case class Configured(r: String) extends Response
+  case class Observed(fileId: String) extends Response
 }
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
@@ -82,5 +82,5 @@ object Result {
 
   sealed trait Response
   case class Configured(r: String) extends Response
-  case class Observed(fileId: String) extends Response
+  case class Observed(fileId: FileId) extends Response
 }
--- a/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/StepSpec.scala
+++ b/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/StepSpec.scala
@@ -75,6 +75,7 @@ class StepSpec extends FlatSpec {
         List(
           Step(
             1,
+            None,
             config,
             false,
             List(
@@ -96,7 +97,7 @@ class StepSpec extends FlatSpec {
       case Sequence.State.Zipper(zipper, status) =>
         status should be (Idle)
         inside (zipper.focus.toStep) {
-          case Step(_, _, _, ex1::ex2::Nil) =>
+          case Step(_, _, _, _, ex1::ex2::Nil) =>
             assert( Execution(ex1).results.length == 3 && Execution(ex2).actions.length == 1)
 
         }
@@ -114,6 +115,7 @@ class StepSpec extends FlatSpec {
         List(),
         Step.Zipper(
           2,
+          None,
           config,
           false,
           List(),
@@ -134,7 +136,7 @@ class StepSpec extends FlatSpec {
       case Sequence.State.Zipper(zipper, status) =>
         status should be (Running)
         inside (zipper.focus.toStep) {
-          case Step(_, _, _, ex1::ex2::Nil) =>
+          case Step(_, _, _, _, ex1::ex2::Nil) =>
             assert( Execution(ex1).actions.length == 2 && Execution(ex2).actions.length == 1)
         }
     }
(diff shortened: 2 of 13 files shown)


andrewwstephens:
When is the file ID assigned?  Would it be possible to display the file ID on the running step?

andrewwstephens:
I think it will be important for the observer to know what file ID to expect since files will not be sequential when running multiple simultaneous sequences.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
```diff
@@ -77,6 +77,10 @@ object Execution {
 sealed trait Result
 
 object Result {
-  case class OK[R](r: R) extends Result
+  case class OK(response: Response) extends Result
   case class Error[E](e: E) extends Result
+
+  sealed trait Response
+  case class Configured(r: String) extends Response
+  case class Observed(fileId: FileId) extends Response
```
cquiroz: I think we are missing one here, there should be a `StartObserve` and contain the `fileId` and an estimate of the running time (That can be blank for now)

cquiroz:
It looks in general fine to me but I'd like that we have an event when we start observe with the fileid. It would solve @andrewwstephens question above

arturog8m:
Right - I think it will be best if the seqexec shows the file ID as soon as it gets it from the DHS.

jdnavarro:
What I initially understood is that the `fileId` could only be obtained at the end of an observation, but if that's not the case what is the earliest point at which the `fileId` could be known? At the beginning of an observation? If it could be known at the beginning of a Step, it would be possible simplify the code. /cc @jluhrs

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
jdnavarro: In the UI, you never have to look here, right? Ultimately, we need the id in `Step.fileId`, isn't it? I'm saying this because creating that case class would be much more problematic than updating the `fileId` directly in the `Step`.

cquiroz: Actually I want to look there, because  the observation buttons should be displayed only when the observation begins, so it is kind of a marker event. It would be ok to the fileId in the step though

jdnavarro: I'm confused. That `Response` is not a UI event, it's the result of action that's launched asynchronously by the engine that when completed is processed again by the engine to create a new application state that is sent to the UI together with a UI event.

I think what you mean is that you need the UI event `StartObserve` here: https://github.com/gemini-hlsw/ocs3/blob/develop/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala#L12

Sorry to be pestering you, but I need to be sure we are in the same page before going ahead.

jluhrs:
The internal view of a Sequence in ordered list of black boxes (Execution) that can be run to produce a value. They are run one after the other. The observation is run as one of those black boxes. The fileId is generated inside the box and published as its output when the black box finishes its execution. To be able to access the fileId is not a trivial task.
One option to make it work would be to replace the Execution by functions that produce an Execution using the result of the previous one. Of course, most of then would be consts, passing always the Execution we already have in place. But with such mechanism we could put the retrieval of the fileId in one Execution and the observation itself in the next one, with the fileId as parameter.

jdnavarro:
In order to create the a valid `fileId`, does the DHS need any parameters that only exist *after* the instrument configuration or TCS configuration?

I was already working on setting the `fileId` at the beginning of a Step (another action in parallel with TCS and the instrument configuration). The `fileId` would be set in the `Step` after the first *Execution*, so when is time to observe, instead of asking the DHS for a `fileId`, it uses this `fileId` already set in the `Step` for the actual observation.

cquiroz:
It shouldn't need, it only requests an id without params

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/StepSpec.scala
```diff
@@ -153,18 +157,18 @@ class StepSpec extends FlatSpec {
...
+  val stepza0: Step.Zipper  = Step.Zipper(1, None, config, List(List(action)), Execution.empty, Nil, (Execution.empty, List(List(action))))
+  val stepza1: Step.Zipper  = Step.Zipper(1, None, config, List(List(action)), Execution(List(result.right)), Nil, (Execution(List(action.left)), List(List(action))))
+  val stepzr0: Step.Zipper  = Step.Zipper(1, None, config, Nil, Execution.empty, List(List(result)), (Execution(List(action.left)), Nil))
+  val stepzr1: Step.Zipper  = Step.Zipper(1, None, config, Nil, Execution(List(result.right, result.right)), Nil, (Execution(List(action.left, action.left)), Nil))
+  val stepzr2: Step.Zipper  = Step.Zipper(1, None, config, Nil, Execution(List(result.right, result.right)), List(List(result)), (Execution(List(action.left)), List(List(action, action))))
+  val stepzar0: Step.Zipper = Step.Zipper(1, None, config, Nil, Execution(List(result.right, action.left)), Nil, (Execution(List(action.left, action.left)), Nil))
+  val stepzar1: Step.Zipper = Step.Zipper(1, None, config, List(List(action)), Execution(List(result.right, result.right)), List(List(result)), (Execution(List(action.left)), List(List(action, action), List(action))))
```
jluhrs: I modified this a bit in my PR to make it easier to adapt to changes to Step.Zipper

jdnavarro: No problem, I'll pick your code when solving the merge conflict.

jdnavarro:
We've been discussing in person with @jluhrs about showing the `fileId` before doing the observation and it seems the task requires more thought, so we think it should be continued in another PR with a separate ticket.

Aside of the `fileId` not showing up earlier, are there any other outstanding issues with this PR? Are we good to merge then?

cquiroz:
That's ok, I think it can be merged

PR #151, opened 2017-01-25
SEQNG-32 Implemented breakpoints in execution engine

jluhrs:


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -91,18 +102,18 @@ package object engine {
               // Final State
               case Some(qs: Sequence.State.Final) =>
                 putS(id)(qs) *> send(q)(finished(id))
-              // Execution completed
+              // Execution completed. Check breakpoint here
               case Some(qs) =>
-                putS(id)(qs) *> send(q)(executing(id))
+                putS(id)(qs) *> (if(qs.getCurrentBreakpoint) switch(q)(id)(SequenceState.Idle)
```
jluhrs: I'm sure there is a more elegant way of writing this.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -8,10 +8,12 @@ import Scalaz._
 /**
   * A list of `Executions` grouped by observation.
   */
-case class Step[+A](id: Int, config: StepConfig, executions: List[List[A]])
+case class Step[+A](id: Step.Id, config: StepConfig, breakpoint: Boolean, executions: List[List[A]])
 
 object Step {
 
+  type Id = Int
```
cquiroz: Is this the position? What happens if we reload the sequence and the steps have changed?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -74,6 +75,7 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
       case engine.Start(_)    => SequenceStart(svs)
       case engine.Pause(_)    => SequencePauseRequested(svs)
       case engine.Load(_, _)  => SequenceLoaded(svs)
+      case engine.Breakpoint(_, _, _) => StepBreakpointChanged(svs)
```
cquiroz: Would you align the `=>`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -177,33 +180,9 @@ object Sequence {
       * Initialize a `State` passing a `Queue` of pending `Sequence`s.
       */
     // TODO: Make this function `apply`?
-    def init(q: Sequence[Action]): State = Initial(q, SequenceState.Idle)
+    def init(q: Sequence[Action]): State = Sequence.Zipper.currentify(q).map(Zipper(_, SequenceState.Idle)).getOrElse(Final(Sequence.empty(q.id), SequenceState.Idle))
```
jdnavarro: Could this line be broken in 2?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: You could create analogous function to `getCurrentBreakPoint` lifted to `Handle`, like `status` in this module I think, but if this only place where you do this, I don't think it's worth the effort.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
jluhrs: For the time being the id is generated internally using the step position. If the sequence is reloaded its internal representation is regenerated from scratch. Therefore, the breakpoints are lost.

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -95,6 +95,7 @@ object Model {
   object SequenceState {
     case object Completed         extends SequenceState
     case object Running           extends SequenceState
+    case object Stopping          extends SequenceState
```
jdnavarro: 😕 We had some discussions while you were away about banning the word `Stop` from the code and rename it all to `Pause`. In any case if this doesn't show up in the UI, I think we'll be fine.

PR #152, opened 2017-01-26
Ui improvements

cquiroz:
This PR contains several updates related to the UI including:

* Abstraction of Table Headers
* Updates to libraries diode and scala.js
* Speed ups of tests

The latests has reduced travis execution time from about 15min to around 8 min


cquiroz:
What glitches?

jdnavarro:
Sometimes, in Firefox the Sequence table kind of shifted out of its container when loading a new sequence. Perhaps this was fixed before and I only noticed it now.

cquiroz:
Ok, I should probably do more development on Firefox than Chrome since we deploy in firefox anyway
Will merge anyway

PR #154, opened 2017-01-30
Fixed bug that made the sequence appear as running in the clients after it was paused.

jluhrs:
The problem was that the sequence was being stopped but when that happened, the event sent to the clients was a NewLogMessage, without the new state of the sequence. I created a new type of event, a generic update, to send the new state to the clients.

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -28,6 +28,9 @@ object Model {
 
     case class SequenceRefreshed(view: SequencesQueue[SequenceView]) extends SeqexecModelUpdate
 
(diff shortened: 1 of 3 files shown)


sraaphorst:
👍

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -28,6 +28,9 @@ object Model {
 
     case class SequenceRefreshed(view: SequencesQueue[SequenceView]) extends SeqexecModelUpdate
 
+    //Generic update. It will probably become useless if we have a special Event for every case.
+    case class SequenceUpdated(view: SequencesQueue[SequenceView]) extends SeqexecModelUpdate
```
cquiroz: We have a `SequenceRefreshed` event, could we reuse it?

PR #155, opened 2017-01-30
Feature/result-error

jdnavarro:
With this PR an action Error is promoted to the Step status so that it shows up in the UI.

Notice the Error type for the Action has only a `String` parameter representing the error message. This was the simplest option for showing the error message, but we will probably need to do more with the errors than just showing the error message, so in the future that message should probably be replaced by `SeqexecFailure` (or similar) and we'll have to add its own UI component.

I left a TODO in the code for that task and will add a ticket for the task if you are OK with it.

error
.


cquiroz:
There are some broken tests

jdnavarro:
I just pushed the fixed the tests I forgot to update.

PR #156, opened 2017-01-30
SEQNG-129 Fixed bug that activated breakpoint in the middle of a Step.

jluhrs:
Now the effect of the breakpoint applies only at the beginning of the Step. Unit test included.
This PR includes a simpler solution to SEQNG-127, and some improvements to the tests.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
+++ b/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
@@ -1,7 +1,8 @@
 package edu.gemini.seqexec.engine
 
 import edu.gemini.seqexec.engine.Event.start
-import edu.gemini.seqexec.model.Model.SequenceState.Idle
+import edu.gemini.seqexec.model.Model.SequenceState
+import edu.gemini.seqexec.model.Model.SequenceState.{Error, Idle}
 import edu.gemini.seqexec.model.Model.{SequenceMetadata, StepConfig}
 
 import scala.Function.const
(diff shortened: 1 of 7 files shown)


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
```diff
@@ -61,6 +62,14 @@ class SequenceSpec extends FlatSpec {
     )
   )
 
+  def runToCompletion(q: scalaz.stream.async.mutable.Queue[Event], s0: EngineState): EngineState = {
+    def isFinished(status: SequenceState): Boolean =
+      status == Idle || status == SequenceState.Completed || status === Error
+
+    q.enqueueOne(start(seqId)).flatMap( _ =>
+       processE(q).drop(1).takeThrough(a => !isFinished(a._2.get(seqId).get.status) ).runLast.eval(s0)).unsafePerformSync.get._2
```
jdnavarro: Didn't know about `.takeThrough`. That will come handy.

jluhrs: Yes. First I tried to use takeUntil, and I spend some time banging my head against the wall trying to understand why the resulting state was not what I expected, until I realized the state I was interested in was not included. Then I found takeThrough, and I was happy.

jdnavarro: I also tried with `takeWhile` some time ago but gave up after much headache. How did you find `takeThrough`? By looking at scalaz-streams source code? Is the rendered API documentation for `scalaz-stream` public somewhere?

PR #157, opened 2017-01-31
SEGNG-121: Link UI for breakpoints with the backend

cquiroz:
This PR connects the breakpoints set on the UI to the backend


jdnavarro:
I was already working on passing the Step Ids to the client because I needed it for the fileId fix, but since you already did it, you save me from some extra work 😄

PR #158, opened 2017-02-01
SEQNG-130: Setup test seqexec servers

cquiroz:
This PR updates the build to support deployment of test Seqexec servers at GN and GS


cquiroz:
They should be in the conf file. Did you try with the task `universal:packageZipTarball` on the projects?

We could probably ignore the 32/64 bit distinction

jdnavarro:
I actually did first a build with xz compression and couldn't see the `app.conf` files, but in the zip build the configuration files are there.

cquiroz:
Removed references to linux64, we are not going to do 32bit releases of the seqexec

PR #159, opened 2017-02-01
SEQNG-128 Fix order of done steps

jdnavarro:
This started as a bug with the `fileId` but it turns out all the *done* steps were in reverse order. After figuring out the root cause of the bug, the fix was trivial.


cquiroz:
Why it seemed ok on the client, because only the done steps were reversed?

jdnavarro:
It looked alright before because the column with the Step number didn't actually refer to the `Step.id` in the engine as I initially thought, the numbers are just an enumeration for the table rows. The `Step.id` wasn't even in the shared model until a recent PR.

PR #160, opened 2017-02-02
SEQNG-76: sbt update to 0.13.13

cquiroz:
After a very frustrating attempt to port to sbt 0.13.13 using AutoPlugins and other techniques in an more elegant way I give up.

So this is the simplest option that work even though our `build.sbt` ends up being over 450 lines long


swalker2m:
I'm not well qualified to evaluate an SBT build, and I know you weren't happy with the result, but it does seem well documented and easy to find where each module is defined when we need to make changes.  Unless @tpolecat has any advice, it looks fine to me.

What is your estimation of how far apart the two builds, `ocs3` and `gem`, are at this point?

cquiroz:
It should be fairly easy to put them together, barring any incompatibilities at the library version level

cquiroz:
In case somebody is interested this is the alternative which I was pursuing using AutoPlugin. I'm not sure which one is worst

https://github.com/gemini-hlsw/ocs3/compare/develop...cquiroz:feature/SEQNG-76

tpolecat:
This looks reasonable to me. I don't understand auto-plugins. I have asked several people to explain them to me and they have all failed.

cquiroz:
I don't think the concept of AutoPlugins (in the limited sense I was using them) was so unclear. What killed it is how random settings work in AutoPlugin and some don't, same with settings in one place or another

Pure magic...

tpolecat:
Yeah it's terrible. Sorry for your suffering.

PR #161, opened 2017-02-02
SEQNG-123: Base implementation of FITS keywords.

jluhrs:
I had to scale down the scope of the task. Now it only include the infrastructure that will allow me to add the keywords.
I separated the writing of the keywords to the DHS from the reading of the values. That way I can test without accessing the EPICS systems.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/resources/DhsHdr.xml
+++ b/modules/edu.gemini.seqexec.server/src/main/resources/DhsHdr.xml
@@ -1,0 +1,1046 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<Records xmlns="http://www.gemini.edu/CaSchema">
+    <Top name="sfs">sfs</Top>
+    <Top name="tcs">tcs</Top>
+    <Top name="m2">m2</Top>
+    <Top name="ag">ag</Top>
+    <Top name="oiwfs">oiwfs</Top>
+    <Top name="gp">gp</Top>
+    <Top name="gsaoi">gsaoi</Top>
+    <Top name="ao">ao</Top>
+    <Top name="pwfs1">pwfs1</Top>
+    <Top name="ws">ws</Top>
+    <Top name="gc">gc</Top>
+    <Top name="sfo">sfo</Top>
+    <Top name="aom">aom</Top>
+    <Top name="myst">myst</Top>
+    <Top name="pwfs2">pwfs2</Top>
+    <Status name="dhshdr::gems">
+        <top>aom</top>
+        <attribute name="adcScLoop">
+            <channel>adcScLoop.VAL</channel>
+            <type>STRING</type>
+            <description>Science ADC Loop state (OPEN|CLOSED)</description>
+        </attribute>
+        <attribute name="ttLoop">
+            <channel>ttLoopStatus.VAL</channel>
+            <type>STRING</type>
+            <description>Status of the tip-tilt loop</description>
+            <top>myst</top>
+        </attribute>
+        <attribute name="flexLoop">
+            <channel>flexLoopStatus.VAL</channel>
+            <type>STRING</type>
+            <description>Status of the flexure loop</description>
+            <top>myst</top>
+        </attribute>
+        <attribute name="adcNgsLoop">
+            <channel>adcNgsLoop.VAL</channel>
+            <type>STRING</type>
+            <description>Ngs ADC Loop state (OPEN|CLOSED)</description>
+        </attribute>
+        <attribute name="odgw4x">
+            <channel>sad:dc:d_gdr_winXo4.VAL</channel>
+            <type>STRING</type>
+            <description>ODGW4 X center (pixels)</description>
+            <top>gsaoi</top>
+        </attribute>
+        <attribute name="astroMode">
+            <channel>slewReset.VALA</channel>
+            <type>STRING</type>
+            <description>Mode (None/Regular/Good/Best)</description>
+            <top>myst</top>
+        </attribute>
+        <attribute name="rZero">
+            <channel>r0.VAL</channel>
+            <type>STRING</type>
+            <description>r0 value at 500nm in cm</description>
+            <top>myst</top>
+        </attribute>
+        <attribute name="odwgSize">
+            <channel>sad:dc:d_gdr_winSizeS.VAL</channel>
+            <type>STRING</type>
+            <description>ODGW size</description>
+            <top>gsaoi</top>
+        </attribute>
+        <attribute name="focLoop">
+            <channel>focLoopStatus.VAL</channel>
+            <type>STRING</type>
+            <description>Status of the Slow Focus loop</description>
+            <top>myst</top>
+        </attribute>
+        <attribute name="odgw2x">
+            <channel>sad:dc:d_gdr_winXo2.VAL</channel>
+            <type>STRING</type>
+            <description>ODGW2 X center (pixels)</description>
+            <top>gsaoi</top>
+        </attribute>
+        <attribute name="lgsLoop">
+            <channel>lgsLoopStatus.VAL</channel>
+            <type>STRING</type>
+            <description>Status of the LGS loop</description>
+            <top>myst</top>
+        </attribute>
+        <attribute name="odgw3x">
+            <channel>sad:dc:d_gdr_winXo3.VAL</channel>
+            <type>STRING</type>
+            <description>ODGW3 X center (pixels)</description>
+            <top>gsaoi</top>
+        </attribute>
+        <attribute name="lgsStrehl">
+            <channel>StrehlEstimate.VAL</channel>
+            <type>STRING</type>
+            <description>Strehl estimate from the LGS in H-band</description>
+            <top>myst</top>
+        </attribute>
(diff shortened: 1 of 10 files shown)


modules/edu.gemini.seqexec.server/src/main/resources/DhsHdr.xml
```diff
@@ -0,0 +1,1046 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
jluhrs: This is the import of the Channel Access definition file used for the DHS keywords in the old Seqexec. Not yet used, but still included for future work.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
```diff
@@ -0,0 +1,35 @@
...
+      get.map(x => k.add(f(name, x)))
+  def buildInt8(get: SeqAction[Byte], name: String ): KeywordBag => SeqAction[KeywordBag]       = buildKeyword(get, name, Int8Keyword)
+  def buildInt16(get: SeqAction[Short], name: String ): KeywordBag => SeqAction[KeywordBag]     = buildKeyword(get, name, Int16Keyword)
+  def buildInt32(get: SeqAction[Int], name: String ): KeywordBag => SeqAction[KeywordBag]       = buildKeyword(get, name, Int32Keyword)
+  def buildFloat(get: SeqAction[Float], name: String ): KeywordBag => SeqAction[KeywordBag]     = buildKeyword(get, name, FloatKeyword)
+  def buildDouble(get: SeqAction[Double], name: String ): KeywordBag => SeqAction[KeywordBag]   = buildKeyword(get, name, DoubleKeyword)
+  def buildBoolean(get: SeqAction[Boolean], name: String ): KeywordBag => SeqAction[KeywordBag] = buildKeyword(get, name, BooleanKeyword)
+  def buildString(get: SeqAction[String], name: String ): KeywordBag => SeqAction[KeywordBag]   = buildKeyword(get, name, StringKeyword)
+
+  private def bundleKeywords(inst: String, ks: Seq[KeywordBag => SeqAction[KeywordBag]]): SeqAction[KeywordBag] = {
```
jluhrs: I'm sure that in a month, I will not understand this code.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import edu.gemini.seqexec.model.dhs.ObsId
+
+import edu.gemini.seqexec.server.DhsClient._
+
+/**
+  * Created by jluhrs on 1/31/17.
+  */
+
+trait TcsKeywordsReader {
```
jluhrs: More values to come. It is not yet used.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
jdnavarro: It may not be pretty, but it can be guessed what it does. Why the the `get` parameter and the result of `build...` are wrapped in a `SeqAction`?

modules/edu.gemini.seqexec.server/src/main/resources/DhsHdr.xml
cquiroz: Is this XML machine generated?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
jluhrs: Because most of the time the action of retrieving a value involves reading from an EPICS channel, and it can fail.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
```diff
@@ -28,19 +29,23 @@ final case class Flamingos2(f2Controller: Flamingos2Controller) extends Instrume
...
 
-  override def observe(config: Config): SeqObserve[DhsClient, ObserveResult] = Reader { client =>
+  val instContributorName = "flamingos2"
+  val dhsInstrumentName = "F2"
+
+  override def observe(config: Config): SeqObserve[(DhsClient, List[Header]), ObserveResult] = Reader { case (dhs, headers) =>
     for {
-      id <- client.createImage(DhsClient.ImageParameters(DhsClient.Permanent, List("flamingos2", "dhs-http")))
+      id <- dhs.createImage(DhsClient.ImageParameters(DhsClient.Permanent, List(instContributorName, "dhs-http")))
+      _  <- headers.map(_.sendBefore(id, dhsInstrumentName)).sequenceU
```
cquiroz: Is this a call to the dhs for each header?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouth.scala
```diff
@@ -30,16 +30,16 @@ object GmosSouth extends Instrument {
...
-      id <- client.createImage(DhsClient.ImageParameters(DhsClient.Permanent, List("gmos", "dhs-http")))
+      id <- dhs.createImage(DhsClient.ImageParameters(DhsClient.Permanent, List("gmos", "dhs-http")))
       _ <- EitherT(Task {
              Log.log(Level.INFO, name + ": starting observation " + id)
              Thread.sleep(5000)
              Log.log(Level.INFO, name + ": observation completed")
              TrySeq(())
            })
-      _ <- client.setKeywords(id, DhsClient.KeywordBag(
+      _ <- dhs.setKeywords(id, DhsClient.KeywordBag(
```
cquiroz: Why are these keywords hardcoded?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
cquiroz: You may need some shapeless there 😄

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
```diff
@@ -9,7 +9,7 @@ import scalaz.concurrent.Task
 trait Instrument extends System {
   // The name used for this instrument in the science fold configuration
   val sfName: String
-  def observe(config: Config): SeqObserve[DhsClient, ObserveResult]
+  def observe(config: Config): SeqObserve[(DhsClient, List[Header]), ObserveResult]
```
cquiroz: I guess we need to link this to the client level request to, to send client headers

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
cquiroz: Did you consider to have a single `get[A](name: String):A` type of method instead?

modules/edu.gemini.seqexec.server/src/main/resources/DhsHdr.xml
jluhrs: It was created running a program over the file that defined the channels in the old Seqexec. Then I had to change the types by hand, because the old format doesn't have type information. So yes, the file is machine generated, but not in the sense that it can be generated at build time, if that is what you were thinking.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
jluhrs: Yes.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouth.scala
jluhrs: No reason. That GMOS code is just a place holder.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
jluhrs: Not here. There is an ObsKeywordsReader trait used to get those values. The concrete must get them from the client.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
jluhrs: Yes, but decided against it. The implementation of `get[A](name: String):A` would still need to switch between all the possible values of name, so it is not less code. And misspelling a name would be a run time error instead of a compile time error.

PR #162, opened 2017-02-03
SEQNG-132 Real values for OCS keywords.

jluhrs:
Still missing the observer/operator names, and the current observing conditions. Those features will require a bit more work, so I'll put them in separate tasks. The main complication is that most of the OCS keywords are set when the sequence is loaded, but the observer/operator names and the current observing conditions are set when the sequence is run. I would need a mechanism to inject those values in the already constructed Actions. Maybe just rebuild the whole sequence?

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -4,6 +4,7 @@ import java.time.LocalDate
 
 import edu.gemini.seqexec.engine.{Action, Result, Sequence, Step}
 import edu.gemini.seqexec.model.Model.SequenceMetadata
+import edu.gemini.model.p1.immutable.Site
 import edu.gemini.seqexec.server.SeqexecFailure.UnrecognizedInstrument
 import edu.gemini.seqexec.server.ConfigUtilOps._
 import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
@@ -16,13 +17,9 @@ import scalaz._
 /**
   * Created by jluhrs on 9/14/16.
   */
-object SeqTranslate {
+class SeqTranslate(site: Site) {
 
-  case class Systems(
-                    dhs: DhsClient,
-                    tcs: TcsController,
-                    flamingos2: Flamingos2Controller
-                    )
+  import SeqTranslate._
 
   implicit def toAction(x: SeqAction[Result.Response]): Action = x.run map {
     case -\/(e) => Result.Error(SeqexecFailure.explain(e))
@@ -33,7 +30,8 @@ object SeqTranslate {
 
     def buildStep(inst: Instrument): Step[Action] = {
       val sys = List(Tcs(systems.tcs), inst)
-      val headers = List(new StandardHeader(systems.dhs, DummyObsKeywordsReader, DummyTcsKeywordsReader))
+      val headers = List(new StandardHeader(systems.dhs,
+        ObsKeywordReaderImpl(config, site.name.replace(' ', '-')), DummyTcsKeywordsReader))
 
       Step[Action](
         i,
@@ -74,3 +72,12 @@ object SeqTranslate {
 
 }
 
+object SeqTranslate {
+  def apply(site: Site) = new SeqTranslate(site)
+
+  case class Systems(
+                      dhs: DhsClient,
+                      tcs: TcsController,
+                      flamingos2: Flamingos2Controller
+                      )
+}
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -4,6 +4,7 @@ import java.time.LocalDate
 
 import edu.gemini.epics.acm.CaService
 import edu.gemini.pot.sp.SPObservationID
+import edu.gemini.model.p1.immutable.Site
 import edu.gemini.seqexec.engine
 import edu.gemini.seqexec.engine.Event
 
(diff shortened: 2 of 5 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -65,6 +66,27 @@ trait ObsKeywordsReader {
   def getTelescope: SeqAction[String]
 }
 
+case class ObsKeywordReaderImpl(config: Config, telescope: String) extends ObsKeywordsReader {
+  override def getObsType: SeqAction[String] = SeqAction(
+        config.getItemValue(new ItemKey(OBSERVE_KEY, OBSERVE_TYPE_PROP)).toString)
```
cquiroz: the indentation is a bit erratic

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -65,6 +66,27 @@ trait ObsKeywordsReader {
...
+
+  override def getObsId: SeqAction[String] = {
+    val v = config.getItemValue(new ItemKey(OCS_KEY, OBSERVATIONID_PROP)).toString.split("-").toList.lastOption.getOrElse("")
+
+    SeqAction(v)
+  }
+
+  override def getObservatory: SeqAction[String] = SeqAction(telescope)
+
+  override def getTelescope: SeqAction[String] = SeqAction(telescope)
```
cquiroz: is this correct? telescope and observatory have the same keyword value?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -33,7 +33,7 @@ object SeqTranslate {
 
     def buildStep(inst: Instrument): Step[Action] = {
       val sys = List(Tcs(systems.tcs), inst)
-      val headers = List(new StandardHeader(systems.dhs, DummyObsKeywordsReader, DummyTcsKeywordsReader))
+      val headers = List(new StandardHeader(systems.dhs, ObsKeywordReaderImpl(config, "Gemini-South"), DummyTcsKeywordsReader))
```
cquiroz: How does this work at GN?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
jluhrs: Yes

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -23,6 +24,8 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
 
   val odbProxy = new ODBProxy(new Peer(settings.odbHost, 8443, null))
 
+  val translator = SeqTranslate(settings.site)
```

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -150,6 +154,10 @@ object SeqexecEngine {
   def apply(settings: Settings) = new SeqexecEngine(settings)
 
   def seqexecConfiguration: Kleisli[Task, Config, Settings] = Kleisli { cfg: Config => {
+      val site = cfg.require[String]("seqexec-engine.site") match {
+        case "GS" => Site.GS
+        case "GN"     => Site.GN
```
cquiroz: Do you want to align the `=>`?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/test/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutesSpec.scala
```diff
@@ -12,27 +12,26 @@ import edu.gemini.seqexec.model.Model.SeqexecEvent.ConnectionOpenEvent
...
 import scalaz.OptionT
 import scalaz.concurrent.Task
 import scalaz.stream.async.mutable.{Queue, Topic}
-
 import org.scalatest.{FlatSpec, Matchers}
 
 class SeqexecUIApiRoutesSpec extends FlatSpec with Matchers with UriFunctions with ModelBooPicklers with CaseInsensitiveStringSyntax {
   val config = AuthenticationConfig(devMode = true, Hours(8), "token", "abc", useSSL = false, LDAPConfig(Nil))
-  val engine = SeqexecEngine(Settings("", LocalDate.now(), "", dhsSim = true, tcsSim = true, instSim = true, gcalSim = true, instForceError = false))
+  val engine = SeqexecEngine(Settings(Site.GS, "", LocalDate.now(), "", dhsSim = true, tcsSim = true, instSim = true, gcalSim = true, instForceError = false))
```
cquiroz: hmm, I guess we should put this as a setting, perhaps on another task

jdnavarro:
This is a similar problem we had with the `fileId`s showing before the observation finishes. We could always replace the whole Sequence if the astronomer name is changed or only the `Action` in the case of `fileId`s. But since I think we'll have more parameters that will be only known when actually executing and not when loading the sequence, I guess it's worth coming with a more general solution for this kind of scenarios.

Of the the solutions @jluhrs proposed for `fileId` the one I like the most that returns a function as a `Result`. That way we can *load* Sequences/Steps/Actions with unset parameters.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/test/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutesSpec.scala
jluhrs: Those are unit tests, I don't think it is necessary to make it configurable. The application uses a setting.

PR #163, opened 2017-02-07
SEQNG-09: Capture exceptions when executing actions.

jluhrs:
Implemented the capture of exception. This was bothering me today, because the test DHS was down, and that caused that the server crashed unceremoniously. 
For some reason the connection timeout exception didn't produce an error message.


jluhrs:
The goal is that if there is an exception triggered inside an Action, the server keeps running.

PR #165, opened 2017-02-10
SEQNG-133 Read TCS keywords from real TCS.

jluhrs:


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -120,10 +122,17 @@ package object engine {
...
+      e.printStackTrace(new PrintWriter(sw))
+      sw.toString
+    }
+
     // Send the expected event when the `Action` is executed
     def act(t: (Action, Int)): Task[Unit] = t match {
       case (action, i) =>
-        action.attempt.map(_.valueOr(e => Result.Error(e.getMessage))).flatMap {
+        action.attempt.map(_.valueOr(e =>
+          Result.Error(if (e.getMessage == null || e.getMessage.isEmpty) stackToString(e) else e.getMessage))).flatMap {
```
jluhrs: I had some bugs that caused exceptions, and for some reason `getMessage` returned `null`, leaving me with no clue of what was happening. I added the stack trace to get more information. That is very useful while we are developing, but shouldn't be left in the final product, at least not in its current form.

PR #167, opened 2017-02-13
SEQNG-134 Implemented Flamingos-2 keywords.

jluhrs:
This commit includes all the Flamingos-2 keywords set by Seqexec. There are two keywords in the old Seqexec that are filled with values read from the instrument. I implemented them, only to find that they are not actually sent to the DHS. I left the code in (except for the part that writes the keywords).


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
```diff
@@ -94,6 +94,11 @@ final class Flamingos2Epics(epicsService: CaService) {
 
   def windowCover: Option[String] = Option(f2State.getStringAttribute("windowCover").value)
 
+  // For FITS keywords
+  def health: Option[String] = Option(f2State.getStringAttribute("INHEALTH").value)
+
+  def state: Option[String] = Option(f2State.getStringAttribute("INSTATE").value)
```
jluhrs: These is one of the changes that I ended not using.

PR #169, opened 2017-02-16
SEQNG-143 Implemented data privacy keywords.

jluhrs:
Also include a change to the errors produced when reading values from the ODB sequence.


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ConfigUtilOps.scala
```diff
@@ -16,7 +16,14 @@ import Scalaz._
  */
 object ConfigUtilOps {
 
-  type ExtractFailure = String // for now
+  sealed trait ExtractFailure
```
jluhrs: I made the errors more specific to distinguish the case when a keyword is not present (which is OK for optional keywords) to the case when there is a type mismatch.

PR #171, opened 2017-02-28
Setters for Operator and Observer names in the server side

jdnavarro:
This is the server part for SEQNG-139. From the client it should be possible to be possible to call the functions wrapping the AJAX requests.

The primary place to store the names is in the Sequence metadata of the engine.

I wasn't sure about the format for the URL routes for the setters, so if you prefer a different scheme I can change it.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
@@ -17,6 +17,8 @@ case class Start(id: Sequence.Id) extends UserEvent
 case class Pause(id: Sequence.Id) extends UserEvent
 case class Load(id: Sequence.Id, sequence: Sequence[Action]) extends UserEvent
 case class Breakpoint(id: Sequence.Id, step: Step.Id, v: Boolean) extends UserEvent
+case class SetOperator(id: Sequence.Id, name: String) extends UserEvent
+case class SetObserver(id: Sequence.Id, name: String) extends UserEvent
 case object Poll extends UserEvent
 case object Exit extends UserEvent
 
@@ -35,11 +37,12 @@ object Event {
 
   def start(id: Sequence.Id): Event = EventUser(Start(id))
   def pause(id: Sequence.Id): Event = EventUser(Pause(id))
+  def load(id: Sequence.Id, sequence: Sequence[Action]): Event = EventUser(Load(id, sequence))
+  def breakpoint(id: Sequence.Id, step: Step.Id, v: Boolean): Event = EventUser(Breakpoint(id, step, v))
+  def setOperator(id: Sequence.Id, name: String): Event = EventUser(SetOperator(id, name))
+  def setObersver(id: Sequence.Id, name: String): Event = EventUser(SetObserver(id, name))
   val poll: Event = EventUser(Poll)
   val exit: Event = EventUser(Exit)
-  def load(id: Sequence.Id, sequence: Sequence[Action]): Event = EventUser(Load(id, sequence))
-  def breakpoint(id: Sequence.Id, step: Step.Id, v: Boolean): Event =
-    EventUser(Breakpoint(id, step, v))
 
   def failed(id: Sequence.Id, i: Int, e: Result.Error): Event = EventSystem(Failed(id, i, e))
   def completed[R<:RetVal](id: Sequence.Id, i: Int, r: OK[R]): Event = EventSystem(Completed(id, i, r))
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
@@ -17,7 +17,7 @@ case class Start(id: Sequence.Id) extends UserEvent
 case class Pause(id: Sequence.Id) extends UserEvent
 case class Load(id: Sequence.Id, sequence: Sequence[Action]) extends UserEvent
 case class Breakpoint(id: Sequence.Id, step: Step.Id, v: Boolean) extends UserEvent
-case class SetOperator(id: Sequence.Id, name: String) extends UserEvent
+case class SetOperator(name: String) extends UserEvent
 case class SetObserver(id: Sequence.Id, name: String) extends UserEvent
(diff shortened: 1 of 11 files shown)


jluhrs:
It look ok, but I am not sure about having the operator in the sequence. The operator is the same for all the current sequences, but the proposed model would allow to have a different operator for each sequence.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
```diff
@@ -17,6 +17,8 @@ case class Start(id: Sequence.Id) extends UserEvent
 case class Pause(id: Sequence.Id) extends UserEvent
 case class Load(id: Sequence.Id, sequence: Sequence[Action]) extends UserEvent
 case class Breakpoint(id: Sequence.Id, step: Step.Id, v: Boolean) extends UserEvent
+case class SetOperator(id: Sequence.Id, name: String) extends UserEvent
```
cquiroz: The operator is not sequence specific, it is global

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -56,6 +56,25 @@ class SeqexecCommandRoutes(auth: AuthenticationService, inputQueue: engine.Event
         newVal <- \/.fromTryCatchNonFatal(bp.toBoolean).fold(e => Task.fail(e), Task.now)
         _      <- se.setBreakpoint(inputQueue, obs, step, newVal)
         resp   <- Ok(s"Set breakpoint in step $step of sequence $obsId")
+
+      } yield resp
+
+    case POST -> Root / obsId / "operator" / name =>
```
cquiroz: The endpoints seem fine, except the fact that operator is not sequence specific

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
jluhrs: Then you don't need to pass the Sequence.Id

jdnavarro:
I placed the operator in the sequence because I considered it was the most conservative approach and the path of least resistance. In order to make it global there are some loose ends that I'm not sure how to solve.

From the  client  perspective, setting the operator name globally is straightforward, but will the client ever need to read the operator name? If so where should I put it? Putting it in the `SequenceView` would be equivalent to setting it in the engine `Sequence` at it is now. I'm not sure how to do it otherwise. I guess we could have yet another parameter being passed all over the code or having another HTTP route just to read the operator, but I'm not sure if that will make the code more *maintainable*.

Also, are we sure we won't need to access the operator name from the engine Sequence at any point? Not even when sending it through the DHS which will have to do soon? If we have code doing something at the Sequence level, as most engine code does, and we need the operator name, we'll have to provide it externally.

I'm not saying it can't be done but since we are already doing most of the heavy lifting at the Sequence level, I wonder if deviating from the usual pattern of having everything we need right there in the Sequence is worth the effort.

If you think it is a must have now, no problem, just let me know where to place the primary place to store the operator name and I'll try to make it flow from there to wherever it's needed.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -138,6 +142,10 @@ object Sequence {
 
     def getCurrentBreakpoint: Boolean
 
+    def setObserver(name: String): State
```
cquiroz: So this is still at the Sequence level, does it mean that the same operator value will be set on each sequence?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -260,12 +281,28 @@ object Sequence {
...
+
+      override def setObserver(name: String): State = observerL.set(self, Some(name))
+
       override val done: List[Step[Result]] = seq.steps
 
       override def mark(i: Int)(r: Result): State = self
 
       override val toSequence: Sequence[Action \/ Result] = seq.map(_.right)
 
+      private val sequenceL: Final @> Sequence[Result] =
```
cquiroz: What's `Final`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -84,6 +92,8 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
       case engine.Pause(_)            => SequencePauseRequested(svs)
       case engine.Load(_, _)          => SequenceLoaded(svs)
       case engine.Breakpoint(_, _, _) => StepBreakpointChanged(svs)
+      case engine.SetOperator(_)   => OperatorUpdated(svs)
```
cquiroz: align?

jluhrs:
I think this is better.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
jdnavarro: `Final` is the case class representing the final state of a `Sequence.State`. It is used when the Zipper is advancing. However this case class was not introduced in this PR, I just needed a Lens for it.

I guess this is confusing when reading the diffs but I think it's more clear when considering the whole file because `Final` is scoped inside the `Sequence.State`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
jdnavarro: Yes, it sets the same operator name for all the loaded Sequences. If you want the operator name in a single place, let me know where to put it because as I mentioned above, it's not obvious to me where to put it.

PR #174, opened 2017-03-02
SEQNG-152 Sent sequence notifications to the ODB.

jluhrs:
Implementation is a bit heterogeneous. The Sequence Start and Sequence Complete notifications are sent inside actions in the first an last steps. The Observation Abort and Observation Pause are sent as reactions to events.
There is an issue with the implementation of the Sequence Complete notification. Because it is an action, it may be posible to pause the sequence before it is executed.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
@@ -204,7 +204,7 @@ class DhsClientSim(date: LocalDate) extends DhsClient {
 
   override def createImage(p: ImageParameters): SeqAction[ImageFileId] =
     EitherT(Task.delay{
-      TrySeq(f"S${date.format(format)}${counter.incrementAndGet()}%04d")
+      TrySeq(f"S${date.format(format)}S${counter.incrementAndGet()}%04d")
     })
 
   override def setKeywords(id: ImageFileId, keywords: KeywordBag, finalFlag: Boolean): SeqAction[Unit] = SeqAction(())
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -28,7 +28,7 @@ class SeqTranslate(site: Site) {
     case \/-(r) => Result.OK(r)
   }
 
-  private def step(systems: Systems, settings: Settings)(obsId: SPObservationID, i: Int, config: Config): SeqexecFailure \/ Step[Action] = {
+  private def step(systems: Systems, settings: Settings)(obsId: SPObservationID, i: Int, config: Config, last: Boolean): SeqexecFailure \/ Step[Action] = {
 
     def buildStep(inst: Instrument, instHeaders: List[Header]): Step[Action] = {
 
@@ -63,7 +63,7 @@ class SeqTranslate(site: Site) {
           _ <- inst.observe(config)(id)
           _ <- headers.map(_.sendAfter(id, inst.dhsInstrumentName)).sequenceU
           _ <- closeImage(id, systems.dhs)
-          - <- sendDataEnd(id)
+          _ <- sendDataEnd(id)
         } yield ObserveResult(id)
       }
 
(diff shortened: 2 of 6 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
```diff
@@ -204,7 +204,7 @@ class DhsClientSim(date: LocalDate) extends DhsClient {
 
   override def createImage(p: ImageParameters): SeqAction[ImageFileId] =
     EitherT(Task.delay{
-      TrySeq(f"S${date.format(format)}${counter.incrementAndGet()}%04d")
+      TrySeq(f"S${date.format(format)}S${counter.incrementAndGet()}%04d")
```
cquiroz: Is the S from `Science`? If so, should it be configurable for non-science?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -70,10 +70,16 @@ class SeqTranslate(site: Site) {
         None,
         config.toStepConfig,
         false,
+        (if(i==0) List(List(toAction(systems.odb.sequenceStart(obsId, "").map(_ => Result.Ignored))))
```
cquiroz: Can you explain this bits a little?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -66,7 +66,7 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
   def requestRefresh(q: engine.EventQueue): Task[Unit] = q.enqueueOne(Event.poll)
 
   def eventProcess(q: engine.EventQueue): Process[Task, SeqexecEvent] =
-    engine.process(q)(engine.initState).map {
+    engine.process(q)(engine.initState).flatMap(x => Process.eval(notifyODB(x))).map {
```
jdnavarro: I can't find `notifyODB`, where is it coming from?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -99,7 +105,7 @@ class SeqTranslate(site: Site) {
     val configs = sequenceConfig.getAllSteps.toList
 
     val steps = configs.zipWithIndex.traverseU {
-      case (c, i) => step(systems, settings)(obsId, i, c)
+      case (c, i) => step(systems, settings)(obsId, i, c, i==(configs.length-1))
```
cquiroz: `i == (configs.length - 1)`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jdnavarro: Never mind, I just found it.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClient.scala
jluhrs: I am not sure what the second is for (Summit/Base ?). But this is DhsClientSim, I don't think it needs more work.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: So, sequenceStart notification is sent at the beginning of the first step in the sequence. Here I know it is the first step, because its index is 0.
SequenceComplete must be sent at the end of the last step. I have the index, but to identify the last one here I need to know the total number of steps. Instead of passing that value, I directly pass a flag.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: See previous comment.

cquiroz: I was commenting the lack of spaces between `i` and `==`

jluhrs: Did I mention how much I hate you ?

cquiroz: good-let-the-hate-flow-through-you

PR #175, opened 2017-03-02
SEQNG-139A: Add Operator and Astronomer name to the client

cquiroz:
This is the first PR for SEQNG-139. It enables the Observer text fiel on the UI at the sequence level. It took quite a bit of effort to get it to play correctly given the way React works, there maybe still some use cases not covered

The field will take the name of the logged in user and will update the backend when loosing the focus.

Still missing:

[ ] Update the backend before run
[ ] Update the backend while running
[ ] More testing
[ ] Complete the Operator field

Here's a screenshot in FF

 
screenshot 2017-03-02 15 52 33


andrewwstephens:
The screenshot looks good!

PR #177, opened 2017-03-07
SEQNG-155A: Add Operator name to the client

cquiroz:
This updates the UI to make the `Operator` field functional. There is still some work to do since the upstream model doesn't provide a single global operator field but this should be good for early testing


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/HeadersSideBar.scala
```diff
@@ -1,28 +1,37 @@
...
-            <.input(
-              ^.`type` :="text",
-              ^.autoComplete :="off"
-            )
+            Label(Label.Props("Operator", "operator")),
+            Input(Input.Props("operator", "operator",
+              p.operator().getOrElse(""),
+              placeholder = "Operator...",
+              disabled = !p.status().isLogged,
+              onBlur = name => p.operator.dispatchCB(UpdateOperator(name))))
```
jdnavarro: What does `onBlur` do?

cquiroz: Good question. `onBlur` is a javascript event that you get when the element looses focus. This is not totally optimal and some extra work is needed to get it working at all times. You'd think that we should use `onChange` which is triggered every time the user presses something but that causes undesirable UI artifacts as you'd update the frontend every time the users presses a key

PR #180, opened 2017-03-09
SEQNG-158 Initial support for resource management

jdnavarro:
Sequences can be now run in parallel when there are no `Resource`s shared among other *running* Sequences. Otherwise the Sequence can't be run.

The primary store for resources is the `Step`, from there the resources needed by a `Sequence` are calculated.

This is an initial implementation, there are still several things to polish. Off the top of my head:

 - Find out at load time resources are needed for each Step. The test use synthetic Sequences with resources hardcoded, but in the real application there is no way to create them.
 - Complete types of resources.
 - Check resources in Pending steps only. Now it takes into account the completed ones too. - Provide some kind of feedback when a attempting to run a Sequence fails because of shared resources.
 - Add flag to indicate which Sequences would be ready for running.
 - More testing.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Resource.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Resource.scala
@@ -1,0 +1,18 @@
+package edu.gemini.seqexec.engine
+
+/**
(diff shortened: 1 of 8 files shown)


cquiroz:
The code doesn't provide much to understand what's going on. Could you briefly explain what does it do?

jdnavarro:
We explained what the code does face to face.

PR #182, opened 2017-03-13
Versions update

cquiroz:
Some libraries updates:

* http4s
* jwt
* squants

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -43,18 +43,19 @@ object Settings {
     val scalaZStream = "0.8.6a"
 
     // Scala libraries
-    val http4s       = "0.15.2a"
-    val squants      = "1.0.0"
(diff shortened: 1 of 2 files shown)


project/Settings.scala
```diff
@@ -43,18 +43,19 @@ object Settings {
     val scalaZStream = "0.8.6a"
 
     // Scala libraries
-    val http4s       = "0.15.2a"
-    val squants      = "1.0.0"
+    val http4s       = "0.15.6a"
```
jdnavarro: I guess there won't be any changes in `http4s` relevant to us yet, right?

cquiroz: Not really. 0.16.0 is a bigger but it uses cats-fs2 so we'll have to see when do we migrate

PR #183, opened 2017-03-14
SEQNG-163 weather keywords

jluhrs:
I also modified the way the default values for FITS keywords are handled. Now they are defined in one place. If the value to be written is an Option, and it is None, then the default value is used.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
@@ -22,7 +22,7 @@ final class GwsEpics private (epicsService: CaService) {
   def windVelocity: Option[Velocity] = Option(state.getDoubleAttribute("windspee").value).map(v => MetersPerSecond(v.doubleValue))
   def airPressure: Option[Pressure] = Option(state.getDoubleAttribute("pressure").value).map(v => Bars(v.doubleValue*Milli))
   def ambientT: Option[Temperature] = Option(state.getDoubleAttribute("tambient").value).map(v => Celsius(v.doubleValue))
-  def health: Option[String] = Option(state.getStringAttribute("health").value)
+  def health: Option[Int] = Option(state.getIntegerAttribute("health").value)
   def dewPoint: Option[Temperature] = Option(state.getDoubleAttribute("dewpoint").value).map(v => Celsius(v.doubleValue))
   def windDirection: Option[Angle] = Option(state.getDoubleAttribute("winddire").value).map(v => Degrees(v.doubleValue))
 
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
@@ -2,12 +2,13 @@ package edu.gemini.seqexec.server
 
 import java.util.logging.Logger
 
-import squants.{Temperature, Velocity}
-import edu.gemini.epics.acm.{CaService, XMLBuilder}
-import squants.motion.{Bars, MetersPerSecond, Pressure}
 import squants.MetricSystem.Milli
+import squants.motion.{Bars, MetersPerSecond, Pressure}
 import squants.space.{Angle, Degrees}
 import squants.thermal.Celsius
+import squants.{Temperature, Velocity}
+
+import edu.gemini.epics.acm.{CaService, XMLBuilder}
 
 /**
   * GwsEpics wraps the non-functional parts of the EPICS ACM library to interact with the Weather Server.
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
@@ -1,0 +1,58 @@
+package edu.gemini.seqexec.server
+
+import java.util.logging.Logger
+
+import squants.{Temperature, Velocity}
+import edu.gemini.epics.acm.{CaService, XMLBuilder}
+import squants.motion.{Bars, MetersPerSecond, Pressure}
+import squants.MetricSystem.Milli
+import squants.space.{Angle, Degrees}
+import squants.thermal.Celsius
+
+/**
+  * GwsEpics wraps the non-functional parts of the EPICS ACM library to interact with the Weather Server.
+  * It has all the objects used to read TCS status values and execute TCS commands.
+  *
+  * Created by jluhrs on 3/10/17.
+  */
+final class GwsEpics private (epicsService: CaService) {
+  private val state = epicsService.getStatusAcceptor("gws::state")
+
+  def humidity: Option[Double] = Option(state.getDoubleAttribute("humidity").value).map(_.doubleValue)
+  def windVelocity: Option[Velocity] = Option(state.getDoubleAttribute("windspee").value).map(v => MetersPerSecond(v.doubleValue))
+  def airPressure: Option[Pressure] = Option(state.getDoubleAttribute("pressure").value).map(v => Bars(v.doubleValue*Milli))
+  def ambientT: Option[Temperature] = Option(state.getDoubleAttribute("tambient").value).map(v => Celsius(v.doubleValue))
+  def health: Option[String] = Option(state.getStringAttribute("health").value)
+  def dewPoint: Option[Temperature] = Option(state.getDoubleAttribute("dewpoint").value).map(v => Celsius(v.doubleValue))
+  def windDirection: Option[Angle] = Option(state.getDoubleAttribute("winddire").value).map(v => Degrees(v.doubleValue))
+
+}
+
+object GwsEpics extends EpicsSystem {
+  val Log = Logger.getLogger(getClass.getName)
+  val CA_CONFIG_FILE = "/Gws.xml"
+  val GWS_TOP = "ws:"
+
+  private var instanceInternal = Option.empty[GwsEpics]
+  lazy val instance: GwsEpics = instanceInternal.getOrElse(
+    throw new Exception("Attempt to reference TcsEpics single instance before initialization."))
+
+  override def init(service: CaService): TrySeq[Unit] = {
+    try {
+      (new XMLBuilder).fromStream(this.getClass.getResourceAsStream(CA_CONFIG_FILE))
+        .withCaService(service)
+        .withTop("gws", GWS_TOP)
+        .buildAll()
+
+        instanceInternal = Some(new GwsEpics(service))
+
+        TrySeq(())
+
+    } catch {
+      case c: Throwable =>
+        Log.warning("GwsEpics: Problem initializing EPICS service: " + c.getMessage + "\n"
+          + c.getStackTrace.mkString("\n"))
+        TrySeq.fail(SeqexecFailure.SeqexecException(c))
+    }
+  }
+}
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
@@ -32,4 +32,15 @@ object Header {
     _   <- hs.setKeywords(id, bag, false)
   } yield ()
 
+
+  object Defaults {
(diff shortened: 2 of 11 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
```diff
@@ -0,0 +1,58 @@
+package edu.gemini.seqexec.server
+
+import java.util.logging.Logger
+
+import squants.{Temperature, Velocity}
```
cquiroz: 👍

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
```diff
@@ -0,0 +1,58 @@
...
+/**
+  * GwsEpics wraps the non-functional parts of the EPICS ACM library to interact with the Weather Server.
+  * It has all the objects used to read TCS status values and execute TCS commands.
+  *
+  * Created by jluhrs on 3/10/17.
+  */
+final class GwsEpics private (epicsService: CaService) {
+  private val state = epicsService.getStatusAcceptor("gws::state")
+
+  def humidity: Option[Double] = Option(state.getDoubleAttribute("humidity").value).map(_.doubleValue)
```
cquiroz: Do these calls handle errors, like channel not found or wrong type?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
```diff
@@ -0,0 +1,58 @@
+package edu.gemini.seqexec.server
+
+import java.util.logging.Logger
+
+import squants.{Temperature, Velocity}
+import edu.gemini.epics.acm.{CaService, XMLBuilder}
+import squants.motion.{Bars, MetersPerSecond, Pressure}
+import squants.MetricSystem.Milli
+import squants.space.{Angle, Degrees}
+import squants.thermal.Celsius
```
cquiroz: We could group the imports e.g. `squants` first then `edu.gemini.*`

cquiroz: Even better you could order them alphabetically

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
```diff
@@ -0,0 +1,58 @@
...
+  def windDirection: Option[Angle] = Option(state.getDoubleAttribute("winddire").value).map(v => Degrees(v.doubleValue))
+
+}
+
+object GwsEpics extends EpicsSystem {
+  val Log = Logger.getLogger(getClass.getName)
+  val CA_CONFIG_FILE = "/Gws.xml"
+  val GWS_TOP = "ws:"
+
+  private var instanceInternal = Option.empty[GwsEpics]
```
cquiroz: I wonder if this could be done without resorting to a var. Could you do an `apply` that takes the `CaService`, initializes the `GwsEpics` and returns something like `Throwable \/ GwsEpics`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
```diff
@@ -32,4 +32,15 @@ object Header {
     _   <- hs.setKeywords(id, bag, false)
   } yield ()
 
+
+  object Defaults {
+    implicit def fromStringOption(v: SeqAction[Option[String]]): SeqAction[String] = v.map(_.getOrElse("No Value"))
```
cquiroz: Where are the defaults defined like this for the FITS files?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/KeywordsReader.scala
```diff
@@ -0,0 +1,9 @@
+package edu.gemini.seqexec.server
+
+/**
+  * Created by jluhrs on 3/13/17.
+  */
+object KeywordsReader {
+  implicit def fromAnyValue[A](v: A): SeqAction[Option[A]] = SeqAction(Some(v))
+  implicit def fromOption[A](v: Option[A]): SeqAction[Option[A]] = SeqAction(v)
```
cquiroz: Nice

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -223,12 +225,26 @@ object SeqexecEngine {
...
       else Task(())
+      val gwsInit = if(gwsKeywords) initEpicsSystem(GwsEpics) else Task(())
 
       tcsInit *>
+        gwsInit *>
         instInit *>
         (for {
           now <- Task(LocalDate.now)
-        } yield Settings(site, odbHost, now, dhsServer, dhsSim, tcsSim, instSim, gcalSim, odbNotifications, tcsKeywords, f2Keywords, instForceError) )
+        } yield Settings( site,
```
cquiroz: `Settings( site`? shouldn't it be `Settings(site`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -42,7 +43,7 @@ trait ObsKeywordsReader {
   def getGeminiQA: SeqAction[String]
   def getPIReq: SeqAction[String]
   def getSciBand: SeqAction[Int]
-  def getRequestedAirMassAngle: Map[String, SeqAction[String]]
+  def getRequestedAirMassAngle: Map[String, SeqAction[Double]]
```
cquiroz: Thanks for catching these

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -208,43 +209,64 @@ class StandardHeader(
...
+  import Header.Defaults._
+  import KeywordsReader._
 
   override def sendBefore(id: ImageFileId, inst: String): SeqAction[Unit] = {
 
-    val p: SeqAction[Double] = for {
-      xoff <- tcsReader.getXOffset
-      yoff <- tcsReader.getYOffset
-      iaa  <- tcsReader.getInstrumentAA
+    val p: SeqAction[Option[Double]] = for {
```
cquiroz: There seems to be some repetition on the calculations of `p`, `q`, etc. Are there any opportunities to reduce duplicated code?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -271,12 +293,58 @@ class StandardHeader(
...
+        {
+          val x = gwsReader.getTemperature.map(_.map(_.toCelsiusScale))
+          buildDouble(x, "TAMBIENT")
+        },
+        {
+          val x = gwsReader.getTemperature.map(_.map(_.toFahrenheitScale))
+          buildDouble(x, "TAMBIEN2")
+        },
+        {
+          val x = gwsReader.getAirPressure.map(_.map(_.to(new PressureUnit {
```
cquiroz: Is this not available in `squants`? If so we could make a PR

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -310,9 +378,9 @@ class StandardHeader(
         case (i, tw) =>
           List(
             buildString(tw.start,    f"REQTWS${i + 1}%02d"),
-            buildDouble(tw.duration, f"REQTWD${i + 1}02d"),
-            buildInt32(tw.repeat,    f"REQTWN${i + 1}02d"),
-            buildDouble(tw.duration, f"REQTWD${i + 1}02d"))
+            buildDouble(tw.duration, f"REQTWD${i + 1}%02d"),
+            buildInt32(tw.repeat,    f"REQTWN${i + 1}%02d"),
+            buildDouble(tw.period,   f"REQTWP${i + 1}%02d"))
```
cquiroz: Thanks again

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
jluhrs: The `getXXXAtribute` methods return `null` if there is a problem. So, if there is a problem, the end result is `None`.

cquiroz: 👍

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
jluhrs: It is a way of handling a singleton that requires initialization parameters.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
jluhrs: Those are the defaults defined for FITS keywords.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: I wanted to align it with the tabs

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
jluhrs: Well, the replication is mostly in the extraction of the values, so maybe we could put a common function that returns a triplet

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
jluhrs: Nops. Squants defines Bar, Pascal, Atmosphere and PSI, but not mmHg nor Tor.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsEpics.scala
jluhrs: To be clear, I will not change this in this PR.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Header.scala
jluhrs: Maybe I didn't understand the question. What do you expect me to do to declare this issue resolved?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
jluhrs: I will not change this in this PR.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
jluhrs: This matter shouldn't prevent for the PR to be approved.

cquiroz: Not at all.

cquiroz:
I'm ok in general with the PR. My only concern is on how the `GwsEpics` is initialized. I think we have tried to avoid constructing objects that could throw an exception in initialization

jluhrs:
@cquiroz Every Epics class is initialized the same way (note to myself: consolidate the initialization common code). Trying to use the singleton before initialization reveals an error in the code, which, IMHO, justifies the use of exceptions.

PR #184, opened 2017-03-15
SEQNG-164 Specify resources for F2 Step

jdnavarro:
This is a very basic way to specify resources for F2. Adding more instruments by just adding more case patterns to `instrumentResources`.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -2,7 +2,7 @@ package edu.gemini.seqexec.server
 
 import edu.gemini.model.p1.immutable.Site
 import edu.gemini.pot.sp.SPObservationID
-import edu.gemini.seqexec.engine.{Action, Result, Sequence, Step}
+import edu.gemini.seqexec.engine.{Action, Result, Resource, Sequence, Step}
 import edu.gemini.seqexec.model.Model.SequenceMetadata
 import edu.gemini.seqexec.model.dhs.ImageFileId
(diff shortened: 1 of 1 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -102,6 +102,11 @@ class SeqTranslate(site: Site) {
 
   }
 
+  def instrumentResources(inst: Instrument): Set[Resource] = inst match {
+    case Flamingos2(_) => Set(Resource.Mount, Resource.ScienceFold, Resource.F2)
```
cquiroz: Does this mean that to use flamingos2 you need the `Mount`, `ScienceFold` and `F2` available?
Is the check done when it is constructed on when it runs?

jdnavarro: That's set when loading a Sequence and set per Step. The `Resource` data types are provisional, we have to come up with a comprehensive list of resources that we'll use for real and carefully pick which resources will be assigned for each kind of Step.

cquiroz: That's a good quesition

jdnavarro: > Will it be possible to specify resources based on the type of observation?

Yes, it can be done easily provided there is some way to guess type of Step, i.e. from the Step configuration or from keywords, when loading a Sequence.

We still have to create manually the *inventory* of types of Steps with their correspondent resources though.

PR #185, opened 2017-03-15
SEQNG-147 Implemented EPICS commands and status for GCAL.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/java/edu/gemini/seqexec/server/gcal/BinaryOnOff.java
```diff
@@ -0,0 +1,9 @@
+package edu.gemini.seqexec.server.gcal;
```
cquiroz: Any reason to do this in java?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
```diff
@@ -42,6 +42,10 @@ object EpicsCommand {
     case \/-(e) => e
   })
 
+  def setParameter[A,T](p: Option[CaParameter[T]], v: A, f: A => T): SeqAction[Unit] =
+    safe(EitherT(Task.delay {
```
cquiroz: What's `safe`?

cquiroz: I guess you can use `SeqAction.either` here

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalEpics.scala
```diff
@@ -0,0 +1,138 @@
...
+/**
+  * Created by jluhrs on 3/14/17.
+  */
+class GcalEpics(epicsService: CaService) {
+
+  import EpicsCommand.setParameter
+  import GcalEpics._
+
+  object shutterCmd extends EpicsCommand {
+    override val cs = Option(epicsService.getCommandSender("gcal::shutter"))
```
cquiroz: I'm pretty sure you're correct here, but I've normally seen only one `:` as an epics separator

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalEpics.scala
```diff
@@ -0,0 +1,138 @@
...
+  def diffuser: Option[String] = Option(state.getStringAttribute("diffuser").value)
+}
+
+object GcalEpics extends EpicsSystem {
+
+  val Log = Logger.getLogger(getClass.getName)
+  val CA_CONFIG_FILE = "/Gcal.xml"
+  val GCAL_TOP = "gc:"
+
+  private var instanceInternal = Option.empty[GcalEpics]
```
cquiroz: Same comment as on the previous PR, I wish we had a cleaner way to initialize this

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalEpics.scala
```diff
@@ -0,0 +1,138 @@
...
+        Log.warning("GcalEpics: Problem initializing EPICS service: " + c.getMessage + "\n"
+          + c.getStackTrace.mkString("\n"))
+        TrySeq.fail(SeqexecFailure.SeqexecException(c))
+    }
+  }
+
+  class EnumAttribute[T <: Enum[T]](sa: CaStatusAcceptor, attrName: String, attrChannel: String, desc: String)(c: Class[T]) {
+    private val attribute: Option[CaAttribute[T]] = Option(sa.addEnum(attrName,
+        GCAL_TOP + attrChannel, c, desc))
+    def apply: Option[T] = attribute.flatMap(v => Option(v.value))
```
cquiroz: Is this in case the enumeration value is `null`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -226,9 +226,11 @@ object SeqexecEngine {
         Nondeterminism[Task].gatherUnordered(List(Flamingos2Epics).map(initEpicsSystem(_)))
       else Task(())
       val gwsInit = if(gwsKeywords) initEpicsSystem(GwsEpics) else Task(())
+      val gcalInit = if(!gcalSim) initEpicsSystem(GcalEpics) else Task(())
```
cquiroz: I think it is slightly better to do `Task.now(())` here

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalEpics.scala
jluhrs: That is not an EPICS channel name, but the name assigned to a Command Sender. Because all the Command Senders are in the same space, the name use that convention to avoid conflicts.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalEpics.scala
jluhrs: v.value can be null if it is not read from the corresponding EPICS channel.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalEpics.scala
cquiroz: 👍

modules/edu.gemini.seqexec.server/src/main/java/edu/gemini/seqexec/server/gcal/BinaryOnOff.java
jluhrs: Yes. I have a call to a java class method (CaStatusServer.addEnum) that requires an java Enum class as a parameter.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
jluhrs: `safe` is defined in the same file. It transforms a SeqAction to capture exceptions and put them in the Task output as a left value.

PR #186, opened 2017-03-15
SEQNG-162: Improve text fields change detection

cquiroz:
This is one of the trickiust UI related PRs so far. It enables the observer and operator field to autodetect when they change and submit to the backend. It is not ideal to do this on every char typed, lest it produce too much traffic, so the interim solution was to submit when the input field looses focus. However this fails if the user stay on the text field

The solution un this PR sets up a timer to detect changes every 2 seconds and submits

Several border cases made this a bit harder but it seems to work satisfactorily and has been tested with multiple clients to do the right thing

As usual for UI PRs the code doesn't tell that much

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/HeadersSideBar.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/HeadersSideBar.scala
@@ -10,7 +10,11 @@ import japgolly.scalajs.react.extra.{ExternalVar, TimerSupport}
 import japgolly.scalajs.react.vdom.prefix_<^._
 import org.scalajs.dom.html.Div
 
+import scalaz.syntax.equal._
+import scalaz.std.string._
+import scalaz.std.option._
 import scala.concurrent.duration._
+import scalaz.Equal
 
 /**
   * Display to show headers per sequence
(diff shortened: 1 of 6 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/HeadersSideBar.scala
```diff
@@ -1,24 +1,44 @@
...
+    def updateState(value: String): Callback =
+      $.modState(_.copy(currentText = Some(value)))
+
+    def submitIfChanged: Callback =
+      ($.state zip $.props) >>= {
+        case (s, p) => Callback.when(s.currentText =/= p.operator())(Callback.empty) // We are not submitting until the backend properly update this property
+      }
+
+    def render(p: Props, s: State): ReactTagOf[Div] = {
+      val operatorEV = ExternalVar(s.currentText.getOrElse(""))(updateState)
```
jdnavarro: What are `ExternalVar`s for?

cquiroz: They are essentially safe vars that can be used to communicate across components. It is a `scalajs-react` facility that in this case lets state flow from the component `InputEV` to its parent` HeadersSideBar`, complementing the strict top-down data flow used by `diode`

It was the only way I found to this without inducing UI glitches, in particular, the input element would loose focus when typing doing the top-down approach

PR #187, opened 2017-03-17
Make EngineState a data type

jdnavarro:
Previously `EngineState` was a type synonym to a map of `Id`s to `Sequence`s. Now it's data type wrapper for that Map. This facilitates adding global values as parameters of that type.

I also renamed the previous `Queue` object to `Engine` where I declared the new data type and removed all the previous code.


jdnavarro:
I forgot to update the code outside of the engine 🤦‍♂️

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -0,0 +1,27 @@
+package edu.gemini.seqexec.engine
+
+import scalaz._
+import Scalaz._
+
+/**
+  * A Map of `Sequence`s.
+  */
+case class Engine[+A](sequences: Map[Sequence.Id, Sequence[A]])
```
cquiroz: Why's that you need variance here?

cquiroz: Is the idea that we'd add more components to engine, like Observing condiitions?

jdnavarro: > Why's that you need variance here?
Theoretically I still struggling to understand subtyping in Scala, so I don't really know why it's required, but in practice I needed it for the Functor instance. We don't use the Functor instance anywhere yet, but I carried it over from the previous `Queue` data type. It may come handy if we want to map over all existing `Execution`s or to create a `Foldable` instance.

> Is the idea that we'd add more components to engine, like Observing conditions?
Yes, I'll do it right away in another PR. When I started I had the operator in mind but it turns out the conditions may be an easier first global parameter.

jluhrs: The next step would be to add another parameter to Engine for the global values, or am I not understanding it?

jdnavarro: I'm adding the conditions parameter to the Engine, but if you prefer we could put all the global values in a single Engine parameter (similar to `SequenceMetadata`). Is that what you have in mind?

jluhrs: Not really. It was not my intention to suggest it should be a single parameter, only that something else will be added.

PR #188, opened 2017-03-20
SEQNG-167 Implemented configuration of GCAL EPICS system.

jluhrs:
Model to represent the GCAL configuration. It can be the configuration read from the real system, or the configuration that is meant to be applied.
There are cases when certain elements of GCAL should be left alone. For that, I opted to make every value in the configuration an Option. If any of the values in a requested configuration is None, then that GCAL element will not be configured. In the read configuration a None means the value cannot be read, or the value read cannot be converted.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalController.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalController.scala
@@ -1,0 +1,87 @@
+package edu.gemini.seqexec.server
+
+import edu.gemini.spModel.gemini.calunit.CalUnitParams.Shutter
+
+/**
+  * Created by jluhrs on 3/15/17.
+  */
+trait GcalController {
+
+  import GcalController._
+
+  def getConfig: SeqAction[GcalConfig]
+
+  def applyConfig(config: GcalConfig): SeqAction[Unit]
+
+}
+
+object GcalController {
+
+  trait LampState
+
+  object LampState {
+
+    case object Off extends LampState
+
+    case object On extends LampState
+
+  }
+
+  final case class ArLampState(self: LampState) extends AnyVal
+
+  final case class CuArLampState(self: LampState) extends AnyVal
+
+  final case class QHLampState(self: LampState) extends AnyVal
+
+  final case class ThArLampState(self: LampState) extends AnyVal
+
+  final case class XeLampState(self: LampState) extends AnyVal
+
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalController.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  final case class CuArLampState(self: LampState) extends AnyVal
+
+  final case class QHLampState(self: LampState) extends AnyVal
+
+  final case class ThArLampState(self: LampState) extends AnyVal
+
+  final case class XeLampState(self: LampState) extends AnyVal
+
+  final case class IrLampState(self: LampState) extends AnyVal
```
jluhrs: Those classes are to avoid confusion when building a configuration. They make it harder to swap them by mistake.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalController.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  type Shutter = edu.gemini.spModel.gemini.calunit.CalUnitParams.Shutter
```
cquiroz: Any reason to create a type alias with the same name?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalControllerEpics.scala
```diff
@@ -0,0 +1,184 @@
...
+    (v: Diffuser) =>
+      v match {
+        case Diffuser.IR => "IR"
+        case Diffuser.VISIBLE => "VISIBLE"
+      }
+  }
+
+  private def getDiffuser: Option[Diffuser] = for {
+    x <- GcalEpics.instance.diffuser
+    y <- decode[String, Option[Diffuser]](x)
```
cquiroz: Nice 👍

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalController.scala
jluhrs: It's there just in case I need to make it a different class.

jluhrs:
I agree it would be good to have tests, but I don't know how to test reading from an external system.

PR #189, opened 2017-03-21
SEQNG-173 Make `Conditions` a global value in Engine

jdnavarro:
This is still not ready to merge because I have to implement the conditions in the server but I wanted to get early feedback at engine level in case I'm doing something wrong.

Notice that now `Conditions` is imported from the ODB modell, so the `SeqexecODB` dependency was added to the `engine`.

`Conditions.WORST`, which under the hood sets all conditions to `ANY`, is the default value but you might want some other default.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
@@ -3,6 +3,8 @@ package edu.gemini.seqexec.engine
 import scalaz._
 import Scalaz._
 
+import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.Conditions
+
 /**
   * A Map of `Sequence`s.
   */
@@ -22,11 +24,12 @@ object Engine {
       Engine(q.sequences.mapValues(_.map(f)))
   }
 
-  case class State(sequences: Map[Sequence.Id, Sequence.State])
+  case class State(conditions: Conditions, sequences: Map[Sequence.Id, Sequence.State])
 
   object State {
 
-    def empty: State = State(Map.empty)
+    // WORST sets the value ANY for every condition
+    def empty: State = State(Conditions.WORST, Map.empty)
 
   }
 
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
@@ -3,7 +3,7 @@ package edu.gemini.seqexec.engine
 import scalaz._
 import Scalaz._
 
-import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.Conditions
+import edu.gemini.seqexec.model.Model.Conditions
 
 /**
   * A Map of `Sequence`s.
@@ -29,7 +29,7 @@ object Engine {
   object State {
 
     // WORST sets the value ANY for every condition
-    def empty: State = State(Conditions.WORST, Map.empty)
+    def empty: State = State(Conditions.worst, Map.empty)
 
   }
 
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -55,7 +55,7 @@ object Model {
   type Operator = String
   type Observer = String
 
-  implicit val equal: Equal[SequenceId] = Equal.equalA[SequenceId]
+  implicit val equalSequenceId: Equal[SequenceId] = Equal.equalA[SequenceId]
 
   sealed trait StepState {
     def canRunFrom: Boolean = false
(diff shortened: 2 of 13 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -22,11 +24,12 @@ object Engine {
       Engine(q.sequences.mapValues(_.map(f)))
   }
 
-  case class State(sequences: Map[Sequence.Id, Sequence.State])
+  case class State(conditions: Conditions, sequences: Map[Sequence.Id, Sequence.State])
```
cquiroz: 👍

cquiroz:
Looks good in general to me, the `Conditions.WORST` may not be correct but it probably doesn't matter given that we are loading them from the ODB.

jdnavarro:
Does it make sense from the user's perspective to have no conditions at all at any point?

cquiroz:
I was wondering about that too, I'd say no, we provide a default which should match the ones on the OT

jluhrs:
As an alternative, we could make the type of the first parameter generic. The question is, does the engine need to now the details of what is in there? I guess the answer is linked to our final solution for how to pass changing parameters to the Actions. If the solution is to have the Actions instanced just before being run, then the instancing function could have the same parameter as input, in which case the engine can be agnostic of its type.

jdnavarro:
> does the engine need to now the details of what is in there?

Right now it doesn't, but at some point it'll have to check, I guess when is about run a Step, whether the actual conditions match the required conditions.

But even if that wasn't the case, I don't see much benefit *right now* on making the first parameter polymorphic.

- Is there ever going to be another Engine which when being called won't have the Conditions as a parameter?

- Do we want to have an entire Engine where everything where the Conditions will be replaced by some other type? If we need to do something with the Conditions, we can manipulate it aside, but do we need to carry the whole Engine when doing so?

- I see the need for polymorphism for the functions representing a *to be resolved at run time* Action, but why they have to be constrained to match the type of the first global parameter? I think they will have to be different in order to work, but even if we make it work with the same type, what do we gain with that?

- We'll also have to carry around that polymorphic parameter everywhere were we do something with the Engine, which is alright, but what do we gain by dealing with it?

In general, when introducing polymorphism, I try to have a very good reason for it, specially in Scala, where you usually have to help the compiler infer the polymorphic types, and here I still don't see it.

In any case, if the need for it arises, we can make it polymorphic later, it's not a big deal.

cquiroz:
We don't care if the actual conditions match the requested. they are independent and both stored on the FITS keywords

jdnavarro:
> We don't care if the actual conditions match the requested. they are independent and both stored on the FITS keywords

We were discussing that with @jluhrs yesterday and from what I understood that if the actual conditions are not good enough for the required discussions the engine should give an error and not run. Don't you agree?

We should have had this discussion on the Hipchat channel, sorry about that.

cquiroz:
I totally trust @jluhrs on this. I had the impression they were not used and only kept for the keywords. I must be mistaken

jluhrs:
@cquiroz Yes, you are. :smile:

jdnavarro:
This is ready for final review.

As I mentioned in the meeting, the client can *set* the conditions but can't read them yet. I'll be working on making the global values readable in a separate PR.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
cquiroz: I guess we'll add the operator there too?

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -26,6 +26,8 @@ object Model {
 
     case class ObserverUpdated(view: SequencesQueue[SequenceView]) extends SeqexecModelUpdate
 
+    case class ConditionsUpdated(view: SequencesQueue[SequenceView]) extends SeqexecModelUpdate
```
cquiroz: So the message is ready but not sent downstream yet?

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -139,6 +141,121 @@ object Model {
...
+    val nominal = Conditions(
+      CloudCover.Percent50,
+      ImageQuality.Percent70,
+      SkyBackground.Percent50,
+      WaterVapor.Any
+    )
+
+    val best = Conditions(
+      // In the ODB model it's 20% but that value it's marked as obsolete
+      // so I took the non-obsolete lowest value.
```
cquiroz: Good, we don't need the 20% here

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -139,6 +141,121 @@ object Model {
...
+      ImageQuality.Percent20,
+      SkyBackground.Percent20,
+      WaterVapor.Percent20
+    )
+
+    implicit val equalConditions: Equal[Conditions] = Equal.equalA[Conditions]
+
+    implicit val showConditions: Show[Conditions] = Show.shows[Conditions] {
+      case Conditions(cc, iq, sb, wv) =>
+        List(cc.shows, iq.shows, sb.shows, wv.shows).intercalate(", ")
```
cquiroz: `intercalate` what's the difference with `mkString`?

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
jdnavarro: `SequenceQueue` doesn't contain the conditions yet.

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
jdnavarro: I just didn't know about `mkString`, is it better here?

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/ModelBooPicklers.scala
```diff
@@ -58,6 +58,32 @@ trait ModelBooPicklers {
     .addConcreteType[ObserverUpdated]
     .addConcreteType[OperatorUpdated]
 
+  implicit val conditionsPickler = compositePickler[Model.Conditions]
```
cquiroz: Now I see, `Conditions` is not a composite you could use a regular pickler

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/services/SeqexecWebClient.scala
```diff
@@ -77,6 +78,17 @@ object SeqexecWebClient extends ModelBooPicklers {
   }
 
   /**
+    * Requests the backend to set the Conditions globally
+    */
+  def setConditions(conditions: Conditions): Future[RegularCommand] = {
+    Ajax.post(
+      url = s"$baseUrl/commands/conditions",
```
cquiroz: This doesn't seem to be a `command` per se perhaps we you use `$baseUrl/conditions`

cquiroz: hmm, that means moving it to the other routes, Just leave it in `commands`

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/ModelBooPicklers.scala
jdnavarro: Yeah, I had to do that anyway in a commit for making the conditions readable.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/services/SeqexecWebClient.scala
jdnavarro: I'm not sure what it takes to be a `command`. This is telling the engine to set/change the conditions. Which criteria should follow to make something a `command`?

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
cquiroz: I don't know, I didn't know about `intercalate` 😄

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/services/SeqexecWebClient.scala
cquiroz: yeah it is a bit fuzzy. Nevermind, I slept too little yesterday 😄

cquiroz:
Looks good to me

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
jdnavarro: The thing about about `intercalate` is that it works for everything that is a `Monoid` including `String`, but in this case `mkString` seems a better choice.

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/ModelBooPicklers.scala
jdnavarro: I'll change it in this PR, it may create issues on the client.

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
cquiroz: Good to know, I guess it is up to you

PR #194, opened 2017-03-22
SEQNG-168 Included GCAL in step configuration.

jluhrs:
I did something different with GCAL. For FLAMINGOS-2, all the configuration parameters are always set. With GCAL, the current configuration is read and compared with the required configuration, and only those elements that are different are configured.
This works as only being tested against its internal GCAL simulator. The tests with the real GCAL are a separate task.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
@@ -1,0 +1,94 @@
+package edu.gemini.seqexec.server
+
+import edu.gemini.seqexec.server.ConfigUtilOps._
+import edu.gemini.seqexec.server.GcalController._
+import edu.gemini.spModel.config2.Config
+import edu.gemini.spModel.gemini.calunit.CalUnitConstants._
+import edu.gemini.spModel.gemini.calunit.CalUnitParams.Lamp
+import edu.gemini.spModel.seqcomp.SeqConfigNames.{CALIBRATION_CONFIG_NAME, CALIBRATION_KEY}
+
+import scala.collection.JavaConversions._
+import scala.Function.const
+import scalaz.Scalaz._
+import scalaz.\/
+
+/**
+  * Created by jluhrs on 3/21/17.
+  */
+final case class Gcal(controller: GcalController, isCP: Boolean) extends System {
+  import Gcal._
+
+  override val name: String = CALIBRATION_CONFIG_NAME
+
+  /**
+    * Called to configure a system, returns a Task[SeqexecFailure \/ ConfigResult]
+    */
+  override def configure(config: Config): SeqAction[ConfigResult] = ^(controller.getConfig, fromSequenceConfig(config, isCP))(diffConfiguration)
+    .flatMap(controller.applyConfig).map(const(ConfigResult(this)))
+}
+
+object Gcal {
+  def explainExtractError(e: ExtractFailure): SeqexecFailure =
+    SeqexecFailure.Unexpected(ConfigUtilOps.explain(e))
+
+  implicit class Recover[T](v: \/[ConfigUtilOps.ExtractFailure, T]) {
+    def recoverWithDefault[R >:T](d: R): TrySeq[R] = v.recoverWith[ConfigUtilOps.ExtractFailure, R] {
+          case ConfigUtilOps.KeyNotFound(_) => d.right[ConfigUtilOps.ExtractFailure]
+          case e@ConfigUtilOps.ConversionError(_, _) => e.left
+        }.leftMap(explainExtractError)
+  }
+
+  def diffConfiguration(from: GcalConfig, to: GcalConfig): GcalConfig = {
+
+    def diff[T](from: Option[T], to: Option[T]): Option[T] = (from, to) match {
+      case (Some(a), Some(b)) if a != b => Some(b)
+      case (None, Some(b))              => Some(b)
+      case _                            => None
+    }
+
+    GcalConfig(
+      diff(from.lampAr, to.lampAr),
+      diff(from.lampCuAr, to.lampCuAr),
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
```diff
@@ -0,0 +1,94 @@
...
+        ar <- arLamp.map(_.map(ArLampState))
+        cuar <- cuarLamp.map(_.map(CuArLampState))
+        thar <- tharLamp.map(_.map(ThArLampState))
+        qh <- qhLamp.map(_.map(QHLampState))
+        xe <- xeLamp.map(_.map(XeLampState))
+        ir <- (if (isCP) irLampCP else irLampMK).map(_.map(IrLampState))
+        sht <- shutter
+        flt <- filter
+        dif <- diffuser
+      } yield if(lamps.isEmpty && List(sht, flt, dif).all(_.isEmpty)) GcalConfig.allOff
```
jluhrs: There are no calibration values, which means we are not in a calibration step. Therefore, GCAL light sources must be turned off.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  def diffConfiguration(from: GcalConfig, to: GcalConfig): GcalConfig = {
+
+    def diff[T](from: Option[T], to: Option[T]): Option[T] = (from, to) match {
+      case (Some(a), Some(b)) if a != b => Some(b)
+      case (None, Some(b))              => Some(b)
+      case _                            => None
+    }
+
+    GcalConfig(
```
cquiroz: Is this done to apply the configuration only when the settings are different?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
```diff
@@ -0,0 +1,94 @@
...
+  */
+final case class Gcal(controller: GcalController, isCP: Boolean) extends System {
+  import Gcal._
+
+  override val name: String = CALIBRATION_CONFIG_NAME
+
+  /**
+    * Called to configure a system, returns a Task[SeqexecFailure \/ ConfigResult]
+    */
+  override def configure(config: Config): SeqAction[ConfigResult] = ^(controller.getConfig, fromSequenceConfig(config, isCP))(diffConfiguration)
```
cquiroz: is `lamps` a `List` right? If so if it is empty the loop will stop in line 79

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalControllerSim.scala
```diff
@@ -1,11 +1,30 @@
...
+    "lampThAr = " + config.lampThAr,
+    "lampXe = " + config.lampXe,
+    "lampIr = " + config.lampIr,
+    "shutter = " + config.shutter,
+    "filter = " + config.filter,
+    "diffuser = " + config.diffuser
+    )
+
+  override def applyConfig(config: GcalConfig): SeqAction[Unit] = SeqAction.either{
+    Log.info("applyConfig: config is\n" + printConfig(config).mkString("\n"))
```
cquiroz: hmm, log messages with new lines

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
jluhrs: Yes.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
jluhrs: `^` is like for Applicative.apply2. It takes two Applicatives and a function with two parameters, extract the values from the Applicatives, and use them with the function, all while keeping the semantics of the Applicative.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalControllerSim.scala
jluhrs: It looks prettier that way.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Gcal.scala
jluhrs: No, `lamps` is not a `List`. It is a `TrySeq[List[Lamp]]`.

jdnavarro: I also would like to start using Applicatives more frequently instead of Monads where possible, but I still don't know which Applicative operators are more idiomatic in Scala. That's the first time I've seen the `^`. Do you usually see it more in the wild than the `|@|` operator? what about `<*>`?

It's a pity the syntax for Applicatives is so weird, they are very handy IMHO.

jluhrs: I haven't seen much Scala code to give an answer. But I think it is cleaner than using `<*>`, because with `<*>` you need to curry the function and put it inside an Applicative with `pure`.

cquiroz: With the little I've seen of advanced FP scala I've seen `|@|` and `<*>` several times, I don't recall `^`

PR #195, opened 2017-03-22
SEQNG-117: Seqexec auth expiring

cquiroz:
The auth tokens on the seqexec are set to expire n hours after login. This has the effect that on a long night you could still be working and be logged out

This PR sets a shorten time to live but it renews it every time there is activity which is more logical in my opinion

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/JwtAuthentication.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/JwtAuthentication.scala
@@ -16,7 +16,7 @@ trait CookiesService {
   def ssl: Boolean
   def ttl: Long
 
-  def buildCookie(token: String): Cookie = {
+  def buildCookie(token: String): Task[Cookie] = Task.delay {
     // if successful set a cookie
     val exp = Instant.now().plusSeconds(ttl)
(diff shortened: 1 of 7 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/JwtAuthentication.scala
```diff
@@ -1,23 +1,63 @@
...
 
+
+trait CookiesService {
+  def name: String
+  def ssl: Boolean
+  def ttl: Long
+
+  def buildCookie(token: String): Cookie = {
+    // if successful set a cookie
+    val exp = Instant.now().plusSeconds(ttl)
```
jdnavarro: This does unsafe IO. I think in this particular case there is no need to correct it because it seems `buildCookie` is only meant to be used locally and the Tasks where it's being called are executed right away or close to it, but it could be a problem if you use it to build more complex Task trees.

It seems that in Scala code by convention time measurements are unsafe, but I'd just would wrap every time measurement in a Task, just to avoid thinking about potential issues down the line.

cquiroz: That's a valid concern. I always forget that time is IO. Should be fairly straightforward to wrap it on a Task

cquiroz:
Updated the code as suggested, wrapping the time read calls in `Task` where needed

PR #198, opened 2017-03-23
Set version number for release

cquiroz:
Update version for today's release


cquiroz:
Yes, we should do tags for releases. I'm not sure what you mean by removing the previous tags. i'd rather keep a tag for each release

jdnavarro:
I made a tag for the commit just before merging the old `engine` branch in case we need it to look up some of the previous code quickly, but it seems we won't longer need it.

I'll delete it myself, don't worry about it.

PR #199, opened 2017-03-28
SEQNG-177 Make Conditions globally readable by the client

jdnavarro:
Now the client can access the conditions by looking them up in `SequencesQueue`.

The tests are now slower, now they take ~22 min instead of 8 min, because after introducing the conditions there are more combinations of Arbitrary `SequenceQueue`s to test the serialization, so any ideas about how to reduce the number of tests while actually including the Conditions in the serializations tests are welcome. I wrote in the commit message of the tests what I already tried.

Another thing to watch out is that now the HTTP response to a Sequence load request is just its id instead of a `SequencesQueue`. Everything seems to be working fine but I'm 100% sure there are no issues related to this change. Including the previously set conditions in the response would have been tricky.

I have the uneasy feeling that I'm forgetting something in this PR but I don't know exactly what it is.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
+++ b/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
@@ -82,7 +82,7 @@ class SequenceSpec extends FlatSpec {
     val q = async.boundedQueue[Event](10)
     val qs0: Engine.State =
       Engine.State(
-        Conditions.worst,
+        Conditions.default,
         Map(
           (seqId,
            Sequence.State.init(
@@ -111,7 +111,7 @@ class SequenceSpec extends FlatSpec {
     val q = async.boundedQueue[Event](10)
     val qs0: Engine.State =
       Engine.State(
-        Conditions.worst,
+        Conditions.default,
         Map(
           (seqId,
            Sequence.State.init(
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -139,7 +139,7 @@ object Model {
     * Represents a queue with different levels of details. E.g. it could be a list of Ids
     * Or a list of fully hydrated SequenceViews
     */
-  case class SequencesQueue[T](queue: List[T])
+  case class SequencesQueue[T](conditions: Conditions, queue: List[T])
 
   // Ported from OCS' SPSiteQuality.java
 
@@ -174,6 +174,8 @@ object Model {
       SkyBackground.Percent20,
       WaterVapor.Percent20
     )
+
+    val default = worst // Taken from ODB
 
     implicit val equalConditions: Equal[Conditions] = Equal.equalA[Conditions]
 
--- a/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
+++ b/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
@@ -11,7 +11,7 @@ import org.scalatest.prop.PropertyChecks
 // Keep the arbitraries in a separate trait to improve caching
 object SharedModelArbitraries {
   import org.scalacheck.Shapeless._
-  val maxListSize = 5
+  val maxListSize = 4
 
   // N.B. We don't want to auto derive this to limit the size of the lists for performance reasons
   def sequencesQueueArb[A](implicit arb: Arbitrary[A]): Arbitrary[SequencesQueue[A]] = Arbitrary {
--- a/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
+++ b/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
@@ -16,8 +16,9 @@ object SharedModelArbitraries {
   // N.B. We don't want to auto derive this to limit the size of the lists for performance reasons
   def sequencesQueueArb[A](implicit arb: Arbitrary[A]): Arbitrary[SequencesQueue[A]] = Arbitrary {
     for {
-      a <- Gen.listOfN[A](maxListSize, arb.arbitrary)
-    } yield SequencesQueue(a)
+      a <- implicitly[Arbitrary[Conditions]].arbitrary
+      b <- Gen.listOfN[A](maxListSize, arb.arbitrary)
+    } yield SequencesQueue(a, b)
   }
 
   implicit val udArb  = implicitly[Arbitrary[UserDetails]]
--- a/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
+++ b/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
@@ -11,7 +11,7 @@ import org.scalatest.prop.PropertyChecks
 // Keep the arbitraries in a separate trait to improve caching
 object SharedModelArbitraries {
   import org.scalacheck.Shapeless._
-  val maxListSize = 4
+  val maxListSize = 2
 
   // N.B. We don't want to auto derive this to limit the size of the lists for performance reasons
   def sequencesQueueArb[A](implicit arb: Arbitrary[A]): Arbitrary[SequencesQueue[A]] = Arbitrary {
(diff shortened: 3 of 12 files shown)


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/SequenceSpec.scala
```diff
@@ -82,7 +82,7 @@ class SequenceSpec extends FlatSpec {
     val q = async.boundedQueue[Event](10)
     val qs0: Engine.State =
       Engine.State(
-        Conditions.worst,
+        Conditions.default,
```
cquiroz: :+1: default is a better name

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -139,7 +139,7 @@ object Model {
     * Represents a queue with different levels of details. E.g. it could be a list of Ids
     * Or a list of fully hydrated SequenceViews
     */
-  case class SequencesQueue[T](queue: List[T])
+  case class SequencesQueue[T](conditions: Conditions, queue: List[T])
```
cquiroz: Ok, now I see why this broke on the client. `SequencesQueue` is/was a simple marker for a list of somehting. This change is fine though. Did it break the client?

modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
```diff
@@ -11,13 +11,14 @@ import org.scalatest.prop.PropertyChecks
...
   import org.scalacheck.Shapeless._
-  val maxListSize = 5
+  val maxListSize = 4
 
   // N.B. We don't want to auto derive this to limit the size of the lists for performance reasons
   def sequencesQueueArb[A](implicit arb: Arbitrary[A]): Arbitrary[SequencesQueue[A]] = Arbitrary {
     for {
-      a <- Gen.listOfN[A](maxListSize, arb.arbitrary)
-    } yield SequencesQueue(a)
+      a <- implicitly[Arbitrary[Conditions]].arbitrary
```
cquiroz: Right, I can see why this exploded testing time. You could reduce `maxListSize` to 2

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
cquiroz: Maybe this should be a more general `SequenceMetadata` to be used when we add `operator`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
```diff
@@ -313,14 +313,21 @@ class WebSocketEventsHandler[M](modelRW: ModelRW[M, (SeqexecAppRootModel.LoadedS
     case ServerMessage(s: SeqexecModelUpdate) =>
       // Replace the observer if not set and logged in
       val observer = value._3.map(_.displayName)
-      val (sequencesWithObserver, effects) = s.view.queue.foldLeft((List.empty[SequenceView], List(Some(Effect(Future(NoAction: Action)): Effect)))) { case ((seq, eff), q) =>
-        if (q.metadata.observer.isEmpty && observer.nonEmpty) {
-          (q.copy(metadata = q.metadata.copy(observer = observer)) :: seq, Some(Effect(Future(UpdateObserver(q, observer.getOrElse("")): Action))) :: eff)
+      val (sequencesWithObserver, effects) =
```
cquiroz: Did something change here or is it just improved formatting?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -106,7 +106,7 @@ class SeqexecUIApiRoutes(auth: AuthenticationService, events: (engine.EventQueue
                     \/.fromTryCatchNonFatal(new SPObservationID(oid))
                       .fold(Task.fail, Task.now)
                 u     <- se.load(inputQueue, obsId)
-                resp  <- u.fold(_ => NotFound(s"Not found sequence $oid"), _ => Ok(SequencesQueue[SequenceId](List(oid))))
+                resp  <- u.fold(_ => NotFound(s"Not found sequence $oid"), _ => Ok(oid))
```
cquiroz: This should break the client. But we could fix it on the next client-side PR

cquiroz:
It seems the conditions can be read but are they pushed with state updates too?

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
jdnavarro: >  Did it break the client?

The whole PR was about fixing what broke when adding the extra parameter. Fortunately, it was all caught at compilation time.

I vaguely remember discussing about squashing all global parameters in one parameter in the engine, so we'll to do something like that when adding the operator.

modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
jdnavarro: Does a length of 2 still provide good test coverage? I'm not implying it doesn't, I'm just saying I'm not sure it does. But if that's good enough for you, it'll do for me 😄 .

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/SeqexecCircuit.scala
jdnavarro: That's formatting only

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
jdnavarro: How does it break? I tested it manually and I couldn't find anything wrong.

cquiroz: Really? Well maybe it doesn't 😄

modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
cquiroz: Not too good coverage really but it is not very critical and don't want very long builds

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
cquiroz: Sure, sounds fine to me

jdnavarro:
> It seems the conditions can be read but are they pushed with state updates too?

It should. The `SequencesQueue` is *constructed* from the engine with every event and the conditions are already there. But now that you made me think about it, is there any input event that doesn't trigger a push of the whole `SequencesQueue` through the websocket?

cquiroz:
I think all of them do

cquiroz:
What's going on with travis?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
jdnavarro: Actually, I was convinced that change must have broken something, but I couldn't find anything.

In any case if we see something weird in the client that you didn't see before, this line should be the prime suspect for potential problems 🎱

jdnavarro:
It seems the serialization tests are close to the limit of the timeout Travis allows with no output to the console. Setting `maxListSize` to 2 should fix the problem.

jdnavarro:
Even with length 2 the tests in travis are still taking around ~20min, so I'm not sure anymore that the Arbitrary instance of `SequencesQueue` is causing the delay.

cquiroz:
would you try with a single fixed condition?

jdnavarro:
Yeah... I already did that, but the tests take the same time in my machine (`develop` is faster locally as well). I think the bottleneck is now `WebSocketEventsHandlerSpec` but I still can't pinpoint the exact test.

I'll eventually find it 😣

jdnavarro:
I couldn't find why this is slower now in travis now. There may be some caching involved because I'm not getting consistent testing times locally.

I'm going to merge and watch out for the next travis builds to see what happens.

PR #200, opened 2017-03-29
SEQNG-172 Made systems configured in step depend on step type.

jluhrs:
The original goal was to not configure TCS for calibration steps. In the road towards that goal I implemented the identification of step types, which I think is the big contribution of this PR.
I am not totally sure about one aspect of the implementation thought. The type of step determines the resources used, the systems that must be configured, and the headers. I created classes for the types, and separate functions to calculate the resources, systems and headers. There is some repeated code between those functions, so maybe they could be consolidated. The other aspect that may be changed is that I think there should be a relationship between resources and systems.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -123,7 +123,7 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     stepType match {
       case CelestialObject(inst) => toInstrumentSys(inst).map(_ :: List(Tcs(systems.tcs), Gcal(systems.gcal, site == Site.GS)))
       case FlatOrArc(inst)       => toInstrumentSys(inst).map(_ :: List(Gcal(systems.gcal, site == Site.GS)))
-      case _ => TrySeq.fail(Unexpected(s"Unsupported step type ${stepType.toString}"))
+      case _                     => TrySeq.fail(Unexpected(s"Unsupported step type ${stepType.toString}"))
     }
   }
 
@@ -147,7 +147,7 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     case CelestialObject(inst) => calcInstHeader(config, inst).map(_ :: commonHeaders(config))
     case FlatOrArc(inst)       => calcInstHeader(config, inst).map(_ :: commonHeaders(config)) //TODO: Add GCAL keywords here
     case DarkOrBias(inst)      => calcInstHeader(config, inst).map(_ :: commonHeaders(config))
-    case st@_                  => TrySeq.fail(Unexpected("Unsupported step type " + st.toString))
+    case st                    => TrySeq.fail(Unexpected("Unsupported step type " + st.toString))
   }
 
 }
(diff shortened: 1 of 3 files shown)


cquiroz:
Can you explain a bit more about the relationship between resources and systems?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -88,45 +81,79 @@ class SeqTranslate(site: Site) {
...
+    stepType match {
+      case CelestialObject(inst) => toInstrumentSys(inst).map(_ :: List(Tcs(systems.tcs), Gcal(systems.gcal, site == Site.GS)))
+      case FlatOrArc(inst)       => toInstrumentSys(inst).map(_ :: List(Gcal(systems.gcal, site == Site.GS)))
+      case _ => TrySeq.fail(Unexpected(s"Unsupported step type ${stepType.toString}"))
+    }
+  }
+
+  private def calcInstHeader(config: Config, inst: Resource.Instrument): TrySeq[Header] = inst match {
+    case Resource.F2 =>  TrySeq(Flamingos2Header(systems.dhs, new Flamingos2Header.ObsKeywordsReaderImpl(config),
+      if(settings.f2Keywords) Flamingos2Header.InstKeywordReaderImpl else Flamingos2Header.DummyInstKeywordReader,
```
cquiroz: I think you know what's wrong in this line 😄

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -142,4 +169,55 @@ object SeqTranslate {
                       gwsKeywords: Boolean
                      )
 
+
+  private sealed trait StepType {
```
cquiroz: Why are these private?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -88,45 +81,79 @@ class SeqTranslate(site: Site) {
...
+    // TODO: Replace Unit by something that can read the State
+    StateKeywordsReader(Unit)
+  ))
+
+
+  private def calcHeaders(config: Config, stepType: StepType): TrySeq[List[Header]] = stepType match {
+    case CelestialObject(inst) => calcInstHeader(config, inst).map(_ :: commonHeaders(config))
+    case FlatOrArc(inst)       => calcInstHeader(config, inst).map(_ :: commonHeaders(config)) //TODO: Add GCAL keywords here
+    case DarkOrBias(inst)      => calcInstHeader(config, inst).map(_ :: commonHeaders(config))
+    case st@_                  => TrySeq.fail(Unexpected("Unsupported step type " + st.toString))
```
cquiroz: The `st@_` looks a bit funny

cquiroz:
Looks very good in general.

jluhrs:
About resources and systems. Resources, as they are now, are used to identify the resources required to run a step. They are kept only as a mean to determine what resources are used by a sequence, and when they are available to run a sequence. Systems on the other hand give the means to interact with external systems. Every System has a method `applyConfig` that will apply a configuration to the real system. Ideally, there should be a one to one relation between Resources and Systems.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: Because they are not (yet) used outside `SeqTranslate`.

cquiroz:
:+:1

cquiroz:
https://github.com/gemini-hlsw/ocs3/pull/200/commits/554142572bceb255d983f9dad3847f8be2fc647e looks beautiful

PR #202, opened 2017-03-30
Allow setting the image quality directly

cquiroz:
This PR adds methods to set the image quality independently of the other conditions. After the review it should be expanded to allow updating the other condition subsets


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -118,6 +121,7 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
       case engine.SetOperator(_)      => OperatorUpdated(svs)
       case engine.SetObserver(_, _)   => ObserverUpdated(svs)
       case engine.SetConditions(_)    => ConditionsUpdated(svs)
+      case engine.SetImageQuality(_)  => ConditionsUpdated(svs)
```
cquiroz: Note that I'm reusing the downstream event. I hope that's ok, it seemed wasteful to create yet another event type

PR #204, opened 2017-04-04
SEQNG-169: GCAL FITS keywords

cquiroz:
Implementation of the Gcal keywords


jluhrs:
Almost there. The one issue is something I forgot to put in the task description. The GCAL keywords are optional, they are written only for calibration steps. Therefore, they should not be in StandardHeader, but in its own Header, and added to the list of headers written for `FlatOrArc` in `calcHeaders`

cquiroz:
should the keywords be sent before or after as per the methods in `Header`?

cquiroz:
Updated according to the PR comments

cquiroz:
@jluhrs Would you be able to validate this with GCAL once merged?

jluhrs:
Yes, I'll create a task to test this and my PR against the real systems.

PR #205, opened 2017-04-05
SEQNG-186: Synchronize operator field across clients

cquiroz:
This PR re enables the operator field on the side bar. It required an update to the backend to move the operator to a higher level along the conditions.

The operator field is getting the same treatment as the observer field and updates will be detected on the background

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
@@ -3,7 +3,7 @@ package edu.gemini.seqexec.engine
 import scalaz._
 import Scalaz._
 
-import edu.gemini.seqexec.model.Model.Conditions
+import edu.gemini.seqexec.model.Model.{Conditions, Operator}
 
 /**
   * A Map of `Sequence`s.
@@ -24,11 +24,11 @@ object Engine {
       Engine(q.sequences.mapValues(_.map(f)))
   }
 
-  case class State(conditions: Conditions, sequences: Map[Sequence.Id, Sequence.State])
+  case class State(conditions: Conditions, operator: Option[Operator], sequences: Map[Sequence.Id, Sequence.State])
 
   object State {
 
-    def empty: State = State(Conditions.default, Map.empty)
+    def empty: State = State(Conditions.default, None, Map.empty)
 
   }
 
--- a/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
+++ b/modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
@@ -17,7 +17,8 @@ object SharedModelArbitraries {
   def sequencesQueueArb[A](implicit arb: Arbitrary[A]): Arbitrary[SequencesQueue[A]] = Arbitrary {
     for {
       b <- Gen.listOfN[A](maxListSize, arb.arbitrary)
-    } yield SequencesQueue(Conditions.default, b)
+      o <- Gen.option(Gen.alphaStr)
+    } yield SequencesQueue(Conditions.default, o, b)
   }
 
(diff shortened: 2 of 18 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -24,11 +24,11 @@ object Engine {
       Engine(q.sequences.mapValues(_.map(f)))
   }
 
-  case class State(conditions: Conditions, sequences: Map[Sequence.Id, Sequence.State])
+  case class State(conditions: Conditions, operator: Option[Operator], sequences: Map[Sequence.Id, Sequence.State])
```
jdnavarro: I'm fine with an extra parameter here, next parameter we add more we could group them all in a single parameter like `metadata` or something like that. @cquiroz I think you already suggested that, am I right?

modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
```diff
@@ -17,7 +17,8 @@ object SharedModelArbitraries {
   def sequencesQueueArb[A](implicit arb: Arbitrary[A]): Arbitrary[SequencesQueue[A]] = Arbitrary {
     for {
       b <- Gen.listOfN[A](maxListSize, arb.arbitrary)
-    } yield SequencesQueue(Conditions.default, b)
+      o <- Gen.option(Gen.alphaStr)
+    } yield SequencesQueue(Conditions.default, o, b)
```
jdnavarro: This is going to increase the number of generated tests by a lot. We'll have to control the number of tests in another PR, but until then I don't think there will be a quick fix to make the tests on Travis pass.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/HeadersSideBar.scala
```diff
@@ -33,30 +35,37 @@ object HeadersSideBar {
...
       $.modState(_.copy(currentText = Some(value)))
 
+    def setupTimer: Callback =
+      // Every 2 seconds check if the field has changed and submit
+      setInterval(submitIfChanged, 2.second)
+
     def submitIfChanged: Callback =
       ($.state zip $.props) >>= {
-        case (s, p) => Callback.empty // Callback.when(s.currentText =/= p.operator())(Callback.empty) // We are not submitting until the backend properly update this property
+        case (s, p) => Callback.when(s.currentText =/= p.model().operator)(updateOperator(~s.currentText))
```
jdnavarro: What's the `~` mean here?

cquiroz: It is the same as `orZero` or if you prefer `getOrElse("")`. I'm not sure if it is the preferred syntax these days but I learned it some time ago

modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/BoopicklingSpec.scala
cquiroz: Agree. I didn't see changes on the time on my local machine in terms of time but yeah, travis keeps failing

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
cquiroz: Yeah, at some moment it could make sense to group them but I'm not sure when that would be. Probably with three params you want to have a separate class class

cquiroz:
@jluhrs Do you mean how it works on the frontend or the backend?

PR #206, opened 2017-04-06
SEQNG-171 Implemented configuration of science fold to direct light to instrument.

jluhrs:
The PR ended I bit more complicated than expected. First, I had to write code to deal with the way the name of the Science Fold position is encoded. And then there is the fact that the AG is configured through TCS, so I can not treat it as an independent system.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ConfigUtilOps.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ConfigUtilOps.scala
@@ -7,7 +7,6 @@ import java.beans.PropertyDescriptor
 
 import scala.reflect.ClassTag
 import scala.collection.breakOut
-
 import scalaz._
 import Scalaz._
 
@@ -23,6 +22,13 @@ object ConfigUtilOps {
   def explain(e: ExtractFailure): String = e match {
     case KeyNotFound(k) => s"Missing config value for key ${k.getPath}"
     case ConversionError(k, msg) => s"Error reading key ${k.getPath}: $msg"
+  }
+
+  def explainExtractError(e: ExtractFailure): SeqexecFailure =
+    SeqexecFailure.Unexpected(ConfigUtilOps.explain(e))
+
+  implicit class TrySeqed[A] private [server] (r: ExtractFailure \/ A) {
+    def asTrySeq: TrySeq[A] = r.leftMap(explainExtractError)
   }
 
   // key syntax: parent / child
@@ -53,5 +59,6 @@ object ConfigUtilOps {
             }(breakOut): Map[String, String])
       }
     }
+
   }
 }
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerEpics.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerEpics.scala
@@ -9,7 +9,9 @@ import squants.time.Seconds
 import collection.JavaConversions._
 import edu.gemini.seqexec.server.TcsController._
 import edu.gemini.epics.acm.{CaAttributeListener, CaService, XMLBuilder}
+import edu.gemini.seqexec.engine.Resource
 import edu.gemini.spModel.core.Wavelength
+import squants.Length
 import squants.space.{Angstroms, Degrees, Microns, Millimeters}
 
 import scalaz._
@@ -173,6 +175,17 @@ object TcsControllerEpics extends TcsController {
       GuiderSensorOptionAO(aoOn)))
   }.getOrElse(TrySeq.fail(SeqexecFailure.Unexpected("Unable to read guider detectors state from TCS.")))
 
+  private def getInstPort(inst: Resource.Instrument): Option[Int] = (inst match {
+    case Resource.GMOS  => TcsEpics.instance.gmosPort
+    case Resource.GSAOI => TcsEpics.instance.gsaoiPort
+    case Resource.F2    => TcsEpics.instance.f2Port
+    case Resource.GPI   => TcsEpics.instance.gpiPort
+    case Resource.NIRI  => TcsEpics.instance.niriPort
+    case Resource.GNIRS => TcsEpics.instance.gnirsPort
+    case Resource.NIFS  => TcsEpics.instance.nifsPort
+    case _              => None
+  }).flatMap(p => if (p == 0) None else Some(p))
+
   // Decoding and encoding the science fold position require some common definitions, therefore I put them inside an
   // object
   private object CodexScienceFoldPosition {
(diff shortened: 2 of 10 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ConfigUtilOps.scala
```diff
@@ -25,6 +24,13 @@ object ConfigUtilOps {
     case ConversionError(k, msg) => s"Error reading key ${k.getPath}: $msg"
   }
 
+  def explainExtractError(e: ExtractFailure): SeqexecFailure =
+    SeqexecFailure.Unexpected(ConfigUtilOps.explain(e))
+
+  implicit class TrySeqed[A] private [server] (r: ExtractFailure \/ A) {
```
cquiroz: Why do you have the `private [server]`.
Perhaps this could be a value class

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerEpics.scala
```diff
@@ -182,34 +195,56 @@ object TcsControllerEpics extends TcsController {
...
+    private def findInstrument(str: String): Option[Resource.Instrument] = instNameMap.find{ case (_, n) => str.startsWith(n.self)}.map(_._1)
 
-    implicit val encodeScienceFoldPosition: EncodeEpicsValue[Position, String] = EncodeEpicsValue((a: Position)
-    => a.source match {
-        case Sky => a.sink.sfName
-        case AO => AO_PREFIX + a.sink.sfName
-        case GCAL => GCAL_PREFIX + a.sink.sfName
-      }
+    implicit val decodeScienceFoldPosition: DecodeEpicsValue[String, Option[ScienceFoldPosition]] = DecodeEpicsValue(
+      (t: String) => if (t.startsWith(PARK_POS)) Parked.some
```
cquiroz: Perhaps this would be clearer as pattern match?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerEpics.scala
```diff
@@ -306,15 +341,20 @@ object TcsControllerEpics extends TcsController {
     _ <- setProbeTrackingConfig(TcsEpics.instance.oiwfsProbeGuideCmd, c.oiwfs.self)
   } yield TrySeq(())
 
+  // Special case: if source is the sky and the instrument is at the bottom port (port 1), the science fold must be parked.
   def setScienceFoldConfig(sfPos: ScienceFoldPosition): SeqAction[Unit] = sfPos match {
     case ScienceFoldPosition.Parked => TcsEpics.instance.scienceFoldParkCmd.mark
-    case p: ScienceFoldPosition.Position => TcsEpics.instance.scienceFoldPosCmd.setScfold(encode(p))
+    case p@ScienceFoldPosition.Position(LightSource.Sky, inst) => getInstPort(inst).flatMap(port =>
+      if (port == 1) TcsEpics.instance.scienceFoldParkCmd.mark.some
```
cquiroz: What's the port here? ISS port? Perhaps the `1` should be made a more meaningful constant

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
```diff
@@ -281,6 +281,9 @@ final class TcsEpics(epicsService: CaService) {
 
   def inPosition: Option[String] = Option(inPositionAttr.value)
 
+  val agInPositionAttr: CaAttribute[java.lang.Double] = tcsState.getDoubleAttribute("agInPosition")
```
cquiroz: Could you convert to scala's Double right away?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
```diff
@@ -340,6 +343,58 @@ final class TcsEpics(epicsService: CaService) {
       }
     })))
 
+  def waitAGInPosition(timeout: Time): SeqAction[Unit] =
+    EpicsCommand.safe(EitherT(Task.async[TrySeq[Unit]]((f) => {
```
cquiroz: What's `safe`?

cquiroz: `Task.async` takes a thread pool or strategy. Are you importing the default one?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
```diff
@@ -340,6 +343,58 @@ final class TcsEpics(epicsService: CaService) {
...
+              }
+
+              f(TrySeq(()).right)
+            }
+          }
+
+          override def onValidityChange(newValidity: Boolean): Unit = {}
+        }
+
+        locked {
```
cquiroz: Why do we need the lock here? Could you explain a bit what's this about?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
jluhrs: Yes.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerEpics.scala
jluhrs: Yes, it is the ISS port

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ConfigUtilOps.scala
jluhrs: I don't know. I copied code written by Shane (the `Extracted` class below)

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerEpics.scala
jluhrs: Pattern match with regular expressions ?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
jluhrs: No, because CaAttribute is a java class.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerEpics.scala
cquiroz: this is not really very important but you could do 
```
t match {
case a if a.startsWith(...)
```

cquiroz: It is debatable which option is clearer

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
jluhrs: Uff, I replicated the code used to wait for the TCS in-position flag. I wrote that like two years ago. I'll need some time to remember how this works.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
cquiroz: I see, perhaps `agInPositionAttr` could be private or inside a block to initialize `agInPosition`

jluhrs: I agree it should be private, but not in a block, because it is not only used in `agInPosition`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
jluhrs: `safe` is a function that takes a SeqAction and returns another SeqAction that will run the inner Task in a safe way, capturing exceptions.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
cquiroz: 👍

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
jluhrs: Ok, so this is java code that deals with the asynchronous waiting of the AG in-position flag. There are two parts, the function that deals with changes on the EPICS channel (in essence an Observer), and the timer that implements the timeout. The observer and the timer are deal inside critical sections (the locked parts), There is also an AtomicInteger used as a flag between the channel observer and the timer. The first one to trigger a result marks the flag to prevent the other one to also generate a result.

cquiroz: The lock makes me a bit nervous though it is confined to the `Task`. If we had many threads calling it should work fine but perhaps there are problems if the same threads gets into the lock section

I'm kind of forgetting how to think about locks so I'll trust you 😄 

Perhaps you could put the same description on your comment in the code

PR #207, opened 2017-04-07
SEQNG-185: Fix for stalled builds

cquiroz:
This PR apparently fixes the issues observed when running builds in travis

This comes after several rounds of testing and with the experience of other projects that have large scala.js code bases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ install:
 before_script:
   - sudo apt-get install -y rpm
 script:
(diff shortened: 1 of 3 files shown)


cquiroz:
Yes it does and that causes troubles with scala.js

cquiroz:
I have re-enabled the requirement of passing tests to merge

cquiroz:
With the last change we are back to builds of 8mins

PR #209, opened 2017-04-11
Responsive UI improvements

cquiroz:
This PR brings some improvements to the UI in different device sizes. Depending on the size of the screens some buttons would get miss-aligned

Also the mobile level view gets too crowded to be usable so I'm turning it into a view-only mode. This is debatable of course and could be changed if there is user demand


cquiroz:
Thanks for taking the time to test it

PR #210, opened 2017-04-13
SEQNG-87: Update auth middleware for http4s

cquiroz:
http4s deprecated the old way to do authentication and in 0.16.x it is gone. I updated http4s to 0.16 and had to deal with this change

This PR adopts the new way to do authentication and let's me remove some code that was incorporated in the project

I'm also tightening the screws in a few places where auth wasn't fully controlled


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -73,33 +77,35 @@ class SeqexecCommandRoutes(auth: AuthenticationService, inputQueue: engine.Event
       } yield resp
 
 
-    case req @ POST -> Root / "conditions" =>
-      req.decode[Conditions] (conditions =>
+    case req @ POST -> Root / "conditions" as _ =>
+      req.req.decode[Conditions] (conditions =>
```
jdnavarro: Has the structure of a `http4s` changed with this release? Why the wrapper?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -21,18 +25,18 @@ import scalaz.concurrent.Task
...
-  val tokenAuthService = JwtAuthentication(auth, auth.config.devMode)
-
   val commands = Commands(se.odbProxy)
+  // Handles authentication
+  val httpAuthentication = new Http4sAuthentication(auth)
 
-  val service = tokenAuthService { GZip { HttpService {
-    case GET  -> Root  / "host" =>
+  val commandServices: AuthedService[UserDetails] = AuthedService {
+    case GET  -> Root  / "host" as _ =>
```
jdnavarro: What the `as _` does?

project/Settings.scala
```diff
@@ -38,11 +38,11 @@ object Settings {
     val scalaJQuery  = "1.0"
 
     // Java libraries
-    val scalaZ       = "7.2.2"
+    val scalaZ       = "7.2.10"
     val scalaZStream = "0.8.6a"
```
jdnavarro: No version bump for `scalaZStream` yet? Do you know if the upgrade is expected for next `http4s` release?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/encoder/BooPickleInstances.scala
```diff
@@ -27,7 +28,7 @@ trait BooPickleInstances {
     }
 
   def booEncoderOf[A](implicit encoder: Pickler[A]): EntityEncoder[A] =
-    EntityEncoder[ByteBuffer].contramap[A] { v =>
-      Pickle.intoBytes(v)
+    EntityEncoder[ByteVector].contramap[A] { v =>
```
jdnavarro: What's the difference between a `ByteBuffer` and a `ByteVector`?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/HttpAuthenticator.scala
```diff
@@ -0,0 +1,63 @@
...
+trait CookiesService {
+  def name: String
+  def ssl: Boolean
+  def ttl: Long
+
+  def buildCookie(token: String): Task[Cookie] = Task.delay {
+    // if successful set a cookie
+    val exp = Instant.now().plusSeconds(ttl)
+    Cookie(name, token, path = "/".some, expires = exp.some, secure = ssl, httpOnly = true)
+  }
```
jdnavarro: I think only `exp` needs to be wrapped with a `Task` and `Cookie` could map over `exp`.

In this case computing the time and creating the cookie is so instantaneous that I don't think it doesn't really matter, but if the computation of `Cookie.apply` had to take some time to complete, I think it would.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
cquiroz: that is an extractor that lets you read the user data doing `as user`. However in this case we don't care who the user is just that we have one so we can use the underscore

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
cquiroz: If you have authentication yes. You don't get a plain request (possibly the tame of the variable should change) and you get an `AuthedRequest` instead that contains the original `Request` in the `req` variable

project/Settings.scala
cquiroz: Probably never, `http4s` is going to be exclusively `cats/fs2` starting in 0.17.x

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/encoder/BooPickleInstances.scala
cquiroz: `ByteBuffer` comes from java nio and its behaviour is complicated as it is mutable. `ByteVector` comes from scodec and it is immutable so `http4s` will prefer to use it instead. `EntityEncoder[ByteBuffer]` will be removed altogether on `0.17.x`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/security/HttpAuthenticator.scala
cquiroz: I see what you mean. Cookie is just a case class. I could change it though

cquiroz:
On the last commit I restored an ability to extend the session life every time the user does a request

PR #211, opened 2017-04-21
SEQNG-189 Implemented access to GMOS EPICS commands and statuses.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/resources/Gmos.xml
```diff
@@ -0,0 +1,660 @@
...
+                <type>STRING</type>
+                <description>Translation Stage Mode</description>
+            </parameter>
+            <!--
+            <parameter name="useElectronicOffsetting">
+                <channel>wfs:followA.K</channel>
+                <type>INTEGER</type>
+                <description>Electronic Offsets</description>
+            </parameter>
+            -->
```
jluhrs: The XML file has limitations. One is that is assumes all parameters are inputs to a CAD. They have a type, but the underlining channel is always of type String. The problem comes is in cases like this, where the parameter type is actually the channel type. The library supports this case, but it needs an additional parameter that is not included in the XML. Therefore, I need to create the corresponding Attribute by hand. I kept definition in the XML as a reminder.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -248,9 +249,7 @@ object SeqexecEngine {
 
     val tcsInit = if(tcsKeywords || !tcsSim) initEpicsSystem(TcsEpics) else Task.now(())
     // More instruments to be added to the list here
-    val instInit = if(f2Keywords || !instSim)
-      Nondeterminism[Task].gatherUnordered(List(Flamingos2Epics).map(initEpicsSystem(_)))
-    else Task.now(())
+    val instInit = Nondeterminism[Task].gatherUnordered(List((f2Keywords, Flamingos2Epics), (gmosKeywords, GmosEpics)).filter(_._1 || !instSim).map(x => initEpicsSystem(x._2)))
```
cquiroz: Why `instSim` is only checked for gmos and not f2?

jluhrs: It is checked for all. The `filter` is applied to all the elements.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosEpics.scala
```diff
@@ -0,0 +1,301 @@
...
+
+  val roi4 = new RoiStatus(dcState, 4)
+
+  val roi5 = new RoiStatus(dcState, 5)
+
+  def ccdXBinning: Option[Int] = Option(dcState.getDoubleAttribute("ccdXBinning").value.toInt)
+
+  def ccdYBinning: Option[Int] = Option(dcState.getDoubleAttribute("ccdYBinning").value.toInt)
+
+  def currentCycle: Option[Int] = Option(dcState.getIntegerAttribute("currentCycle").value.toInt)
```
jluhrs: It is there to avoid a warning from sbt.

PR #214, opened 2017-04-26
SEQNG-195: Automatic load and unload of sequences from ODB queue.

jluhrs:
The mechanism implemented is a periodic polling. Every 10 seconds Seqexec retrieves the list of sequences in the ODB queue and compares it with the sequences loaded. New sequences in the list will be loaded, and sequences not in the list will be unloaded.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -10,12 +10,15 @@ import edu.gemini.seqexec.engine.{Engine, Event, EventSystem, Executed, Failed, 
 
 import scalaz._
 import Scalaz._
-import scalaz.concurrent.Task
-import scalaz.stream.Process
-import knobs._
+import scalaz.concurrent.{Strategy, Task}
+import scala.concurrent.duration._
+import scalaz.stream.{DefaultScheduler, Process, wye}
+import scalaz.stream.wye._
+import scalaz.stream.time._
 import edu.gemini.seqexec.model.Model._
 import edu.gemini.seqexec.model.Model.SeqexecEvent._
 import edu.gemini.spModel.core.Peer
+import knobs.Config
 
 /**
   * Created by jluhrs on 10/7/16.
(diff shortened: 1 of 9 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -76,13 +79,14 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
   def setCloudCover(q: engine.EventQueue, cc: CloudCover): Task[SeqexecFailure \/ Unit] =
     q.enqueueOne(Event.setCloudCover(cc)).map(_.right)
 
-  // TODO: Add seqId: SPObservationID as parameter
   def setSkipMark(q: engine.EventQueue, id: SPObservationID, stepId: edu.gemini.seqexec.engine.Step.Id): Task[SeqexecFailure \/ Unit] = ???
 
   def requestRefresh(q: engine.EventQueue): Task[Unit] = q.enqueueOne(Event.poll)
 
+  def seqQueueRefreshProcess(q: engine.EventQueue): Process[Task, Event]= awakeEvery(10.seconds)(Strategy.DefaultStrategy, DefaultScheduler).map(_ => Event.getState(refreshSequenceList(q)))
```
cquiroz: Should the time be configurable?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -192,6 +202,14 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
     // TODO: Implement willStopIn
     SequenceView(seq.id, seq.metadata, st, engineSteps(seq), None)
   }
+
+  private def refreshSequenceList(q: engine.EventQueue): Engine.State => Task[Unit] = {
+    st: Engine.State => odbProxy.queuedSequences().flatMap(seqs => EitherT(Nondeterminism[Task].gatherUnordered{
```
cquiroz: This is quite hard to follow, perhaps it could be split for better readability? Could you run this inside a for comprehension?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: I forgot to do that. I'll put it right now.

PR #215, opened 2017-04-27
Send distinct engine event when trying to run a Sequence where some resource are busy

jdnavarro:
Now instead of doing nothing, it explicitly sends an event indicating that the Sequence couldn't be run because some resources are busy.

The *Run* button in the UI still appears greyed out after this event but I don't think it's worth fixing because I think we shouldn't even show the *Run* button when there are busy resources (to be done in another PR).

This event will be used only when trying to run the sequence through the HTTP API.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -57,11 +57,11 @@ package object engine {
     */
   def switch(q: EventQueue)(id: Sequence.Id)(st: SequenceState): Handle[Unit] =
     resources.flatMap(
-      ores => getS(id).flatMap {
+      other => getS(id).flatMap {
         case Some(seq) =>
           if (st === SequenceState.Running)
             // No resources being used by other running sequences
-            if (seq.toSequence.resources.intersect(ores).isEmpty)
+            if (seq.toSequence.resources.intersect(other).isEmpty)
               putS(id)(Sequence.State.status.set(seq, st))
             // Some resources are being used
             else send(q)(busy(id))
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -64,14 +64,7 @@ package object engine {
             if (seq.toSequence.resources.intersect(ores).isEmpty)
               putS(id)(Sequence.State.status.set(seq, st))
             // Some resources are being used
-            else
-              send(q)(
-                failed(
-                  id,
-                  0,
-                  Result.Error("Unable to run Sequence, some of the resources needed are in use")
-                )
-              )
+            else send(q)(busy(id))
           else putS(id)(Sequence.State.status.set(seq, st))
         case None => unit
       }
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -57,16 +57,17 @@ package object engine {
...
-        seq =>
-        if (st === SequenceState.Running)
-          // No resources being used by other running sequences
-          if (seq.toSequence.resources.intersect(ores).isEmpty)
-            Sequence.State.status.set(seq, st)
-          // Some resources are being used, sequence unchanged
-          else seq
-        else Sequence.State.status.set(seq, st)
-      )
+      ores => getS(id).flatMap {
```
cquiroz: ores?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -255,32 +256,33 @@ package object engine {
...
 
     def handleUserEvent(ue: UserEvent): Handle[Unit] = ue match {
-      case Start(id)               => log("Output: Started") *> rollback(q)(id) *>
+      case Start(id)               => log("Engine: Started") *> rollback(q)(id) *>
         switch(q)(id)(SequenceState.Running) *> send(q)(Event.executing(id))
-      case Pause(id)               => log("Output: Paused") *> switch(q)(id)(SequenceState.Stopping)
-      case Load(id, seq)           => log("Output: Sequence loaded") *> load(id, seq)
-      case Unload(id)              => log("Output: Sequence unloaded") *> unload(id)
-      case Breakpoint(id, step, v) => log("Output: breakpoint changed") *>
+      case Pause(id)               => log("Engine: Paused") *> switch(q)(id)(SequenceState.Stopping)
```
cquiroz: This is unrelated but it seems that `log` prints to stdout. We should fix that

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -153,6 +153,8 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
       case engine.Completed(_, _, _)     => NewLogMessage("Action completed")
       case engine.PartialResult(_, _, _) => SequenceUpdated(svs)
       case engine.Failed(_, _, _)        => NewLogMessage("Action failed")
+      // TODO: Does it need a different event?
+      case engine.Busy(_)                => SequenceRefreshed(svs)
```
cquiroz: I'd say yes, the reason for another event is that we can react in a special way

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: Yes, there is ticket has been there for a while in the backlog.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: It comes from *other resource* but it's a very short lived value so I think it's not that important to convey any meaning, it could have just been called *x*.

PR #216, opened 2017-04-28
SEQNG-80 EPICS tops are now read from app.conf

jluhrs:


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/resources/app.conf
```diff
@@ -49,4 +49,5 @@ seqexec-engine {
     gcalKeywords = false
     instForceError = false
     odbQueuePollingInterval = "10 seconds"
+    tops = "tcs=tcs:, ao=ao:, gm=gm:, gc=gc:, gws= ws:, m2=m2:, oiwfs=oiwfs:, ag=ag:, f2=f2:"
```
jluhrs: I decided against making a parameter for each one because I'm lazy. This format as the drawback that one typo may break all the tops.

cquiroz:
How is that this change speeds up startup?

jluhrs:
@cquiroz The problem you saw is that Seqexec was trying to connect to channels in the TCS simulator instead of the real TCS. That was because it had hardcoded the value "tc1:" as the prefix to all the TCS channels. This PR makes all prefixes configurable from app.conf, and sets the correct value for the TCS prefix. Now it should connect very quickly to each TCS channel without timeouts.

PR #222, opened 2017-05-05
SEQNG-83: Epics configuration for systems

cquiroz:
This has been bothering me for a while. EPICS gets initialized from the system environment variable, but when it is not present we get an NPE

This PR adds a configuration variable to set it explicitly. If not present the system will revert to the environment variable

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -5,8 +5,8 @@ import java.time.LocalDate
 import edu.gemini.epics.acm.CaService
 import edu.gemini.pot.sp.SPObservationID
 import edu.gemini.model.p1.immutable.Site
-import edu.gemini.seqexec.{engine, server}
-import edu.gemini.seqexec.engine.{Engine, Event, EventSystem, Executed, Failed, Result, Sequence}
+import edu.gemini.seqexec.engine
+import edu.gemini.seqexec.engine.{Engine, Event, EventSystem, Executed, Failed}
 
 import scalaz._
 import Scalaz._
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -107,9 +107,7 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
         obsId <- safeGetObsId(id)
         _ <- systems.odb.obsAbort(obsId, e.msg)
       } yield ()
-      case (EventSystem(Executed(id)), st) if st.sequences.get(id).map(
-        _.status === SequenceState.Idle
-      ).getOrElse(false) => for {
+      case (EventSystem(Executed(id)), st) if st.sequences.get(id).exists(_.status === SequenceState.Idle) => for {
```
jdnavarro: I know this is carried over from the previous code, but `EventSystem(Executed(id))` can be written as `Event.executed(id)`.  Now that I think about it, maybe we should make `EventSystem` private to the engine package, it only makes sense internally.

cquiroz: Good to know, I only adjusted a bit the code formatting, will change that

cquiroz: I can't do `Event.executed(id)` here as we are inside a pattern match here, right?

jdnavarro: Oh, you are right, I forgot about how pattern matching works in Scala. I guess it could be done implementing `unapply` for `Event`s, but it's not worth doing it in this PR. I'll try to implement it at some point when dealing with the engine Events.

cquiroz:
It seems the travis fix didn't help here. Should we merge this anyway?

jdnavarro:
Fine with me but since this PR includes some EPICS configuration, perhaps @jluhrs should have a look before merging.

PR #226, opened 2017-05-09
Updated ocs libraries to 2017A 1.1.6

cquiroz:
Updates the version to the latest available


sraaphorst:
I'm guessing this is something I didn't know about that should be done every time we do a formal OCS update?

cquiroz:
At least when the serialization breaks

PR #228, opened 2017-05-10
SEQNG-209 Implemented FLAMINGOS-2 darks.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
```diff
@@ -85,18 +86,27 @@ object Flamingos2 {
     case _                    => WindowCover.OPEN
   }
 
+  implicit def grismFromSPDisperser(d: Disperser): Grism = d match {
```
cquiroz: Do you get a warning here. It was kind of frowned upon to do this kind of direct conversion though in this case it seems fine

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
```diff
@@ -85,18 +86,27 @@ object Flamingos2 {
...
+
   def ccConfigFromSequenceConfig(config: Config): TrySeq[CCConfig] = ( for {
+      obsType <- config.extract(OBSERVE_KEY / OBSERVE_TYPE_PROP).as[String]
       // WINDOW_COVER_PROP is optional. If not present, then window cover position is inferred from observe type.
-      p <- config.extract(INSTRUMENT_KEY / WINDOW_COVER_PROP).as[WindowCover] match {
-            case a: \/-[WindowCover] => a
-            case _         => config.extract(OBSERVE_KEY / OBSERVE_TYPE_PROP).as[String]
-                              .map(windowCoverFromObserveType)
-          }
+      p <- config.extract(INSTRUMENT_KEY / WINDOW_COVER_PROP).as[WindowCover].recover(PartialFunction((x:ConfigUtilOps.ExtractFailure) => windowCoverFromObserveType(obsType)))
```
cquiroz: So we try to read it from the sequence but if not we try to guess it from the observation type?

jluhrs: Yes.

PR #230, opened 2017-05-11
SEQNG-224: Fixed the type of some EPICS channels.

jluhrs:
A couple of channels were created as String parameters, but when trying to use them they were searched as Int parameters. That caused the NullPointerException.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
@@ -48,7 +48,7 @@ final class Flamingos2Epics(epicsService: CaService) {
   object configCmd extends EpicsCommand {
     override val cs = Option(epicsService.getCommandSender("flamingos2::config"))
 
-    val useElectronicOffsetting = cs.map(_.getInteger("useElectronicOffsetting"))
+    val useElectronicOffsetting = cs.map(_.addInteger("useElectronicOffsetting", "wfs:followA.K", "Enable electronic Offsets", false))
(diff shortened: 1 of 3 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
```diff
@@ -48,7 +48,7 @@ final class Flamingos2Epics(epicsService: CaService) {
   object configCmd extends EpicsCommand {
     override val cs = Option(epicsService.getCommandSender("flamingos2::config"))
 
-    val useElectronicOffsetting = cs.map(_.getInteger("useElectronicOffsetting"))
+    val useElectronicOffsetting = cs.map(_.addInteger("useElectronicOffsetting", "wfs:followA.K", "Enable electronic Offsets", false))
```
cquiroz: Why are we adding an integer channel here? What does it mean?

jluhrs: It is because of a mix of facts:
- The EPICS channel is of type LONG (equivalent to an integer)
- It is a parameter, but it is not a CAD input.
- CAD inputs are always Strings
- `CaParameter`s can have a type different from String, even if they connect to a CAD input (conversions are made internally). When the parameter is created, it must specify if it is a CAD input or not.
- Parameters defined in an XML file don't have that option. It is assumed that they are always CAD inputs

The bottom line is that the `CaParameter` for `useElectronicOffsetting` must be build in code.

cquiroz: Ok 👍

PR #232, opened 2017-05-16
SEQNG-227: Read the title from the ODB

cquiroz:
This PR includes a very small client to read an observation name from the ODB using the odbbrowser API

We may need to move this to another package on the next subtask

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/odbclient/ODBClient.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/odbclient/ODBClient.scala
@@ -3,6 +3,8 @@ package edu.gemini.seqexec.web.server.odbclient
 import scalaz.Kleisli
 import knobs.Config
 import scalaz.concurrent.Task
+import edu.gemini.spModel.core.SPProgramID
+import edu.gemini.seqexec.model.Model.SequenceId
 import org.http4s.client.blaze._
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/odbclient/ODBClient.scala
```diff
@@ -0,0 +1,38 @@
...
+case class ODBClient(config: ODBClientConfig) {
+  val httpClient = PooledHttp1Client()
+  // Entity Decoder for xml
+  implicit val decoder = scalaxml.xml()
+  def observationTitle(id: SPProgramID, obsId: SequenceId): Task[String] = {
+    val baseUri = s"${config.odbHost}/odbbrowser/observations"
+    val target = Uri.fromString(baseUri).toOption.get +?("programReference", id.stringValue)
+    httpClient.expect[Elem](target).map { xml =>
+      for {
+        x <- xml \\ "observations" \ "observation"
```
jdnavarro: Are these operators coming from `blaze` or are they `xml` specific?

cquiroz: These are from `scala-xml` which is part of core scala since forever. It basically takes the root, finds any node `observations` then one node down `observation`, and filters on the id

cquiroz:
Note that I put this on the web server package but most likely I'll need to move it to the engine, or make another module

PR #233, opened 2017-05-17
Fixed problem with sequence execution stalling.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosEpics.scala
```diff
@@ -22,7 +24,7 @@ class GmosEpics(epicsService: CaService) {
     val stageMode = cs.map(_.getString("stageMode"))
     def setStageMode(v: String): SeqAction[Unit] = setParameter(stageMode, v)
 
-    val useElectronicOffsetting = cs.map(_.addInteger("useElectronicOffsetting", "wfs:followA.K", "Enable electronic Offsets", false))
+    val useElectronicOffsetting = cs.map(_.addInteger("useElectronicOffsetting", GMOS_TOP + "wfs:followA.K", "Enable electronic Offsets", false))
```
cquiroz: Why is this changing only here and not in other vals on `GmosEpics`?

jluhrs: The other vals call getXXX, which retrieves an already existing parameter. `useElectronicOffsetting` is special: it cannot be build from the XML file. Here it is being built in the call to addInteger. I think the confusion is that this one includes the EPICS channel name in its parameters, which was the one missing the prefix.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -201,9 +201,11 @@ package object engine {
       _.map { seq =>
         seq match {
           case Sequence.State.Final(_,_) => unit
-          case _          => Nondeterminism[Task].gatherUnordered (
-            seq.current.actions.zipWithIndex.map (act)
-            ).liftM[HandleStateT] *> send (q) (executed (id) )
+          case _                         => Task.delay {
+            (Nondeterminism[Task].gatherUnordered (
+              seq.current.actions.zipWithIndex.map (act)
+            ) *> q.enqueueOne(executed (id) )).unsafePerformAsync(x => ())
```
jdnavarro: I've been bitten several times by Tasks getting stuck that when using `unsafePerformAsync` made it work. In almost every case there was an extra Task wrapping the true Task that needs to be removed. The double Task can be hard to spot because the type checker can't catch it. Here, at a glance, I can't find where it is though.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -122,7 +122,9 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
     )
     val u = t.flatMapF{
       case (err::errs, None) => q.enqueueOne(Event.logMsg(SeqexecFailure.explain(err))).map(_.right)
-      case (errs, Some(seq)) => ((if(errs.isEmpty) Task(()) else q.enqueueAll(errs.map(e => Event.logMsg(SeqexecFailure.explain(e))))) *> q.enqueueOne(Event.load(seqId.stringValue(), seq))).map(_.right)
+      case (errs, Some(seq)) => (if(errs.isEmpty) Task(()) else q.enqueueAll(errs.map(e => Event.logMsg(SeqexecFailure.explain(e))))) *> Task.delay {
```
cquiroz: I think instead of `Task(())` you may want to use `Task.now(())`

cquiroz:
This may or may be not relevant but looking at the code I see this

https://github.com/gemini-hlsw/ocs3/blob/develop/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala#L120

There we are using `Task.apply` instead of `Task.delay`. There is a subtle difference between them especially in terms of asynchrony

PR #235, opened 2017-05-18
SEQNG-234: AmpCount should be written as a string

cquiroz:
The AmpCount values should be written as a String according to the seqexec LUT

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
@@ -45,9 +45,10 @@ object GmosControllerEpics extends GmosSouthController {
   implicit val ampGainSettingEncoder: EncodeEpicsValue[AmpGainSetting, String] = EncodeEpicsValue(_.value.toString)
 
   implicit val ampCountEncoder: EncodeEpicsValue[AmpCount, String] = EncodeEpicsValue {
-    case AmpCount.THREE  => "BEST"
-    case AmpCount.SIX    => "ALL"
-    case AmpCount.TWELVE => "SECONDARY"
+    // gmosAmpCount.lut
+    case AmpCount.THREE  => ""
+    case AmpCount.SIX    => "BEST"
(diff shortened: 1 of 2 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
```diff
@@ -44,10 +44,10 @@ object GmosControllerEpics extends GmosSouthController {
...
   implicit val ampGainSettingEncoder: EncodeEpicsValue[AmpGainSetting, String] = EncodeEpicsValue(_.value.toString)
 
-  implicit val ampCountEncoder: EncodeEpicsValue[AmpCount, Int] = EncodeEpicsValue {
-    case AmpCount.THREE  => 3
-    case AmpCount.SIX    => 6
-    case AmpCount.TWELVE => 12
+  implicit val ampCountEncoder: EncodeEpicsValue[AmpCount, String] = EncodeEpicsValue {
+    case AmpCount.THREE  => "BEST"
+    case AmpCount.SIX    => "ALL"
+    case AmpCount.TWELVE => "SECONDARY"
```
jluhrs: Contents of gmosAmpCount.lut:
#OT amp count	GMOS amp mode
Six				BEST
Twelve			ALL

jluhrs: THREE is not an option in OT.

cquiroz: hmm, I missunderstood. THREE is in the OCS code, maybe hidden in the UI?

sraaphorst: Yes, I just worked on this, and THREE is an option in the OCS code, but only for, IIRC, E2V detectors. I think now that we are using Hamamatsu CCDs for both GMOS-N and GMOS-S, THREE may be obsolete, but I am not certain.

PR #236, opened 2017-05-22
Http4s update

cquiroz:
This PR updates http4s to version 0.16-M2. This version should be more stable when stopping the server but it bring some API changes


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -110,7 +113,7 @@ object WebServerLauncher extends ServerApp with LogInitialization {
   /**
     * Reads the configuration and launches the web server
     */
-  override def server(args: List[String]): Task[Server] = {
+  override def process(args: List[String]): Process[Task, Unit] = {
```
jdnavarro: Is this the new recommended way of running the application? Is this related to the possibility of cleaning up before exiting?

cquiroz: Yes, that's exactly right. It was found that in many cases cleanup didn't proceed correctly. I don't think we do any cleanup but we might want to do it properly

PR #238, opened 2017-05-24
SEQNG-193: GMOS FITS keywords

cquiroz:
This PR adds support for the GMOS FITS keywords baring the N&S related ones. Most of the logic comes from the original TCL code


cquiroz:
Travis passed, maybe #237 did help after all

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosEpics.scala
```diff
@@ -256,20 +250,20 @@ object GmosEpics extends EpicsSystem[GmosEpics] {
...
+    val ccdXsize: Option[CaParameter[Integer]] = cs.map(_.getInteger(s"ccdXsize$i"))
     def setCcdXsize1(v: Integer): SeqAction[Unit] = setParameter(ccdXsize, v)
 
-    val ccdYsize = cs.map(_.getInteger(s"ccdYsize$i"))
+    val ccdYsize: Option[CaParameter[Integer]] = cs.map(_.getInteger(s"ccdYsize$i"))
     def setCcdYsize1(v: Integer): SeqAction[Unit] = setParameter(ccdYsize, v)
   }
 
-  class RoiStatus(sa: CaStatusAcceptor, i: Int) {
+  case class RoiStatus(sa: CaStatusAcceptor, i: Int) {
```
jluhrs: Why a case class?

cquiroz: Just In general I tend to use case classes unless there is a very good reason not to, you get a bunch of goodies, including proper equals and hashCode

PR #239, opened 2017-05-25
Better implicit conversions

cquiroz:
While doing #238 I noted that we do implicit conversions. They are not so much recommended nowadays given that they lead to unexpected actions, see [JavaConverters and JavaConversions](https://stackoverflow.com/questions/8301947/what-is-the-difference-between-javaconverters-and-javaconversions-in-scala)

I also added a little typeclass to make the conversion more generic


jluhrs:
I agree that to make explicit the use of default values when there is no real value available for a keyword is better.

cquiroz:
`orDefault` is just a name I picked up. I could have use `orZero` to be consistent with scalaz but our "Zero" is 9999 so it seemed odd

PR #240, opened 2017-05-25
SEQNG-238 Created unit test to prove that event processing is not blocked by Actions.

jluhrs:
The test tries to run an sequence with an Action that blocks, and is awaken by a later event. So if the event is not run, the sequence never ends, causing the test to fail.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
+++ b/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
@@ -1,12 +1,14 @@
 package edu.gemini.seqexec.engine
 
-import Result._
+import java.util.concurrent.Semaphore
+
 import Event._
 import org.scalatest.FlatSpec
 
-import edu.gemini.seqexec.model.Model.SequenceState.{Idle, Error}
+import edu.gemini.seqexec.model.Model.SequenceState.{Error, Idle}
 import edu.gemini.seqexec.model.Model.{Conditions, SequenceMetadata, SequenceState, StepConfig}
 
+import scala.concurrent.duration._
 import scalaz.syntax.apply._
 import scalaz.syntax.foldable._
 import scalaz.Nondeterminism
@@ -114,9 +116,10 @@ class packageSpec extends FlatSpec {
   val qs2 = Engine.State(Conditions.default, None, qs1.sequences + (seqId2 -> qs1.sequences(seqId1)))
   val qs3 = Engine.State(Conditions.default, None, qs2.sequences + (seqId3 -> seqG))
 
+  def isFinished(status: SequenceState): Boolean =
+    status == Idle || status == edu.gemini.seqexec.model.Model.SequenceState.Completed || status === Error
+
   def runToCompletion(q: scalaz.stream.async.mutable.Queue[Event], s0: Engine.State): Engine.State = {
-    def isFinished(status: SequenceState): Boolean =
-      status == Idle || status == edu.gemini.seqexec.model.Model.SequenceState.Completed || status === Error
 
     q.enqueueOne(start(seqId)).flatMap(
       _ => processE(q).drop(1).takeThrough(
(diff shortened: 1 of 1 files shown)


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -197,4 +200,41 @@ class packageSpec extends FlatSpec {
...
+          )
+        )
+      ) ) ) )
+    )
+
+    val result = Nondeterminism[Task].both(
+        q.enqueueOne(start(seqId)) *> Task.apply(startedFlag.acquire) *>
+          q.enqueueOne(Event.getState{_ => Task.delay{finishFlag.release}}),
+        processE(q).drop(1).takeThrough(a => !isFinished(a._2.sequences(seqId).status)).run.eval(qs)
+      ).timed(5.seconds).unsafePerformSyncAttempt
```
jluhrs: In general, I would avoid having timeouts in the tests. But in this case the wait only happens in the bad path, so I think it's OK.

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -1,12 +1,14 @@
 package edu.gemini.seqexec.engine
 
-import Result._
+import java.util.concurrent.Semaphore
+
 import Event._
 import org.scalatest.FlatSpec
 
-import edu.gemini.seqexec.model.Model.SequenceState.{Idle, Error}
+import edu.gemini.seqexec.model.Model.SequenceState.{Error, Idle}
```
cquiroz: IDEA sorts alphabetically

cquiroz:
Unrelated but it seems travis is happy again. I'll reenable the requirement of a passing build

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -197,4 +200,41 @@ class packageSpec extends FlatSpec {
...
+        SequenceMetadata("GMOS", None),
+        List(
+          Step(
+            1,
+            None,
+            config,
+            Set(Resource.GMOS),
+            breakpoint = false,
+            List(
+              List(Task.apply{
```
jluhrs: This worries me a bit. The test would not work if I used `delay` instead of `apply`. Which means that the engine is not robust to the kind of Tasks passed as Actions.

cquiroz: And you have mixed Task.apply and Task.delay in the test

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -197,4 +200,41 @@ class packageSpec extends FlatSpec {
       ).unsafePerformSync._2.get._2.sequences(seqId3).status === SequenceState.Running
     )
   }
+
+  "engine" should "keep processing input messages regardless of how long Actions take" in {
+    val q = async.boundedQueue[Event](10)
+    val startedFlag = new Semaphore(0)
```
cquiroz: Why the semaphores?

cquiroz:
I don't really understand enough about the engine to say much about this. The test code though looks too complex for me to parse

jdnavarro:
> I don't really understand enough about the engine to say much about this. The test code though looks too complex for me to parse

I think I have a decent understanding of the engine but still I may be too dumb to understand what's going on with this test. What are we trying to solve here? The blocking when more than 10 sequences are loaded? If that's not the case what's exactly the blocking problem we are trying to solve here?

We should be able to reproduce the issue first, or if we can't because we need the instrument, we need a precise description of the circumstances under which this happens, and what exactly happens when it blocks. We can't be trying to solve problems that we are not 100% sure we really have.

I thought the test with an action that hangs was to make sure the engine handles the errors in its children Tasks properly. I was expecting that would be done after fixing the issue, so that we don't regress again.

jluhrs:
This is not about the problem with a bounded queue. It is about the problem of long `Action`s blocking the processing of events, as seeing in the tests with FLAMINGOS-2. This test check that that situation does not occur.
The test works like this: The initial state has a single `Sequence`, which has one step with one `Execution`, composed of only one `Action`. The `Action` uses two `Semaphore`s, When it runs, it signals the first semaphore (`startedFlag`) , and waits on the second one (`finishFlag`). The engine is run in parallel with a second `Task`. The second `Task` put a `Start` event in the event queue to start the execution of the sequence, and then waits on `startedFlag`. After that, it puts a `GetState` in the queue, which is an event that carries a state dependent `Task` that is run by the engine. This particular `Task` signals `finishFlag`, allowing the `Action` to finish. Because the `GetState` event is put in the queue only after the Action has signaled `startedFlag`, we are sure that the event is put when the Action is in progress. If some problem in the engine prevented it to process the following events, `GetState` would never run, Action would never end, and the execution of the engine in the test will end because of a timeout.
The test pass in the current implementation of the engine. I tried it after reverting the change that introduced that ugly `unsafePerformAsync`, and then it failed.

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
jluhrs: Yes, but I used `apply` only on the `Task`s that block.

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
jluhrs: See my explanation below.

PR #244, opened 2017-05-31
SEQNG-165: Port UI to the new scalajs-react library

cquiroz:
This PR brings the UI to use the new scalajs-react library 1.0 which is a  fairly large upgrade. Thus this PR is fairly large. 

On an inspection of the UI it seems all features are working but there could be some details here and there not being treated correctly

One thing left out as it changed quite a bit is the autoscrolling when the sequence advances. That will be addressed in a separate PR


cquiroz:
Yes, it is somewhat mechanical, after all the underlying React library is the same so it has to endup calling the same js. But it is internally simpler and with fewer quirks. 

I played with the UI but certainly if you can check it, it would be very useful

cquiroz:
You shouldn't see differences :smile: That's the idea it is only a library port

PR #246, opened 2017-06-01
SEQNG-241 Pass events as an output instead of relying on Queue.enqueueOne

jluhrs:
This PR is to have an initial review. I still need to fix the tests.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
@@ -19,10 +19,11 @@ object Engine {
 
   def empty[A]: Engine[A] = Engine(Map.empty)
 
-  implicit val engineFunctor = new Functor[Engine] {
-    def map[A, B](q: Engine[A])(f: A => B): Engine[B] =
-      Engine(q.sequences.mapValues(_.map(f)))
-  }
+// This fails to compile with the error "not found: type $anon"
+//  implicit val engineFunctor = new Functor[Engine] {
+//    def map[A, B](q: Engine[A])(f: A => B): Engine[B] =
+//      Engine(q.sequences.mapValues(_.map(f)))
+//  }
 
   case class State(conditions: Conditions, operator: Option[Operator], sequences: Map[Sequence.Id, Sequence.State])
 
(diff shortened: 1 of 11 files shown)


cquiroz:
I've tried to follow the changes but I'd probably need much more understanding on the details of the implementation to comment. So I'd rather pass on judging this PR, perhaps @jdnavarro has a much better grasp on this

As a general comment I'm worried about ignoring tests. The engine has too few tests already for my tests and ignoring more is probably not a good idea. I understand that your are trying to make all the tests run, right?

jdnavarro:
> As a general comment I'm worried about ignoring tests. The engine has too few tests already for my tests and ignoring more is probably not a good idea. I understand that your are trying to make all the tests run, right?

I'm also waiting for the tests to be updated in order to help me understand how it will work now.

jluhrs:
The PR is now in a good state. Tests were fixed.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -19,10 +19,11 @@ object Engine {
 
   def empty[A]: Engine[A] = Engine(Map.empty)
 
-  implicit val engineFunctor = new Functor[Engine] {
-    def map[A, B](q: Engine[A])(f: A => B): Engine[B] =
-      Engine(q.sequences.mapValues(_.map(f)))
-  }
+// This fails to compile with the error "not found: type $anon"
```
cquiroz: So we don't need to rewrite this?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -273,80 +302,73 @@ package object engine {
...
   /** Traverse a process with a stateful computation. */
-  private def mapEvalState[F[_]: Monad: Catchable, A, S, B](
-    fs: Process[F, A], s: S, f: A => StateT[F, S, B]
-  ): Process[F, B] = {
-    def go(fs: Process[F, A], s: S): Process[F, B] =
-      Process.eval(fs.unconsOption).flatMap {
-        case None         => Process.halt
-        case Some((h, t)) => Process.eval(f(h).run(s)).flatMap {
-          case (s, a) => Process.emit(a) ++ go(t, s)
+  private def mapEvalState[A, S, B](
```
cquiroz: Very nice. As an extended challenge you could abstract the types for `Option` and `Task` 😄

cquiroz:
after an explanation live of the code I think I now understand it much better, nice job

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jluhrs: I thought of not using Option, but pass an empty Process when no one was needed. Maybe I should? The code could be a bit cleaner.

PR #247, opened 2017-06-02
SEQNG-243: Anonymize sensitive data on client feed

cquiroz:
We are hiding some data, specifically the observation name, for anonymous users. However this is enforced at the UI level but the data is present on the WebSocket stream. This PR adds a filter so that anonymous and logged in users can get different data

Note that I started using [Monocle](http://julien-truffaut.github.io/Monocle/) for lenses. @tpolecat and @swalker2m, please let me know if that's ok

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,7 @@ lazy val edu_gemini_seqexec_web_server = project.in(file("modules/edu.gemini.seq
   .enablePlugins(BuildInfoPlugin)
   .settings(commonSettings: _*)
   .settings(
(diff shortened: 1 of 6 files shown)


swalker2m:
> Note that I started using Monocle for lenses. @tpolecat and @swalker2m, please let me know if that's ok

I think Monocle is quite good.

cquiroz:
@jdnavarro I'd say we should use them at least on new code. If you want to port existing that's ok too but let's focus on the features we are developing at the moment

PR #248, opened 2017-06-05
SEQNG-247: Redirect http calls to https

cquiroz:
This is a simple but useful improvement. If you hit the seqexec by mistake using http, it will redirect you to the https secure site

The code is ready but it needs a small change to the firewall rules on each machine that will be requested to ITS


cquiroz:
Right, we could do it in a frontend like nginx but we don't have it. I don't want to make deployment more complex considering we only have a few clients

Perhaps on the OT we'd put a frontend for the static files

PR #250, opened 2017-06-06
SEQNG-248 Update ocs to 2017B 1.1.1

jdnavarro:


cquiroz:
I think this is missing uploading the new libraries to the maven-repo

PR #251, opened 2017-06-06
SEQNG-249 Rewrote automatic load to remove calls to unsafePerformAsync

jluhrs:
Used the feature implemented in SEQNG-241 to remove the use of Queue.enqueueOne and the calls to unsafePerformAsync.
I  paid more attention to the managing of errors when loading sequences, and I think I may have fixed SEQNG-237.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -130,28 +130,24 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
     }).run.map(_ => i)
   }
 
-  def load(q: EventQueue, seqId: SPObservationID): Task[SeqexecFailure \/ Unit] = {
+  def load(q: EventQueue, seqId: SPObservationID): Task[SeqexecFailure \/ Unit] = loadEvents(seqId).flatMapF(q.enqueueAll(_).map(_.right)).run
+
+  def loadEvents(seqId: SPObservationID): SeqAction[List[Event]] = {
     val t: EitherT[Task, SeqexecFailure, (List[SeqexecFailure], Option[Sequence[Action]])] = for {
-      odbSeq       <- EitherT(Task(odbProxy.read(seqId)))
+      odbSeq       <- EitherT(Task.delay(odbProxy.read(seqId)))
       progIdString <- EitherT(Task.delay(odbSeq.extract(OCS_KEY / InstConstants.PROGRAMID_PROP).as[String].leftMap(ConfigUtilOps.explainExtractError)))
       progId       <- EitherT.fromTryCatchNonFatal(Task.now(SPProgramID.toProgramID(progIdString))).leftMap(e => SeqexecFailure.SeqexecException(e): SeqexecFailure)
       name         <- EitherT(odbClient.observationTitle(progId, seqId.toString).map(_.leftMap(ConfigUtilOps.explainExtractError)))
     } yield translator.sequence(translatorSettings)(seqId, odbSeq, name)
 
-    val u = t.flatMapF{
-      case (err :: _, None)  => q.enqueueOne(Event.logMsg(SeqexecFailure.explain(err))).map(_.right)
-      case (errs, Some(seq)) =>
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -130,28 +130,24 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
     }).run.map(_ => i)
   }
 
-  def load(q: EventQueue, seqId: SPObservationID): Task[SeqexecFailure \/ Unit] = {
+  def load(q: EventQueue, seqId: SPObservationID): Task[SeqexecFailure \/ Unit] = loadEvents(seqId).flatMapF(q.enqueueAll(_).map(_.right)).run
+
+  def loadEvents(seqId: SPObservationID): SeqAction[List[Event]] = {
     val t: EitherT[Task, SeqexecFailure, (List[SeqexecFailure], Option[Sequence[Action]])] = for {
-      odbSeq       <- EitherT(Task(odbProxy.read(seqId)))
+      odbSeq       <- EitherT(Task.delay(odbProxy.read(seqId)))
```

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -228,17 +224,20 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
-  private def refreshSequenceList(q: EventQueue): Engine.State => Task[Unit] = (st: Engine.State) => {
+  private def refreshSequenceList(q: EventQueue): Engine.State => Task[Option[Process[Task, Event]]] = (st: Engine.State) => {
 
     val seqexecList = st.sequences.keys.toSeq.map(v => new SPObservationID(v))
 
-    def loads(odbList: Seq[SPObservationID]): Seq[Task[SeqexecFailure \/ Unit]] = odbList.diff(seqexecList).map(id => load(q, id))
+    def loads(odbList: Seq[SPObservationID]): Task[List[Event]] = odbList.diff(seqexecList).toList.map(id => loadEvents(id)).sequenceU.map(_.flatten).run.map(_.valueOr( r => List(Event.logMsg(SeqexecFailure.explain(r)))))
 
-    def unloads(odbList: Seq[SPObservationID]): Seq[Task[SeqexecFailure \/ Unit]] = seqexecList.diff(odbList).map(id => unload(q, id))
+    def unloads(odbList: Seq[SPObservationID]): Seq[Event] = seqexecList.diff(odbList).map(id => unload(id))
```
cquiroz: Why this  returns `Seq[Event]` unlike `loads` that returns `Task[List[Event]]`. I'd think they should be isomorphic

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -228,17 +224,20 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
 
-    odbProxy.queuedSequences().flatMap(seqs => EitherT(Nondeterminism[Task].gatherUnordered(loads(seqs) ++
-      unloads(seqs)).map(_.sequenceU))
-    ).run.map(_ => ())
+    val x = odbProxy.queuedSequences().flatMapF(seqs => loads(seqs).map(ee => (ee ++ unloads(seqs)).right)).run
+    val y = x.map(_.valueOr(r => List(Event.logMsg(SeqexecFailure.explain(r)))))
+    y.map{ee =>
+      if(ee.isEmpty) None
```
cquiroz: I think you can do `ee option Process.emitAll...`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: `loads` reads data from an external source, that is why each load is in a `Task`. It can also fail, therefore the Task output is in a `\/`.
The case of `unloads` is different. The computation does not have side effects, and it cannot fail. Therefore, you can just output the result as it is.

PR #252, opened 2017-06-07
SEQNG-236 Added new FLAMINGOS-2 filters.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2ControllerEpics.scala
```diff
@@ -87,8 +87,8 @@ object Flamingos2ControllerEpics extends Flamingos2Controller {
       case Filter.JH      => "JH_G0809"
       case Filter.HK      => "HK_G0806"
       case Filter.DARK    => "DK_G0807"
-      case Filter.K_BLUE  => ???
-      case Filter.K_RED   => ???
+      case Filter.K_BLUE  => "K-blue_G0814"
```
cquiroz: I wonder if we should use constants from the OCS model rather than directly the strings, WDYT?

jluhrs: Those strings are not in the OCS model.

cquiroz: Are they from a LUT?

jluhrs: The Tcl Seqexec has them in the code.

PR #255, opened 2017-06-13
SEQNG-146 Load sequences with executed steps from the ODB

jdnavarro:
I still have to take care of Steps marked as skipped and write some tests, but the most important feature of this PR is already there. I could confirm manually from the UI that it can load an ongoing Sequences.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
@@ -189,7 +189,7 @@ object Sequence {
       * Initialize a `State` passing a `Queue` of pending `Sequence`s.
       */
     // TODO: Make this function `apply`?
-    def init(q: Sequence[Action]): State = Sequence.Zipper.currentify(q).map(Zipper(_, SequenceState.Idle))
+    def init(q: Sequence[Action \/ Result]): State = Sequence.Zipper.currentify(q).map(Zipper(_, SequenceState.Idle))
       .getOrElse(Final(Sequence.empty(q.id), SequenceState.Idle))
 
     /**
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
@@ -27,7 +27,7 @@ object Sequence {
 
   def empty[A](id: Id): Sequence[A] = Sequence(id, SequenceMetadata("", None, ""), Nil)
 
-  implicit val SequenceFunctor = new Functor[Sequence] {
+  implicit val sequenceFunctor = new Functor[Sequence] {
     def map[A, B](fa: Sequence[A])(f: A => B): Sequence[B] =
       Sequence(fa.id, fa.metadata, fa.steps.map(_.map(f)))
   }
@@ -122,15 +122,15 @@ object Sequence {
       separate(seq).flatMap {
         case (pending, done)   => pending match {
           case Nil             => None
-          case stepP :: stepPs =>
-            Step.Zipper.currentify(stepP).map(
-              Zipper(seq.id, seq.metadata, stepPs, _, done)
+          case s :: ss =>
+            Step.Zipper.currentify(s).map(
+              Zipper(seq.id, seq.metadata, ss, _, done)
             )
         }
       }
 
-    // MonadPlus' `separate` would be good if we wanted to separate Actions or Results,
-    // but here we want only Steps.
+    // We would use MonadPlus' `separate` if we wanted to separate Actions or
+    // Results, but here we want only Steps.
     private def separate(seq: Sequence[Action \/ Result]): Option[(List[Step[Action]], List[Step[Result]])] = {
 
       seq.steps.foldLeftM[Option, (List[Step[Action]], List[Step[Result]])]((Nil, Nil))(
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
@@ -1,9 +1,11 @@
 package edu.gemini.seqexec.engine
 
-import edu.gemini.seqexec.model.Model.{SequenceMetadata, SequenceState}
+import edu.gemini.seqexec.model.Model.{SequenceMetadata, SequenceState, StepState}
+import edu.gemini.seqexec.engine.Step._
 
 import scalaz._
 import Scalaz._
+import scalaz.concurrent.Task
 
 /**
   * A list of `Step`s grouped by target and instrument.
@@ -116,6 +118,52 @@ object Sequence {
           )
       }
 
+    def zipper(seq: Sequence[Action \/ Result]): Option[Zipper] =
+      separate(seq).flatMap {
+        case (pending, done)   => pending match {
+          case Nil             => None
+          case stepP :: stepPs =>
+            Step.Zipper.currentify(stepP).map(
+              Zipper(seq.id, seq.metadata, stepPs, _, done)
+            )
+        }
+      }
+
+    // MonadPlus' `separate` would be good if we wanted to separate Actions or Results,
+    // but here we want only Steps.
+    private def separate(seq: Sequence[Action \/ Result]): Option[(List[Step[Action]], List[Step[Result]])] = {
+
+      seq.steps.foldLeftM[Option, (List[Step[Action]], List[Step[Result]])]((Nil, Nil))(
+        (acc, step) =>
+        if (Step.status(step) === StepState.Pending)
+          Some(
+            acc.leftMap(
+              _ :+ step.map(
+                _.fold(
+                  identity,
+                  // It should never happen
+                  _ => Task(Result.Error("Inconsistent status"))
+                )
+              )
+            )
+          )
+        else if (Step.status(step) === StepState.Completed)
+          Some(
+            acc.rightMap(
+              _ :+ step.map(
+                _.fold(
+                  // It should never happen
+                  _ => Result.Error("Inconsistent status"),
+                  identity
+                )
+              )
+            )
+          )
+        else None
+      )
+
+    }
+
     val focus: Zipper @> Step.Zipper =
       Lens.lensu((s, f) => s.copy(focus = f), _.focus)
 
@@ -178,8 +226,8 @@ object Sequence {
         (qs, s) => (
           qs match {
             // TODO: Isn't there a better way to write this?
-            case Zipper(st, _)  => Zipper(st, s)
-            case Final(st, _)   => Final(st, s)
+            case Zipper(st, _) => Zipper(st, s)
+            case Final(st, _)  => Final(st, s)
           }
           ),
         _.status
@@ -189,7 +237,7 @@ object Sequence {
       * Initialize a `State` passing a `Queue` of pending `Sequence`s.
       */
     // TODO: Make this function `apply`?
-    def init(q: Sequence[Action \/ Result]): State = Sequence.Zipper.currentify(q).map(Zipper(_, SequenceState.Idle))
+    def init(q: Sequence[Action \/ Result]): State = Sequence.Zipper.zipper(q).map(Zipper(_, SequenceState.Idle))
       .getOrElse(Final(Sequence.empty(q.id), SequenceState.Idle))
 
     /**
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
@@ -14,6 +14,7 @@ case class Step[+A](
   config: StepConfig,
   resources: Set[Resource],
   breakpoint: Boolean,
+  skip: Boolean,
   executions: List[List[A]]
 )
 
@@ -71,6 +72,7 @@ object Step {
     config: StepConfig,
     resources: Set[Resource],
     breakpoint: Boolean,
+    skip: Boolean,
     pending: List[Actions],
     focus: Execution,
     done: List[Results],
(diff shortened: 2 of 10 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -47,12 +47,11 @@ object Step {
...
-      // At least an Action in this Step errored.
-      // TODO: These errors for empty cases should be enforced at the type level
-      if (step.executions.isEmpty
-            || step.executions.all(_.isEmpty)
-      ) StepState.Error("This should never happen, please submit a bug report")
+    ).map(StepState.Error).getOrElse
+      // It's possible to have a Step with empty executions when a completed
+      // Step is loaded from the ODB.
+      if (step.executions.isEmpty || step.executions.all(_.isEmpty)
+      ) StepState.Completed
```
cquiroz: Funny indentation

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -126,7 +125,7 @@ class packageSpec extends FlatSpec {
   it should "be in Running status after starting" in {
     val p = Process.eval(Task.now(start(seqId)))
     val qs = process(p)(qs1).take(1).runLast.unsafePerformSync.get._2
-    assert(qs.sequences(seqId).status === SequenceState.Running)
+    assert(qs.sequences(seqId).status == SequenceState.Running)
```
cquiroz: Why not to continue with the typesafe equal?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -15,7 +15,6 @@ import edu.gemini.seqexec.web.server.http4s.encoder._
 import edu.gemini.spModel.core.SPBadIDException
 
 import org.http4s._
-import org.http4s.server.syntax._
```
cquiroz: 👍

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -106,7 +135,19 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     val instName = configs.headOption.map(extractInstrumentName).getOrElse("Unknown instrument")
 
     steps match {
-      case (errs, ss) => (errs, if(ss.isEmpty) None else Some(Sequence[Action](obsId.stringValue(), SequenceMetadata(instName, None, name), ss)))
+      case (errs, ss) => (
+        errs,
+        if (ss.isEmpty)
```
cquiroz: maybe you can do `ss.isEmpty option ...`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -260,7 +258,7 @@ object SeqexecEngine {
                       odbQueuePollingInterval: Duration)
   val defaultSettings = Settings(Site.GS,
     "localhost",
-    LocalDate.of(2017, 1,1),
+    LocalDate.of(2017, 1, 1),
```
cquiroz: Why do me have a date in the settings?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -63,23 +68,41 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
...
-        List(
-          sys.map(x => toAction(x.configure(config).map(y => Result.Configured(y.sys.name)))),
-          List(toAction(observe(config).map(x => Result.Observed(x.dataId))))
-        )
-        ++
-        (if(last) List(List(toAction(systems.odb.sequenceEnd(obsId).map(_ => Result.Ignored))))
-        else List())
-      )
+      extractStatus(config) match {
+        case "ready" =>
```
cquiroz: Is this when a step has been already observed according to the odb?

cquiroz: It maybe clearer to replace `"ready"` with a constant like `StepObserved` or something like that

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
jdnavarro: For some reason, now the compiler couldn't infer that `SequenceState` is an instance of *Equal*, no idea why. I try fiddling but couldn't figure it out. If you know how to fix it, please let me know.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jdnavarro: This is the `String` that needs to be extracted from the configuration keywords that tells that the Step is pending. Do you mean to have something like an alias like `val StepObserved = "ready"`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
cquiroz: Right, @jluhrs why do we have a date here?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
cquiroz: Yes, perhaps there is even some constant on spModel. My rationale is that `ready` doesn't tell you much but some other constant could be clearer?

jdnavarro: I did that commit some time ago, but now that you mention it, I vaguely remember looking for some constants in spModel but couldn't finding them. I'm not sure if it was for this case though, so I'll check again.

cquiroz: Not every value has a constant on spModel, so we could just create one locally

jdnavarro: I found `STATUS_PROP` for the key, but I can't find any constants for the values. Where should I place the constants? In the same file where is being used? Is there any file in the `ocs3` project for that?

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
swalker2m: Perhaps it's just that `Equal` is invariant.  What if you add an explicit type annotation `(SequenceState.Running: SequenceState)` or maybe add a constructor that produces a `SequenceState.Running` but typed as `SequenceState`?

jdnavarro: 😕 I tried the explicit types but still getting the same error. This really puzzles me because this PR didn't touch directly anything related to `SequenceState` and previously there was no issue. This is the concrete error:

```
[error] /Users/jnavarro/Develop/ocs3/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala:128: type mismatch;
[error]  found   : edu.gemini.seqexec.model.Model.SequenceState
[error]  required: ?{def ===(x$1: ? >: edu.gemini.seqexec.model.Model.SequenceState): ?}
[error] Note that implicit conversions are not applicable because they are ambiguous:
[error]  both method ToEqualOps in trait ToEqualOps of type [F](v: F)(implicit F0: scalaz.Equal[F])scalaz.syntax.EqualOps[F]
[error]  and method convertToEqualizer in trait TripleEquals of type [T](left: T)packageSpec.this.Equalizer[T]
[error]  are possible conversion functions from edu.gemini.seqexec.model.Model.SequenceState to ?{def ===(x$1: ? >: edu.gemini.seqexec.model.Model.SequenceState): ?}
[error]     assert((qs.sequences(seqId).status: SequenceState) === (SequenceState.Running: SequenceState))
[error]                                       ^
[error] /Users/jnavarro/Develop/ocs3/modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala:128: value === is not a member of edu.gemini.seqexec.model.Model.SequenceState
[error]     assert((qs.sequences(seqId).status: SequenceState) === (SequenceState.Running: SequenceState))
```

jdnavarro: I also tried fiddling with the imports but still couldn't fix it. Do the imports look alright to you?

swalker2m: I see.  Scala refuses to do an implicit conversion when there are two options that might apply in scope.  One is the `scalaz.EqualOps` and the other `TripleEquals` I'm not familiar with.  Perhaps there are imports that could be tightened to avoid picking up `TripleEquals`?

swalker2m: I didn't notice this was a test case.  I have had the same issue because all the test libraries want to define `===` too.  I just punt and use `==` in test cases too, but maybe you could explicitly hide TripleEquals in an import if you want.

cquiroz: Right, I think you can add a trait to solve it, something like http://doc.scalatest.org/1.8/org/scalatest/NonImplicitAssertions$.html

jdnavarro: > Right, I think you can add a trait to solve it, something like http://doc.scalatest.org/1.8/org/scalatest/NonImplicitAssertions$.html

This works! This makes me realize that previously we were using the `===` coming scalatest, not from scalaz, so now I wonder how the scalatest `===` checks for equality.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
cquiroz: Not really. `STATUS_PRO` isn't very clear either. Perhaps we just make a local constant

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
jluhrs: One of the attributes of `Settings` is a date, therefore `defaultSettings` needs a date. `defaultSettings` is used on the tests as a base instance of Settings, from which a new one is built with specific values for some attributes.

jluhrs: The date is used by the `DhsSimulator`

cquiroz: ok, that makes sense

cquiroz: Perhaps the date should be hardcoded into the simulator?

jluhrs: The idea was that the DHS simulator is passed the current date, so you can have labels like S20170614S0001 on June 14th, 2017. But if we read the current date inside the simulator, then you have a side effect. Actually, it was like that at some point, but was changed to its current form.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -116,6 +118,52 @@ object Sequence {
...
+      seq.steps.foldLeftM[Option, (List[Step[Action]], List[Step[Result]])]((Nil, Nil))(
+        (acc, step) =>
+        if (Step.status(step) === StepState.Pending)
+          Some(
+            acc.leftMap(
+              _ :+ step.map(
+                _.fold(
+                  identity,
+                  // It should never happen
+                  _ => Task(Result.Error("Inconsistent status"))
```
jluhrs: Why there is a Task here?

jdnavarro: An `Action` is required there by the types, but that code branch should never be reached because I already filtered for all the lefts.

I acknowledge this code sucks, I tried to just get it to work quickly.

jdnavarro:
I think now this is now ready for the final review before merging. Since last review, aside of taking care of some of the comments, I added support for loading `skipped` Steps.

I tested it manually but there are still not automatic tests. I think that should go in a different PR because it will take some time to create ODB test fixtures or at least some kind of data structure that mock the Sequences coming from the ODB. I guess we could probably *steal* something from the current `ocs`.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jdnavarro: I had a commit where I used `option` instead of the `if...else` but removed it at the last moment because with that specific commit nothing was getting loaded. I didn't get into details why it was happening.

PR #258, opened 2017-06-19
Fixed FITS keywords

jluhrs:
Fixed two problems:
Default values were applied even when not needed.
Guide values had to be converted to expected Strings.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/resources/Tcs.xml
+++ b/modules/edu.gemini.seqexec.server/src/main/resources/Tcs.xml
@@ -1391,12 +1391,6 @@
             <type>DOUBLE</type>
             <description>A&amp;G in position</description>
         </attribute>
-        <attribute name="aoName">
-            <channel>aoName</channel>
-            <type>STRING</type>
-            <description>AO fold position</description>
-            <top>ag</top>
-        </attribute>
         <attribute name="port3Label">
             <channel>port:port3</channel>
             <type>STRING</type>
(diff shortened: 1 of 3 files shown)


modules/edu.gemini.seqexec.server/src/main/resources/Tcs.xml
```diff
@@ -1391,12 +1391,6 @@
             <type>DOUBLE</type>
             <description>A&amp;G in position</description>
         </attribute>
-        <attribute name="aoName">
-            <channel>aoName</channel>
-            <type>STRING</type>
-            <description>AO fold position</description>
-            <top>ag</top>
-        </attribute>
```
jluhrs: I took this out because it was dupliceted.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -381,6 +381,12 @@ class StandardHeader(
       val windowsCount = buildInt32(SeqAction(timingWindows.length), "NUMREQTW")
       sendKeywords(id, inst, hs, windowsCount :: windows)
     }
+    
+    def decodeGuide(v: StandardGuideOptions.Value): String = v match {
+      case StandardGuideOptions.Value.park   => "parked"
+      case StandardGuideOptions.Value.guide  => "guiding"
+      case StandardGuideOptions.Value.freeze => "frozen"
+    }
```
jdnavarro: Is `StandardGuideOptions.Value` an enumerated type?

jluhrs: Yes

jdnavarro: So I guess the `_.toString` method doesn't match the what needs to be sent. How is it different? Why not fix it upstream?

PR #259, opened 2017-06-19
SEQNG-263 OBJECT keyword now has value Dark for Manual Darks.

jluhrs:


cquiroz:
The code seems fine but I don't understand why this change. Could you expand the description of the PR?

jluhrs:
Well, there was a time when the value coming from the ODB was "Dark", before automatic calibrations was introduced. That value was written as the keyword OBJECT (the keyword has the name of the sky object being observed, or the calibration type). Now the value coming from the ODB is "Manual Dark", but there was the requirement to keep the value "Dark" for the keyword OBJECT.

jluhrs:
This was a behavior of the old Seqexec that I miss to translate.

PR #261, opened 2017-06-20
Engine logging

jdnavarro:
Not very configurable yet but better than a `println`, specially when testing with the instrument.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -5,6 +5,7 @@ import java.io.{PrintWriter, StringWriter}
 import edu.gemini.seqexec.engine.Event._
 import edu.gemini.seqexec.engine.Result.{PartialVal, RetVal}
 import edu.gemini.seqexec.model.Model.{CloudCover, Conditions, ImageQuality, SequenceState, SkyBackground, WaterVapor}
+import java.util.logging.{Logger => JLogger}
 
 import scalaz._
 import Scalaz._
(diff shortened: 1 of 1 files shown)


jdnavarro:
> What would it take to support other levels?

We could create an ADT for different logging levels, but I'm pretty sure there must be a way to get the logging level values from the Java Logging API.

The problem is that I'm not familiar with how logging is usually done in functional Scala using the Java logging facilities. If you can point to any example I could try to ape it.

jluhrs:
:+1: :+1: :+1:

jdnavarro:
BTW, all the logging messages in the engine as of now would still go into `info` even if we support different levels.

jluhrs:
I just remembered, the reason we didn't push to improve logging in Seqexec was that Shane was working on the logging for the ODB, and we didn't want to duplicate the effort.

cquiroz:
I suppose that in this case you could just add more methods for e.g. `error`, `warn` etc

jdnavarro:
@jluhrs Right, there were some discussions about centralizing all logging into a separate logging facility.

cquiroz:
In any case, great to see the `println` gone

jdnavarro:
> I suppose that in this case you could just add more methods for e.g. error, warn etc

I can add `logWarn` as an example of how to add more levels, but still, this should be considered all provisional :)

swalker2m:
> I just remembered, the reason we didn't push to improve logging in Seqexec was that Shane was working on the logging for the ODB, and we didn't want to duplicate the effort.

Oh?

jdnavarro:
> > I just remembered, the reason we didn't push to improve logging in Seqexec was that Shane was working on the logging for the ODB, and we didn't want to duplicate the effort.

> Oh?

If I remember correctly I think it was discussed only as a possibility when talking about the `ObsLog` rewrite.

I believe the intention was to have some kind of logging separate from the database and have some kind of message filtering to create different views depending on where or by whom the messages need to be seen.

But this was all discussed as something to think about, nothing was written in stone 🗿.

PR #265, opened 2017-06-22
SEQNG-264 Fixed some problems with GMOS interactions.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
```diff
@@ -274,13 +274,15 @@ object GmosControllerEpics extends GmosSouthController {
...
   }
 
+  val PixelsToMicrons = 15.0
+
   def setCCConfig(cc: CCConfig): SeqAction[Unit] = for {
     _ <- setFilters(cc.filter)
     _ <- setDisperser(cc.disperser)
     _ <- setFPU(cc.fpu)
     _ <- CC.setStageMode(encode(cc.stage))
     // TODO Is DTaX channel a double?
```
cquiroz: I guess we can remove the commend

PR #266, opened 2017-06-26
SEQNG-270: Typelevel compiler

cquiroz:
This PR uses the typelevel compiler on ocs3. I found troubles compiling the JS side, while it compiles files it produces invalid js that fails on tests and on creating an executable js for the browser

Hence, I'm only use typelevel for the JVM side until this can be fixed. I'm planning to create a minimal demo to submit for a fix

Using the typelevel compiler requires some code changes especially on terms of more strict match exhaustiveness

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalControllerEpics.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalControllerEpics.scala
@@ -22,6 +22,7 @@ object GcalControllerEpics extends GcalController {
       v match {
         case BinaryOnOff.OFF => LampState.Off
         case BinaryOnOff.ON  => LampState.On
+        case _               => sys.error("Cannot happen")
       }
   }
 
@@ -30,6 +31,7 @@ object GcalControllerEpics extends GcalController {
       v match {
         case LampState.Off => BinaryOnOff.OFF
         case LampState.On  => BinaryOnOff.ON
+        case _             => sys.error("Cannot happen")
       }
   }
 
@@ -38,7 +40,7 @@ object GcalControllerEpics extends GcalController {
       v match {
         case "OPEN"  => Some(Shutter.OPEN)
         case "CLOSE" => Some(Shutter.CLOSED)
-        case _ => None
+        case _       => None
       }
   }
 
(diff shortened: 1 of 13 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalControllerEpics.scala
```diff
@@ -22,6 +22,7 @@ object GcalControllerEpics extends GcalController {
       v match {
         case BinaryOnOff.OFF => LampState.Off
         case BinaryOnOff.ON  => LampState.On
+        case _               => sys.error("Cannot happen")
```
cquiroz: @tpolecat This one surprised me. `BinaryOnOff` is a Java enumeration with only two values, however the typelevel compiler puts a warning that the match is not fully exhaustive. The fix above makes it work but IMHO it shouldn't be necessary

cquiroz:
Did you try launching the application? The issue with scala.js is not about compiling but the produced js is invalid

jdnavarro:
I just run the application and everything seems normal.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalControllerEpics.scala
tpolecat: What is the error and which options have you enabled? There are the two that TLS adds:

```scala
"-Xstrict-patmat-analysis",      // more accurate reporting of failures of match exhaustivity
"-Xlint:strict-unsealed-patmat", // warn on inexhaustive matches against unsealed traits
```

Do you get this behavior if you turn the one or both off?

It's definitely a bug if it doesn't understand Java enums.

cquiroz: The error produced is

```
[error] It would fail on the following input: (x: edu.gemini.seqexec.server.gcal.BinaryOnOff forSome x not in (OFF, ON))
[error]       v match {
```

I'm not using those compiler options yet. I'll harmonize them when we move to 2.12

At the moment we have:

```scala
    val scalacOptions = Seq(
      "-unchecked",
      "-deprecation",
      "-encoding", "UTF-8", // Keep on the same line
      "-feature",
      "-language:existentials",
      "-language:higherKinds",
      "-language:implicitConversions",
      "-target:jvm-1.8",
      "-Xlint",
      "-Xlint:-stars-align",
      "-Xfatal-warnings",
      "-Ywarn-unused-import"
    )
```

PR #267, opened 2017-06-29
SEQNG-271: Update to scala 2.12

cquiroz:
This PR updates ocs3 to use scala 2.12 with especially constructed dependencies from ocs3. The changes are mostly to the build but there are also small updates to the codes as scala 2.12 is a bit more strict

The ocs version is set to 2017B.1.1.2 but it should work with 2017B.1.1.1 installations of spdb


jdnavarro:
It compiles fine in my 💻

cquiroz:
@jdnavarro  Thanks for checking

cquiroz:
Thanks to @swalker2m work we can now compile on 2.12 and talk to an odb running with 2.11.

Some notes:

* You probably need to nuke your ivy cache due to the way the versions are published on ocs. you can do that by `rm -Rf ~/.ivy2/cache/edu.gemini.ocs/`
* This new bundle isn't deployed yet so this will only work on local development but we should deploy this to production soon

If somebody could check this it would be great before merging

swalker2m:
Where does it actually pick up the new `.jar`? Do you need to bring over `spModel.io` as well since it uses that?

cquiroz:
They come from our public maven repo on github. I already published the new libraries, including `spModel.io` there

jdnavarro:
I could compile it again. I also removed my `~/.ivy2/cache...` directory.

cquiroz:
Great, it prabably was enough to remove only ivy's `edu.gemini.ocs` dir but it will do

PR #268, opened 2017-06-29
Remove seqexec terminal application

cquiroz:
The terminal webapp hasn't been really used and we have a direct terminal for telnet on `gem`. It is a nice experiment but at the moment, it seems this module only makes the build slower. 

This PR removes it but we can leave it if people think it is good to keep


jdnavarro:
Would it be worth git tagging the previous commit so that we can find the terminal code quickly in case we need it in the future?

cquiroz:
That's a good idea, if we want to reuse it in the future

PR #270, opened 2017-07-03
Fixed some issues with FITS keywords

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -388,51 +388,51 @@ class StandardHeader(
     }
 
     sendKeywords(id, inst, hs, List(
+      buildString(obsObject.orDefault, "OBJECT"),
       buildString(obsReader.getObsType, "OBSTYPE"),
       buildString(obsReader.getObsClass, "OBSCLASS"),
       buildString(obsReader.getGemPrgId, "GEMPRGID"),
       buildString(obsReader.getObsId, "obsid"),
       buildString(obsReader.getDataLabel, "DATALAB"),
+      buildString(stateReader.getObserverName, "OBSERVER"),
```
cquiroz: What's `stateReader`?

jluhrs: It is a case class that has all the values that can be changed from the client, like the operator's name, or the observer's name.

cquiroz: Nice, would this solve the issue that the names are not being applied to the FITS keywords?

PR #272, opened 2017-07-03
SEQNG-280: Slow rendering of updates

cquiroz:
In the last Ui refactoring the way updates were rendered was simplified and the code made simpler to follow. However, as a result the Ui got much slower due to the amount of re renderings needed.

This PR fixes that with some added complexity. There are still opportunities for performance improvements but the most critical issues has been solved

This PR also fixes SEQNG-274 which was a side effect of the way the whole screen was re-rendered

As usual the code changes are fairly criptic


jdnavarro:
I tried loading the usual dummy Sequence F2 but this is what I got:

seqng-280

PR #273, opened 2017-07-04
SEQNG-281: Display on the queue the selected tab

cquiroz:
This PR indicates on the queue table what's the selected tab on the sequence area. It also has some UI improvements for queue tables with man entries

This change is motivated by difficulties on seeing the current state when there are many sequences on the queue.

Here's a screenshot with one sequence running and another being displayed

screenshot 2017-07-04 17 04 57


cquiroz:
sorry, that's a bad jar on the maven repo, let's see if we can fix it

PR #275, opened 2017-07-07
SEQNG-290: Errors loading partially executed sequences

cquiroz:
I've been tracking a bug where the metadata of a sequence isn't properly sent to the client when a sequence is partially executed. This is not a full solution and I'm presenting this only to start the conversation

This is an ad-hoc fix for one of the cases when this happens. When a partially run sequence is added to the queue the metadata doesn't include instrument nor name. There is another case not fixed but it should have the same root cause

I'm fairly sure there is a deeper cause for this but I can't really follow the code 

WDYT?

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -7,6 +7,7 @@ import edu.gemini.pot.sp.SPObservationID
 import edu.gemini.model.p1.immutable.Site
 import edu.gemini.seqexec.engine
 import edu.gemini.seqexec.engine.{Action, Engine, Event, EventSystem, Executed, Failed, Result, Sequence}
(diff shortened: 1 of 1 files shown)


jdnavarro:
Let me have a look. The metadata should be somehow set when loading the Sequence.

cquiroz:
Thanks for checking

I'm very unhappy with my solution, it is totally ad-hoc and I found another event showing the same problem. This PR was really to expose the issue but I wouldn't expect to merge it as it is now but rather found a general solution

cquiroz:
We probably should just inform the users that we can't load partially executed sequences but they can still test running "unexecuted" sequences

jdnavarro:
> I found another event showing the same problem.

Which one is it? I only know of `Load` coming with a `Sequence` in the event. We should identify the rest no matter what. You can file a ticket and take care of going systematically through all the events.

jdnavarro:
We could close this PR. The issue is deeper than it seems but whatever it is I don't think it's a problem of how the Sequence is *viewed*.

cquiroz:
Yeah, let's close this

PR #276, opened 2017-07-10
SEQNG-284: Let it crash

jluhrs:
Runtime errors are not captured anymore, letting the application crash.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
@@ -65,9 +65,8 @@ trait EpicsSystem[T] {
 }
 
 object EpicsCommand {
-  def safe[A](a: SeqAction[A]): SeqAction[A] = EitherT(a.run.attempt.map {
-    case -\/(t) => SeqexecException(t).left
-    case \/-(e) => e
+  def safe[A](a: SeqAction[A]): SeqAction[A] = EitherT(a.run.handle {
+    case e: Exception => SeqexecException(e).left
   })
 
   def setParameter[A,T](p: Option[CaParameter[T]], v: A, f: A => T): SeqAction[Unit] =
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/package.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/package.scala
@@ -31,7 +31,9 @@ package object server {
   }
 
   implicit class SeqActionOps[A](a: SeqAction[A]) {
-    def runSeqAction: TrySeq[A] = a.run.unsafePerformSyncAttempt.leftMap[SeqexecFailure](SeqexecException).join
+    def runSeqAction: TrySeq[A] = a.run.handle{
+      case e: Exception => SeqexecException(e).left
+    }.unsafePerformSync
   }
 
   implicit class MoreDisjunctionOps[A,B](ab: A \/ B) {
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/package.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/package.scala
@@ -1,7 +1,6 @@
 package edu.gemini.seqexec
 
 import edu.gemini.seqexec.engine.Event
-import edu.gemini.seqexec.server.SeqexecFailure.SeqexecException
 
 import scalaz._
 import Scalaz._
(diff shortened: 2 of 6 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
```diff
@@ -65,9 +65,8 @@ trait EpicsSystem[T] {
 }
 
 object EpicsCommand {
-  def safe[A](a: SeqAction[A]): SeqAction[A] = EitherT(a.run.attempt.map {
-    case -\/(t) => SeqexecException(t).left
-    case \/-(e) => e
+  def safe[A](a: SeqAction[A]): SeqAction[A] = EitherT(a.run.handle {
+    case e: Exception => SeqexecException(e).left
```
cquiroz: In the description you talk about Runtime Errors but this will catch all of them, is that correct?

jdnavarro:
Perhaps I'm not getting the full picture of how errors are going to be handled in general but if I'm not wrong, all actions, in practice, are executed with the `recover` method which is converting *all* exceptions to a `SeqexecFailure`. So how is not the same as what we had before?

The only difference in behavior I can notice is that there is no attempt to print a stacktrace and that if there is any `Exception` while creating an `Action`, not while executing the action itself which is normally where you would expect them,  the `Exception` will go unhandled.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/package.scala
```diff
@@ -31,7 +31,9 @@ package object server {
   }
 
   implicit class SeqActionOps[A](a: SeqAction[A]) {
-    def runSeqAction: TrySeq[A] = a.run.unsafePerformSyncAttempt.leftMap[SeqexecFailure](SeqexecException).join
+    def runSeqAction: TrySeq[A] = a.run.handle{
+      case e: Exception => SeqexecException(e).left
+    }.unsafePerformSync
```
cquiroz: My impression is that having the `unsafe...` call here is a bit problematic for callers which don't know that you do. I'd rather remove it or rename the method to `unsafeRunSeqAction`. WDYT?

cquiroz:
I have the same impression than @jdnavarro that we are catching all exceptions

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/package.scala
jdnavarro: I'm not sure where `runSeqAction` is being called. Perhaps we can remove it if we are not using it anywhere.

jdnavarro:
Perhaps it's easier to start with a concrete case of fatal error that can be easily tested manually. For example, what is the expected behavior of the seqexec if you kill the ODB at any moment?

cquiroz:
IMHO these scenarios should be subjected to automatic testing though I admit they maybe a bit hard to setup

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
jluhrs: Yes. Strictly speaking, Throwables of the subclass Exception will be caught and passed as a (bad) result, while Throwables of the subclass Error will cause the application to terminate. That group includes errors like OutOfMemory and StackOverflow.

jluhrs:
IMHO killing the ODB should not kill the Seqexec. It should inform of the connection problem, and operate normally once the ODB is back.

cquiroz:
I agree like for instruments, if the ODB disappears for a while the seqexec should recover when it returns

jluhrs:
`runSeqAction` is not used anywhere. I'll remove it

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
jdnavarro: I see, I thought `Exception` was at the top of the error hierarchy. I didn't know Java differentiates between `Error`s and `Exception`s. 👍

jdnavarro:
> IMHO killing the ODB should not kill the Seqexec. It should inform of the connection problem, and operate normally once the ODB is back.

Agreed, it should try to reconnect after a certain interval of time, similar to how @cquiroz does it in the UI with the websocket. Right now I think it just stops and prints failure in whatever Step it was, so we could capture the `Exception` the ODBProxy creates (or whatever way of reporting the ODB it's down) and handle the reconnection.

We could use that example as a canonical way of handling expected exceptions in the Seqexec.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
cquiroz: Right, I was thinking about it the other way around

PR #277, opened 2017-07-11
SEQNG-292: Hide the observation control buttons for F2

cquiroz:
We had the buttons earlier to check the UI but this would be confusing for the users. Let's hide them


jdnavarro:
Are these the buttons that go in the rows representing the Steps?

PR #280, opened 2017-07-12
SEQNG-196: Display running observation

cquiroz:
This PR enables that when the page is loaded and a sequence is running, it will be immediately selected. As usual for UI PRs the code tells little of what actually happens


jdnavarro:
It should be redirected to the instrument URL when landing on the root URL and the instrument is running, right? I'm not getting that behavior 😟

cquiroz:
hmm, yeah that should be the right behavior

cquiroz:
I added a few changes to rebase on master and the bug that danny found. This could be reviewed again

PR #282, opened 2017-07-17
Reordered keywords to match old Seqexec.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalKeywordReader.scala
```diff
@@ -32,16 +32,22 @@ object GcalKeywordsReaderImpl extends GcalKeywordReader {
...
+    def ir   = GcalEpics.instance.lampIr().map{
+      case BinaryOnOff.ON => "IRhigh"
+      case _              => "IRlow"
+    }
+    def qh   = GcalEpics.instance.lampQH().filter(_ == BinaryOnOff.ON) *> "QH".some
+    def thAr = GcalEpics.instance.lampThAr().filter(_ == BinaryOnOff.ON) *> "ThAr".some
+    def xe   = GcalEpics.instance.lampXe().filter(_ == BinaryOnOff.ON) *> "Xe".some
+
+    Stream(ar, xe, cuAr, thAr, qh, ir).flatten.headOption
   }
```
jluhrs: I'm not sure what I was thinking when I first implemented this. This new implementation matches the old Seqexec code.

jluhrs: The use of `def`s and a `Stream` is to make it stop on the first positive value.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -205,22 +205,25 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
...
+  private val gcalHeader: Header = new GcalHeader(
     systems.dhs,
     if (settings.gcalKeywords) GcalKeywordsReaderImpl else DummyGcalKeywordsReader)
 
   private def calcHeaders(config: Config, stepType: StepType): TrySeq[List[Header]] = stepType match {
-    case CelestialObject(inst) => calcInstHeader(config, inst).map(_ :: commonHeaders(config))
-    case FlatOrArc(inst)       => calcInstHeader(config, inst).map(f => f :: gcalHeaders :: commonHeaders(config)) // TODO: Add GCAL keywords here
-    case DarkOrBias(inst)      => calcInstHeader(config, inst).map(_ :: commonHeaders(config))
+    case CelestialObject(inst) => calcInstHeader(config, inst).map(List(commonHeaders(config), gwsHeaders, _))
+    case FlatOrArc(inst)       => calcInstHeader(config, inst).map(List(commonHeaders(config), gcalHeader, gwsHeaders, _))
```
jluhrs: All this so I could put GCAL keywords between the standard keywords and GWS keywords

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalKeywordReader.scala
cquiroz: maybe you can do `ar.orElse(xe).orElse(cuAr)...`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GwsHeader.scala
```diff
@@ -0,0 +1,56 @@
...
+import edu.gemini.seqexec.server.Header._
+import edu.gemini.seqexec.server.Header.Implicits._
+
+/**
+  * Created by jluhrs on 7/17/17.
+  */
+class GwsHeader(hs: DhsClient, gwsReader: GwsKeywordReader) extends Header {
+  override def sendBefore(id: ImageFileId, inst: String): SeqAction[Unit] = {
+    gwsReader.getHealth.flatMap{
+      case Some(0) => sendKeywords(id, inst, hs, List(
```
cquiroz: What's the 0 there? Maybe we could use a meaningful constant?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GcalKeywordReader.scala
jluhrs: Is it not the same ?

cquiroz: It is the same but it seems lees work. Is it a scalaz-stream or o regular Stream?

swalker2m: Also, the `flatten` version, if I'm reading it correctly, makes use of an implicit conversion of `Option` to `Iterable` which is discouraged and would be flagged as an error in the `gem` build.

jluhrs: Ok, Shane's argument convinced me. But something worries me. Is that conversion the same that allows us to `flatMap` a `List` with a function that produce `Option`, and have as a result a `List` of only the inner values of the `Some`s, with the `None`s removed? Because if that is the case, we need to change the code in several parts.

swalker2m: Correct, that's what's happening.  You can still do it though if you map the `Option` results as in `lst.flatMap { a => f(a).toList } `.

PR #284, opened 2017-07-20
SEQNG-302: Exception when loading a sequence without steps

cquiroz:
After #283 we can now load empty sequences, however running them produces an exception and it puts the engine on an infinite loop

This PR fixes both problems and also removes the buttons on the UI to run empty sequences

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -136,7 +136,6 @@ package object engine {
   def setCloudCover(cc: CloudCover): HandleP[Unit] =
     modify(st => Engine.State(st.conditions.copy(cc = cc), st.operator, st.sequences))
 
-
   /**
     * Load a Sequence
     */
@@ -194,13 +193,13 @@ package object engine {
           case SequenceState.Running =>
             seq.next match {
               // Empty state
-              case None =>
+              case None                           =>
                 send(finished(id))
               // Final State
               case Some(qs: Sequence.State.Final) =>
                 putS(id)(qs) *> send(finished(id))
               // Execution completed. Check breakpoint here
-              case Some(qs) =>
+              case Some(qs)                       =>
                 putS(id)(qs) *> (if(qs.getCurrentBreakpoint) switch(id)(SequenceState.Idle)
                                  else send(executing(id)))
             }
@@ -214,7 +213,6 @@ package object engine {
     * It also updates the `State` as needed.
     */
   private def execute(id: Sequence.Id): HandleP[Unit] = {
-
     // Send the expected event when the `Action` is executed
     // It doesn't catch run time exceptions. If desired, the Action as to do it itself.
     def act(t: (Action, Int)): Process[Task, Event] = t match {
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -229,18 +227,20 @@ package object engine {
     getS(id).flatMap(
       _.map { seq =>
         seq match {
-          case Sequence.State.Final(_,_) => unit
-          case _                         => {
+          case Sequence.State.Final(_, _) =>
+            // The sequence is marked as completed here
+            putS(id)(seq) *> send(finished(id))
```
cquiroz: I'm a bit unsure that this is the correct solution. But otherwise the engine gets into an infinite loop when calling run on an empty sequence

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -210,6 +211,8 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
       // TODO: Calculate the whole status here and remove `Engine.Step.status`
       // This will be easier once the exact status labels in the UI are fixed.
       seq.steps.map(viewStep) match {
+        // The sequence could be empty
+        case Nil => Nil
         // Find first Pending Step when no Step is Running and mark it as Running
         case steps if (st === SequenceState.Running || st === SequenceState.Stopping) && steps.all(_.status =/= StepState.Running) =>
           val (xs, (y :: ys)) = splitWhere(steps)(_.status === StepState.Pending)
```
cquiroz: If `steps` is empty this throws a `MatchError` killing the engine completely

swalker2m: Seems like the compiler should have caught this.  Do you guys use `-Xfatal-warnings`?

cquiroz: We do but trying to split an empty List isn't flagged as a warning. Perhaps with the more advanced compiler flags on gem

swalker2m: This is what we're using in `gem` and I think it is helping to keep the code cleaner:

https://github.com/gemini-hlsw/gem/blob/master/build.sbt#L83-L134

cquiroz: Yes, I have a partial port of that though some of the options create troubles with the scala.js part requiring some partition of the modules

jdnavarro:
I'm bit late to the discussion but from what I understood from the initial requirements is that a Sequence must have at least 1 Step and I'm pretty sure initially it was enforced by the type system. When does it make sense for a Sequence to have no Steps?

In any case I'm glad you took care of the issue because it would have been difficult to fix it for me, so thanks for this PR.

cquiroz:
I think that requirement is true in the sense that you need to have steps to run a sequence but the OT can perfectly have sequences without steps. I don't think it was enforced by the type system

PR #285, opened 2017-07-24
Engine experiments for potential improvements

jdnavarro:
This is still half-baked but I left it as it is today because I had already spent quite some time and I think it's good enough to concretize the ideas I put forward some time ago.

I started by trying to get rid of `Process`s internally and use only `Task`s but found it too difficult to separate that aspect of the other improvements I initially proposed, so almost everything is in there.

The previous `scalaz-stream` `Process` is now `fs2` `Stream`. I started using the current `fs2` `Task`s but it's going be deprecated in favor of the `cats-effect` `Effect` type class which is what I ended up using instead. `scalaz` `Task` are quite different from what is being used in `fs2`/`cats-effect`. Asynchronicity is now explicit, you have to explicitly write when you want to fork, and although I agree this is a better approach in general, our asynchronous code would need a major rewrite even if we only focus on the `fs2` migration and ignore the rest of the improvements.

**This PR is not intended for merging**, only to identify what could be extracted into some kind of separate component that could be simpler instead of keep writing *ad-hoc* fixes to the current engine which is making it increasingly more difficult to understand (at least for me).

### DIFF
--- a/modules/engine-prototype/build.sbt
+++ b/modules/engine-prototype/build.sbt
@@ -1,0 +1,24 @@
+import Dependencies._
+
+lazy val root = (project in file(".")).
+  settings(
+    inThisBuild(List(
+      organization := "edu.gemini",
+      scalaVersion := "2.12.1",
+      scalacOptions ++= Seq(
+        "-feature",
+        "-language:higherKinds",
+        "-language:existentials"
+      ),
+      version      := "0.1.0-SNAPSHOT"
+    )),
+    name := "engine",
+    libraryDependencies ++= Seq(
+      cats,
+      catsEffect,
+      fs2,
+      monocle,
+      monocleMacro,
+      scalaTest % Test
+    )
+  )
--- a/modules/engine-prototype/src/main/scala/engine/Execution.scala
+++ b/modules/engine-prototype/src/main/scala/engine/Execution.scala
@@ -1,0 +1,17 @@
+package engine
+
+import cats.effect.Effect
+
+object Execution {
+
+  def configureTCS[F[_]](implicit F: Effect[F]): F[Either[Failure, OK]]  = ???
+
+  def configureInst[F[_]](name: String)(implicit F: Effect[F]): F[Either[Failure, OK]]  = ???
+
+  def observe[F[_]](implicit F: Effect[F]): F[Option[String]] = ???
+
+  type Failure = Unit
+
+  type OK = Unit
+
+}
--- a/modules/engine-prototype/src/main/scala/engine/Resource.scala
+++ b/modules/engine-prototype/src/main/scala/engine/Resource.scala
@@ -1,0 +1,32 @@
+package engine
+
+/**
+  * A Seqexec resource represents any system that can be only used by one single agent.
+  *
+  */
+sealed trait Resource
+object Resource {
+
+  case object P1 extends Resource
+  case object OI extends Resource
+  // Mount and science fold cannot be controlled independently. Maybe in the future.
+  // For now, I replaced them with TCS
+  //  case object Mount extends Resource
+  //  case object ScienceFold extends Resource
+  case object TCS extends Resource
+  case object Gcal extends Resource
+  case object Gems extends Resource
+  case object Altair extends Resource
+
+  trait Instrument extends Resource
+  object Instrument {
+    case object GMOS extends Instrument
+    case object F2 extends Instrument
+    case object GSAOI extends Instrument
+    case object GPI extends Instrument
+    case object NIRI extends Instrument
+    case object NIFS extends Instrument
+    case object GNIRS extends Instrument
+  }
+
+}
--- a/modules/engine-prototype/src/main/scala/engine/Sequence.scala
+++ b/modules/engine-prototype/src/main/scala/engine/Sequence.scala
@@ -1,0 +1,91 @@
+package engine
+
+import scala.concurrent.ExecutionContext
+
+import cats.{FlatMap, Functor}
+import cats.data.NonEmptyList
+import cats.effect.Effect
+import cats.effect.implicits._
+import cats.implicits._
+
+import fs2.Stream
+import fs2.async
+import fs2.async.mutable.Signal
+
+import monocle.{Lens, Prism}
+import monocle.macros.GenLens
+
+sealed trait Sequence
+object Sequence {
+
+  sealed trait F2 extends Sequence
+  object F2 {
+
+    final case class Ready(
+      done: List[Step.F2.Done],
+      pending: List[Step.F2.Pending],
+      current: Step.F2.Current
+    ) extends F2
+
+    final case class Done(done: NonEmptyList[Step.F2.Done]) extends F2
+
+  }
+
+  final case class State(sequence: Sequence, status: Status)
+
+  object State {
+
+    val sequence: Lens[State, Sequence] = GenLens[State](_.sequence)
+
+    val status: Lens[State, Status] = GenLens[State](_.status)
+
+    val current: Prism[State, Step.F2.Current] = ???
+
+    val standard: Prism[State, Step.F2.Current.Pending.Standard] = ???
+
+    val ready: Prism[State, Step.F2.Current.Pending] = ???
+
+    // Thread safe mutable but careful need to be careful not to block for very long.
+    case class Mutable[F[_]](signal: Signal[F, Sequence.State]) {
+
+      def getStatus(implicit F: Functor[F]): F[Status] = signal.get.map(_.status)
+
+      def setStatus(s: Status)(implicit F: Functor[F]): F[Sequence.State] =
+        signal.modify(State.status.set(s)).map(_.now)
+
+
+      def getSequence(implicit F: Functor[F]): F[Sequence] = signal.get.map(_.sequence)
+
+      def setSequence(seq: Sequence)(implicit F: Functor[F]): F[Sequence.State] =
+        signal.modify(State.sequence.set(seq)).map(_.now)
+
+
+      val getState: F[Sequence.State] = signal.get
+
+      def setState(st: Sequence.State): F[Unit] = signal.set(st)
+
+
+      // is it worth making Sequence.State.Monad a righteous Monad?
+      def withState(f: Sequence.State => F[Sequence.State])(implicit F: FlatMap[F]): F[Sequence.State] =
+        signal.get.flatMap(f)
+
+
+      def modify(f: Sequence.State => Sequence.State)(implicit F: Functor[F]): F[Sequence.State] =
+        signal.modify(f).map(_.now)
+
+    }
+
+    object Mutable {
+
+      def stream[F[_]](st0: Sequence.State)(handle: (Mutable[F]) => F[Unit])(implicit F: Effect[F], ec: ExecutionContext): Stream[F, Sequence.State] =
+        Stream.force(
+          async.signalOf[F, Sequence.State](st0).flatMap(sig =>
+            async.fork(handle(Mutable(sig))) *> sig.discrete.pure[F]
+          )
+        )
+
+    }
+
+  }
+
+}
--- a/modules/engine-prototype/src/main/scala/engine/Status.scala
+++ b/modules/engine-prototype/src/main/scala/engine/Status.scala
@@ -1,0 +1,19 @@
+package engine
+
+sealed trait Status
+object Status {
+  case object Running extends Status
+  case object Waiting extends Status
+  case object Finished extends Status
+  case object Failed extends Status
+
+
+//   whenRunning(sig)(f: ) = sig.get.flatMap(st =>
+//     sig.get.flatMap(st =>
+//       st.status match {
+//         case Status.Running => fa
+//         case _ => F.pure(Unit)
+//       }
+// )
+
+}
--- a/modules/engine-prototype/src/main/scala/engine/Step.scala
+++ b/modules/engine-prototype/src/main/scala/engine/Step.scala
@@ -1,0 +1,251 @@
+package engine
+
+import scala.concurrent.ExecutionContext
+
+import cats.implicits._
+import cats.effect.Effect
+
+sealed trait Step
+object Step {
+
+  // Whatever is shared among all Steps. Notice it doesn't // extend Step
+  // because it's meant to be used only internally
+  case class Core(
+    id: String,
+    config: Core.Configuration,
+    breakpoint: Boolean
+  )
+
+  object Core {
+    case class Configuration()
+  }
+
+  sealed trait F2 extends Step
+  object F2 {
+
+    case class Done(core: Core, fileId: String) extends F2
+
+    sealed trait Pending extends F2
+    object Pending {
+      case class Standard(core: Core) extends Pending
+      case class Flat(core: Core) extends Pending
+      case class Dark(core: Core) extends Pending
+    }
+
+    sealed trait Ongoing extends F2
+    object Ongoing {
+      final case class Standard(core: Core, tcsConfigured: Boolean, instConfigured: Boolean) extends Ongoing
+    }
+
+    sealed trait Failed extends F2
+    object Failed {
+      final case class Standard(core: Core, tcsFailed: Boolean, instFailed: Boolean, observationFailed: Boolean) extends F2
+      // TODO: final case class Standard(core: Core, tcsFailure: Option[Failure], instFailure: Option[Failure]) extends F2
+    }
+
+    sealed trait Current extends F2 {
+
+      // Accessor to the core Step
+      val core: Core = ???
+
+    }
+
+    object Current {
+      // Only a Pending Step in Current can be executed
+      sealed trait Pending extends Current {
+
+        def execute[F[_]](m: Sequence.State.Mutable[F])(implicit F: Effect[F], ec: ExecutionContext): F[Unit]
+
+      }
+
+      object Pending {
+
+        case class Standard(step: F2.Pending.Standard) extends Pending {
+
+          def execute[F[_]](m: Sequence.State.Mutable[F])(implicit F: Effect[F], ec: ExecutionContext): F[Unit] = {
+
+            def configureTCS: F[Boolean] =
+              Execution.configureTCS.flatMap {
+                case Right(_) =>
+                  m.modify(
+                    Sequence.State.current.modify {
+                      case Pending.Standard(s) => Ongoing.Standard(
+                        F2.Ongoing.Standard(s.core, tcsConfigured = true, instConfigured = false)
+                      )
+                      case Ongoing.Standard(s) => Ongoing.Standard(
+                        F2.Ongoing.Standard(s.core, tcsConfigured = true, instConfigured = true)
+                      )
+                      case Failed.Standard(s) => Failed.Standard(
+                        F2.Failed.Standard(
+                          s.core,
+                          tcsFailed = false,
+                          instFailed = true,
+                          observationFailed = true
+                        )
+                      ) // Instrument already failed
+                    }
+                  ) *> true.pure[F]
+
+                case Left(_) =>
+                  m.modify(
+                    Sequence.State.current.modify {
+                      case Failed.Standard(s) => Failed.Standard(
+                        F2.Failed.Standard(
+                          s.core,
+                          tcsFailed = true,
+                          instFailed = true,
+                          observationFailed = false
+                        )
+                      )
+                      // Instrument didn't fail yet
+                      case s => Failed.Standard(
+                        F2.Failed.Standard(
+                          s.core,
+                          tcsFailed = true,
+                          instFailed = false,
+                          observationFailed = false
+                        )
+                      )
+                    }
+                  ) *> false.pure[F]
+
+              }
+
+            def configureInst: F[Boolean] =
+              Execution.configureInst("F2").flatMap {
+                case Right(_) =>
+                  m.modify(
+                    Sequence.State.current.modify {
+                      case Pending.Standard(s) => Ongoing.Standard(
+                        F2.Ongoing.Standard(s.core, tcsConfigured = false, instConfigured = true)
+                      )
+                      case Ongoing.Standard(s) => Ongoing.Standard(
+                        F2.Ongoing.Standard(s.core, tcsConfigured = true, instConfigured = true)
+                      )
+                      case Failed.Standard(s) => Failed.Standard(
+                        F2.Failed.Standard(
+                          s.core,
+                          tcsFailed = true,
+                          instFailed = false,
+                          observationFailed = false
+                        )
+                      ) // Instrument already failed
+                    }
+                  ) *> true.pure[F]
+
+                case Left(_) =>
+                  m.modify(
+                    Sequence.State.current.modify {
+                      case Failed.Standard(s) => Failed.Standard(
+                        F2.Failed.Standard(
+                          s.core,
+                          tcsFailed = true,
+                          instFailed = true,
+                          observationFailed = false
+                        )
+                      )
+                      // TCS didn't fail yet
+                      case s => Failed.Standard(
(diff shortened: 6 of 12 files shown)


modules/engine-prototype/build.sbt
```diff
@@ -0,0 +1,24 @@
...
+    name := "engine",
+    libraryDependencies ++= Seq(
+      cats,
+      catsEffect,
+      fs2,
+      monocle,
+      monocleMacro,
+      scalaTest % Test
+    )
+  )
```
jdnavarro: This was a personal exercise in order to learn `sbt`. I used the default template to create a project. I'm not proposing anything regarding the build.

modules/engine-prototype/src/main/scala/engine/Execution.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  def configureInst[F[_]](name: String)(implicit F: Effect[F]): F[Either[Failure, OK]]  = ???
+
+  def observe[F[_]](implicit F: Effect[F]): F[Option[String]] = ???
+
+  type Failure = Unit
+
+  type OK = Unit
+
+}
```
jdnavarro: These functions are at the boundary with the underlying EPICS subsystems . Equivalent to what we call `Action`s in the current seqexec. They are supposed to be part of the *mini-DSL* to define Step executions. Obviously, they'll be shared among the different Steps so that's why they are defined in a separate object.

modules/engine-prototype/src/main/scala/engine/Resource.scala
```diff
@@ -0,0 +1,32 @@
...
+    case object GMOS extends Instrument
+    case object F2 extends Instrument
+    case object GSAOI extends Instrument
+    case object GPI extends Instrument
+    case object NIRI extends Instrument
+    case object NIFS extends Instrument
+    case object GNIRS extends Instrument
+  }
+
+}
```
jdnavarro: Directly ported from the current seqexec.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+import fs2.async
+import fs2.async.mutable.Signal
+
+import monocle.{Lens, Prism}
+import monocle.macros.GenLens
+
+sealed trait Sequence
+object Sequence {
+
+  sealed trait F2 extends Sequence
```
jdnavarro: A Sequence is an ADT supporting different instruments: `Sequence = Sequence.F2 | Sequence.GMOS | ...`. This way we ensure we can create Sequences that have Steps that are supported by the Sequence. For example, you shouldn't be able to create a Sequence with configurations for GMOS and F2.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+  object F2 {
+
+    final case class Ready(
+      done: List[Step.F2.Done],
+      pending: List[Step.F2.Pending],
+      current: Step.F2.Current
+    ) extends F2
+
+    final case class Done(done: NonEmptyList[Step.F2.Done]) extends F2
```
jdnavarro: We the `Done` case class because it doesn't make sense to have a Step under execution when all Steps are done.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+sealed trait Sequence
+object Sequence {
+
+  sealed trait F2 extends Sequence
+  object F2 {
+
+    final case class Ready(
+      done: List[Step.F2.Done],
+      pending: List[Step.F2.Pending],
+      current: Step.F2.Current
```
jdnavarro: Notice the Steps types are different for each position. `Current` will contain the Step under execution with information pertaining to the actions being taken. `Done` and `Pending` are almost the same but still distinct as so that it's not possible to mess things up.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+    ) extends F2
+
+    final case class Done(done: NonEmptyList[Step.F2.Done]) extends F2
+
+  }
+
+  final case class State(sequence: Sequence, status: Status)
```
jdnavarro: This is the top level state in the prototype. In reality it would have to be a Map of Sequences as it is now in the seqexec but kept things simple.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+
+    val sequence: Lens[State, Sequence] = GenLens[State](_.sequence)
+
+    val status: Lens[State, Status] = GenLens[State](_.status)
+
+    val current: Prism[State, Step.F2.Current] = ???
+
+    val standard: Prism[State, Step.F2.Current.Pending.Standard] = ???
+
+    val ready: Prism[State, Step.F2.Current.Pending] = ???
```
jdnavarro: Too lazy to bother implementing the `Prism` but it should be relatively straightforward with the monocle macros.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+
+    // Thread safe mutable but careful need to be careful not to block for very long.
+    case class Mutable[F[_]](signal: Signal[F, Sequence.State]) {
```
jdnavarro: The underlying concurrency is `fs2` `Signal` but in the rest of the code we shouldn't mess directly with `Signal`s,  instead we create the wrappers that match more closely the terminology we use in the Seqexec, simpler more specialized signatures and also by having all these State manipulation in an object with conspicuous name like `Mutable` we make sure you should watch out what you do when using this functions.

In reality the danger refers that we may block for too long (something that also have to be careful in our current seqexec), there are no fundamental dangerous operations with race conditions or deadlocks.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+        signal.modify(State.sequence.set(seq)).map(_.now)
+
+
+      val getState: F[Sequence.State] = signal.get
+
+      def setState(st: Sequence.State): F[Unit] = signal.set(st)
+
+
+      // is it worth making Sequence.State.Monad a righteous Monad?
+      def withState(f: Sequence.State => F[Sequence.State])(implicit F: FlatMap[F]): F[Sequence.State] =
```
jdnavarro: This is the most *dangerous* operation because if the function we pass as argument takes too long we may block for too long. I think this is only necessary when switching the Status from Waiting to Running so that nothing can be changed in-between, but I can't think of other situations where we needed this level of atomicity.

In the rest of functions since it's not possible to pass an arbitrary effectful function I find it hard to do anything that would block for too long.

We have to deal with these situations in the current Seqexec, the difference here is that we have more awareness of the potential danger.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+        signal.get.flatMap(f)
+
+
+      def modify(f: Sequence.State => Sequence.State)(implicit F: Functor[F]): F[Sequence.State] =
+        signal.modify(f).map(_.now)
+
+    }
```
jdnavarro: I still wanted to add `withRunningStatus`, `withWaitingStatus` and other similar functions, to make it easier to write execution Steps without having to write code that inspects/manipulates the State directly where the Step execution is defined.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
```diff
@@ -0,0 +1,91 @@
...
+
+    object Mutable {
+
+      def stream[F[_]](st0: Sequence.State)(handle: (Mutable[F]) => F[Unit])(implicit F: Effect[F], ec: ExecutionContext): Stream[F, Sequence.State] =
+        Stream.force(
+          async.signalOf[F, Sequence.State](st0).flatMap(sig =>
+            async.fork(handle(Mutable(sig))) *> sig.discrete.pure[F]
+          )
+        )
```
jdnavarro: This is the way to initialize and run a *Mutable* State.

It takes the initial State (the one that should come when first loaded), the top level action that will do something with that *Mutable* State and returns the `Stream` that is meant to be consumed by the client after serializing and whatever is needed.

This is only meant to be run once at the start of the application.

modules/engine-prototype/src/main/scala/engine/Status.scala
```diff
@@ -0,0 +1,19 @@
...
+
+
+//   whenRunning(sig)(f: ) = sig.get.flatMap(st =>
+//     sig.get.flatMap(st =>
+//       st.status match {
+//         case Status.Running => fa
+//         case _ => F.pure(Unit)
+//       }
+// )
+
```
jdnavarro: I should have restored this one and put it in `Mutable`.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+sealed trait Step
+object Step {
+
+  // Whatever is shared among all Steps. Notice it doesn't // extend Step
+  // because it's meant to be used only internally
+  case class Core(
+    id: String,
+    config: Core.Configuration,
+    breakpoint: Boolean
+  )
```
jdnavarro: This should have been private to the `Step` object but I had some issues with scoping that I don't remember exactly. This contains all the shared information of all the Steps. All Steps have Core Step as a composite value.

Now that I'm writing this I realize `breakpoint` shouldn't go here, it only makes sense in the Pending Steps.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+
+  object Core {
+    case class Configuration()
+  }
+
+  sealed trait F2 extends Step
+  object F2 {
+
+    case class Done(core: Core, fileId: String) extends F2
```
jdnavarro: Notice a Done F2 Step must have a `fileId`, we are deliberately omitting `Option`.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+
+    sealed trait Pending extends F2
+    object Pending {
+      case class Standard(core: Core) extends Pending
+      case class Flat(core: Core) extends Pending
+      case class Dark(core: Core) extends Pending
+    }
```
jdnavarro: These case classes represent the different types of Steps with different execution definitions. They are only here as an example, but they should be the same for Pending, Ongoing, etc...

BTW, it's not still clear how to better represent this deeply nested ADT. Perhaps `Step.F2.Flat.Pending` instead of `Step.F2.Pending.Flat`. The reason for the first option is that in a Sequence the pending Steps can be of any execution type like `Dark`, `Flat` but need to be always Pending.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+
+    sealed trait Ongoing extends F2
+    object Ongoing {
+      final case class Standard(core: Core, tcsConfigured: Boolean, instConfigured: Boolean) extends Ongoing
```
jdnavarro: These are just dummy parameters to make a point, but we could put anything there to represent a Step under execution. For example, an estimated timestamp of completion for progress bars that could be updated if the estimation taken from the EPICS channels changes.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    sealed trait Failed extends F2
+    object Failed {
+      final case class Standard(core: Core, tcsFailed: Boolean, instFailed: Boolean, observationFailed: Boolean) extends F2
+      // TODO: final case class Standard(core: Core, tcsFailure: Option[Failure], instFailure: Option[Failure]) extends F2
+    }
```
jdnavarro: Here I was just exploring the possibility of having a distinct Failed Step but I'm not sure if we need that level of detail when something fails. The biggest advantage, as opposed to just a specific error, is that we still keep the information of whatever was right in that Step until it failed.

Also with failed Steps we can create specific renderers in the UI for a particular failed Step, but again, I don't know if this is overkill and with the error message is just enough.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+
+    sealed trait Current extends F2 {
+
+      // Accessor to the core Step
+      val core: Core = ???
```
jdnavarro: Lots of accessors like this would have to be written in order to use such a deeply nested ADTs, so I'm still unsure if there is a better way to encode the ADTs.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+
+    }
+
+    object Current {
+      // Only a Pending Step in Current can be executed
+      sealed trait Pending extends Current {
+
+        def execute[F[_]](m: Sequence.State.Mutable[F])(implicit F: Effect[F], ec: ExecutionContext): F[Unit]
```
jdnavarro: Only the current Step when in Pending can be executed. This ensures we can only execute when in the right position.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+                          s.core,
+                          tcsFailed = true,
+                          instFailed = false,
+                          observationFailed = false
+                        )
+                      )
+                    }
+                  ) *> false.pure[F]
+
+              }
```
jdnavarro: This is very verbose because I was too lazy to factor out the state modifications/observations into *Mutable* (I think I did this before that) and didn't create the lens for changing the parameters. But this is essentially what definining an `Action` should look like now.

The returned boolean is used to know if it failed or not to facilitate the continuation, but in reality a result with more info could be used instead.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+                        F2.Failed.Standard(
+                          s.core,
+                          tcsFailed = false,
+                          instFailed = true,
+                          observationFailed = false
+                        )
+                      )
+                    }
+                  ) *> false.pure[F]
+              }
```
jdnavarro: This is duplicated from the other action. Again, didn't have time to do something about it, but most of the code could be factored out with lens and `Mutable` functions.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+
+              // XXX: Check running status
+              both(configureTCS, configureInst).flatMap {
+              case (true, true) =>
```
jdnavarro: Execute the first 2 actions in parallel and after they are done decide what to do. In this case both where successful.

I would still have to check for the Status before going ahead but with a wrapper function like `withRunningStatus` it should be straightforward.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+                  // XXX: Check running status
+                  Execution.observe.flatMap {
+                    case Some(fileId) => execute(m) // XXX: Shift current focus
```
jdnavarro: This is incomplete, but after the result it should update the State, shift the current Step, put the done Step in the Done Steps, etc... and at the end make the recursive call to continue the execution.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+                            F2.Failed.Standard(
+                              s.core,
+                              tcsFailed = false,
+                              instFailed = false,
+                              observationFailed = true
+                            )
+                          )
+                        )
+                      ).void
+                  }
```
jdnavarro: In the case of failure the state is updated with the failed state and no recursive call is made so it stops. Again changing the Status is missing.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+      sealed trait Failed extends Current
+      object Failed {
+        case class Standard(step: F2.Failed.Standard) extends Failed
+      }
+    }
+
+  }
+
+  sealed trait GMOS extends Step
+  case object GMOS
```
jdnavarro: This is just a placeholder to indicate there are other Step types for other instruments.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+//         )
+//       )
+//       case Right(d) => m.setSequence(
+//         F2((d :: done), remainder, None)
+//       ).flatMap(_.sequence.execute(m))
+//     }
+//   }
+//   case Some(_) => F.pure(Unit) // Event: Tried execute Sequence with an already ongoing step.
+// }
+// }
```
jdnavarro: This is an early attempt with Current Steps where I'm doing some State shifting. I didn't delete it because I was going to recycle the shifting part in the new code.

modules/engine-prototype/src/main/scala/engine/Step.scala
```diff
@@ -0,0 +1,251 @@
...
+//      r2 <- t2
+//    } yield (r1, r2)).flatMap {
+//     case (Right(_), Right(_)) =>
+//         Execution.observe.map {
+//           case Right(_) => ??? // Right(Done())
+//           case Left(_) => ??? // Left(Failed())
+//         }
+//     case _ => ??? // F.pure(Left(Failed()))
+//   }
+// }
```
jdnavarro: This is an experiment of concurrency in `fs2`. Notice now how the forking needs to be made explicit, in this case with `async.start`.

Something similar would have to be done for the unimplemented `both`, so that's why I left it commented out.

modules/engine-prototype/src/main/scala/engine/package.scala
```diff
@@ -0,0 +1,46 @@
...
+import cats.effect.Effect
+
+import fs2.Stream
+import fs2.async
+import fs2.async.mutable.Queue
+
+package object engine {
+
+  def run[F[_]](in: Queue[F, Event])(st0: Sequence.State)(implicit F: Effect[F], ec: ExecutionContext): Stream[F, Sequence.State] =
+    Sequence.State.Mutable.stream[F](st0: Sequence.State)(m => forever(reader(in)(m)))
```
jdnavarro: This is the very top level entry point to the engine. You pass a initial State a queue from where to read input user events and you obtain a Stream of States. This should be the only function consumer code should depend on.

modules/engine-prototype/src/main/scala/engine/package.scala
```diff
@@ -0,0 +1,46 @@
...
+    in.dequeue1.flatMap {
+      case Event.Start => m.withState(st0 =>
+        st0.status match {
+          case Status.Waiting => {
+            val st1 = Sequence.State.status.set(Status.Running)(st0)
+            st1 match {
+              // Execute only when the State is in ready position. If it's not
+              // it's either already executing or it doesn't make sense to execute,
+              // like when it's done.
+              case Sequence.State.ready(p) => async.fork(p.execute(m)) *> F.pure(st1)
```
jdnavarro: Before returning the *reaction* to the start event we launch another async `Task` (`Effect` or whatever) that takes care of executing and is not responsible of handling user input. The current threads goes back to listening incoming input events.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
jluhrs: I agree in general with the idea, but I think building the ADT like this will be long and tedious. Maybe we could declare `Sequence` as
```
sealed trait Sequence[I<:Resource.Instrument]
```
and then
```
object Sequence {
  type F2 = Sequence[Resource.Instrument.F2]
}
```

jluhrs:
I think this design tries to solve to many things at the same time. The current design of Seqexec separates the problem in two main parts, the execution engine and the seqexec engine. The first one deals with managing the execution of sequences, while the second one deals with building the sequences and  translating results for the clients. That's why I have tried to keep the execution engine agnostic respect to the details of what is executed. I think this proposal goes in the opposite direction.

modules/engine-prototype/src/main/scala/engine/Execution.scala
jdnavarro: They are scary at first, but not that bad after you get used to read them. Normal `IO` would be the primary instance of those type classes, but you are right, you could run this code with a fake `IO` type that captures all side effects for testing without doing actually any real IO.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
jdnavarro: I agree the way ADTs are written in the code is cumbersome but are subtypes the only option to make this kind of ADTs easier to write. Aren't there more options?

Writing this kind of ADTs in Haskell is trivial so I'm pretty sure there must be some other reasonable ways to encode nested ADTs easily in Scala. I'm missing reading more examples of encoding Scala ADTs in the wild, what we have here is not that special.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
jdnavarro: If you generalize the types, this signature matches the `flatMap` method of a Monad, so you could use this function to define a Monad instance if the laws hold, which I think they do.

In this case having the name `withState` instead of `flatMap` and the more specialized types may be more *user friendly* for writing consumer code.

Also, if this function is not used carefully you could block longer than necessary, so a different name than `flatMap` makes it's more noticeable.

But I don't know, I don't have a strong opinion for either case, I'd be fine with `flatMap`.

modules/engine-prototype/src/main/scala/engine/Sequence.scala
jdnavarro: I guess so, but to be honest I just put the `ec` wherever the typechecker complained. Everybody seems to be using the default one (global something is called, right) that seems to make the `async.fork` actually asynchronous in almost all possible scenarios.

jdnavarro:
> I'd like to get at least some benefits out of this exercise. Could we consider porting the current code to fs2/cats-effect regardless? The use of ADTs is certainly welcome as well.

We could port it all to `fs2` as it is but doesn't solve the problem it triggered it all. And the opposite is true, you could port this prototype to `scalaz-stream`. Even `Signal`s is available `scalaz-streams`.

The thing is that this wasn't actually an exercise to learn about fs2/cats-effect, that was just a side-effect (pun intended) of the primary goal which was to show what could be improved in the engine so that it would be, at least for me, easier to understand and maintain, because right now I'm struggling to contribute meaningfully to the current seqexec since it's becoming increasingly complex for me to follow.

I don't know what could be extracted from this exercise because it's all related, the shared state doesn't make sense without splitting the top level thread, the execution continuation and getting rid of the transformers is not possible without sharing state with something like Signal, and the ADTs expect the focus to continue the execution.

This was an attempt at showing what would be simpler for me. But it turns out that what's simpler for me it's more complicated for @jluhrs . And it's alright, it's understandable, this is all highly subjective and ultimately @jluhrs is the main responsible of the project.

Let's just ditch it all and move on. I think we can find with @arturog8m some other project where I could be more productive. There are plenty of things to do at Gemini.

PR #286, opened 2017-07-25
SENQG-285: Match scala compiler params of the gem project

cquiroz:
This PR updates the scalac params to match the `gem` project. This revealed several cases where the code was either not in the correct style (e.g. unused params) or suspicious (many cases of unused non-Unit return values)

I had to separate the scala.js facades into its own project to give them a special set of scalac params that would make it fail otherwise

It is probably good idea to review the changes carefully. In some cases they highlight code that may not be fully correct

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -278,12 +278,12 @@ package object engine {
     /**
       * Log info lifted into Handle.
       */
-    def info(msg: String): HandleP[Unit] = pure({(logger.info(msg), None); ()})
+    def info(msg: String): HandleP[Unit] = pure((logger.info(msg), None)).void
 
     /**
       * Log warning lifted into Handle.
       */
-    def warning(msg: String): HandleP[Unit] = pure({(logger.warning(msg), None); ()})
+    def warning(msg: String): HandleP[Unit] = pure((logger.warning(msg), None)).void
   }
 
   /**
@@ -386,8 +386,8 @@ package object engine {
     modify(st => Engine.State(st.conditions, st.operator, st.sequences.updated(id, s)))
 
   // For debugging
-  def printSequenceState(id: Sequence.Id): HandleP[Option[Unit]] =
-    getSs(id)((qs: Sequence.State) => {Task.now(println(qs)).liftM[HandleStateT]; ()})
+  def printSequenceState(id: Sequence.Id): HandleP[Unit] =
+    getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT]).void
 
   // The `Catchable` instance of `Handle`` needs to be manually written.
   // Without it it's not possible to use `Handle` as a scalaz-stream process effects.
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -243,7 +243,7 @@ package object engine {
   }
 
   private def getState(f: Engine.State => Task[Option[Process[Task, Event]]]): HandleP[Unit] =
-    get.flatMap(s => HandleP(f(s).liftM[HandleStateT].map((Unit, _))))
+    get.flatMap(s => HandleP[Unit](f(s).liftM[HandleStateT].map(((), _))))
 
   /**
     * Given the index of the completed `Action` in the current `Execution`, it
@@ -278,12 +278,12 @@ package object engine {
     /**
       * Log info lifted into Handle.
       */
-    def info(msg: String): HandleP[Unit] = pure((logger.info(msg), None))
+    def info(msg: String): HandleP[Unit] = pure({(logger.info(msg), None); ()})
 
     /**
       * Log warning lifted into Handle.
       */
-    def warning(msg: String): HandleP[Unit] = pure((logger.warning(msg), None))
+    def warning(msg: String): HandleP[Unit] = pure({(logger.warning(msg), None); ()})
   }
 
   /**
@@ -356,7 +356,7 @@ package object engine {
 
   def pure[A](a: A): HandleP[A] = Applicative[HandleP].pure(a)
 
-  private val unit: HandleP[Unit] = pure(Unit)
+  private val unit: HandleP[Unit] = pure(())
 
   val get: HandleP[Engine.State] =
     MonadState[Handle, Engine.State].get.toHandleP
@@ -386,7 +386,8 @@ package object engine {
     modify(st => Engine.State(st.conditions, st.operator, st.sequences.updated(id, s)))
 
   // For debugging
-  def printSequenceState(id: Sequence.Id): HandleP[Option[Unit]] = getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT])
+  def printSequenceState(id: Sequence.Id): HandleP[Option[Unit]] =
+    getSs(id)((qs: Sequence.State) => {Task.now(println(qs)).liftM[HandleStateT]; ()})
 
   // The `Catchable` instance of `Handle`` needs to be manually written.
   // Without it it's not possible to use `Handle` as a scalaz-stream process effects.
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
@@ -17,13 +17,17 @@ trait EpicsCommand {
 
   protected val cs: Option[CaCommandSender]
 
-  def post: SeqAction[Unit] = safe(EitherT(Task.async[TrySeq[Unit]](f => cs.map(_.postCallback {
-    new CaCommandListener {
-      override def onSuccess(): Unit = f(TrySeq(()).right)
+  def post: SeqAction[Unit] = safe(EitherT(
+    Task.async[TrySeq[Unit]](f => {
+      cs.foreach(_.postCallback {
+        new CaCommandListener {
+          override def onSuccess(): Unit = f(TrySeq(()).right)
 
-      override def onFailure(cause: Exception): Unit = f(cause.left)
+          override def onFailure(cause: Exception): Unit = f(cause.left)
+        }
+      } )
+      ()
     }
-  } ).getOrElse(SeqexecFailure.Unexpected("Unable to trigger command.").left)
   ) ) )
 
   def mark: SeqAction[Unit] = safe(EitherT(Task.delay {
@@ -73,7 +77,7 @@ object EpicsCommand {
     safe(SeqAction.either {
       p.map(_.set(f(v)).right).getOrElse(SeqexecFailure.Unexpected("Unable to set parameter.").left)
     } )
-    
+
   def setParameter[T](p: Option[CaParameter[T]], v: T): SeqAction[Unit] =
     safe(SeqAction.either {
       p.map(_.set(v).right).getOrElse(SeqexecFailure.Unexpected("Unable to set parameter.").left)
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
@@ -17,18 +17,22 @@ trait EpicsCommand {
 
   protected val cs: Option[CaCommandSender]
 
-  def post: SeqAction[Unit] = safe(EitherT(
-    Task.async[TrySeq[Unit]](f => {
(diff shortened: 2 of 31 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -275,12 +275,12 @@ package object engine {
     /**
       * Log info lifted into Handle.
       */
-    def info(msg: String): HandleP[Unit] = pure((logger.info(msg), None))
+    def info(msg: String): HandleP[Unit] = pure({(logger.info(msg), None); ()})
```
swalker2m: Might also work to use `scalaz.Functor.void` if necessary as in `pure((logger.info(msg), None)).void`.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -283,12 +283,15 @@ object SeqexecEngine {
...
+  private def decodeTops(s: String): Map[String, String] =
+    s.split("=|,").grouped(2).collect {
+      case Array(k, v) => k.trim -> v.trim
+    }.toMap
 
   private def initSmartGCal(smartGCalHost: String, smartGCalLocation: String): Task[Unit] = {
     // SmartGCal always talks to GS
     val peer = new Peer(smartGCalHost, 8443, Site.GS)
-    Task.delay(Paths.get(smartGCalLocation)) >>= {p => Task.delay(SmartGcal.initialize(peer, p))}
+    Task.delay(Paths.get(smartGCalLocation)) >>= {p => Task.delay(SmartGcal.initialize(peer, p)) *> Task.now(()) }
```
swalker2m: Similarly, I think you can `.void` the result if you don't need it here?

Also, I'm curious why this isn't just a `map`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
cquiroz: Thanks, that seems to work and looks much cleaner

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
```diff
@@ -17,13 +17,17 @@ trait EpicsCommand {
...
+  def post: SeqAction[Unit] = safe(EitherT(
+    Task.async[TrySeq[Unit]](f => {
+      cs.foreach(_.postCallback {
+        new CaCommandListener {
+          override def onSuccess(): Unit = f(TrySeq(()).right)
+
+          override def onFailure(cause: Exception): Unit = f(cause.left)
+        }
+      } )
+      ()
```
cquiroz: This one baffled me. I guess we don't care about the result of the `async` call?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
```diff
@@ -86,9 +86,9 @@ object GmosControllerEpics extends GmosSouthController {
     case s            => DC.setShutterState(encode(s))
   }
 
-  private def roiNumUsed(s: RegionsOfInterest): Int = s match {
-    case RegionsOfInterest(\/-(rois)) => rois.length
-    case RegionsOfInterest(-\/(b))    => 1
+  private def roiNumUsed(s: RegionsOfInterest): Int = s.rois match {
```
cquiroz: This change maybe an error on the compiler. The previous match produced a warning on `non-exhaustive` matching checks

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosHeader.scala
```diff
@@ -182,7 +182,7 @@ object GmosHeader {
     override def dcName: SeqAction[String] = SeqAction(Header.StrDefault)
     override def detectorType: SeqAction[String] = SeqAction(Header.StrDefault)
     override def detectorId: SeqAction[String] = SeqAction(Header.StrDefault)
-    override def exposureTime: SeqAction[Double] = SeqAction(Header.IntDefault)
+    override def exposureTime: SeqAction[Double] = SeqAction(Header.DoubleDefault)
```
cquiroz: @jluhrs Is this correct?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -138,8 +138,7 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
       case _         => false
     }
 
-  def sequence(settings: Settings)
-              (obsId: SPObservationID, sequenceConfig: ConfigSequence, name: String):
+  def sequence(obsId: SPObservationID, sequenceConfig: ConfigSequence, name: String):
```
cquiroz: This is also interesting, settings is not used in this method

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -353,7 +353,7 @@ package object engine {
 
   def pure[A](a: A): HandleP[A] = Applicative[HandleP].pure(a)
 
-  private val unit: HandleP[Unit] = pure(Unit)
+  private val unit: HandleP[Unit] = pure(())
```
jdnavarro: I'm still confused about when to use `()` and `Unit` in Scala. Is `Unit` supposed to be used only at the type level and `()` at the value level? Is it a stylistic thing or is there more to it?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -383,7 +383,8 @@ package object engine {
     modify(st => Engine.State(st.conditions, st.operator, st.sequences.updated(id, s)))
 
   // For debugging
-  def printSequenceState(id: Sequence.Id): HandleP[Option[Unit]] = getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT])
+  def printSequenceState(id: Sequence.Id): HandleP[Unit] =
+    getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT]).void
```
jdnavarro: 👏 Are long lines now warned somehow by the compiler?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouthController.scala
```diff
@@ -85,7 +85,7 @@ object GmosSouthController {
         case (BuiltinROI.CUSTOM, r) if r.nonEmpty => \/.right(new RegionsOfInterest(\/.right(r)) {})
         case _ => \/.left(Unexpected("Inconsistent values for GMOS regions of interest"))
       }
-    def unapply(r: RegionsOfInterest): Option[BuiltinROI \/ List[ROI]] = Some(r.rois)
+    def unapply(r: RegionsOfInterest): Some[BuiltinROI \/ List[ROI]] = Some(r.rois)
```
jdnavarro: `Some` in the type signature? Why is that?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jdnavarro: I think those `Settings` were intended as a placeholder but I might be wrong.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
swalker2m: Correct, `Unit` is the type of which there is one value `()`.

cquiroz: Right, `Unit` is the type. Apparently this causes confusions on haskellers where both are the same

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
cquiroz: No, I just split them whet they seem too long. I'm not sure if gem's scalatyle settings flag them

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouthController.scala
cquiroz: It was an experiment actually to see if the pattern matching worked. I should probably revert it to `Option`

jdnavarro: I'm not saying it's wrong, maybe using the constructor in the type signature in `unapply` makes more sense. I was just curious to know why it was in the type signature.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
tpolecat: Above you were using the companion object of `trait Unit`, called `Unit`, which is not of type `Unit` … it's type is the unrelated `Unit.type`. In case you weren't already confused.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
tpolecat: I don't think this is right. You're constructing a task and then throwing it away.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouthController.scala
tpolecat: It's a particular trick with `unapply`, to signal to the compiler that the pattern is irrefutable; it always matches.

tpolecat:
This looks good other than the async task thing.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
tpolecat: (We discussed on Gitter).

cquiroz:
i added changes to `post` suggested by @tpolecat (thanks a lot) and rebased it. @jluhrs could you check if the latest version makes sense to you

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosHeader.scala
jluhrs: Yes

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: I think what happened was that settings was converted to a class parameter, but not removed from the function.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Header.scala
```diff
@@ -17,7 +17,7 @@ import scalaz.concurrent.Task
   * Created by jluhrs on 2/10/17.
   */
 
-class Flamingos2Header(hs: DhsClient, f2ObsReader: Flamingos2Header.ObsKeywordsReader, f2Reader: Flamingos2Header.InstKeywordsReader, tcsKeywordsReader: TcsKeywordsReader) extends Header {
+class Flamingos2Header(hs: DhsClient, f2ObsReader: Flamingos2Header.ObsKeywordsReader, tcsKeywordsReader: TcsKeywordsReader) extends Header {
```
jluhrs: Ups, I just noticed I forgot to put the FITS keywords that use the values provided by `f2Reader`. This cannot be removed. The keywords must be added instead.

cquiroz: If this is reverted the code won't compile with the new params. I suggest to add them in another PR

jluhrs: Forget my comment. There are two keywords defined in the old Seqexec that have values set but never sent to the DHS.

PR #287, opened 2017-07-25
Actions are now realized just before executing.

jluhrs:
This change allows to have the Actions change to take into account the current values of parameters like the observer name or the observing conditions.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -27,7 +27,7 @@ package object engine {
     * systems.
     */
   type Action = Kleisli[Task, ActionMetadata, Result]
-
+  def fromTask(t: Task[Result]): Action = new Action(_ => t)
   /**
     * An `Execution` is a group of `Action`s that need to be run in parallel
     * without interruption. A *sequential* `Execution` can be represented with
@@ -232,11 +232,10 @@ package object engine {
           case Sequence.State.Final(_, _) =>
             // The sequence is marked as completed here
             putS(id)(seq) *> send(finished(id))
-          case _                          => {
+          case _                          =>
             val u = seq.current.actions.zipWithIndex.map(x => act(x, ActionMetadata(st.conditions, st.operator, seq.toSequence.metadata.observer)))
             val v = merge.mergeN(Process.emitAll(u)) ++ Process(executed (id))
             HandleP.fromProcess(v)
-          }
         }
       }.getOrElse(unit)
     )
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -2,7 +2,7 @@ package edu.gemini.seqexec
 
 import edu.gemini.seqexec.engine.Event._
 import edu.gemini.seqexec.engine.Result.{PartialVal, RetVal}
-import edu.gemini.seqexec.model.Model.{CloudCover, Conditions, ImageQuality, SequenceState, SkyBackground, WaterVapor}
+import edu.gemini.seqexec.model.Model.{CloudCover, Conditions, ImageQuality, SequenceState, SkyBackground, WaterVapor, Operator, Observer}
 import java.util.logging.{Logger => JLogger}
 
 import scalaz._
@@ -15,6 +15,7 @@ package engine {
   object HandleP {
     def fromProcess(p: Process[Task, Event]): HandleP[Unit] = HandleP(Applicative[Handle].pure[(Unit, Option[Process[Task, Event]])](((), Some(p))))
   }
+  case class ActionMetadata(conditions: Conditions, operator: Option[Operator], observer: Option[Observer])
 }
 
 package object engine {
@@ -25,7 +26,7 @@ package object engine {
     * This represents an actual real-world action to be done in the underlying
     * systems.
     */
-  type Action = Task[Result]
+  type Action = Kleisli[Task, ActionMetadata, Result]
 
   /**
     * An `Execution` is a group of `Action`s that need to be run in parallel
@@ -136,6 +137,7 @@ package object engine {
   def setCloudCover(cc: CloudCover): HandleP[Unit] =
     modify(st => Engine.State(st.conditions.copy(cc = cc), st.operator, st.sequences))
 
+
   /**
     * Load a Sequence
     */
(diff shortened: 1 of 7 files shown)


cquiroz:
Am I correct that this PR doesn't yet fix the part of writing the keywords?

jdnavarro:
I guess this might have been discussed in previous meetings I missed, but would you mind reminding what's this PR for?

jdnavarro:
> I guess this might have been discussed in previous meetings I missed, but would you mind reminding what's this PR for?

Never mind writing it here. I already got it.

jluhrs:
@cquiroz You are correct, this PR is just about the functionality that will allow to implement the OBSERVER and OPERATOR keywords.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -15,6 +15,7 @@ package engine {
   object HandleP {
     def fromProcess(p: Process[Task, Event]): HandleP[Unit] = HandleP(Applicative[Handle].pure[(Unit, Option[Process[Task, Event]])](((), Some(p))))
   }
+  case class ActionMetadata(conditions: Conditions, operator: Option[Operator], observer: Option[Observer])
```
cquiroz: Could this mechanism be used to change the sequence on the fly if e.g. the user changed some parameters of the sequence

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/ExecutionSpec.scala
```diff
@@ -8,7 +8,7 @@ import scalaz.concurrent.Task
 class ExecutionSpec extends FlatSpec with Matchers {
 
   val ok: Result = Result.OK(Result.Observed("dummyId"))
-  val action: Action = Task(ok)
+  val action: Action = new Action(_ => Task(ok))
```
cquiroz: Maybe you can add a constructor to do this, it seems repeated in several places

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: Nothing wrong with this but now we have 2 places in the engine where the conditions, the operator, and the observer are defined and would have to be kept in sync somehow.

Should we remove the other places where those parameters are defined and when needed globally obtain it on-the-fly from the underlying Actions? How do we handle a change in any of those parameters when updated from the UI?

For the record, when I worked on making the operator global for the engine I was against it but had to made it nevertheless, so I prefer the metadata as it is in this PR.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -25,7 +26,7 @@ package object engine {
     * This represents an actual real-world action to be done in the underlying
     * systems.
     */
-  type Action = Task[Result]
+  type Action = Kleisli[Task, ActionMetadata, Result]
```
jdnavarro: Why not `case class Action(metadata: ActionMetadata, task: Task[Result]`)? 

I like the `Kleisli` because it's a nice pattern to avoid passing read-only parameters all over the place but I don't see in this PR any code reading the metadata using the `Kleisli` so where are you expecting to use it?

In practice the `Kleisli` represents the Reader monad transformer and I got the impression we wanted to avoid them. I'm fine with monad transformers, so does this mean we can start using them without constraints?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jluhrs: Maybe, but then there will be little left for Action.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -213,27 +215,28 @@ package object engine {
...
-      _.map { seq =>
+    get.flatMap(st => st.sequences.get(id).map { seq =>
         seq match {
           case Sequence.State.Final(_, _) =>
             // The sequence is marked as completed here
             putS(id)(seq) *> send(finished(id))
-          case _                          =>
-            val u = seq.current.actions.zipWithIndex.map(act)
+          case _                          => {
+            val u = seq.current.actions.zipWithIndex.map(x => act(x, ActionMetadata(st.conditions, st.operator, seq.toSequence.metadata.observer)))
```
jdnavarro: I can see that this is only set when the current `Execution` actually being executed, so what's in the `ActionMetadata` before, in the pending Executions when the Sequence is loaded?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
cquiroz: Unconstrained use of transformers? That's reading too much into this 😄 
I'd rather use other techniques as we have seen that transformers make the code very hard to follow

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
cquiroz: That's a good point, I agree they should be in only one place

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jdnavarro: Right, maybe unconstrained is too *strong* of a word. But it's still not clear to me what criteria to use when deciding if it's worth introducing transformers or not.

One easy rule of thumb I was trying to follow is that it's fine as long as it's only 1 layer, but in this case it's not clear this meets the rule because the `Kleisli` Task will be used in code that will be already under in `Handle` which is monad transformer which is made of `StateT and `Task`, so we'll have some code that hidden under the synonyms will be a StateT/Kleisli/Task monad stack.

Again, this is fine with me, but just wanted to make sure it's worth introducing the `Kleisli` if we can do the same with a simple case class.

cquiroz: That's a good rule of thumb. 1 layer is readable enough and not worth going `Free` and 2 or more starts to get unreadabale

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jluhrs: `ActionMetadata` only exists as a way to pass parameters to Action. Its contents come from two places in the engine state, the global state, and the sequence metadata. Its life is short.

PR #289, opened 2017-07-27
SEQNG-299 Carried values for operator, observer and conditions to the FITS keywords.

jluhrs:
Now the values set in the UI end up in the FITS.


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -271,7 +271,9 @@ object Sequence {
 
       override def getCurrentBreakpoint: Boolean = zipper.focus.breakpoint && zipper.focus.done.isEmpty
 
-      override def setObserver(name: String): State = observerL.set(self, Some(name))
+      // I put a guard against blank values, although Observer is forced to have
+      // a default value at an upper level.
+      override def setObserver(name: String): State = observerL.set(self, if(name.isEmpty || name.forall(_.isSpaceChar)) None else Some(name))
```
cquiroz: You can do on the condition `(name.isEmpty...) option name)`

PR #293, opened 2017-07-31
SEQNG-279: Use the typelevel compiler for both jvm and js

cquiroz:
I finally found how to make the typelevel compiler and scala.js to play along


cquiroz:
No, that's pretty much it. it compiles, tests run and the UI looks sane

PR #294, opened 2017-08-01
SEQNG-289: Excessive operator update messages

cquiroz:
Sorry for the size of this PR. This started noting that the observer update messages were sent more than once and snowballed to a large refactoring of the UI.

This PR contains changes in the following areas:

* Pass the user to the engine to let it e.g. log who's changing the observer or who's executing a sequence
* Split of the UI into sections containing the minimal amount of model to be displayed. This is more efficient at the cost of more code complexity
* The split allows us to fix an annoying bug where the focus is lost on the text field when changes in other places occur

There are a couple of things to fix but I have dabbled on this for too long


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
```diff
@@ -17,22 +20,36 @@ case class EventSystem(se: SystemEvent) extends Event
...
-case class SetConditions(conditions: Conditions) extends UserEvent
-case class SetImageQuality(iq: ImageQuality) extends UserEvent
-case class SetWaterVapor(wv: WaterVapor) extends UserEvent
-case class SetSkyBackground(wv: SkyBackground) extends UserEvent
-case class SetCloudCover(cc: CloudCover) extends UserEvent
-case object Poll extends UserEvent
-case class GetState(f: (Engine.State) => Task[Option[Process[Task, Event]]]) extends UserEvent
-case class Log(msg: String) extends UserEvent
+sealed trait UserEvent {
+  def user: Option[UserDetails]
```
jluhrs: Shouldn't it be `val` ?

cquiroz: Often on `trait`s we use `def` rather than `Val` to let implementations decide

PR #295, opened 2017-08-02
SEQNG-297 Fixed ordering of keywords.

jluhrs:


jluhrs:
Yes. This is the change I had to left out because it caused the HTTP proxy to crash.

PR #296, opened 2017-08-03
SEQNG-308: Failure to set observer and operator to the empty string

cquiroz:
This is a small change to let users set the operator as the empty string unlike the current situation where the empty string is set as `None`

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
@@ -199,8 +199,8 @@ case class StateKeywordsReader(conditions: Conditions, operator: Option[Operator
   def encodeCondition(c: Int): String = if(c == 100) "Any" else s"$c-percentile"
 
   // TODO: "observer" should be the default when not set in state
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/StandardHeader.scala
```diff
@@ -199,8 +199,8 @@ case class StateKeywordsReader(conditions: Conditions, operator: Option[Operator
   def encodeCondition(c: Int): String = if(c == 100) "Any" else s"$c-percentile"
 
   // TODO: "observer" should be the default when not set in state
-  def getObserverName: SeqAction[String] = SeqAction(observer.getOrElse("observer"))
-  def getOperatorName: SeqAction[String] = SeqAction(operator.getOrElse("ssa"))
+  def getObserverName: SeqAction[String] = SeqAction(observer.filter(_.nonEmpty).getOrElse("observer"))
+  def getOperatorName: SeqAction[String] = SeqAction(operator.filter(_.nonEmpty).getOrElse("ssa"))
```
jdnavarro: I guess these are the default keyword values when no operator or observer are defined, is that right?

cquiroz: That is my understanding from reading the code

PR #299, opened 2017-08-08
SEQNG-233 Implement of GMOS-N support.

jluhrs:
My main goal was to add support for GMOS-N reusing as much as possible of the existing GMOS-S code.
This implementation uses (and abuses) dependent types.
One of the main problems of this implementation is that `trait`s that depend on dependent types cannot be properly sealed (or at least I couldn't find a way). For example, in this code
```
trait A {
    type X
    trait U
    case class V(v: X) extends U
}
```
A `match` on a value of type `U` has to match `V` for all possible values of `X`. Another problem is that if `A` defines a method `foo(v: V)` , then an implementation of `A` that defines a concrete type for `X` can still have its implementation of `foo` receive a value `V` for a different `X`.

I plan to reorganize the code with the following considerations:
* The instrument code can be functionally divided in four parts: the configuration model, the conversion from the ODB sequence to the configuration model, the application of the configuration model, and the conversion of the configuration model to EPICS values.
* A big portion of the GMOS-S and GMOS-N configuration model is common to both, as it is the conversion from the ODB sequence.
* The object responsible for the application of the configuration model can be swapped (EPICS or Sim)
* The application of the configuration at the EPICS level is common to both instruments, as the types required by the EPICS channels.
* But the conversion to EPICS parameters is not completely the same. Specifically, the parameters that differ in the configuration model need a different conversion, although they all convert to the same types.

---

I am happier with the latest implementation. I factored out all the differences between GMOS-S and GMOS-N while keeping the common code together.
There is use of abstract inner types. I had some troubles with the syntax, specially with the differences between using `T#A` and `t.A`, and in some parts I just used what worked.


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
```diff
@@ -21,15 +17,16 @@ import scalaz.Scalaz._
...
   private val Log = Logger.getLogger(getClass.getName)
 
+  import GmosController._
   import EpicsCodex._
 
   val CC = GmosEpics.instance.configCmd
   val DC = GmosEpics.instance.configDCCmd
 
-  override def getConfig: SeqAction[GmosSouthConfig] = ???
+  override def getConfig: SeqAction[GmosConfig] = ???
```
cquiroz: What do we need to implement this?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
```diff
@@ -167,110 +163,44 @@ object GmosControllerEpics extends GmosSouthController {
...
-        _ <- beam.fold(SeqAction.void)(b => CC.setInBeam(encode(b)))
-      } yield ()
+      fpuName.fold(SeqAction.void)(CC.setFpu) *>
+        beam.fold(SeqAction.void)(b => CC.setInBeam(encode(b)))
     }
 
     cc match {
-      case UnknownFPU          => SeqAction.void
-      case BuiltInFPU(fpu)     => builtInFPU(fpu)
+      case this.BuiltInFPU(fpu)     => builtInFPU(fpu)
```
cquiroz: Alignment?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouthController.scala
```diff
@@ -1,95 +1,8 @@
...
-        case _ => \/.left(Unexpected("Inconsistent values for GMOS regions of interest"))
-      }
-    def unapply(r: RegionsOfInterest): Some[BuiltinROI \/ List[ROI]] = Some(r.rois)
-  }
-
-  final case class DCConfig(t: ExposureTime, b: BiasTime, s: ShutterState, r: CCDReadout, bi: CCDBinning, roi: RegionsOfInterest)
-
-  final case class GmosSouthConfig(cc: CCConfig, dc: DCConfig)
-
+trait GmosSouthController extends GmosController {
```
cquiroz: nice

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -47,7 +47,7 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
       if (settings.instForceError) Flamingos2ControllerSimBad
       else Flamingos2ControllerSim
     } else Flamingos2ControllerEpics,
-    if (settings.instSim) GmosSouthControllerSim else GmosControllerEpics
+    if (settings.instSim) GmosSouthControllerSim else GmosSouthControllerEpics
```
cquiroz: I guess you need to take into account the site here

jluhrs: You are right

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
jluhrs: A lot of work to read every value from EPICS and create a model from them. It is not used right now, but having it will allow to implement a more efficient configuration, that only changes what needs to change.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
jluhrs: I hate you.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouthController.scala
jluhrs: This part I like. Maybe a better way to use it is to pass it around as a type parameter that defines other types. So maybe define a `trait GmosDependantController` that as the not-common types, and have the common types in a `GmosController` defined as  `GmosController[T<:GmosDependantController]`

jluhrs:
@cquiroz You mean, like with the real instrument?

cquiroz:
At last loading from the ODB

jluhrs:
@cquiroz Yes, I loaded a sequence from the ODB, and run it.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosController.scala
```diff
@@ -0,0 +1,101 @@
...
+  */
+trait GmosController {
+  import GmosController._
+
+  type Filter
+  type FPU
+  type GmosStageMode
+  type Disperser
+
+  case class BuiltInFPU(fpu: FPU) extends GmosFPU
```
cquiroz: I'm not an expert on path dependent types but this maybe doable if you were to declare `GmosFPU` to take the type of `FPU` as a type param?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Resource.scala
```diff
@@ -19,7 +19,8 @@ object Resource {
   case object Gems extends Resource
   case object Altair extends Resource
   trait Instrument extends Resource
-  case object GMOS extends Instrument
+  case object GMOS_S extends Instrument
```
cquiroz: This has changed a bit upstream, so you'd need to rebase. I defined GMOS-N to get the UI to work

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosController.scala
```diff
@@ -0,0 +1,159 @@
...
+    override type GmosStageMode = edu.gemini.spModel.gemini.gmos.GmosNorthType.StageModeNorth
+    override type Disperser = edu.gemini.spModel.gemini.gmos.GmosNorthType.DisperserNorth
+  }
+
+  type NorthConfigTypes = Config[NorthTypes]
+  val northConfigTypes = new NorthConfigTypes
+
+  // This is a trick to allow using a type from a class parameter define the type of another type parameter
+  class GmosConfig[T<:SiteDependentTypes](val cc: Config[T]#CCConfig, val dc: DCConfig, val c: Config[T]) {
+    def this(c: Config[T])(cc: c.CCConfig, dc: DCConfig) = this(cc, dc, c)
```
cquiroz: neat, do you need both constructors to be public?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerEpics.scala
```diff
@@ -167,125 +152,59 @@ object GmosControllerEpics extends GmosSouthController {
...
-        _ <- beam.fold(SeqAction.void)(b => CC.setInBeam(encode(b)))
-      } yield ()
+      fpuName.fold(SeqAction.void)(CC.setFpu) *>
+        beam.fold(SeqAction.void)(b => CC.setInBeam(beamEncoder.encode(b)))
     }
 
     cc match {
-      case UnknownFPU          => SeqAction.void
-      case BuiltInFPU(fpu)     => builtInFPU(fpu)
+      case cfg.BuiltInFPU(fpu)     => builtInFPU(fpu)
```
cquiroz: Alignment?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouthControllerEpics.scala
```diff
@@ -0,0 +1,97 @@
...
+      case FPU.NS_3        => ("NS1.0arcsec".some, InBeam.some)
+      case FPU.NS_4        => ("NS1.5arcsec".some, InBeam.some)
+      case FPU.NS_5        => ("NS2.0arcsec".some, InBeam.some)
+      case FPU.CUSTOM_MASK => (none, none)
+    }
+    (r._1, r._2.map(GmosControllerEpics.beamEncoder.encode))
+  } }
+
+  override val filter: EncodeEpicsValue[SouthTypes#Filter, (String, String)] = EncodeEpicsValue{
+    case Filter.Z_G0343       => ("Z_G0343", "open2-8")
```
cquiroz: I bet you want to align these on `=>`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -204,7 +206,7 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
   private def calcInstHeader(config: Config, inst: Resource.Instrument): TrySeq[Header] = inst match {
     case Resource.F2   =>  TrySeq(Flamingos2Header(systems.dhs, new Flamingos2Header.ObsKeywordsReaderImpl(config),
       if (settings.tcsKeywords) TcsKeywordsReaderImpl else DummyTcsKeywordsReader))
-    case Resource.GMOS =>
+    case Resource.GMOS_S|Resource.GMOS_N =>
```
cquiroz: spaces around ` | `?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Resource.scala
jluhrs: I rebased, but didn't find any difference.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosController.scala
jluhrs: Not really.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosSouthControllerEpics.scala
jluhrs: I prefer to respect the choice of the original author.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Resource.scala
cquiroz: Ok

PR #300, opened 2017-08-08
SEQNG-296A: Fix smart gcal exposure times

cquiroz:
This doesn't fix SEQNG-296 but this should be fixed nonetheless. For F2 the exposure time is always read from the observation while for calibrations it should be read from `calibration:exposureTime`

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
@@ -9,12 +9,12 @@ import edu.gemini.spModel.gemini.flamingos2.Flamingos2.Filter
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2.ReadoutMode
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2.Reads
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2.WindowCover
-import edu.gemini.spModel.obscomp.InstConstants.{DARK_OBSERVE_TYPE, OBSERVE_TYPE_PROP}
+import edu.gemini.spModel.obscomp.InstConstants.{ARC_OBSERVE_TYPE, CAL_OBSERVE_TYPE, BIAS_OBSERVE_TYPE, FLAT_OBSERVE_TYPE, SCIENCE_OBSERVE_TYPE, DARK_OBSERVE_TYPE, OBSERVE_TYPE_PROP}
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2._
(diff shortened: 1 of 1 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
```diff
@@ -91,24 +91,38 @@ object Flamingos2 {
   }
 
   private def disperserFromObserveType(observeType: String, d: Disperser): Grism = observeType match {
-    case DARK_OBSERVE_TYPE    => Grism.Dark
-    case _                    => d
+    case DARK_OBSERVE_TYPE => Grism.Dark
+    case _                 => d
   }
 
+  private def exposureTime(config: Config): ExtractFailure \/ Duration =
```
swalker2m: My understanding is that you have `instrument:exposureTime`, `calibration:exposureTime`, and `observe:exposureTime`.  The `instrument` and `calibration` values won't necessarily agree but the "right" value to use will always be in `observe:exposureTime` which should match the one that corresponds to the step type.  In other words, I think this may be doing more work than necessary.

cquiroz: Ok, I see. The current seqexec apparently uses `calibration:exposureTime` and that's the reason for this change. But if it is not necessary I can remove it

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2.scala
```diff
@@ -91,24 +91,38 @@ object Flamingos2 {
...
+      q       <- config.extract(INSTRUMENT_KEY / DECKER_PROP).as[Decker]
+      r       <- fpuConfig(config)
+      s       <- config.extract(INSTRUMENT_KEY / FILTER_PROP).as[Filter]
+      t       <- config.extract(INSTRUMENT_KEY / LYOT_WHEEL_PROP).as[LyotWheel]
+      u       <- config.extract(INSTRUMENT_KEY / DISPERSER_PROP).as[Disperser].map(disperserFromObserveType(obsType, _))
 
     } yield CCConfig(p, q, r, s, t, u) ).leftMap(e => SeqexecFailure.Unexpected(ConfigUtilOps.explain(e)))
 
   def dcConfigFromSequenceConfig(config: Config): TrySeq[DCConfig] = ( for {
-    p <- config.extract(OBSERVE_KEY / EXPOSURE_TIME_PROP).as[java.lang.Double].map(x => Duration(x, SECONDS))
```
swalker2m: It does look to me like this should work.

cquiroz:
Closing this given @swalker2m comments

PR #301, opened 2017-08-09
SEQNG-311: Set the observer for newly scheduled observations

cquiroz:
This fixes a small bug where the operator is set for the first set of sequences but not for newly queued sequences


jdnavarro:
I tried it in my machine setting the operator for new loaded sequences is seems to be working fine. How did you realize it wasn't setting the operator correctly?

cquiroz:
checking the logs you can see some missing updates

PR #302, opened 2017-08-09
SEQNG-272: Use scalastyle to match the gem project

cquiroz:
This PR establish the use of scalastyle along the settings used on the gem project. There are some refactorings (mostly breaking up large methods/files) to be able to pass as much of the rules as possible

In a few places a scalastyle ignore annotation has been used for example for large objects containing many type definitions

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -387,7 +387,7 @@ package object engine {
 
   // For debugging
   def printSequenceState(id: Sequence.Id): HandleP[Unit] =
-    getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT]).void
+    getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT]).void // scalastyle:ignore
 
   // The `Catchable` instance of `Handle`` needs to be manually written.
   // Without it it's not possible to use `Handle` as a scalaz-stream process effects.
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClientHttp.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClientHttp.scala
@@ -1,0 +1,143 @@
+package edu.gemini.seqexec.server
+
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+import java.util.concurrent.atomic.AtomicInteger
+
+import argonaut._
+import Argonaut._
+import edu.gemini.seqexec.model.dhs.ImageFileId
+import edu.gemini.seqexec.server.DhsClient.{ImageParameters, KeywordBag}
+import org.apache.commons.httpclient.HttpClient
+import org.apache.commons.httpclient.methods.{EntityEnclosingMethod, PostMethod, PutMethod}
+
+import scala.io.Source
+import scalaz.concurrent.Task
+import scalaz.EitherT
+import scalaz._
+
+/**
+  * Implementation of DhsClient that interfaces with the real DHS over the http interface
+  */
+class DhsClientHttp(val baseURI: String) extends DhsClient {
+  import DhsClientHttp._
+
+  // Connection timeout, im milliseconds
+  val timeout = 10000
+
+  implicit def errorDecode: DecodeJson[Error] = DecodeJson[Error]( c => for {
+      t   <- (c --\ "type").as[ErrorType]
+      msg <- (c --\ "message").as[String]
+    } yield Error(t, msg)
+  )
+
+  implicit def obsIdDecode: DecodeJson[TrySeq[ImageFileId]] = DecodeJson[TrySeq[ImageFileId]](c => {
+    val r = c --\ "response"
+    val s = (r --\ "status").as[String]
+    s.flatMap {
+      case "success" =>
+        (r --\ "result").as[String].map(TrySeq(_))
+      case "error"   =>
+        (r --\ "errors").as[List[Error]].map(l =>
+          TrySeq.fail[ImageFileId](SeqexecFailure.Unexpected(l.mkString(", ")))
+        )
+      case r         =>
+        DecodeResult.fail(s"Unknown response: $r", s.history.getOrElse(CursorHistory.empty))
+    }
+  } )
+
+  implicit def unitDecode: DecodeJson[TrySeq[Unit]] = DecodeJson[TrySeq[Unit]]( c => {
+    val r = c --\ "response"
+    val s = (r --\ "status").as[String]
+    s flatMap {
+      case "success" =>
+        DecodeResult.ok(TrySeq(()))
+      case "error"   =>
+        (r --\ "errors").as[List[Error]].map(
+          l => TrySeq.fail[Unit](SeqexecFailure.Unexpected(l.mkString(", "))))
+      case r         =>
+        DecodeResult.fail(s"Unknown response: $r", s.history.getOrElse(CursorHistory.empty))
+    }
+  } )
+
+  implicit def imageParametersEncode: EncodeJson[DhsClient.ImageParameters] = EncodeJson[DhsClient.ImageParameters]( p =>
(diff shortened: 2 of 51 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -387,7 +387,7 @@ package object engine {
 
   // For debugging
   def printSequenceState(id: Sequence.Id): HandleP[Unit] =
-    getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT]).void
+    getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[HandleStateT]).void // scalastyle:ignore
```
jdnavarro: What does it complain about here?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClientHttp.scala
```diff
@@ -0,0 +1,143 @@
...
+      TrySeq(f"S${date.format(format)}S${counter.incrementAndGet()}%04d")
+    })
+
+  override def setKeywords(id: ImageFileId, keywords: KeywordBag, finalFlag: Boolean): SeqAction[Unit] = SeqAction(())
+
+}
+
+object DhsClientSim {
+  def apply(date: LocalDate): DhsClient = new DhsClientSim(date)
+}
```
jdnavarro: Is this only a file rename?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -379,5 +380,6 @@ object SeqexecEngine {
 
     }
   }
+  // scalastyle:on
```
jdnavarro: This is a large block where scalastyle is off. I'm assume there are several warnings but can't see what could it be.

scalastyle-config.xml
```diff
@@ -0,0 +1,75 @@
...
+ </check>
+
+  <check level="error" class="org.scalastyle.scalariform.EmptyClassChecker" enabled="true"/>
+  <check level="error" class="org.scalastyle.scalariform.ForBraceChecker" enabled="true"/>
+  <!-- <check level="error" class="org.scalastyle.scalariform.ImportGroupingChecker" enabled="true"/> -->
+  <!-- <check level="error" class="org.scalastyle.scalariform.MultipleStringLiteralsChecker" enabled="true"/> -->
+  <check level="error" class="org.scalastyle.scalariform.NoCloneChecker" enabled="true"/>
+  <check level="error" class="org.scalastyle.scalariform.NoFinalizeChecker" enabled="true"/>
+  <check level="error" class="org.scalastyle.scalariform.NoWhitespaceAfterLeftBracketChecker" enabled="true"/>
+  <!-- <check level="error" class="org.scalastyle.scalariform.NoWhitespaceBeforeLeftBracketChecker" enabled="true"/> -->
```
jdnavarro: Do we get into too much trouble if we enable this one?

scalastyle-config.xml
```diff
@@ -0,0 +1,75 @@
...
+  </parameters>
+ </check> -->
+ <check level="error" class="org.scalastyle.scalariform.UppercaseLChecker" enabled="true"></check>
+ <check level="error" class="org.scalastyle.scalariform.SimplifyBooleanExpressionChecker" enabled="true"></check>
+ <check level="error" class="org.scalastyle.scalariform.MethodLengthChecker" enabled="true">
+  <parameters>
+   <parameter name="maxLength"><![CDATA[50]]></parameter>
+  </parameters>
+ </check>
+ <!-- <check level="error" class="org.scalastyle.file.NewLineAtEofChecker" enabled="true"></check> -->
```
jdnavarro: I think no one in the team uses an editor that doesn't append a newline to the end of the file, but it'd be nice to enable it just in case.

scalastyle-config.xml
```diff
@@ -0,0 +1,75 @@
...
+  <parameters>
+   <parameter name="regex"><![CDATA[println]]></parameter>
+  </parameters>
+ </check>
+ <check level="error" class="org.scalastyle.scalariform.NumberOfTypesChecker" enabled="true">
+  <parameters>
+   <parameter name="maxTypes"><![CDATA[30]]></parameter>
+  </parameters>
+ </check>
+ <!-- <check level="error" class="org.scalastyle.scalariform.CyclomaticComplexityChecker" enabled="true">
```
jdnavarro: This was a bugger when I was using scalastyle on my own. It's not that we need to enable it but maybe we need to be more careful about about very large functions.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
cquiroz: About `println` we have a couple more in other parts of the code

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/DhsClientHttp.scala
cquiroz: yes, split to have less types on one file

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
cquiroz: Yeah, i wasn't able to break it off neatly

scalastyle-config.xml
cquiroz: the scalastyle config comes directly from the gem project. We could certainly change things and this particular setting would be fine with me

scalastyle-config.xml
cquiroz: I guess this is sort of replaced by the max length of functions

jluhrs:
Wait, does this mean that if I approve this PR and it is merged before mine, I'll need to change my PR to comply with this forced style?

jluhrs:
I am not sure about taking out internal functions. I agree that it looks better, but it relax a meaningful restriction over the function, that it only makes sense for the container function.

cquiroz:
No don't worry it will be easy:D in truth the style check doesn't touch things like indentation, etc it is mostly about shorter methdos and documenting types, etc

cquiroz:
If you point to specific functions I'll revert moving the functions back inside the body and mark them as scalastyle off
I agree that in some cases is an it forced

cquiroz:
Javier pointed out that in some cases internal functions have been moved out to accommodate  the rule of functions no longer than 50 lines. If that's the preferred style I'd rather break the rule and do what's right from a style point of view.

Let me know if you want me to revert some of those

jdnavarro:
> Javier pointed out that in some cases internal functions have been moved out to accommodate the rule of functions no longer than 50 lines. If that's the preferred style I'd rather break the rule and do what's right from a style point of view.

> Let me know if you want me to revert some of those

Maybe the 50 lines limit is too restrictive, but we had some very long internal functions that may be worth splitting and make them standalone. I don't know where the limit should be... 200?

swalker2m:
I don't really have a strong opinion about this and the limit is kind of arbitrary I suppose, but really long functions are probably a bad sign.  I guess I'd prefer to keep them short even if it means breaking out a few internal functions.

scalastyle-config.xml
tpolecat: Sometimes I like to align things on brackets so I turned it off. I didn't put much thought into it.

scalastyle-config.xml
tpolecat: It would be nice if we could flag any method with more than one `if`.

tpolecat:
Method length is a good heuristic. I would suggest a number closer to 10 than to 50. The gem project has about 12,000 lines of Scala and only has 4 methods longer than 20 lines.

cquiroz:
one annoying thing is that the check complains also if a class definition is longer than 50 lines

tpolecat:
> one annoying thing is that the check complains also if a class definition is longer than 50 lines

Really? Hm. We have plenty of those in gem and no warnings.

PR #308, opened 2017-08-16
SEQNG-288: Completed observations should remain

cquiroz:
This PR changes the logic on the UI to keep completed sequences on the sequences table when completed though they get removed from the queue table as seen below

screenshot 2017-08-16 15 10 28

This was tricky because the state is removed from the server so we have to keep an extra copy on the client. This also mean that the completed sequence will disappear on a page refresh but that's likely ok

I added a small improvement to the tabs displaying the current running step

screenshot 2017-08-16 16 39 07


jluhrs:
When is the completed sequence removed (besides that on a page refresh) ?

cquiroz:
it is removed when you select another sequence of the same instrument

PR #310, opened 2017-08-18
SEQNG-319: Use sbt-git to generate the application version

cquiroz:
With this PR I propose that we use [sbt-git](https://github.com/sbt/sbt-git) to generate the version number. This is based on tags and what is provided by [git-describe](https://git-scm.com/docs/git-describe)

The benefit is that our public version will always match what software version on github as we often forget to update the version number or even release uncommited changes. The UI gets the actual version number too

screenshot 2017-08-18 08 29 12

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,6 +25,20 @@ parallelExecution in (ThisBuild, Test) := false
 
 cancelable in Global := true
 
+// Settings to use git to define the version of the project
+enablePlugins(GitVersioning)
+
+git.useGitDescribe := true
+
+git.formattedShaVersion := git.gitHeadCommit.value map { sha => s"v$sha" }
+
+git.uncommittedSignifier in ThisBuild := Some("UNCOMMITTED")
+
+enablePlugins(GitBranchPrompt)
+
+//////////////
+// Projects
+//////////////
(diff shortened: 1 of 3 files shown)


swalker2m:
Every semester there are tweaks to the program model, and of course the clients often are incompatible with previous versions of the model.  For that reason we've always included the corresponding semester in the version number.  (Though maybe not for the seqexec, I'm not sure.)

I think we'd like to get away from having giant releases aligned with semester, but nevertheless it is on semester boundaries that filters are added or removed, etc.  It seems like some way to indicate the semester that the release corresponds to is still useful?

cquiroz:
sbt-git allows you to customize the version or use tags so the semester could be put in there. My main aim with this change is to make sure that we don't use the same version when we are in between tags, like it is often happening in between releases

swalker2m:
Thinking about it a bit, given that we're using webapps and the user has no choice about which version to use, there's no danger that they will try to use the wrong version.  Maybe semester isn't so important anymore.  As you pointed out, we should probably try to capture the times when filters, etc. are valid in the model.

cquiroz:
One thing to consider is when we have multiple projects we may need to have multiple versions for each. That should be doable with `sbt-git` but would still require some changes

PR #315, opened 2017-08-24
Automate header creation on compile

cquiroz:
This change to the build will add the header to files without it on compile


jdnavarro:
Which header is this?

cquiroz:
The one at the top of each file with the copyright. This change will add it automatically to files without it, like it is done on the gem project

PR #316, opened 2017-08-28
SEQNG-323: Add a UI element to display log messages

cquiroz:
This PR supports displaying the last N log messages on the UI. Note that the current control is a simple text area, that will be improved in future updates, this PR is about enabling the basic functionality

Another issue is that some of the messages are more relevant to the server than the client, we should improve this in future PRs as well.

screenshot 2017-08-28 11 35 00


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/FixedLengthBuffer.scala
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2016-2017 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package edu.gemini.web.common
+
+import scalaz.{Show, Equal, Functor}
+import scalaz.syntax.std.boolean._
+
+object FixedLengthBuffer {
+  private final case class FixedLengthBufferImpl[A](maxLength: Int, data: Vector[A]) extends FixedLengthBuffer[A] {
```
jdnavarro: I've seen this pattern `...Impl` a lot in Scala, but I'm still not sure what's for? Is it to hide the default constructors and create your own custom *smart* constructors?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/FixedLengthBuffer.scala
```diff
@@ -0,0 +1,69 @@
...
+    // Sanity check
+    require(maxLength >= data.length)
+    require(maxLength >= 0)
+
+    def append(element: A): FixedLengthBuffer[A] = {
+      if (data.length == maxLength && data.length >= 0) {
+        FixedLengthBufferImpl[A](maxLength, data.tail :+ element)
```
jdnavarro: So this means when maximum buffer length is reached the first messages are deleted, right? Nice 👍

cquiroz: Yes we keep only N elements to avoid killing the memory on the client

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.shared/src/main/scala/edu/gemini/seqexec/web/common/FixedLengthBuffer.scala
cquiroz: exactly. since the `Impl` class is private you cannot see so you have to go through the `apply` or `unsafe...` constructors. For this particular class it means we can restrict the `maxLength` to only positive values

Now that you mention it, there is a slightly cleaner technique with `abstract class`. i'll try that before merging

PR #321, opened 2017-09-04
SEQNG-355: Focus lost when running a sequence

cquiroz:
This PR fixes a bug that takes the focus away on the observer text field when a sequence is running. Every time a step completes the text field would be re-rendered and the focus lost.

This PR fixes this by making the lenses more specific. I also did a relatively large refactoring to the circuit and handler class which are starting to be too large


cquiroz:
The build passes now. I couldn't reproduce the error locally nor on the remote server

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/SequenceArea.scala
```diff
@@ -55,36 +55,35 @@ object SequenceStepsTableContainer {
...
-  final case class Props(site: SeqexecSite, p: ModelProxy[InstrumentStatusFocus]) {
+  final case class Props(site: SeqexecSite, p: ModelProxy[InstrumentTabContentFocus]) {
     protected[sequence] val connect = SeqexecCircuit.connect(SeqexecCircuit.statusAndStepReader(p().instrument))
   }
 
   private val component = ScalaComponent.builder[Props]("SequenceTabContent")
     .stateless
     .render_P { p =>
-      val InstrumentStatusFocus(instrument, active, id, _) = p.p()
+      val InstrumentTabContentFocus(instrument, active, sequenceSelected) = p.p()
```
cquiroz: This is the actual fix for the bug. The last unused param on `InstrumentStatusFocus` is not used but it can change triggering a re-render

PR #322, opened 2017-09-05
SEQNG-359: ODB doesn't send engineering parameters for FLAMINGOS-2 observations

cquiroz:
This PR brings the new ocs library versions done by @swalker2m to fix this on the seqexec client library

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ODBProxy.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ODBProxy.scala
@@ -31,7 +31,7 @@ class ODBProxy(val loc: Peer, cmds: ODBProxy.OdbCommands) {
 
   def host(): Peer = loc
   def read(oid: SPObservationID): SeqexecFailure \/ ConfigSequence =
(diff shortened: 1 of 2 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ODBProxy.scala
```diff
@@ -31,7 +31,7 @@ class ODBProxy(val loc: Peer, cmds: ODBProxy.OdbCommands) {
 
   def host(): Peer = loc
   def read(oid: SPObservationID): SeqexecFailure \/ ConfigSequence =
-    SeqExecService.client(loc).sequence(oid).leftMap(SeqexecFailure.ODBSeqError)
+    SeqExecService.client(loc).sequence(oid).map(_.config).leftMap(SeqexecFailure.ODBSeqError)
```
swalker2m: Is this change part of it?

cquiroz: Yes, I changed the return type of `sequence` from `ConfigSequence` to `SeqexecSequence` which contains more data including the sequence name and the already executed file ids. Those will be used in a future PR but this config sequence needs to be extracted here in any case

PR #323, opened 2017-09-05
SEQNG-312: Display file id of already executed steps

cquiroz:
Thanks to #322 we are extracting more information about a sequence from the ODB. In this PR we use this to obtain the title of the sequence (thus avoiding an extra call to the odb browser) and we also read the file names of previously executed steps

Thus we can display de file id of previous executions when loading a sequence like

screenshot 2017-09-05 15 43 03

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -189,7 +189,7 @@ lazy val edu_gemini_seqexec_server = project
           Knobs,
           OpenCSV,
           Log4s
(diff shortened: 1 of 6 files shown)


swalker2m:
Hmm, steps 2 - 5 all have the same filename?

cquiroz:
Yes, i was surprised too but it has a good explanation. I often test these running a step and then restarting the server which as a result resets the file counter
We simulate the DHS in development with a simple counter

PR #327, opened 2017-09-12
SEQNG-122: Get fileId in the engine state as soon as the observation starts

cquiroz:
This PR is to request comments as this is not properly working right now

The idea on this PR is to change the engine to read the file id at the beginning of the observation, and then send an extra event to the client at the start of the observation


cquiroz:
I found the bug and this is ready for review. Here i'm attaching a screenshot

screenshot 2017-09-12 14 16 21

PR #328, opened 2017-09-13
Implemented command to cancel a pause request

jluhrs:
I also added some tests, and fixed some bugs that the same tests uncovered.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -90,34 +90,21 @@ package object engine {
 
   /**
     * Changes the `Status` and returns the new `Queue.State`.
-    *
-    * It also takes care of initiating the execution when transitioning to
-    * `Running` `Status`.
     */
   def switch(id: Sequence.Id)(st: SequenceState): HandleP[Unit] =
-    resources.flatMap(
-      other => getS(id).flatMap {
-        case Some(seq) =>
-          if (st === SequenceState.Running)
-            // No resources being used by other running sequences
-            if (seq.toSequence.resources.intersect(other).isEmpty)
-              putS(id)(Sequence.State.status.set(seq, st))
-            // Some resources are being used
-            else send(busy(id))
-          else putS(id)(Sequence.State.status.set(seq, st))
-        case None => unit
-      }
-    )
+    modifyS(id)(s => Sequence.State.status.set(s, st))
 
   def start(id: Sequence.Id): HandleP[Unit] =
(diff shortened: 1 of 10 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -90,25 +90,29 @@ package object engine {
 
   /**
     * Changes the `Status` and returns the new `Queue.State`.
-    *
-    * It also takes care of initiating the execution when transitioning to
-    * `Running` `Status`.
     */
   def switch(id: Sequence.Id)(st: SequenceState): HandleP[Unit] =
```
jluhrs: `switch` was used for all sequence state changes. It add a special behavior for going to `Running`. Now that I added special ehavior for other state transitions, I replaced the use of `switch` for specialized functions `start`, `pause` and `cancelPause`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -90,25 +90,29 @@ package object engine {
...
       other => getS(id).flatMap {
         case Some(seq) =>
-          if (st === SequenceState.Running)
-            // No resources being used by other running sequences
-            if (seq.toSequence.resources.intersect(other).isEmpty)
-              putS(id)(Sequence.State.status.set(seq, st))
-            // Some resources are being used
+          // No resources being used by other running sequences
+          if (seq.status === SequenceState.Idle)
+            if(seq.toSequence.resources.intersect(other).isEmpty)
```
cquiroz: Maybe it is worth to make a method doing the whole `seq.toSequence...`

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/StepSpec.scala
```diff
@@ -185,18 +195,189 @@ class StepSpec extends FlatSpec {
...
+        )
+      )
+
+    val qss = process(Process.eval(Task.now(Event.pause(seqId, user))))(qs0).runLog.unsafePerformSync.map(_._2)
 
+    assert(qss.length == 1)
+    inside (qss.headOption.flatMap(_.sequences.get(seqId))) {
+      case Some(Sequence.State.Zipper(zipper, status)) =>
+        inside (zipper.focus.toStep) {
+          case Step(_, _, _, _, _, _, ex1::ex2::Nil) =>
```
cquiroz: i suppose there are other possible cases, Will the fail the test?. is that assured by `inside`?

PR #332, opened 2017-09-15
SEQNG-118 Implemented engine event to support stop and abort observe

jluhrs:
The idea is that the engine doesn't know how to stop or abort an exposure. That is provided by the upper layer. But the engine still must be aware, to stop the sequence in case the function provided by the upper layer doesn't work. That takes care of limit cases, like when the stopObserve is sent to the instrument just when the instrument finishes the exposure.


jluhrs:
Yes.

PR #334, opened 2017-09-28
SEQNG-100 Implemented GMOS Stop Observe command.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerSim.scala
```diff
@@ -3,28 +3,49 @@
...
+  private val tic = 200
+  private def observeTic(obsid: ImageFileId, remain: Int): Task[\/[SeqexecFailure, ImageFileId]] =
+    if(remain < tic) Task {
+      Log.info(s"Simulate Gmos $name observation completed")
+      TrySeq(obsid)
+    }
+    else Task.delay(stopFlag.get).flatMap( stop => Task.delay(abortFlag.get).flatMap( abort =>
+      if(stop) Task(TrySeq.fail(SeqexecFailure.Execution("Exposure stopped by user.")))
+      else if(abort) Task(TrySeq.fail(SeqexecFailure.Execution("Exposure aborted by user.")))
+      else Task{ Thread.sleep(tic.toLong)} *> observeTic(obsid, remain-tic)))
```
jluhrs: I could have used a for comprehension, but I was worried about tail recursion.

swalker2m: This isn't tail recursive anyway.  Add a `@tailrec` annotation (which is a good idea anyway).  The last branch of your `if/else` expression returns the result of `*>` which is passed the result of the recursive call.

cquiroz: I'm not sure what you mean by being worried about tail recursion in this case

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerSim.scala
```diff
@@ -3,28 +3,49 @@
...
+
   override def observe(obsid: ImageFileId): SeqAction[ImageFileId] = EitherT( Task {
     Log.info(s"Simulate taking Gmos $name observation with label " + obsid)
-    Thread.sleep(5000)
-    Log.info(s"Simulate Gmos $name observation completed")
-    TrySeq(obsid)
-  } )
+    stopFlag.set(false)
```
cquiroz: Should these actions be atomic? What happens if setting `abortFlag` fails. That would lean `stopFlag` set

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -65,8 +65,11 @@ class SeqexecCommandRoutes(auth: AuthenticationService, inputQueue: server.Event
       } yield resp
 
     case POST -> Root / obsId / stepId / "stop" as _ =>
-      // TODO call stop on the engine
-      Ok(s"Stop requested for $obsId on step $stepId")
+      for {
+        obs  <- \/.fromTryCatchNonFatal(new SPObservationID(obsId)).fold(e => Task.fail(e), Task.now)
```
cquiroz: Thanks for adding this one

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GmosControllerSim.scala
jluhrs: How could `set` they fail?

cquiroz: is this setting an epics channel's value?

jluhrs: No, `stopFlag` is an `AtomicInteger`. This is `GmosControllerSim`.

cquiroz: ok, in principle they should be both be changed at once, but this is only for simulation so it should be ok

PR #337, opened 2017-10-04
SEQNG-370: Pause button must be kept visible, and reflect if a pause was requested

cquiroz:
This PR lets the user to cancel a pause command and return to a running state

screenshot 2017-10-04 12 47 33


jluhrs:
I didn't find references to the `Pausing` state. How do you know a pausing is in process?

PR #343, opened 2017-10-10
SEQNG-380 CAR pause value is now properly processed.

jluhrs:
Added the paused state for EPICS commands. This PR includes only the changes in module ACM. I had to make some changes on `seqexec.server` to make it compile, but those changes have to be revisited when proper behavior for paused exposures is implemented.

### DIFF
--- a/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
+++ b/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
@@ -130,4 +130,18 @@ public final class CaCommandMonitorImpl implements CaCommandMonitor {
         }
     }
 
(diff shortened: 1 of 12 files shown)


modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
```diff
@@ -130,4 +130,18 @@ public synchronized void completeFailure(Exception cause) {
         }
     }
 
+    public synchronized void completePause() {
+        lock.lock();
```
cquiroz: Should the `unlock` be done on a `try/finally` block? in case some call throws?

jluhrs: `unlock` doesn't throw checked exceptions.

PR #346, opened 2017-10-11
SEQNG-379: Link abort/stop buttons on the UI

cquiroz:
Small update to the stop/abort buttons to update properly with the sequence state

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/StepsControlButtons.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/StepsControlButtons.scala
@@ -16,6 +16,8 @@ import edu.gemini.seqexec.model.operations._
 import edu.gemini.seqexec.web.client.actions.{RequestAbort, RequestStop}
 import edu.gemini.seqexec.web.client.circuit.SeqexecCircuit
 
+import scalaz.syntax.equal._
+
 /**
  * Contains the control buttons like stop/abort at the row level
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/StepsControlButtons.scala
```diff
@@ -32,10 +34,10 @@ object StepsControlButtons {
     Callback(SeqexecCircuit.dispatch(RequestAbort(id, stepId)))
 
   def handleStop(id: SequenceId, stepId: Int): ScalazReact.ReactST[CallbackTo, State, Unit] =
-    ST.retM(requestStop(id, stepId)) >> ST.mod(_.copy(stopRequested = true)).liftCB
+    ST.retM(requestStop(id, stepId)) >> ST.mod(_.copy(stopRequested = true, abortRequested = true)).liftCB
```
jluhrs: Why are both true?

cquiroz: This is local state, to disable the buttons until a response is heard from the backend. Both true means that `Stop` and `Abort` buttons will be disabled until the response of either is received

Basically it means you can't do `Abort` while `Stopping` and viceversa

jluhrs: Ah, ok.

PR #348, opened 2017-10-13
SEQNG-246: Use scalajs-bundler

cquiroz:
We have been loading javascript dependencies using `webjars` instead of directly from `npm`. This method is being deprecated in favor of using `scalajs-bundler` to load `npm` modules.

This has several benefits:
* Direct access to the latest versions
* Better downstream management of dependencies
* We can get a reduced version of semantic-ui modules

And the most important of all, this lets us use `react` components directly. They are only supported via `npm` modules

This PR has been very long as the intersection of `sbt` and `npm` is complicated. There are some possible improvements but they will be done later on when possible

The interaction with `semantic/jquery` took quite a bit of effort and this leads me to attempt to use `semantic/react` instead in the future

as a result of this PR several other PRs were sent to upstream projects so this could be considered a win for OSS 😄

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -122,28 +122,6 @@ lazy val edu_gemini_seqexec_web_server = project.in(file("modules/edu.gemini.seq
   )
   .dependsOn(edu_gemini_seqexec_web_shared_JVM, edu_gemini_seqexec_server, edu_gemini_web_server_common)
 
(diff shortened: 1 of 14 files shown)


cquiroz:
Tested on seqexec-gs-test

PR #349, opened 2017-10-16
Send stop and abort only if exposure is in progress.

jluhrs:


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -15,11 +16,14 @@ import scalaz.concurrent.Task
 import scalaz.stream.{Process, Sink, merge}
 
 package engine {
+
+
```
cquiroz: double empty line?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -242,6 +245,16 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     }
   }
 
+  // I cannot use a sealed trait as base, because I cannot have all systems in one source file (too big),
+  // so either I use an unchecked notation, or add a default case that throws an exception.
```
cquiroz: interesting problem, I think this solution is fine

PR #352, opened 2017-10-19
Optimize react in production

cquiroz:
This PR contains a small update to further minimize react.js in production reducing the overall lib size from 411k to 335k.

It also fixes a bug introduced with the update of `scalajs-bundler` to `0.9.0`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.log
 *.iml
 .DS_Store
+*.pem
 
 # sbt specific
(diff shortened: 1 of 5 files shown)


cquiroz:
The client side tests are causing troubles since we started using `scalajs-bundler` due to a much slower step of creating the test runner. Given there are only a few tests I'm disabling them until I manage to split the client module into two to make testing more reasonable

PR #357, opened 2017-10-30
SEQNG-397: Send observation state to the clients

cquiroz:
This PR calculates the current observation status on each step out of the executions and adds it to the state sent to the clients

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -297,12 +297,12 @@ object SeqexecEngine {
   def splitAfter[A](l: List[A])(p: (A => Boolean)): (List[A], List[A]) =
     l.splitAt(l.indexWhere(p) + 1)
 
-  def kindToResult(kind: ActionType): List[Resource] = kind match {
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -329,20 +329,50 @@ object SeqexecEngine {
...
+   * Overall pending status for a step
+   */
+  private def stepConfigStatus(step: StepAR): List[(Resource, ActionStatus)] =
     engine.Step.status(step) match {
       case StepState.Pending => pendingConfigStatus(step.executions)
       case _                 => configStatus(step.executions)
     }
 
+  protected[server] def observeStatus(executions: List[List[engine.Action \/ engine.Result]], configStatus: List[(Resource, ActionStatus)]): ActionStatus = {
+    def containsPartial(e: List[engine.Action \/ engine.Result]): Boolean =
```
cquiroz: This is a special case. A `Partial` is a `Result` but we should make the state of the Observation as `Running`
@jluhrs Do you think this is right. It is a bit weird that `Partial` is a `Result`

jluhrs: I agree that it is weird. I had a similar problem recently with this.

PR #367, opened 2017-11-14
SEQNG-414: Target name leakage

cquiroz:
I noted that we are displaying the target name to anonymous users through the step configuration. This PR will hide that information on the UI for anonymous users

It contains also improvements to the model removing some String usage

Here's a screenshot of a hidden element

screenshot 2017-11-13 23 23 03

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -164,8 +164,8 @@ object Model {
       eachViewL       ^|->  // each sequence on the queue
       obsNameL              // sequence's observation name
 
-    // Composite lens to find the target name
-    val sequenceTargetNameL: Traversal[SeqexecEvent, TargetName] =
+    // Composite lens to find the step config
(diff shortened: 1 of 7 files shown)


modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -128,19 +157,84 @@ object Model {
...
+    // Composite lens to find the target name on telescope
+    val telescopeTargetNameL: Traversal[SeqexecEvent, TargetName] =
+      sequenceConfigL                                 ^|-?  // configuration of the step
+      configTargetNameL(SystemName.telescope, "Base:name")       // on the configuration find the target name
 
     implicit val equal: Equal[SeqexecEvent] = Equal.equalA
   }
 
-  type SystemName = String
+  // The system name in ocs is a string but we can represent the important ones as an ADT
```
jluhrs: Is this one of those cases of types in ocs that are passed as String and then converted to other types in ocs3?

cquiroz: almost all in ocs is a string. This just makes it simpler to work with with some risk that we are missing a type

PR #380, opened 2017-11-24
SEQNG-402: Triggered observe in GMOS from IS, not DC, added endObserve notification

jluhrs:
GMOS require the endObserve notification to capture dtaZEnd, so I added it. Also, dhsZStart was not captured because I was using gm:dc:observe instead of gm:observe. That was fixed too.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -70,7 +70,8 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
 
     def notifyObserveStart: SeqAction[Unit] = otherSys.map(_.notifyObserveStart).sequenceU.map(_ => ())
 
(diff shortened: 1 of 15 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -70,7 +70,8 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
 
     def notifyObserveStart: SeqAction[Unit] = otherSys.map(_.notifyObserveStart).sequenceU.map(_ => ())
 
-    def notifyObserveEnd: SeqAction[Unit] = otherSys.map(_.notifyObserveEnd).sequenceU.map(_ => ())
+    // endObserve must be sent to the instrument too.
+    def notifyObserveEnd: SeqAction[Unit] = (inst +: otherSys).map(_.notifyObserveEnd).sequenceU.map(_ => ())
```
cquiroz: Why do we discard the result of `notifyObserveEnd`?

jluhrs: Because it is a list of Units

PR #381, opened 2017-11-24
SEQNG-403: Put reading of GCAL values for keywords in Tasks.

jluhrs:
Values were read on startup only. Now they should work.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gcal/GcalKeywordReader.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gcal/GcalKeywordReader.scala
@@ -28,11 +28,11 @@ object DummyGcalKeywordsReader extends GcalKeywordReader {
 
 object GcalKeywordsReaderImpl extends GcalKeywordReader {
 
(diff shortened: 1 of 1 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gcal/GcalKeywordReader.scala
```diff
@@ -28,11 +28,11 @@ object DummyGcalKeywordsReader extends GcalKeywordReader {
 
 object GcalKeywordsReaderImpl extends GcalKeywordReader {
 
-  def getDiffuser: SeqAction[Option[String]] = GcalEpics.instance.diffuser
+  def getDiffuser: SeqAction[Option[String]] = SeqAction(GcalEpics.instance.diffuser)
```
cquiroz: I don't get why the type of `getDiffuser` didn't change if you're wrapping it on a `SeqAction`

jluhrs: Because it was being implicitly converted before.

cquiroz: ah ok. this is much better

PR #385, opened 2017-11-27
SEQNG-422: Split settings into columns

cquiroz:
It seems that we have many settings, probably too many to display in a single column. This PR adds some refactoring and puts the offset and type onto their own columns. More columns will be added as part of SEQNG-367

screenshot 2017-11-27 15 11 17


jluhrs:
What about steps where the offset is not shown, like Darks?

cquiroz:
There is no diferenece at the moment for different types. It would show offsets 0,0

PR #386, opened 2017-11-28
SEQNG-325: Support level filtering on the client

cquiroz:
On this PR the log table is color coded according to the level and we let the user filter levels with a few controls. This can be better seen on the video below

https://monosnap.com/file/4drLagEJAzukV2qID0ekGABzLLzBa4#

Note that the filters are reset every time we reload. Should we make them persistent?


arturog8m:
Looks nice. If we were going to make the filters persistent, I would do that as a property associated to the user, so she gets her settings whenever she logs in. However, I think that's probably not a high priority task unless is almost zero effort. 

I wonder if the filtering can be done with standard checkboxes? I believe that would be even more clear to the user, instead of the sliders. Opinions?

cquiroz:
Yes, keeping it persistent should be on a user basis. I'd say it is lower priority and to do it right we need to add some sort of db which makes it a lot bigger task

An intermediate solution is to keep it persistent per user on a browser db. That is much simpler but it won't follow you to other browsers

PR #388, opened 2017-11-29
SEQNG-423: Display guiding per step

cquiroz:
This PR adds a column to show the guiding state as on/off. The icons are up for discussion though I agreed on these ones on discussion with science

As part of this PR I moved the `SeqexecEvent` class hierarchy to another file as the `Model` object is getting very large

Here's a screenshot

screenshot 2017-11-29 14 59 04


jluhrs:
I think the icon makes sense.

PR #391, opened 2017-12-05
SEQNG-425: Show coadds x exptime

cquiroz:
This PR adds a column with coadds and exposure time. The formatting depends on the instrument with Gmos using no decimals, F2, GNIRS and NIFS with 1 decimal and the rest with 2 decimals

coadds is shown only for instrument supporting it

Some screenshots:

F2
screenshot 2017-12-04 21 15 00

GMOS:
screenshot 2017-12-04 21 15 08

and finally an example with coadds. This is to demonstrate the code only, GMOS won't show the coadds in general
Fake GMOS with coadds:
screenshot 2017-12-04 20 56 18


andrewwstephens:
That looks good!  Alternatively, rather then hard-coding the display of coadds for some instruments, we could just display the coadds if coadds > 1.

PR #392, opened 2017-12-05
SENQG-420: Support copying log lines to clipboard

cquiroz:
This is a facility to support copying log lines to the clipboard. An extra button was added to the last column that can be used to copy the line to the clipboard

Do you think this is useful? Should we support selecting n lines and copy them all at once?

Here it is the feature in action
https://monosnap.com/file/MD6y69T6T9td4kHYJBfQXfy73rEUbl#

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/LogArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/LogArea.scala
@@ -8,6 +8,7 @@ import diode.react.ModelProxy
 import edu.gemini.seqexec.model.Model.{SeqexecSite, ServerLogLevel}
 import edu.gemini.seqexec.model.events.SeqexecEvent.ServerLogMessage
 import edu.gemini.seqexec.web.client.semanticui.elements.checkbox.Checkbox
+import edu.gemini.seqexec.web.client.semanticui.elements.icon.Icon.IconCopy
 import edu.gemini.seqexec.web.client.model.GlobalLog
 import edu.gemini.web.common.FixedLengthBuffer
 import japgolly.scalajs.react._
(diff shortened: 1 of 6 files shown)


arturog8m:
Nice! And yes, copying multiple lines would be useful. I would put that in the list of things to do for future enhancements.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/LogArea.scala
```diff
@@ -76,17 +94,30 @@ object LogArea {
...
+      // TODO use local tiems
+      val toCsv = s"${row.timestamp}, ${row.level}, ${row.msg}"
+      CopyLogToClipboard(toCsv)
+    }
+
     val columns = List(
       Column(Column.props(200, "timestamp", label = "Timestamp", disableSort = true)),
       Column(Column.props(80, "level", label = "Level", disableSort = true)),
-      Column(Column.props(size.width.toInt - 200 - 80, "msg", label = "Message", disableSort = true, flexGrow = 1))
+      Column(Column.props(size.width.toInt - 200 - 80 - 37, "msg", label = "Message", disableSort = true)),
```
arturog8m: What are all those magic numbers?

jluhrs:
Yes. Usually you need to copy several lines to get a sequence of events.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/LogArea.scala
cquiroz: So the column's widths are defined in pixel and the message column get the rest to fill width. I suppose it would be better to have them as constants. I'll change that

andrewwstephens:
Does the copy include timestamps?

cquiroz:
Yes, it is the same than in the table in CSV format like;:

2017-12-05T12:38:34.801Z, INFO, Engine: Polling current state

cquiroz:
Note that I need to combine this PR with #393 to get the timestamp in local time

PR #393, opened 2017-12-05
SEQNG-418: Display log times on local time

cquiroz:
This PR updates the log table to display the times in local (as in the server's local) time

here's an example. Feel free to comment especially about the format, e.g. should we use the month name or number? Should we have miliseconds?

screenshot 2017-12-05 13 45 20


arturog8m:
I think the format you suggest is good. W/respect to showing ms - I don't know - as the internal clock accuracy on the server is within a few milliseconds of NTP, I would show just one decimal place there if we are going to show it at all.

cquiroz:
The formatter doesn't have a simple way to display a single digit. I can approximate the seconds to make it always make it rounded though

PR #394, opened 2017-12-05
SEQNG-439: Give a placeholder target name for calibrations

cquiroz:
We are reading the target name from the target value of the first science step of a sequence. This makes daycalibrations to look like they don't have a target. Which normally doesn't matter but it looks a bit strange.

This PR changes that to use the word `DAYCAL` for those cases

Here's a screenshot:
screenshot 2017-12-05 14 34 10

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
@@ -96,6 +96,7 @@ object Model {
   type StepId = Int
   type ObservationName = String
   type TargetName = String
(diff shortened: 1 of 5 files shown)


arturog8m:
Perhaps an icon would look nicer? I agree that this is an improvement.

cquiroz:
An icon is a good idea, let me try finding something suitable

andrewwstephens:
That looks good!  I'm partial to text, as it shrinks nicely for those of us who want a more compact view.

andrewwstephens:
Would "Daytime Calibration" fit?

cquiroz:
Replaced by `Daytime Calibration` in bold/italics

screenshot 2017-12-05 23 13 55

PR #398, opened 2017-12-07
Action pause and resume

jluhrs:
This PR modifies the definition of Action, adding a state. Execution is no longer a collection of Actions and Results, but a collection of Actions in (possibly) different states.
Actions can now end in a Paused state. In that case the Sequence is considered as still running. A new event was introduced, ActionResume, which provide a Task that is run as the continuation of the paused Action.

This will be used to implement the observe pause/resume. As I envision it, the pause will directly send a pause command to the instrument, causing the currently running observe Action to end with a Pause result. The resume will put an ActionResume. Its Task will send the resume to the instrument, and wait for the completion of the observation.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
@@ -3,7 +3,6 @@
 
 package edu.gemini.seqexec.engine
 
-import edu.gemini.seqexec.model.ActionType
 import edu.gemini.seqexec.model.Model.Resource
 
 import scalaz._
@@ -14,30 +13,24 @@ import Scalaz._
   * information about which `Action`s have been completed.
   *
   */
-final case class Execution(execution: List[Action \/ Result]) {
+final case class Execution(execution: List[Action]) {
 
   import Execution._
 
   val isEmpty: Boolean = execution.isEmpty
 
-  val actions: Actions = {
-    def lefts[L, R](xs: List[L \/ R]): List[L] = xs.collect { case -\/(l) => l }
-    lefts(execution)
-  }
+  val actions: List[Action] = execution.filter(_.state === Action.Idle)
 
-  val results: Results = {
-    def rights[L, R](xs: List[L \/ R]): List[R] = xs.collect { case \/-(r) => r }
-    rights(execution)
-  }
+  val results: List[Action] = execution.filter(Action.finished)
 
   /**
     * Calculate `Execution` `Status` based on the underlying `Action`s.
     *
     */
   def status: Status =
-    if (execution.forall(_.isLeft)) Status.Waiting
+    if (execution.forall(_.state === Action.Idle)) Status.Waiting
     // Empty execution is handled here
-    else if (execution.all(finished)) Status.Completed
+    else if (finished(this)) Status.Completed
     else if (isEmpty) Status.Completed
     else Status.Running
 
@@ -45,8 +38,8 @@ final case class Execution(execution: List[Action \/ Result]) {
     * Obtain the resulting `Execution` only if all actions have been completed.
     *
     */
-  val uncurrentify: Option[Results] =
-    (execution.nonEmpty && execution.all(_.isRight)).option(results)
+  val uncurrentify: Option[Actions] =
+    (execution.nonEmpty && finished(this)).option(results)
 
   /**
     * Set the `Result` for the given `Action` index in `Current`.
@@ -54,7 +47,9 @@ final case class Execution(execution: List[Action \/ Result]) {
     * If the index doesn't exist, `Current` is returned unmodified.
     */
   def mark(i: Int)(r: Result): Execution =
-    Execution(PLens.listNthPLens(i).setOr(execution, r.right, execution))
+    Execution(PLens.listNthPLens[Action](i).modg(a => a.copy(state = actionStateFromResult(r)), execution).getOrElse(execution))
+
+  def start(i: Int): Execution = Execution(PLens.listNthPLens[Action](i).modg(a => a.copy(state = Action.Started), execution).getOrElse(execution))
 }
 
 object Execution {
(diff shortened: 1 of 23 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
```diff
@@ -64,71 +59,54 @@ object Execution {
...
-
-  def finished(ar: Action \/ Result): Boolean =
-    ar match {
-      case (-\/(_)) => false
-      case (\/-(r)) => r match {
-        case Result.Partial(_, _) => false
-        case _                    => true
-      }
-    }
+    (as.nonEmpty && as.forall((_.state === Action.Idle))).option(Execution(as))
```
cquiroz: Will `forall` be false if empty? If so the first check is not needed

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -17,57 +17,42 @@ import monocle.macros.GenLens
...
+      case _                   => false
     }
 
     // Find an error in the Step
-    step.findLeft(Execution.errored).flatMap(
-      // Get the message if there is one
-      _.fold(_ => None, _.errMsg)
+    step.executions.flatten.find(Action.errored).flatMap { x => x.state match {
+      case Action.Failed(Result.Error(msg)) => msg.some
+      case _ => None
```
cquiroz: alignment?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -134,10 +119,9 @@ object Step {
         breakpoint,
         skip,
         // TODO: Functor composition?
-        done.map(_.map(_.right)) ++
+        done ++
```
cquiroz: Maybe this can be on the same line

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -134,10 +119,9 @@ object Step {
         breakpoint,
         skip,
         // TODO: Functor composition?
```
cquiroz: I guess this TODO doesn't apply anymore

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -22,7 +23,52 @@ package engine {
...
   }
   final case class ActionMetadata(conditions: Conditions, operator: Option[Operator], observer: Option[Observer])
-  final case class Action(kind: ActionType, gen: Kleisli[Task, ActionMetadata, Result])
+  object ActionMetadata {
+    val default: ActionMetadata = ActionMetadata(Conditions.default, None, None)
+  }
+
+  final case class Action(
+    kind: ActionType,
+    gen: ActionGen,
```
cquiroz: `ActionGen`?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -22,7 +23,52 @@ package engine {
...
+
+    final case class Failed(e: Error) extends ActionState
+
+    def errored(ar: Action): Boolean = ar.state match {
+      case Action.Failed(_) => true
+      case _ => false
+    }
+
+    def finished(ar: Action): Boolean = ar.state match {
+      case Action.Failed(_) => true
```
cquiroz: Is it so that it's finished if it failed?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -223,27 +270,28 @@ package object engine {
...
-    def act(t: (Action, Int), cx: ActionMetadata): Process[Task, Event] = t match {
-      case (Action(_, gen), i) =>
-        Process.eval(gen(cx)).flatMap {
-          case r@Result.OK(_)         => Process(completed(id, i, r))
-          case r@Result.Partial(_, c) => Process(partial(id, i, r)) ++ act((c, i), cx)
-          case e@Result.Error(_, _)   => Process(failed(id, i, e))
-        }
-    }
+  // Send the expected event when the `Action` is executed
+  // It doesn't catch run time exceptions. If desired, the Action as to do it itself.
```
cquiroz: typo: `the Action has`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -326,12 +386,14 @@ package object engine {
       case Poll                        => Logger.debug("Engine: Polling current state")
       case GetState(f)                 => getState(f)
       case ActionStop(id, f)           => Logger.debug("Engine: Action stop requested") *> actionStop(id, f)
+      case ActionResume(id, i, cont)   => Logger.debug("Engine: Action resume requested") *> actionResume(id, i, cont)
```
cquiroz: Will these generate front side events too?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -337,36 +347,38 @@ object SeqexecEngine {
...
       case StepState.Pending => pendingConfigStatus(step.executions)
       case _                 => configStatus(step.executions)
     }
 
-  protected[server] def observeStatus(executions: List[List[engine.Action \/ engine.Result]], configStatus: List[(Resource, ActionStatus)]): ActionStatus = {
-    def containsPartial(e: List[engine.Action \/ engine.Result]): Boolean =
-      e.rights.exists {
-        case Partial(FileIdAllocated(_), _) => true
-        case _                              => false
+  //protected[server] def observeStatus(executions: List[List[engine.Action]]): ActionStatus = executions.flatten.find(x => x.kind === ActionType.Observe).map(x => actionStateToStatus(x.state)).getOrElse(ActionStatus.Pending)
```
cquiroz: Should we delete this line?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTableContainer.scala
```diff
@@ -135,12 +135,16 @@ object StepsTableContainer {
       case ActionStatus.Pending   => "gray"
       case ActionStatus.Running   => "yellow"
       case ActionStatus.Completed => "green"
+      case ActionStatus.Paused    => "orange"
```
cquiroz: nice, I can delegate UI tasks now 😄

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
jluhrs: `> List[Int]().forall(_>1)
res3: Boolean = true`
The reason is that `forall` is the complement of `exists`. It is the same for `Option`. `None.forall(f)` is always `true`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
jluhrs: I. Hate. You.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jluhrs: It is an alias for `Kleisli[Task, ActionMetadata, Result]`, the function that generates the `Task` that is the body of the `Action`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jluhrs: Yes. In this context, `finished` means "It was running, but it is not running anymore." It is different from completed.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
jluhrs: Eventually, yes. It has to cause a refresh of the step display.

PR #400, opened 2017-12-18
Add support for scalameta and scalafix

cquiroz:
This PR adds scalameta and scalafix to the build. It dosen't have any direct benefits but it lets me use the LSP for scala provided by scalameta

@tpolecat @swalker2m WDYT?

it seems to make the build a tad slower but not too bad

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -3,6 +3,7 @@ import sbt.Keys._
 import sbt._
 import de.heikoseeberger.sbtheader.HeaderPlugin.autoImport._
 import wartremover.WartRemover.autoImport._
+import scalafix.sbt.ScalafixPlugin.autoImport._
 
 /**
   * Define tasks and settings used by module definitions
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -70,12 +70,14 @@ lazy val edu_gemini_seqexec_web = project.in(file("modules/edu.gemini.seqexec.we
   .settings(commonSettings: _*)
   .enablePlugins(AutomateHeaderPlugin)
   .enablePlugins(GitBranchPrompt)
+  .disablePlugins(RevolverPlugin)
```
cquiroz: Thanks to sbt magic, the revolver plugin gets confused when adding scalafix 🤷‍♂️

cquiroz:
they both feed an atom plugin, there is a vscode one too
https://github.com/laughedelic/atom-ide-scala

Anyway i'm going to sit on this PR and keep using it locally until i'm sure it's reliable enough

cquiroz:
This has been obsoleted by scalameta developments

PR #403, opened 2017-12-19
SEQNG-450 Changed monitoring procedure for observe.

jluhrs:
Before, the monitoring would check the gm:apply and gm:dc:observeC fields to check when the command started and was completed (or paused, or failed). It also checked the consistency of the CLID fields, and would treat a change on the CLID as an error. In other words, it treated the observe as an atomic command; a new command issued while the observe was in progress was treated as an error. This caused problems for issuing commands abort, stop or pause.
The solution implemented is to treat the observe command as a command with a delayed status. The command execution is monitored normally with the records gm:apply and gm:applyC, and then the value of gm:dc:observeC is monitored to find when the observation finishes.

### DIFF
--- a/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
+++ b/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
@@ -25,8 +25,7 @@ import java.util.logging.Logger;
  */
 public class CaObserveSenderImpl implements CaApplySender {
 
-    private static final Logger LOG = Logger.getLogger(CaObserveSenderImpl.class
-            .getName());
+    private static final Logger LOG = Logger.getLogger(CaObserveSenderImpl.class.getName());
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -25,8 +25,7 @@
  */
 public class CaObserveSenderImpl implements CaApplySender {
```
cquiroz: Just a general question. Would it be worth converting these to scala?

jluhrs: I don't know. At some point we have to deal with Java code, because of JCA.

jluhrs:
It is a bit complicated to add tests, because they need to interact with an EPICS system. Right now we have such system, implemented in class `TestSimulator`. It lacks an implementation of `apply`s records.

PR #404, opened 2017-12-20
SEQNG-451 Produce a SeqexecUpdated event when actions fail.

jluhrs:
Fix for problem with step not being updated in client when action failed and it was a single action.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
@@ -315,7 +315,7 @@ package object engine {
 
   def partialResult[R<:PartialVal](id: Sequence.Id, i: Int, p: Result.Partial[R]): HandleP[Unit] = modifyS(id)(_.mark(i)(p))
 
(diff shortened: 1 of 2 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
```diff
@@ -39,7 +39,10 @@ object Event {
   def actionResume(id: Sequence.Id, i: Int, c: Task[Result]): Event = EventUser(ActionResume(id, i, c))
   def logMsg(msg: String): Event = EventUser(Log(msg))
 
-  def failed(id: Sequence.Id, i: Int, e: Result.Error): Event = EventSystem(Failed(id, i, e))
+  def failed(id: Sequence.Id, i: Int, e: Result.Error): Event = {
+    System.out.println(s"Error in sequnece $id, step $i: ${e.msg}")
```
cquiroz: Should this go into the log instead?

cquiroz: There is a typo on `sequence`

PR #406, opened 2017-12-21
SEQNG-353 Implemented observe resume command for GMOS.

jluhrs:
Send "continue" command to GMOS when the user presses the resume button. It checks that the sequence is running and the observe Action is active.
The resumed observation monitoring is the same that for non resumed observations, including the detection of  stop or abort commands.
The "pause" and "continue" command were tested with the real instrument.

### DIFF
--- a/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
+++ b/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
@@ -451,7 +451,7 @@ public class CaObserveSenderImpl implements CaApplySender {
                         case IDLE: return new WaitObserveStart(cm);
                         case BUSY: return new WaitObserveCompletion(cm, getStopMark(), getAbortMark());
                         case ERROR: {
-                            failCommandWithCarError(cm);
+                            failCommandWithObserveCarError(cm);
                             return IdleState;
                         }
                         case PAUSED: {
@@ -520,7 +520,7 @@ public class CaObserveSenderImpl implements CaApplySender {
                     return new WaitObserveCompletion(cm, getStopMark(), getAbortMark());
                 }
                 case ERROR: {
-                    failCommandWithCarError(cm);
+                    failCommandWithObserveCarError(cm);
                     return IdleState;
                 }
                 case PAUSED: {
@@ -574,7 +574,7 @@ public class CaObserveSenderImpl implements CaApplySender {
                     return IdleState;
                 }
                 case ERROR:{
-                    failCommandWithCarError(cm);
+                    failCommandWithObserveCarError(cm);
                     return IdleState;
                 }
                 case PAUSED: {
@@ -694,7 +694,7 @@ public class CaObserveSenderImpl implements CaApplySender {
                     return IdleState;
                 }
                 case ERROR:{
-                    failCommandWithCarError(cm);
+                    failCommandWithObserveCarError(cm);
                     return IdleState;
                 }
                 case PAUSED: {
@@ -822,6 +822,21 @@ public class CaObserveSenderImpl implements CaApplySender {
             String msg = null;
             try {
                 msg = car.getOmssValue();
+            } catch (CAException | TimeoutException e) {
+                LOG.warning(e.getMessage());
+            }
+            cm.completeFailure(new CaCommandError(msg));
+        } );
+    }
+
+    private void failCommandWithObserveCarError(final CaCommandMonitorImpl cm) {
+        // I found that if I try to read OMSS or MESS from the same thread that
+        // is processing a channel notifications, the reads fails with a
+        // timeout. But it works if the read is done later from another thread.
+        executor.execute(() -> {
+            String msg = null;
+            try {
+                msg = observeCar.getOmssValue();
             } catch (CAException | TimeoutException e) {
                 LOG.warning(e.getMessage());
             }
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
@@ -77,7 +77,7 @@ object Execution {
   def actionStateFromResult(r: Result): (Action.State => Action.State) = s => r match {
     case Result.OK(x)         => s.copy(runState = Action.Completed(x))
     case Result.Partial(x, _) => s.copy(partials = x :: s.partials )
-    case Result.Paused        => s.copy(runState = Action.Paused)
+    case Result.Paused(c)     => s.copy(runState = Action.Paused(c))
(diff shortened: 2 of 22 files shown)


modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -829,6 +829,21 @@ private void failCommandWithCarError(final CaCommandMonitorImpl cm) {
         } );
     }
 
+    private void failCommandWithObserveCarError(final CaCommandMonitorImpl cm) {
+        // I found that if I try to read OMSS or MESS from the same thread that
```
cquiroz: Do you thik that's happening at CAS level?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Execution.scala
```diff
@@ -77,7 +77,7 @@ object Execution {
   def actionStateFromResult(r: Result): (Action.State => Action.State) = s => r match {
     case Result.OK(x)         => s.copy(runState = Action.Completed(x))
     case Result.Partial(x, _) => s.copy(partials = x :: s.partials )
-    case Result.Paused        => s.copy(runState = Action.Paused)
+    case Result.Paused(c)        => s.copy(runState = Action.Paused(c))
```
cquiroz: hmm, alignment?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/UserEvent.scala
```diff
@@ -47,6 +47,11 @@ case object Poll extends UserEvent {
 final case class GetState(f: (Engine.State) => Task[Option[Process[Task, Event]]]) extends UserEvent {
   val user: Option[UserDetails] = None
 }
+// Generic event to put a function in the main Process process, which takes an
+// action depending on the current state of a specific sequence
+final case class GetSeqState(id: Sequence.Id, f: (Sequence.State) => Option[Process[Task, Event]]) extends UserEvent {
```
cquiroz: You may not need the parenthesis around `Sequence.State`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -18,6 +18,8 @@ import scalaz.stream.{Process, Sink, merge}
 
 package engine {
 
+  import edu.gemini.seqexec.engine.Result.PauseContext
```
cquiroz: i think this import is redundant

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -172,7 +172,7 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
             skip = false,
             List(
               List(fromTask(ActionType.Undefined,
-              Task(Result.Paused)))
+              Task(Result.Paused(new Result.PauseContext {} ))))
```
cquiroz: Why are we using an anonymous class here? Could we create a class implementing `PauseContext`?

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -188,7 +188,10 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
   }
 
   "action state" should "change to Paused if output is Paused" in {
-    assert(actionPause.map(_.sequences(seqId).current.execution.forall(_.state.runState === Action.Paused)).getOrElse(false))
+    assert(actionPause.map(_.sequences(seqId).current.execution.forall{_.state.runState match {
```
cquiroz: the match for `Action.Paused` seems a common idiom. Perhaps it could be a method to do `runState.isPaused`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -83,24 +82,29 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
...
         headers(ctx).reverseMap(_.sendAfter(id, inst.dhsInstrumentName)).sequenceU *>
         closeImage(id, systems.dhs) *>
         SeqAction.fail(SeqexecFailure.Execution(s"Observation $id stopped by user."))
-      case ObserveCommand.Aborted => sendObservationAborted(obsId, id) *>
+      val abortTail: SeqAction[Result] = sendObservationAborted(obsId, id) *>
         SeqAction.fail(SeqexecFailure.Execution(s"Observation $id aborted by user."))
-      //TODO: Implement pause handling
-      case ObserveCommand.Paused  => SeqAction(Result.Paused)
+
+      r match {
```
cquiroz: nice

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -253,6 +257,26 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     } )
   )
 
+  def resumeObserve(seqId: Sequence.Id)(seqState: Sequence.State): Option[Process[Task, Event]] = {
+    val observeIndex: Option[(ObserveContext, Int)] = seqState.current.execution.zipWithIndex.find(_._1.kind === ActionType.Observe).flatMap{ x => { x match {
```
cquiroz: i think the `flatMap { x => x match` can be reduced to `flatMap { case ...`

modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
jluhrs: I am not sure. This comment is from like 3 years ago.

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
jluhrs: Because it is only for tests.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
jluhrs: Oh, yes. I started with that, but then expanded things to catch a syntax error, and forgot to go back.

PR #410, opened 2017-12-22
SEQNG-455 Modified Step and Sequence state.

jluhrs:
The biggest change was in the Sequence states.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
@@ -204,6 +204,25 @@ object Sequence {
         case Final(st, _)  => Final(st, s)
       })
 
+    def isRunning(st: State): Boolean = SequenceState.isRunning(st.status)
+
+    def userStopRequested(st: State): Boolean = SequenceState.userStopRequested(st.status)
+
+    def anyStopRequested(st: State): Boolean = st.status match {
+      case SequenceState.Running(u, i) => u || i
+      case _                           => false
+    }
+
+    def userStopSet(v: Boolean): State => State = status.modify{
+      case r@SequenceState.Running(_, _) => r.copy(userStop = v)
+      case r                             => r
+    }
+
+    def internalStopSet(v: Boolean): State => State = status.modify{
+      case r@SequenceState.Running(_, _) => r.copy(internalStop = v)
+      case r                             => r
+    }
+
     /**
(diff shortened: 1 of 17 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -204,6 +204,25 @@ object Sequence {
         case Final(st, _)  => Final(st, s)
       })
 
+    def isRunning(st: State): Boolean = SequenceState.isRunning(st.status)
```
jluhrs: Added several utility functions here.

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -186,7 +187,7 @@ object Model {
     }
     case object Completed extends StepState
     case object Skipped extends StepState
-    final case class Error(msg: String) extends StepState {
+    final case class Failed(msg: String) extends StepState {
```
jluhrs: Changed to `Failed` to be consistent.

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -243,18 +244,35 @@ object Model {
...
-    case object Stopping          extends SequenceState
     case object Idle              extends SequenceState
-    case object Paused            extends SequenceState
-    final case class Error(msg: String) extends SequenceState
+    case object Stopped           extends SequenceState
+    final case class Running(userStop: Boolean, internalStop: Boolean) extends SequenceState
+    object Running {
+      val init: Running = Running(false, false)
+    }
+    final case class Failed(msg: String) extends SequenceState
```
jluhrs: Changed to `Failed` to be consistent.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
cquiroz: Maybe these could be converted to `ops` via an implict class, but it's low prio

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
cquiroz: 👍

PR #411, opened 2017-12-26
Generated size reduction

cquiroz:
I've been struggling with slow compilation times and very large generated js size. In particular the size of the unoptimized generated js means we spend a long time compiling  and creates issues to run it quickly on firefox. 

After some research I found one issue with how the macros and implicits for boopickle can lead to an explosion of generated code. Turns out that explicitly defining a couple of missing implicits and changing the order are helpful.

Thanks to this we went from:

| subject  |Before |After|
| ------------- | ------------- |---------|
| Generated classes  | 12K  | 200 |
| Unoptimized JS  | 39M  | 12M |
| Optimized JS | 5.9M | 3.2M |

This PR would alleviate a bit my problems but there are still optimizations to do on this area

A moral of the story is that simple details like the order of defining implicits especially when using macros can have bad effects even if the code compiles right

Also travis times has been reduced in about 2 minutesa

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/ModelBooPicklers.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/ModelBooPicklers.scala
@@ -4,6 +4,7 @@
 package edu.gemini.seqexec.model
 
 import boopickle.Default._
+// import boopickle.PicklerHelper
 import edu.gemini.seqexec.model.Model._
 import edu.gemini.seqexec.model.events.SeqexecEvent
 import edu.gemini.seqexec.model.events.SeqexecEvent._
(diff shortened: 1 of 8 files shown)


modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/ModelBooPicklers.scala
```diff
@@ -15,8 +15,50 @@ import java.time.Instant
...
   * them explicitly
   */
-@SuppressWarnings(Array("org.wartremover.warts.ImplicitParameter", "org.wartremover.warts.NonUnitStatements", "org.wartremover.warts.Throw", "org.wartremover.warts.OptionPartial"))
-trait ModelBooPicklers {
+@SuppressWarnings(Array("org.wartremover.warts.Equals", "org.wartremover.warts.PublicInference", "org.wartremover.warts.ImplicitParameter", "org.wartremover.warts.NonUnitStatements", "org.wartremover.warts.Throw", "org.wartremover.warts.OptionPartial"))
+object ModelBooPicklers {
+  //**********************
+  // IMPORTANT The order of the picklers is very relevant to the generated size
+  // add them with care
+  //**********************
```
swalker2m: Any guidelines on how to order them correctly?  For the uninitiated (like myself) it's difficult to spot the problem.

swalker2m:
That's a very impressive reduction in size, by the way!

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/ModelBooPicklers.scala
cquiroz: Maybe I should add a comment about it. Essentially the object tree should be in the right order, with the subcomponents classes first (Like in here WaterVapor) and the composites (like sequence later). The issue is that if you don't do that, the automatic derivation kicks in instead and it goes many times leading to this explosion of classes

cquiroz:
I guess we should try to avoid macros. Having said that, in this case writing the picklers by hand is not very simple so I'm happy to keep them

PR #417, opened 2018-01-04
SEQNG-437: Shorten web page

cquiroz:
There is a request to make the page fit on the console ops without a scroll bar. this PR does it by manipulating the css through some magic and reducing the height of the log area. This is how it looks on the consoles:

screenshot 2018-01-03 14 09 58

And on the emulator with some notes:
screenshot 2018-01-04 17 17 31

This change has a drawback on larger screen as it wastes space. So the central part with the steps will grow and shrink (to some extent) on window changes. This can be appreciated on the following video

https://monosnap.com/file/0UVbfezmevwrpOOsdj9b6FzRPjyZiZ

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/SeqexecStyles.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/SeqexecStyles.scala
@@ -94,8 +94,13 @@ object SeqexecStyles extends scalacss.StyleSheet.Inline {
   )
 
   val stepsListPane: StyleA = style (
-    maxHeight(24.3.em),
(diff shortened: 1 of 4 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/SeqexecStyles.scala
```diff
@@ -94,8 +94,18 @@ object SeqexecStyles extends scalacss.StyleSheet.Inline {
   )
 
   val stepsListPane: StyleA = style (
-    maxHeight(24.3.em),
-    minHeight(24.3.em)
+    (maxHeight :=! "calc(100vh - 52.5em)"),
```
cquiroz: Dark magic!

andrewwstephens:
+1 Looks good!

PR #418, opened 2018-01-05
Set paused state for Observations in client model

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -373,35 +373,20 @@ object SeqexecEngine {
...
-      executions.filter { e =>
-        val (a, r) = separateActions(e).bimap(_.map(_.kind), _.map(_.kind))
-        a.contains(ActionType.Observe) || r.contains(ActionType.Observe)
-      }.map {
-        case e if containsObserve(e) => ActionStatus.Completed
-        case _                       => ActionStatus.Running
-      }.headOption.getOrElse(ActionStatus.Pending)
-    } else ActionStatus.Pending
-  }
+  protected[server] def observeStatus(executions: List[List[engine.Action]]): ActionStatus =
```
jluhrs: The previous change (that added an explicit state to Action) really simplified this.

cquiroz:
This is looking would. it seems the step state doesn't change, right?

jluhrs:
That's right. If an exposure is paused, the step is considered still active. In that state the exposure is said to be "on-hold", which I think conveys better the idea that it is waiting to be resumed.

PR #420, opened 2018-01-09
Implemented timeout for GMOS and F2 exposure

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
```diff
@@ -48,6 +49,8 @@ trait EpicsCommand {
       cs.map(_.mark().right).getOrElse(SeqexecFailure.Unexpected("Unable to mark command.").left)
     })
   )
+
+  def setTimeout(t: Time): SeqAction[Unit] = EpicsUtil.setTimeout(cs.map(_.getApplySender), t)
```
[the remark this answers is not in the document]
cquiroz: I see `Time` comes from `squants`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/flamingos2/Flamingos2.scala
```diff
@@ -34,13 +35,15 @@ final case class Flamingos2(f2Controller: Flamingos2Controller) extends Instrume
...
-    fileId => f2Controller.observe(fileId).map(_ => ObserveCommand.Success)
+    fileId => f2Controller.observe(fileId, calcObserveTimeout(config)).map(_ => ObserveCommand.Success)
   }
 
   override def configure(config: Config): SeqAction[ConfigResult] =
     fromSequenceConfig(config).flatMap(f2Controller.applyConfig).map(_ => ConfigResult(this))
 
   override def notifyObserveEnd: SeqAction[Unit] = f2Controller.endObserve
+
+  override def calcObserveTimeout(config: Config) = config.extract(OBSERVE_KEY / EXPOSURE_TIME_PROP).as[java.lang.Double].map(x => Seconds(x.toDouble)).getOrElse(Seconds(360))
```
cquiroz: I thought `EXPOSURE_TIME_PROP` would be the requested exposure time. Does this mean the timeout is the same as the requested exposure time?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -282,4 +291,8 @@ object GmosControllerEpics {
     case InBeam    => "IN-BEAM"
   }
 
+  val defaultTimeout: Time = Seconds(60)
```
cquiroz: Usually constants have capitalized casing, like `DefaultTimeout`

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/flamingos2/Flamingos2.scala
jluhrs: No. I then add a readout timeout. But now I see I should have added that value here to make it consistent with the function name.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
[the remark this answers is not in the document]
jluhrs: I am on the fence here. I thought it was better to use squants.Time, but then the original implementation of GMOS uses Duration for the exposure time. And the timeout is then used to, well, implement a timeout, with functions that use Duration or something similar.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: OK, I'll change it.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/flamingos2/Flamingos2.scala
jluhrs: In second thought, what I should change is the name of the function, because the time added to calculate the timeout depends on the underline Controller.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -291,8 +291,8 @@ object GmosControllerEpics {
     case InBeam    => "IN-BEAM"
   }
 
-  val defaultTimeout: Time = Seconds(60)
-  val readoutTimeout: Time = Seconds(300)
-  val configTimeout: Time = Seconds(600)
+  val DefaultTimeout: Time = Seconds(60)
+  val ReadoutTimeout: Time = Seconds(300)
+  val ConfigTimeout: Time = Seconds(600)
```
andrewwstephens: Does this mean that the instrument configuration timeout is 10 minutes?

PR #425, opened 2018-01-11
SEQNG-472 Removed breakpoint control from first runnable step in client.

jluhrs:


modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/Model.scala
```diff
@@ -234,8 +234,8 @@ object Model {
 
       def file: Option[String] = None
 
-      def canSetBreakpoint: Boolean = s.status match {
-        case StepState.Pending | StepState.Skipped | StepState.Paused => s.id > 0
+      def canSetBreakpoint(i: Int, firstRunnable: Int): Boolean = s.status match {
+        case StepState.Pending | StepState.Skipped | StepState.Paused => i > firstRunnable
```
jluhrs: It shouldn't rely on `s.id` being the position in the list, or even it being an Int. That is why I now pass the step position as `i`

PR #428, opened 2018-01-12
SEQNG-472 Separated built-in ROI configuration for GMOS North and South.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gmos/GmosNorthControllerEpics.scala
```diff
@@ -92,6 +94,14 @@ object GmosNorthEncoders extends GmosControllerEpics.Encoders[NorthTypes] {
     case Disperser.B600_G5303  => "B600+_G5303"
     case Disperser.R150_G5306  => "R150+_G5306"
   }
+
+  override val builtInROI: EncodeEpicsValue[BuiltinROI, Option[ROIValues]] = EncodeEpicsValue {
+    case BuiltinROI.FULL_FRAME       => ROIValues.fromInt(xStart = 1, xSize = 6144, yStart = 1, ySize = 4224)
```
cquiroz: Given that these are always know to work you could do an `unsafeFromInt: ROIValues` method and call it that way. I don't know how much work it is to carrry the `Option` up the stack

jluhrs: The Option is not really that much work.

PR #429, opened 2018-01-15
SEQNG-476 Force MASKNAME to None if FPU is OUT-OF-BEAM.

jluhrs:
The value for MASKNAME is the current mask, or None, depending on the location of the FPU.


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gmos/GmosHeader.scala
```diff
@@ -52,11 +52,14 @@ final case class GmosHeader(hs: DhsClient, gmosObsReader: GmosHeader.ObsKeywords
         buildInt32(rv.ySize, s"DETRO${i}YS")
       )
   }.toList
+  
+  private val InBeam: Int = 0
```
cquiroz: Why not a boolean? is it an `Int` in epics?

jluhrs: Yes, it is an Int in EPICS.

PR #437, opened 2018-01-23
Artifacts upload

cquiroz:
This is a bit of an experiment to upload artifacts to s3 when we merge PRs

Some notes:
* We don't really have an s3 gemini account so i'm using mine
* I'm adding a stage of building the optimized js and upload those too
* The uploaded files are public

as usual with travis PRs this may need a bit of experimenting to get right

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - npm install jsdom
   - pip install --user awscli
 script:
-- sbt headerCheck test:headerCheck scalastyle test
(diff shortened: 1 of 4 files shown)


cquiroz:
I agree, its quite cheap and i don't mind hosting it myself but we should try to get an account. But it may take a long time to get one 😄 

@arturog8m what do you think about getting a gemini hlsw account?

cquiroz:
i'm going to close this, it seems we can get the same result with a just the weigh-in script

PR #439, opened 2018-01-25
SEQNG-364: Add sounds for end of step

cquiroz:
This PR adds a "beep" at the end of each observation. It maybe annoying so we could consider having the option to disable it in the future

Here's how it works in action

https://monosnap.com/file/KJavnm13MxJWeZMNnP3BQW9Z2DuwoG#


jluhrs:
El sonido sería mucho más amable al oído si tuviera una envolvente distinta. Encontré un código que hace eso:
`    now = context.currentTime;
    this.param.cancelScheduledValues(now);
    this.param.setValueAtTime(0, now);
    this.param.linearRampToValueAtTime(1, now + this.attackTime);
    this.param.linearRampToValueAtTime(0, now + this.attackTime + this.releaseTime);`

Lo más importante es que el ataque no sea un escalón.

arturog8m:
That is annoying indeed. Can't ship that as is, sorry.

arturog8m:
I meant that in reference to the audio in the video @cquiroz shared. Can we get a sample of what @jluhrs is proposing? Can we use the same audio that the current seqexec has?

jluhrs:
Why not use an audio file?

cquiroz:
I may try your code. The audio API is not very easy to use and get the timings right.
Audio file could be used as well. Do you have a sample?

In all honesty I don't like the whole concept, and this won't properly work on Chrome or iOS in any case

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/handlers.scala
```diff
@@ -492,6 +493,20 @@ object handlers {
...
     }
 
+    val stepCompletedMessage: PartialFunction[Any, ActionResult[M]] = {
+      case ServerMessage(e @ StepExecuted(obsId, sv)) =>
+      val curStep =
+        for {
+            obs      <- sequenceViewT.find(_.id === obsId)(e)
+            curSIdx  <- obs.runningStep.map(_._1)
+            curStep  <- sequenceStepT.find(_.id === curSIdx)(obs)
+            if curStep.observeStatus === ActionStatus.Pending && curStep.status === StepState.Running
```
cquiroz: @jluhrs This is odd to me. I'd expect to check for `curStep.observeStatus === ActionStatus.Completed` but the state from the server at the end of the observation is `Pending`. Any thoughts?

jluhrs: I think there is a bug.

cquiroz:
Changed to use a wav file. Here's a sample

https://monosnap.com/file/vGlhb5nxPI5LxPn0asVDTbVTtrTcMH#

PR #440, opened 2018-01-27
SEQNG-474 Implemented skip function.

jluhrs:
The skip function is now functional. The current implementation makes the skip override the breakpoints, so it is desirable to have the UI indicate this. There is also an issue with the current step indicator in the UI if that step is marked to be skipped.

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
@@ -46,7 +46,7 @@ object Sequence {
     done: List[Step]
   ) {
 
-    private val (toSkip, remaining): (List[Step], List[Step]) = pending.span(_.skipMark)
+    private val (toSkip, remaining): (List[Step], List[Step]) = pending.span(_.skipMark.self)
 
     /**
       * Runs the next execution. If the current `Step` is completed it adds the
@@ -59,14 +59,11 @@ object Sequence {
     val next: Option[Zipper] =
       focus.next match {
         // Step completed
-        case None      => {
-          remaining match {
-            case Nil             => None
-            case stepp :: stepps =>
-              (Step.Zipper.currentify(stepp) |@| focus.uncurrentify) (
-                (curr, stepd) => Zipper(id, metadata, stepps, curr, (done :+ stepd) ::: toSkip.map(_.copy(skipped = true)))
-              )
-          }
+        case None      => remaining match {
+          case Nil             => None
+          case stepp :: stepps => (Step.Zipper.currentify(stepp) |@| focus.uncurrentify) (
+            (curr, stepd) => Zipper(id, metadata, stepps, curr, (done :+ stepd) ::: toSkip.map(_.copy(skipped = Step.Skipped(true))))
+          )
         }
         // Current step ongoing
         case Some(stz) => Some(Zipper(id, metadata, pending, stz, done))
@@ -75,12 +72,12 @@ object Sequence {
     def rollback: Zipper = this.copy(focus = focus.rollback)
 
     def skips: Option[Zipper] = {
-      if (focus.skipMark) {
+      if (focus.skipMark.self) {
         remaining match {
           case Nil => None
           case stepp :: stepps =>
             (Step.Zipper.currentify(stepp) |@| focus.skip.some) (
-              (curr, stepd) => Zipper(id, metadata, stepps, curr, (done :+ stepd) ::: toSkip.map(_.copy(skipped = true)))
+              (curr, stepd) => Zipper(id, metadata, stepps, curr, (done :+ stepd) ::: toSkip.map(_.copy(skipped = Step.Skipped(true))))
             )
         }
       }
@@ -94,8 +91,8 @@ object Sequence {
       */
     val uncurrentify: Option[Sequence] =
       if (remaining.isEmpty)
-        if(focus.skipMark) Sequence(id, metadata, (done :+ focus.skip) ::: toSkip.map(_.copy(skipped = true))).some
-        else focus.uncurrentify.map(x => Sequence(id, metadata, (done :+ x) ::: toSkip.map(_.copy(skipped = true))))
+        if(focus.skipMark.self) Sequence(id, metadata, (done :+ focus.skip) ::: toSkip.map(_.copy(skipped = Step.Skipped(true)))).some
+        else focus.uncurrentify.map(x => Sequence(id, metadata, (done :+ x) ::: toSkip.map(_.copy(skipped = Step.Skipped(true)))))
       else None
 
     /**
@@ -288,14 +285,14 @@ object Sequence {
 
       override def setBreakpoint(stepId: Step.Id, v: Boolean): State = self.copy(zipper =
         zipper.copy(pending =
-          zipper.pending.map(s => if(s.id == stepId) s.copy(breakpoint = v) else s)))
+          zipper.pending.map(s => if(s.id == stepId) s.copy(breakpoint = Step.BreakpointMark(v)) else s)))
 
       override def setSkipMark(stepId: Step.Id, v: Boolean): State = self.copy(zipper =
-        if(zipper.focus.id == stepId) zipper.copy(focus = zipper.focus.copy(skipMark = v))
+        if(zipper.focus.id == stepId) zipper.copy(focus = zipper.focus.copy(skipMark = Step.SkipMark(v)))
         else zipper.copy(pending =
-          zipper.pending.map(s => if(s.id == stepId) s.copy(skipMark = v) else s)))
+          zipper.pending.map(s => if(s.id == stepId) s.copy(skipMark = Step.SkipMark(v)) else s)))
 
-      override def getCurrentBreakpoint: Boolean = zipper.focus.breakpoint && zipper.focus.done.isEmpty
+      override def getCurrentBreakpoint: Boolean = zipper.focus.breakpoint.self && zipper.focus.done.isEmpty
 
       override def setObserver(name: Observer): State = observerL.set(name.some)(self)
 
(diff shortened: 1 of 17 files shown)


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -46,6 +46,8 @@ object Sequence {
     done: List[Step]
   ) {
 
+    private val (toSkip, remaining): (List[Step], List[Step]) = pending.span(_.skipMark)
```
jluhrs: `pendings` is separated into the leading steps to be skipped, and the rest.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -57,27 +59,43 @@ object Sequence {
     val next: Option[Zipper] =
       focus.next match {
         // Step completed
-        case None      =>
-          pending match {
+        case None      => {
+          remaining match {
```
jluhrs: Look into `remaining` instead of `pending`, effectively skipping the steps in `toSkip`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -57,27 +59,43 @@ object Sequence {
...
             case Nil             => None
```
jluhrs: The caller to `next` will produce a `Final` state when the result is `None`. The construction of `Final` takes care of marking the last steps as skipped.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -57,27 +59,43 @@ object Sequence {
...
               )
           }
+        }
         // Current step ongoing
         case Some(stz) => Some(Zipper(id, metadata, pending, stz, done))
       }
 
     def rollback: Zipper = this.copy(focus = focus.rollback)
 
+    def skips: Option[Zipper] = {
```
jluhrs: The steps are normally skipped inside `next`. This function takes care of skipping the leading steps when starting the execution of a sequence.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -57,27 +59,43 @@ object Sequence {
     val next: Option[Zipper] =
       focus.next match {
         // Step completed
-        case None      =>
-          pending match {
+        case None      => {
```
cquiroz: that `{` is optional

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
cquiroz: Wouldn't `span` stop at the first step to skip. What happens if you have multiple skipped steps?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -23,31 +23,40 @@ final case class Step(
...
+  skipped: Boolean,
+  skipMark: Boolean,
   executions: List[List[Action]]
 )
 
 object Step {
 
   type Id = Int
 
+  def step(id: Int,
```
cquiroz: This is a bit strange. Wouldn't it be better to call this `apply`? Or a more descriptive name

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -333,4 +315,86 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
...
+    )
+
+    val sf = runToCompletion(s0)
+
+    inside (sf.map(_.sequences(seqId).done.map(Step.status))) {
+      case Some(stepSs) => assert(stepSs === List(StepState.Skipped, StepState.Completed, StepState.Completed))
+    }
+  }
+
+  it should "skip steps marked to be skipped in the middle of the sequence." in {
```
cquiroz: What about a test with several skipped steps?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -110,8 +110,9 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     } yield Result.Partial(FileIdAllocated(id), Kleisli(_ => doObserve(id).run.map(_.toResult)))
   }
 
-  private def step(obsId: SPObservationID, i: Int, config: Config, last: Boolean, datasets: Map[Int, ExecutedDataset]): TrySeq[Step] = {
-    def buildStep(inst: InstrumentSystem, sys: List[System], headers: Reader[ActionMetadata,List[Header]], resources: Set[Resource]): Step = {
+  // Introduced class StepBuilder to avoid warning for an excessively long function.
```
cquiroz: if this is the only reason for a builder i'd rather mark it for scalastyle to not check it

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTableContainer.scala
```diff
@@ -277,7 +276,7 @@ object StepsTableContainer {
       }
 
     private def classSet(step: Step): List[(String, Boolean)] = List(
-      "disabled" -> (step.skip || step.status === StepState.Completed),
+      "disabled" -> (step.status === StepState.Completed),
```
cquiroz: I'm not sure about this change. It gives the same background to the skipped states than the regular steps. Perhaps `disabled` isn't appropriate as we use it for completed steps but we should mark it somehow differently

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -23,31 +23,40 @@ final case class Step(
   config: StepConfig,
   resources: Set[Resource],
   breakpoint: Boolean,
-  skip: Boolean,
+  skipped: Boolean,
+  skipMark: Boolean,
```
swalker2m: It's tedious, but I think worthwhile in the end to define types for each of these `Boolean`s.

cquiroz: I agree

swalker2m: I don't know much about the seqexec model so pardon my ignorance.  Can a `Step` be a breakpoint and yet skipped and a `skipMark`?  I'm just wondering if the model is permitting combinations that don't make sense or shouldn't occur in reality.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
jluhrs: It stops at the first element that does not fulfill the condition. All the elements in the leading `List` are assured to pass the test, while the first element of the tailing list is assured to not pass the test.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
jluhrs: The point is to create a `Step` with default values for the flags. Maybe I should call it `init`?

modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
jluhrs: Ok, I added one test for that.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
jluhrs: About the first comment, I agree. I even had a bug because I confused the values.

jluhrs: A step can be marked to be skipped an have a breakpoint, but when run, the skip overrides the breakpoint. Currently the UI allows to set them independently. As long as the step is not run, the user can set the skip mark on a step that has a breakpoint. If he/she then unsets the skip mark, the breakpoint should remain.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTableContainer.scala
jluhrs: Skipped steps and successfully completed steps are similar, in that both kinds are already run and have are in a final state.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Step.scala
```diff
@@ -18,32 +18,36 @@ import monocle.macros.GenLens
...
 object Step {
 
   type Id = Int
 
-  def step(id: Int,
-    fileId: Option[FileId],
-    config: StepConfig,
-    resources: Set[Resource],
-    executions: List[List[Action]]): Step = Step(id, fileId, config, resources, false, false, false, executions)
+  final case class BreakpointMark(self: Boolean) extends AnyVal
```
cquiroz: Maybe we switch to an ADT on a later PR

PR #441, opened 2018-01-29
SEQNG-503: Update current step mark when leading steps are marked to be skipped

cquiroz:
Small UI updates to better display skipped states. it properly calculates which is the next step to run on the presence of skips


jluhrs:
I don't think this takes care of the case when there are some steps already executed, the sequence is stopped, and the user sets a skip mark in the current step.

PR #442, opened 2018-01-29
SEQNG-504: Show breakpoints in steps to be skipped in a disabled state

cquiroz:
This PR shows breakpoints on skipped states as gray instead of brown, as:

screenshot 2018-01-29 12 04 47


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTableContainer.scala
```diff
@@ -260,7 +260,11 @@ object StepsTableContainer {
           )
         ),
         <.td(
-          if (step.breakpoint) SeqexecStyles.breakpointTrOn else SeqexecStyles.breakpointTrOff,
+          (step.breakpoint, step.skip) match {
+            case (true, true) => SeqexecStyles.breakpointTrOnSkipped
```
cquiroz: This would certainly be more readable if we had actual objects rather than booleans

cquiroz:
this won't be needed after all

PR #444, opened 2018-01-31
SEQNG-493 GNIRS configuration model.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/GnirsController.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def stopObserve: SeqAction[Unit]
+
+  def abortObserve: SeqAction[Unit]
+
+}
+
+object GnirsController {
+
+  sealed trait Mode
```
jluhrs: After analysing the old Seqexec code I think I can merge the acquisitionMirror and crossDispersed parameters into one. This will help managing interdependencies between those parameters. For example, crossDispersed settings don't have any effect if acquisitionMirror is In.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/GnirsController.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  case object Acquisition extends Mode
+  sealed class Spectrography protected (val disperser: Disperser) extends Mode
```
cquiroz: you may want to make it `abstract` if the constructor is private

PR #446, opened 2018-01-31
SEQNG-506 Mixing skips and breakpoints on leading steps was confusing the engine

jluhrs:


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -46,7 +46,7 @@ object Sequence {
     done: List[Step]
   ) {
 
-    private val (toSkip, remaining): (List[Step], List[Step]) = pending.span(st => st.skipMark.self && !st.breakpoint.self)
+    private val (toSkip, remaining): (List[Step], List[Step]) = pending.span(st => st.skipMark.self)
```
jluhrs: Checking for breakpoints was a special case for `next`, so I moved it to that function.

PR #447, opened 2018-01-31
SEQNG-451 Log error messages from failed actions.

jluhrs:
Also, errors while loading sequences are now logged as warnings, and errors in the observe commands are logged as errors.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -179,8 +179,8 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
     } yield translator.sequence(seqId, odbSeq)
 
     t.map {
-      case (err :: _, None)  => List(Event.logMsg(SeqexecFailure.explain(err)))
-      case (errs, Some(seq)) => Event.load(seqId.stringValue, seq) :: errs.map(e => Event.logMsg(SeqexecFailure.explain(e)))
+      case (err :: _, None)  => List(Event.logDebugMsg(SeqexecFailure.explain(err)))
+      case (errs, Some(seq)) => Event.load(seqId.stringValue, seq) :: errs.map(e => Event.logDebugMsg(SeqexecFailure.explain(e)))
       case _                 => Nil
     }
   }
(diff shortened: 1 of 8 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -207,7 +207,10 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
       case engine.GetState(_)            => NullEvent
       case engine.GetSeqState(_, _)      => NullEvent
       case engine.ActionStop(_, _)       => ActionStopRequested(svs)
-      case engine.Log(msg)               => NewLogMessage(msg)
+      case engine.LogDebug(msg)          => NewLogMessage(msg)
```
cquiroz: This is not needed. We should send them to `NullEvent`. How the log works is that we capture log events at the logger level using log4s/slf4j and those are sent to the client as events of type `ServerLogMessage`
In fact we may remove `NewLogMessage` altogether

cquiroz:
Did you test with the UI?

jluhrs:
Yes, I tested with the UI. The "unsupported instrument" when loading sequences is a bit annoying, but can be filtered.

cquiroz:
"unsupported instrument" should likely be debug level

jluhrs:
Then I would need to put all the load errors at debug level. Which kind of make sense, because load errors come from software problems, like an instrument not yet supported, or a typo in a key name, not something that the user can fix.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -255,13 +258,13 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
     def loads(odbList: Seq[SPObservationID]): Task[List[Event]] =
-      odbList.diff(seqexecList).toList.map(id => loadEvents(id)).sequenceU.map(_.flatten).run.map(_.valueOr( r => List(Event.logMsg(SeqexecFailure.explain(r)))))
+      odbList.diff(seqexecList).toList.map(id => loadEvents(id)).sequenceU.map(_.flatten).run.map(_.valueOr( r => List(Event.logDebugMsg(SeqexecFailure.explain(r)))))
 
     def unloads(odbList: Seq[SPObservationID]): Seq[Event] =
       seqexecList.diff(odbList).map(id => unloadEvent(id))
 
     val x = odbProxy.queuedSequences.flatMapF(seqs => loads(seqs).map(ee => (ee ++ unloads(seqs)).right)).run
-    val y = x.map(_.valueOr(r => List(Event.logMsg(SeqexecFailure.explain(r)))))
+    val y = x.map(_.valueOr(r => List(Event.logWarningMsg(SeqexecFailure.explain(r)))))
```
jluhrs: An error here means there was a problem connecting to the ODB, therefore it merits a warning to the user.

PR #448, opened 2018-01-31
GNIRS controller simulator

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/InstrumentControllerSim.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.log4s.getLogger
+import squants.Time
+
+import scala.annotation.tailrec
+import scalaz.{EitherT, Show}
+import scalaz.concurrent.Task
+import scalaz.syntax.show._
+import scalaz.syntax.std.boolean._
+
+class InstrumentControllerSim(name: String) {
```
cquiroz: One issue I often have with the simulators is that times are sometimes too short. This is sometimes a problem while doing development where you want e.g. observe to take longer to see the effects.

I normally just change the code locally, but perhaps there should be configurable waits on each of the actions so i'd only need to modify those values without needing to add the delays

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/InstrumentControllerSim.scala
```diff
@@ -0,0 +1,104 @@
...
+    pauseFlag.set(false)
+    stopFlag.set(false)
+    abortFlag.set(false)
+    observeTic(false, false, false, expTime.millis.toInt,
+      (expTime.value > 0.0).option(expTime.toMilliseconds.toInt + 2 * tic))
+  } )
+
+  def applyConfig[C: Show](config: C): SeqAction[Unit] = EitherT( Task {
+    Log.info(s"Simulate applying $name configuration ${config.shows}")
+    Thread.sleep(1000)
```
cquiroz: As an example of my previous comment. 1s is sometimes too short for development. And other methodse, e.g. `stopObserve` have no delay

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/GnirsControllerSim.scala
```diff
@@ -0,0 +1,35 @@
...
+import edu.gemini.seqexec.server.gnirs.GnirsController.GnirsConfig
+import edu.gemini.seqexec.server.{InstrumentControllerSim, ObserveCommand, SeqAction}
+import squants.Time
+
+import scalaz.Show
+
+object GnirsControllerSim extends GnirsController {
+
+  implicit val configShow: Show[GnirsConfig] = Show.shows { config =>
+    s"(${config.cc.mode}, ${config.cc.camera}, ${config.cc.centralWavelength}, ${config.cc.decker}, " +
```
cquiroz: Maybe it's simpler to have a list of the components and do `mkString`

cquiroz:
I like this a lot. it saves a lot of common code. I'd request to add those configurable delays though

jluhrs:
One of the changes I made was to make `observe` take `expTime` to complete. The only drawback is that tiimeout simulation now is useless.

jluhrs:
So, you want to make the delays configurable? From where should those values come? app.conf?

cquiroz:
We could just put them in the code as Constants we can easily change during development. The conf is already cluttered

PR #455, opened 2018-02-04
SEQNG-518 Implemented workaround for non-compliant CAR in GEM5 systems.

jluhrs:
GEM5 systems use a definition for the CAR values that is different from later systems. I had to implement a workaround.
I put a new flag in the definition of a Apply record in the Schema file. it is used by XMLBuilder to automatically use the right definition for CAR records.


cquiroz:
What's GEM5? Looks fine in any case

jluhrs:
Historically, the EPICS versions and Gemini support packages were bundled into GEM releases. GEM5 is an old one, I think the oldest in use.

PR #456, opened 2018-02-04
Use a custom tzdb

cquiroz:
This is an experimental option to reduce the code size on the client by only including the data for our relevant timezones. Note this is using a SNAPSHOT of `scala-java-time` I expect it to go final after some more testing


cquiroz:
Note that with this change we reduce the js size in full 2MB

PR #457, opened 2018-02-05
SEQNG-514: Add details to log

cquiroz:
This PR adds to the log the following:

* Show the hostname of a user when he/she logs in
* Show the observation id when taking an observation

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/flamingos2/Flamingos2ControllerSim.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/flamingos2/Flamingos2ControllerSim.scala
@@ -49,6 +49,7 @@ final case class Flamingos2ControllerSimBad(failAt: Int) extends Flamingos2Contr
   private val counter: AtomicInteger = new AtomicInteger(0)
 
   override def observe(fileId: ImageFileId, expTime: Time): SeqAction[ImageFileId] = EitherT( Task {
+    Log.info(s"Simulating Flamingos-2 observe with fileId: $fileId")
     Thread.sleep(5000)
     TrySeq(fileId)
   } )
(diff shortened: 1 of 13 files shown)


jluhrs:
We can avoid passing `obsId` all around if the log messages were generated in `SeqTranslate`, like this:
``
`
def doObserve(id: ImageFileId): SeqAction[Result] =
      for {
        d   <- dataId
        _   <- sendDataStart(obsId, id, d)
        _  <- notifyObserveStart
        _  <- headers(ctx).map(_.sendBefore(id, inst.dhsInstrumentName)).sequenceU
        _ <- EitherT(Task(Log.info(s"Start $inst observation $obsid with label $fileId").right))
        r   <- inst.observe(config)(id)
        _ <- EitherT(Task(Log.info(s"Completed $inst observation $obsid with label $fileId").right))
        ret <- observeTail(id, d)(r)
      } yield ret
```

**Comment by cquiroz** [comment]
Sounds good, would we remove the logs on the instruments then?

**Comment by jluhrs** [comment]
Yes. We can do the same for other log messages. But we must keep the ones from the simulator, to know that the instrument operation is actually simulated.

**Comment by cquiroz** [comment]
Updated according to the requested changes

File: modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/flamingos2/Flamingos2ControllerSim.scala [hunk 1]
```diff
@@ -50,11 +48,9 @@ final case class Flamingos2ControllerSimBad(failAt: Int) extends Flamingos2Contr
 
   private val counter: AtomicInteger = new AtomicInteger(0)
 
-  override def observe(obsid: ImageFileId, expTime: Time): SeqAction[ImageFileId] = EitherT( Task {
-    Log.info("Taking Flamingos-2 observation with label " + obsid)
```
jluhrs: `Flamingos2ControllerSimBad` should put a log message to indicate that we are using a simulated F2.

PR #458, opened 2018-02-06
Implemented GNIRS Instrument class

jluhrs:
The translation of the GNIRS configuration to the actual EPICS parameters is a bit complicated. I tried to define an internal configuration model that minimized the number of parameters, while keeping its expressiveness to the same level as the current Seqexec.
Conversions are spread between the classes `Gnirs` and `GnirsControllerEpics`. The criteria I used is that if I could reduce the number of parameters in the model, I did the conversion in `Gnirs`. But if the input parameter was already part of the model, then I put the translation inside `GnirsControllerEpics`.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
@@ -225,46 +225,51 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
       })
   }
 
-  private def deliverObserveCmd(seqState: Sequence.State, cmd: Option[Process[Task, Event]]): Option[Process[Task, Event]] = {
+  private def deliverObserveCmd(seqState: Sequence.State,
+                                f: ObserveControl => Option[SeqAction[Unit]]): Option[Process[Task, Event]] = {
     def isObserving(v: Action): Boolean = v.kind === ActionType.Observe && (v.state.runState match {
       case Action.Started               => true
       case _                            => false
     })
-    if(seqState.current.execution.exists(isObserving)) cmd
-    else none
+
+    toInstrumentSys(seqState.toSequence.metadata.instrument).toOption.flatMap(x => f(x.observeControl)).flatMap {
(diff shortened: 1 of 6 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqTranslate.scala
```diff
@@ -225,46 +225,51 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
...
-        case _      => Event.nullEvent
-      }))
+    }
+
+    deliverObserveCmd(seqState, f).orElse(abortPaused(seqId)(seqState))
+  }
+
+  def pauseObserve(seqState: Sequence.State): Option[Process[Task, Event]] = {
+    def f(oc: ObserveControl): Option[SeqAction[Unit]] = oc match {
+      case OpticControl(_, _, PauseObserveCmd(pause), _, _, _) => Some(pause)
```
cquiroz: I guess IR instruments cannot pause

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/Gnirs.scala
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private def getDecker(config: Config, slit: SlitWidth, woll: WollastonPrism, xdisp: CrossDispersed): ConfigUtilOps.ExtractFailure\/GnirsController.Decker =
+    config.extract(INSTRUMENT_KEY / DECKER_PROP).as[Decker].orElse{
+      for {
+        pixScale <- config.extract((INSTRUMENT_KEY / PIXEL_SCALE_PROP)).as[PixelScale]
+      } yield xdisp match {
+        case CrossDispersed.LXD => Decker.LONG_CAM_X_DISP
+        case CrossDispersed.SXD => Decker.SHORT_CAM_X_DISP
+        case _                  => {
```
cquiroz: This `{` is not needed

jluhrs: I know, but prevents Idea from messing the indentation.

PR #459, opened 2018-02-07
Create GNIRS Step from Sequence values

jluhrs:
Included several changes made after testing with the real instrument. GNIRS showed several bugs in the code, specially in the ACM assumptions on the order of events.
Also, I had to abandon my intention of triggering all commands from the top level apply record, because I found the corresponding CAR record does not have its CLID connected.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/resources/Gnirs.xml
+++ b/modules/edu.gemini.seqexec.server/src/main/resources/Gnirs.xml
@@ -3,10 +3,10 @@
     <Top name="nirs">nirs</Top>
     <Apply name="nirs::apply">
         <top>nirs</top>
-        <apply>apply</apply>
-        <car>applyC</car>
+        <apply>cc:apply</apply>
+        <car>cc:applyC</car>
         <gem5>true</gem5>
-        <description>IS Primary Apply Record </description>
+        <description>CC Apply Record </description>
         <command name="nirs::config">
             <description>Setup a Configuration</description>
             <parameter name="cover">
@@ -55,7 +55,7 @@
                 <description>Camera</description>
             </parameter>
             <parameter name="gratingMode">
-                <channel>cc:gratingPos</channel>
+                <channel>cc:gratingPosCad.A</channel>
                 <type>STRING</type>
                 <description>Grating Mode</description>
             </parameter>
@@ -85,6 +85,13 @@
                 <description>Filter2</description>
             </parameter>
         </command>
+    </Apply>
+    <Apply name="nirs::dcapply">
+        <top>nirs</top>
+        <apply>dc:apply</apply>
+        <car>dc:applyC</car>
+        <gem5>true</gem5>
+        <description>DC Apply Record </description>
         <command name="nirs::dcconfig">
             <description>Setup a DC Configuration</description>
             <parameter name="lowNoise">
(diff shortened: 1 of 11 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/Gnirs.scala
```diff
@@ -132,14 +135,14 @@ object Gnirs {
 
 
   private def autoFilter(w: Length): GnirsController.Filter2 = {
-    val table = Map(
+    val table = List(
```
jluhrs: What was I thinking ?!

modules/edu.gemini.seqexec.server/src/main/resources/Gnirs.xml
```diff
@@ -85,6 +85,13 @@
                 <description>Filter2</description>
             </parameter>
         </command>
+    </Apply>
+    <Apply name="nirs::dcapply">
```
cquiroz: Why's the top `nirs`?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/Gnirs.scala
```diff
@@ -132,14 +135,14 @@ object Gnirs {
...
       GnirsController.Filter2.J -> 1.42,
       GnirsController.Filter2.H -> 1.86,
-      GnirsController.Filter2.J -> 2.70,
+      GnirsController.Filter2.K -> 2.70,
       GnirsController.Filter2.L -> 4.30,
       GnirsController.Filter2.M -> 6.0
-    ).mapValues(_.nanometers)
+    ).map{ case (f, w) => (f, w.nanometers) }
 
     table.foldRight[GnirsController.Filter2](GnirsController.Filter2.XD){ case (t, v) => if(w < t._2) t._1 else v}
```
cquiroz: Why do you need to do this?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/Gnirs.scala
```diff
@@ -175,9 +178,8 @@ object Gnirs {
     case SlitWidth.SW_5        => GnirsController.SlitWidth.Slit0_45.some
     case SlitWidth.SW_6        => GnirsController.SlitWidth.Slit0_68.some
     case SlitWidth.SW_7        => GnirsController.SlitWidth.Slit1_00.some
+    case SlitWidth.PUPIL_VIEWER => GnirsController.SlitWidth.PupilViewer.some
```
cquiroz: I bet the misalignment bothers you too

modules/edu.gemini.seqexec.server/src/main/resources/Gnirs.xml
jluhrs: Gemini NIRS

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/Gnirs.scala
jluhrs: I don't need it, I just wanted to have everything as the proper physical quantity, but was too lazy as to add the `.nanometers` to each value.

PR #460, opened 2018-02-08
Several fixes for GNIRS

jluhrs:
This PR includes several fixes for GNIRS:
* Changed the channels for parameters *exposure time* and *coadds* to be the same used by the ISD. (Why the ISD shows the input parameters and not the applied values is beyond me).
* Fixed `NullPointerException` in `CaApplySenderImp`
* Implemented an ugly workaround to an issue with the AG in-position (should work on a better solution).
* Incremented timeouts for waiting for telescope an AG in-position.
* Read *coadds* from the right property.
* Implemented support for `Enum` parameters in Command Sender. I thought I needed it for GNIRS, but at the end it was not necessary.

### DIFF
--- a/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
+++ b/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
@@ -206,7 +206,7 @@ final class CaApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements Ca
         @Override
         public State onApplyValChange(Integer val) {
             if (val > 0) {
-                if (carClid != null && carClid.equals(val)) {
+                if (carVal != null && carClid != null && carClid.equals(val)) {
                     if (carVal.isError()) {
                         failCommandWithCarError(cm);
                         return IdleState;
--- a/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
+++ b/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
@@ -206,7 +206,7 @@ final class CaApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements Ca
         @Override
         public State onApplyValChange(Integer val) {
             if (val > 0) {
-                if (carVal != null && carClid != null && carClid.equals(val)) {
+                if (carVal != null && val.equals(carClid)) {
                     if (carVal.isError()) {
                         failCommandWithCarError(cm);
                         return IdleState;
--- a/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
+++ b/modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
@@ -297,7 +297,7 @@ public class CaObserveSenderImpl<C extends Enum<C> & CarStateGeneric> implements
         @Override
         public CaObserveSenderImpl.State onApplyValChange(Integer val) {
             if (val > 0) {
-                if (carClid != null && carClid.equals(val) && carState != null) {
+                if (val.equals(carClid) && carState != null) {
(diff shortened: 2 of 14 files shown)


modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
```diff
@@ -206,7 +206,7 @@ private WaitPreset(CaCommandMonitorImpl cm, CarStateGeneric carVal, Integer carC
         @Override
         public State onApplyValChange(Integer val) {
             if (val > 0) {
-                if (carClid != null && carClid.equals(val)) {
+                if (carVal != null && carClid != null && carClid.equals(val)) {
```
cquiroz: I think you can replace `carClid != null && carClid.equals(val)` by `val.equals(carClid)`

modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -276,36 +276,36 @@ public CaCommandMonitor postCallback(CaCommandListener callback) {
...
-            this.observeCarVal = observeCarVal;
+            this.observeCarState = observeCarState;
         }
 
         @Override
         public CaObserveSenderImpl.State onApplyValChange(Integer val) {
             if (val > 0) {
-                if (carClid != null && carClid.equals(val) && carVal != null) {
-                    if (carVal.isError()) {
+                if (carClid != null && carClid.equals(val) && carState != null) {
```
cquiroz: Same comment as above about simplifying the null check

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -226,6 +226,12 @@ object GnirsControllerEpics extends GnirsController {
     _ <- GnirsEpics.instance.abortCmd.post
   } yield ()
 
+  private def removePartName(s: String) = {
```
cquiroz: Why this?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -369,8 +370,8 @@ object TcsControllerEpics extends TcsController {
 
   private def setM2Guide(c: M2GuideConfig): SeqAction[Unit] = TcsEpics.instance.m2GuideCmd.setState(encode(c))
 
-  private val tcsTimeout = Seconds(30)
-  private val agTimeout = Seconds(30)
+  private val tcsTimeout = Seconds(60)
```
cquiroz: I guess this will apply to all instruments, Why was it changed?

modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
jluhrs: Doh!

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/GnirsControllerEpics.scala
jluhrs: If you read the current filter, GNIRS gives you something like H_G5534. I need to remove the trailing code before comparing.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/tcs/TcsControllerEpics.scala
jluhrs: It was changed because I just found today that the science fold can take 45[s] to reach a position. That affects also `tcsTimeout`.

PR #461, opened 2018-02-08
SEQNG-497 Implemented GNIRS FITS keywords.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/GnirsEpics.scala
```diff
@@ -174,14 +174,30 @@ class GnirsEpics(epicsService: CaService, tops: Map[String, String]) {
...
+  def centralWavelength: Option[Double] = Option(state.getDoubleAttribute("centralWavelength").value).map(_.toDouble)
 
-  def gratingTilt: Option[Double] = Option(state.getStringAttribute("grattilt").value).map(_.toDouble)
+  def gratingTilt: Option[Double] = Option(state.getDoubleAttribute("grattilt").value).map(_.toDouble)
 
   def nirscc: Option[String] = Option(state.getStringAttribute("nirscc").value)
 
   def gratingOrder: Option[Int] = Option(state.getIntegerAttribute("gratord").value).map(_.toInt)
 
+  def filter1Eng: Option[Int] = Option(state.getIntegerAttribute("fw1_eng").value).map(_.toInt)
```
cquiroz: I have the impression that reading the value is a side effect, as we read it from the EPICS channel. Is that correct?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/tcs/TcsKeywordReader.scala
```diff
@@ -258,4 +271,16 @@ object TcsKeywordsReaderImpl extends TcsKeywordsReader {
   override def getOiwfsFreq: SeqAction[Option[Double]] = TcsEpics.instance.oiwfsIntegrationTime.map(calcFrequency)
 
   override def getCarouselMode: SeqAction[Option[String]] = TcsEpics.instance.carouselMode
+
+  override def getGnirsInstPort: SeqAction[Option[Int]] = TcsEpics.instance.gnirsPort
```
cquiroz: Are the ports constants or do we read them from the A&G?

cquiroz:
Looks fine in general but it got me thinking about some parts of the code calling EPICS that we may not be wrapping properly on `Task`. What do you think?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/gnirs/GnirsEpics.scala
jluhrs: The value is cached, so there is no network activity in this call. But, it breaks referential transparency.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/tcs/TcsKeywordReader.scala
jluhrs: They are read from the AG.

jluhrs:
I had the same thought while I was coding. Maybe we should review this in the future.

jluhrs:
BTW, it is wrapped in a `Task`, that is why every keyword construction is a `SeqAction`. Maybe it is not at the right level?

PR #462, opened 2018-02-08
SEQNG-521 Special behavior for GNIRS focus configuration.

jluhrs:
GNIRS focus has to be reset every time some of the componentes move. The implementation delays the application of the parameters, allowing to check if there is any at all. The same implementation allows now avoid applying an empty configuration. I thing the same technique should be applied to the other systems, once refined.


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/EpicsUtil.scala
```diff
@@ -249,6 +249,6 @@ object EpicsUtil {
   }
 
   def smartSetParam[A: Equal](v: A, get: => Option[A], set: SeqAction[Unit]): SeqAction[Unit] =
-    get.flatMap(g => (g =/= v).option(set)).getOrElse(SeqAction.void)
```
jluhrs: This was a bug. If `get` returns `None`, then the parameter has to be set.

PR #470, opened 2018-02-13
Update README.md

sraaphorst:
Removing the no-longer-functioning calling commands.

This file probably requires more changes than that, but I'm still an ocs3 newb.


cquiroz:
it already exists but the documentation isn't correct. For starters the wire protocol isn't json anymore, instead we use boopickle. Then the endpoints have changed and some of the commands are gone

Finally there is no mention of authentication

In summary the documentation should be rewritten

PR #471, opened 2018-02-13
SEQNG-457: Tooltips not visible

cquiroz:
This PR fixes a bug where the button popups would be hidden in many cases. The original popups were pure CSS but presented this problem

In this PR i'm adding a new Popup component and use it in the correct places. The component is very basic as i'm planning to switch to the proper react components from `react-semantic-ui` in the near term

The change can be observed in the image below:
screencast 2018-02-13 17-11-18

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
@@ -20,7 +20,7 @@ object Popup {
   private val component = ScalaComponent.builder[Props]("Popup")
     .stateless
     .renderPC{($, p, c) =>
-      // This is slightly unsafe though in most case we'll be passing a ReactElement here
+      // This is in principle unsafe but we are only allowing Elements on the constructor
       VdomElement($.propsChildren.only().asInstanceOf[ReactElement])
     }
     .componentDidMount(ctx =>
@@ -35,5 +35,5 @@ object Popup {
       }
     ).build
 
-  def apply(p: Props, children: VdomNode): Unmounted[Props, Unit, Unit] = component(p)(children)
+  def apply(p: Props, children: VdomElement): Unmounted[Props, Unit, Unit] = component(p)(children)
 }
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
@@ -6,31 +6,34 @@ package edu.gemini.seqexec.web.client.semanticui.elements.popup
 import japgolly.scalajs.react.component.Scala.Unmounted
 import japgolly.scalajs.react.{Callback, ScalaComponent}
 import japgolly.scalajs.react.vdom.html_<^._
+import japgolly.scalajs.react.raw.ReactElement
 
 /**
-  * Produces a dropdown menu, similar to a combobox
+  * Produces a popup using javascript
+  * This is a barebones component waiting for the proper react component to be made available
   */
-
-@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements", "org.wartremover.warts.AsInstanceOf"))
 object Popup {
   @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
   final case class Props(selector: String, content: String)
 
   private val component = ScalaComponent.builder[Props]("Popup")
     .stateless
-    .renderPC((_, p, c) => <.div(c))
+    .renderPC{($, p, c) =>
+      // This is slightly unsafe though in most case we'll be passing a ReactElement here
+      VdomElement($.propsChildren.only().asInstanceOf[ReactElement])
+    }
     .componentDidMount(ctx =>
       Callback {
         // Enable menu on Semantic UI
         import org.querki.jquery.$
         import edu.gemini.web.client.facades.semanticui.SemanticUIPopup._
 
-        println($(ctx.getDOMNode).find(ctx.props.selector).length)
-        $(ctx.getDOMNode).find(ctx.props.selector).popup(
+        $(ctx.getDOMNode).popup(
           JsPopupOptions.content(ctx.props.content)
         )
       }
     ).build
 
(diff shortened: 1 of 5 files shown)


sraaphorst:
I'm a little confused by the `Button` class:
https://github.com/gemini-hlsw/ocs3/blob/develop/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/button/Button.scala

Why are `NotAnimated` buttons `<button.../>`, but animated buttons are `<div.../>` and seem much simpler? Perhaps I am not understanding the concept of the animation?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
```diff
@@ -0,0 +1,39 @@
...
+  * This is a barebones component waiting for the proper react component to be made available
+  */
+@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements", "org.wartremover.warts.AsInstanceOf"))
+object Popup {
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
+  final case class Props(selector: String, content: String)
+
+  private val component = ScalaComponent.builder[Props]("Popup")
+    .stateless
+    .renderPC{($, p, c) =>
```
sraaphorst: What is the significance of the `$` here? I see it a few places and I'm unsure what it signifies.

cquiroz:
@sraaphorst Do you refer to this line?
https://github.com/gemini-hlsw/ocs3/blob/develop/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/button/Button.scala#L98

If i recall correctly animated buttons require to use a div but in general we want to use `<button>` as the proper markup for a Button. This is very much related to the way SemanticUI works

https://semantic-ui.com/elements/button.html#animated

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
cquiroz: the use of `$` is just a convention used in `scalajs-react`. It's probably a bad idea as you get easily confused with `jQuery`'s `$`.

In any case, in this particular call `$` is a reference to the react component which sometimes you need to access. In the most common case you only do `.renderP` and use the `Props` but this is a special case

the `$` reference lets me access the children and render them directly avoiding wrapping them in another markup

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
```diff
@@ -0,0 +1,39 @@
...
+        import edu.gemini.web.client.facades.semanticui.SemanticUIPopup._
+
+        $(ctx.getDOMNode).popup(
+          JsPopupOptions.content(ctx.props.content)
+        )
+      }
+    ).build
+
+  def apply(p: Props, children: VdomElement): Unmounted[Props, Unit, Unit] = component(p)(children)
+}
```
sraaphorst: Could you explain a little bit what is happening here? Do we know why the original CSS solution wasn't working?

So now, for each button (or UI element) with a popup, we create a stateless `ScalaComponent`.
When is the `Callback` called?  And if you could explain lines 32-33, that would be helpful :-).

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
```diff
@@ -0,0 +1,39 @@
...
+  private val component = ScalaComponent.builder[Props]("Popup")
+    .stateless
+    .renderPC{($, p, c) =>
+      // This is in principle unsafe but we are only allowing Elements on the constructor
+      VdomElement($.propsChildren.only().asInstanceOf[ReactElement])
+    }
+    .componentDidMount(ctx =>
+      Callback {
+        // Enable menu on Semantic UI
+        import org.querki.jquery.$
```
cquiroz: As yo see here `$` is also a reference to `jQuery`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
sraaphorst: Okay, that makes sense. Thanks for the explanation! It is much appreciated.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/semanticui/elements/popup/Popup.scala
cquiroz: Yes, the original pure CSS solution wasn't able to display beyond the target leading to clipped popups.

in line 38. `children` is a reference to what we are wrapping, typically a button, (note that i'm limiting this to only one child)

`SemanticUI` uses jQuery. the code in line 32 creates a `jQuery` reference to the rendered node (In this case the child) and attaches a popup with the `popup` call (a `jQuery` extension) and it also passes some configuration. In particular the text associated with the popup

in general all the code using `SemanticUI` and `jQuery` is fairly low level and hacky. This will improve once i complete the facades for `Semantic UI React`

cquiroz: `Callback` is called by `scalajs-react`. It is slightly similar to `IO` in `cats` or `scalaz`. It lets us return pure functions from the lifecycle methods and it is pervasive on `scalajs-react`

sraaphorst: That helps a lot. Thanks!

sraaphorst:
So I'm unclear: do we have a full implementation of an animated button at this point?

The Semantic UI link will be extremely helpful, and I'm looking forward to learning more about all of this.

cquiroz:
@sraaphorst We do. The difference is that the top level tag has to be different for animated buttons. But this is transparent for the caller

PR #476, opened 2018-02-15
Added debug messages to observe monitoring code.

jluhrs:


modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -259,6 +267,8 @@ public CaCommandMonitor postCallback(CaCommandListener callback) {
     }
 
     private interface State {
+        String signature();
```
cquiroz: Nice name !

PR #477, opened 2018-02-16
Make icon vals lazy to reduce fastOpt size

cquiroz:
I learned that I can reduce the `fastOpt` size by making some `vals` `lazy vals`. This shouldn't affect the full optimized size but it should reduce `fastOpt` size in around 1M

I'm also copying the way we call `weigh_in` from gem and will start tracking the `fullOpt` size too


cquiroz:
The reduction was somewhat smaller, around 200K but still worth it for such a tiny change

swalker2m:
Why does using a `lazy val` reduce size?

cquiroz:
I examined the generated js. If you don't use a val it will be optimized away in full opt `lazy` or not `lazy`. But on `fastOpt`, unused non lazy vals contain their initialization, while unused `lazy val` get set to `null`.

Thus you save some bytes. I have a bunch of Icons (all the ones avaliable on SUI, around 400) defined as `vals`. We use only a few so the rest is optimized away in `fullOpt` but it somewhat less efficiently in `fastOp`

As to the deep reason why this works this way is a question for the `scala.js` guys

swalker2m:
Is there a reason you wouldn't instead only define icons that you actually use?  It isn't a big deal it just strikes me as odd to use a language feature that is about something else altogether for the side-effect of reducing code size.

cquiroz:
Yes, that could be a choice, only define what we need, but I built this as a lib originally.

I'm also surprised about the difference on optimization of lazy in this case but the semantics are the same in terms of behavior, especially on js which is single threaded.

PR #479, opened 2018-02-20
SEQNG-531 Added execution queues to the Seqexec server.

jluhrs:
First implementation of execution queues. They are just a list of sequence ids. The server has the capabilities to add and remove a sequence id from the queue, and to change its position.
The biggest change is that, in order to add the execution queues to the server state, it was modified to accept an arbitrary user data of a parametric type.


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/StepSpec.scala
```diff
@@ -265,7 +271,7 @@ class StepSpec extends FlatSpec {
         )
       )
 
-    val qss = process(Process.eval(Task.now(Event.pause(seqId, user))))(qs0).runLog.unsafePerformSync.map(_._2)
+    val qss =executionEngine. process(Process.eval(Task.now(Event.pause(seqId, user))))(qs0).runLog.unsafePerformSync.map(_._2)
```
cquiroz: Missing space after `=`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -3,24 +3,21 @@
...
 import scalaz._
 import Scalaz._
 import scalaz.concurrent.Task
-import scalaz.stream.{Process, Sink, merge}
+import scalaz.stream.Process
 
 package engine {
 
-  final case class HandleP[A](run: Handle[(A, Option[Process[Task, Event]])])
+  final case class HandleP[A, D](run: Handle[(A, Option[Process[Task, Event[D]]]), D])
```
cquiroz: Can you document these types?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -3,36 +3,374 @@
...
+import scalaz.{Applicative, Kleisli, MonadState, StateT}
+import scalaz._
+import Scalaz._
+import scalaz.stream.{Process, merge}
 
-/**
-  * A Map of `Sequence`s.
-  */
-final case class Engine(sequences: Map[Sequence.Id, Sequence])
+class Engine[D] {
```
cquiroz: Where do `sequences` live now?

cquiroz:
This would be much easier to review without moving the code across files. Now I see a lot of added code and lot of removed code. I wish github were smarter about this

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -3,36 +3,374 @@
...
+    mapEvalState[Event[D], Engine.State[D], (Event[D], Engine.State[D])](input, qs, (e: Event[D]) => runE(e).run)
+
+  // Functions for type bureaucracy
+
+  private def pure[A](a: A): HandleP[A, D] = Applicative[({type L[T] = HandleP[T, D]})#L].pure(a)
+
+  private val unit: HandleP[Unit, D] = pure(())
+
+  private val get: HandleP[Engine.State[D], D] =
+    MonadState[({type L[T] = Handle[T, D]})#L, Engine.State[D]].get.toHandleP
```
cquiroz: You may not need `MonadState` here. It should be possible to do the `get`, `modify` calls directly on `StateT`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -3,36 +3,374 @@
...
+  // For debugging
+  def printSequenceState(id: Sequence.Id): HandleP[Unit, D] =
+    getSs(id)((qs: Sequence.State) => Task.now(println(qs)).liftM[({type L[M[_], A] = HandleStateT[M, A, D]})#L]).void // scalastyle:ignore
+
+}
+
+object Engine {
 
-  final case class State(conditions: Conditions, operator: Option[Operator], sequences: Map[Sequence.Id, Sequence.State])
+  final case class State[D](userData: D, conditions: Conditions, operator: Option[Operator], sequences: Map[Sequence.Id, Sequence.State])
```
cquiroz: Will `userData` contain the calibrations queue?

cquiroz: Do we really need to be polymorphic here? I suppose in most cases the `UserData` will be a well known structure. I question this due to the complexity introduced.

Also, should/could we put conditions/operator inside `D`?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/UserEvent.scala
```diff
@@ -17,66 +17,71 @@ import scalaz.syntax.std.option._
 /**
   * Events generated by the user.
   */
-sealed trait UserEvent {
+sealed trait UserEvent[+D] {
```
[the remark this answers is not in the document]
jluhrs: Because if you do that, then you have `final case class Foo[D] extends UserEvent`. If you then match on a value of type `UserEvent`, you have to consider all the possible `Foo` for every possible `D` (which does not have a restriction, so it can be anything)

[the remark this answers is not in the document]
jluhrs: Is the difference between only `Foo[D]` can be of type `UserEvent[D]`, and any `Foo[A]` can be of type `UserEvent`.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -106,30 +103,30 @@ package object engine {
     * It's named `Handle` after `fs2.Handle` in order to give a hint in a future
     * migration.
     */
-  type Handle[A] = HandleStateT[Task, A]
+  type Handle[A, D] = HandleStateT[Task, A, D]
   // Helper alias to facilitate lifting.
-  type HandleStateT[M[_], A] = StateT[M, Engine.State, A]
+  type HandleStateT[M[_], A, D] = StateT[M, Engine.State[D], A]
```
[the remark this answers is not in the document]
jluhrs: That is old code, I'm not sure why it exists. I guess it was introduced to reduce the number of type parameters and avoid using type lambdas in some places.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
jluhrs: Parameter `sequences` was not used. The one used has always lived in `Engine.State`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
jluhrs: Yes, `userData` contains the calibration queue, and anything else we could need in the future. I want to put conditions/operator inside, but for that we need a mechanism to pass them to inside the Actions.

jluhrs:
For a long time now my idea for the design of the Seqexec server has been to isolate the process of executing the sequences from the rest of the Seqexec code, that's why I put the execution engine in its own package, `seqexec.engine`. The code specific to the Seqexec service goes into `seqexec.server`, which uses `seqexec.engine` to actually execute the sequences.
I wanted to keep `seqexec.engine` as simple as possible, and consolidate its implementation. It is a complex code. That is the main reason I would like to reach a point where it could be just a black box we use and don't have to change. But for several reasons the development has not gone as I planned. For a while now `seqexec.engine` has being modified for almost every new feature we have added to Seqexec, mainly because it is the core of the application, the "game engine", and adding things there was quicker that building an abstraction. But the fact is that for most of the things added to `seqexec.engine`, the engine does not care. They are pieces of data that are passed through, and are not needed for the actual execution or execution control.
This PR is my first step in trying to remedy this situation. I hope in the future to also remove the observing conditions, the file id, the operator name, the observer name, the Action result type, and the type of  the Action generator input. That follows a principle of keeping the code agnostic of the types used unless it really need to know them. The main advantages are that it will simplify the code in `seqexec.engine`, and future changes (and there will be future changes, a lot of them) will not affect it, which, as I said before, is a complex code that is bette rleft alone.
That is my plan. The introduction of the type parameter `D` is one possible realization of my design, but there are other alternatives. The ugly side of introducing a parametric type `D` is that the code of `seqexec.engine` as several interconnected types, and the parametric type has to go in all of them.

cquiroz:
I understand that point. And maybe that's worth the added complexity. I guess eventually the same engine could be used e.g. for the TCC and then having a neutral type would be an advantage.

I'd like you to progress on this before merging. Making sure that e.g. putting `Conditions` and `Observer` on `D`, can be done with materializing `D`

jluhrs:
One alternative would be to introduce type aliases at the package level, and have all the rest of types use the aliases. The bad thing is that it keeps a dependency in `seqexec.engine` on external types it does not really care, but it is very localized and easy to change.
I wish there was a way to have a parametric package, where you classes that use a package set the types, and then every class in the package uses those types

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Event.scala
```diff
@@ -11,47 +11,48 @@ import scalaz.concurrent.Task
 import scalaz.stream.Process
 
 /**
-  * Anything that can go through the Event Queue.
+  * Anything that can go through the Event[D] Queue.
   */
-sealed trait Event
-final case class EventUser(ue: UserEvent) extends Event
-final case class EventSystem(se: SystemEvent) extends Event
+sealed trait Event[+D]
```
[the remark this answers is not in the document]
jluhrs: So we can have object Events.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -5,23 +5,102 @@ package edu.gemini.seqexec.engine
...
+  implicit def engineInstance: Catchable[Handle] =
+    new Catchable[Handle] {
+      def attempt[A](a: Handle[A]): Handle[Throwable \/ A] = a.flatMap(
+        x => Catchable[Task].attempt(Applicative[Task].pure(x)).liftM[HandleStateT]
+      )
+      def fail[A](err: Throwable): Handle[A] = Catchable[Task].fail[A](err).liftM[HandleStateT]
+    }
+
+  /*
+   * HandleP is a Process which has as a side effect a State machine inside a Task, which can produce other
```
cquiroz: Thanks!

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -5,23 +5,102 @@ package edu.gemini.seqexec.engine
...
+   * Making it final causes an error: "The outer reference in this type test cannot be checked at run time"
+   */
+  case class HandleP[A](run: Handle[(A, Option[Process[Task, EventType]])])
+  object HandleP {
+    def fromProcess(p: Process[Task, EventType]): HandleP[Unit] = {
+      HandleP[Unit](Applicative[Handle].pure[(Unit, Option[Process[Task, EventType]])](((), Some(p))))
+    }
+  }
+
+  implicit def handlePInstances: Applicative[HandleP] with Monad[HandleP] = new Applicative[HandleP] with Monad[HandleP] {
```
cquiroz: You only need to define `Monad`. `Applicative` is implicit with `Monad`

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/package.scala
```diff
@@ -7,30 +7,20 @@ import edu.gemini.seqexec.engine.Result.{Error, PartialVal, PauseContext, RetVal
...
-  final case class HandleP[A, D](run: Handle[(A, Option[Process[Task, Event[D]]]), D])
-  object HandleP {
-    def fromProcess[D](p: Process[Task, Event[D]]): HandleP[Unit, D] = HandleP[Unit, D](Applicative[Handle[?, D]].pure[(Unit, Option[Process[Task, Event[D]]])](((), Some(p))))
-  }
   final case class ActionMetadata(conditions: Conditions, operator: Option[Operator], observer: Option[Observer])
   object ActionMetadata {
     val default: ActionMetadata = ActionMetadata(Conditions.default, None, None)
   }
 
+  trait ActionMetadataGenerator[T] {
```
cquiroz: Could you document this and the `generate` method?

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -124,16 +143,17 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
 
   def eventProcess(q: EventQueue): Process[Task, SeqexecEvent] =
-    executeEngine.process(wye(q.dequeue, seqQueueRefreshProcess)(mergeHaltBoth))(Engine.State.empty[ExecutionQueues](Map("Calibration Queue" -> Nil))).flatMap(x => Process.eval(notifyODB(x))).map {
+    executeEngine.process(wye(q.dequeue, seqQueueRefreshProcess)(mergeHaltBoth))(Engine.State.empty[EngineMetadata](EngineMetadata.default)).flatMap(x => Process.eval(notifyODB(x))).map {
       case (ev, qState) =>
         toSeqexecEvent(ev)(
           SequencesQueue(
-            qState.conditions,
-            qState.operator,
+            (Engine.State.userDataL ^|-> EngineMetadata.conditionsL).get(qState),
```
cquiroz: nice

modules/edu.gemini.seqexec.server/src/test/scala/edu/gemini/seqexec/server/SeqexecEngineSpec.scala
```diff
@@ -155,128 +156,184 @@ class SeqexecEngineSpec extends FlatSpec with Matchers {
...
+
+  }
+
+  "SeqexecEngine setWaterVapor" should "set Water Vapor condition" in {
+    val wv = WaterVapor.Percent80
+    val s0 = Engine.State.empty(EngineMetadata.default)
+    val q = async.boundedQueue[executeEngine.EventType](10)
+    val sf = advanceN(q, s0, seqexecEngine.setWaterVapor(q, wv, UserDetails("", "")), 2)
+
+    inside(sf.map((Engine.State.userDataL ^|-> EngineMetadata.conditionsL ^|-> Conditions.wv).get(_))) {
```
cquiroz: inside?

jluhrs: Yes, from scalatest. It is used all over the tests.

PR #481, opened 2018-02-21
SEQNG-540: Added code to display disperser column for GMOS-S

sraaphorst:
This adds, for GMOS-S, a column in the steps table for the disperser, with the form:

`${dispersername} @ ${centralWavelength} nm`

screen shot 2018-02-19 at 6 08 14 pm

### DIFF
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
@@ -158,4 +158,30 @@ object enumerations {
       "CUSTOM_MASK" -> "Custom Mask"
     )
   }
+
+  object disperser {
+    val GmosSDisperser: Map[String, String] = Map(
+      "MIRROR"      -> "Mirror",
+      "B1200_G5321" -> "B1200",
+      "R831_G5322"  -> "R831",
+      "B600_G5323"  -> "B600",
+      "R600_G5324"  -> "R600",
+      "R400_G5325"  -> "R400",
+      "R150_G5326"  -> "R150"
+    )
+
+    // TODO: Do we still need to include the obsolete filters here?
+    // TODO: For now, omitting. See GmosNorthType.DisperserNorth if needed.
+    // TODO: If needed, the reverse mapping will not be well-defined,
+    // TODO: e.g. B600 maps to B600_G5303 and B600_G5307.
+    val GmosNDisperser: Map[String, String] = Map(
+      "MIRROR"      -> "Mirror",
+      "B1200_G5301" -> "B1200",
+      "R831_G5302"  -> "R831",
+      "B600_G5307"  -> "B600",
+      "R600_G5304"  -> "R600",
+      "R400_G5305"  -> "R400",
+      "R150_G5308"  -> "R150"
+    )
+  }
 }
--- a/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
+++ b/modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
@@ -170,10 +170,6 @@ object enumerations {
       "R150_G5326"  -> "R150"
     )
 
-    // TODO: Do we still need to include the obsolete filters here?
-    // TODO: For now, omitting. See GmosNorthType.DisperserNorth if needed.
-    // TODO: If needed, the reverse mapping will not be well-defined,
-    // TODO: e.g. B600 maps to B600_G5303 and B600_G5307.
     val GmosNDisperser: Map[String, String] = Map(
       "MIRROR"      -> "Mirror",
       "B1200_G5301" -> "B1200",
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepSettings.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepSettings.scala
@@ -6,7 +6,7 @@ package edu.gemini.seqexec.web.client.components.sequence.steps
 import edu.gemini.seqexec.model.Model.{FPUMode, Instrument, SequenceId, Step, StepType, StepState}
 import edu.gemini.seqexec.web.client.actions.{NavigateSilentTo, FlipSkipStep, FlipBreakpointStep}
 import edu.gemini.seqexec.model.enumerations
-import edu.gemini.seqexec.web.client.circuit.{SeqexecCircuit, ClientStatus, StepsTableFocus}
+import edu.gemini.seqexec.web.client.circuit.{ClientStatus, SeqexecCircuit, StepsTableFocus}
 import edu.gemini.seqexec.web.client.components.SeqexecStyles
 import edu.gemini.seqexec.web.client.model.Pages
 import edu.gemini.seqexec.web.client.lenses._
@@ -27,6 +27,7 @@ import scalaz.syntax.show._
 import scalaz.syntax.std.option._
 import scalaz.std.anyVal._
 import scalaz.std.string._
+
 
 /**
  * Component to display an icon for the state
@@ -187,6 +188,42 @@ object FilterCell {
       <.div(
         SeqexecStyles.componentLabel,
         filter.getOrElse("Unknown"): String
+      )
+    }
+    .build
+
+  def apply(p: Props): Unmounted[Props, Unit, Unit] = component(p)
+}
+
+/**
+ * Component to display the disperser and wavelength
+ */
+object DisperserCell {
+  final case class Props(s: Step, i: Instrument)
+
+  private val component = ScalaComponent.builder[Props]("DisperserCell")
+    .stateless
+    .render_P { p =>
+
+      val nameMapper: Map[String, String] = p.i match {
+        case Instrument.GmosS => enumerations.disperser.GmosSDisperser
+        case Instrument.GmosN => enumerations.disperser.GmosNDisperser
+        case _                => Map.empty
+      }
+
+      val disperser = for {
+        disperser <- instrumentDisperserO.getOption(p.s)
+      } yield nameMapper.getOrElse(disperser, disperser)
+      val centralWavelength = instrumentDisperserLambdaO.getOption(p.s)
+
+      // Formatter
+      val displayedText = (disperser, centralWavelength) match {
+        case (Some(d), Some(w)) => f"$d @ $w%.0f nm"
+        case _                  => "Unknown"
+      }
+      <.div(
+        SeqexecStyles.componentLabel,
+        displayedText
       )
     }
     .build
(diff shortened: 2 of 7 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -165,28 +169,36 @@ object StepsTable {
...
           Column(Column.props(ColWidths.IdxWidth, "idx", label = "Step", disableSort = true, flexShrink = 0, cellRenderer = stepIdRenderer)).some,
           p.steps.map(i => Column(Column.props(ColWidths.StateWidth, "state", label = "Control", flexShrink = 1, flexGrow = 4, disableSort = true, cellRenderer = stepProgressRenderer(i, p)))),
           offsetColumn.filter(_ => offsetVisible),
-          p.steps.map(i => Column(Column.props(ColWidths.ExposureWidth, "exposure", label = "Exposure", flexShrink = 0, disableSort = true, className = SeqexecStyles.centeredCell.htmlClass, cellRenderer = stepExposureRenderer(i.instrument)))).filter(_ => exposureVisible),
+          p.steps.map(i => Column(Column.props(ColWidths.ExposureWidth, "exposure", label = "Exposure", flexShrink = 0, disableSort = true, className = SeqexecStyles.centeredCell.htmlClass, cellRenderer = stepExposureRenderer(i.instrument)))).filter(_ => exposureVisible)
+        ) ++
+          (if (hasDisperser) List(
```
sraaphorst: There has to be a much, much better way of doing this. @cquiroz, I think we should discuss this and see what you think the best plan is.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepSettings.scala
```diff
@@ -192,6 +193,42 @@ object FilterCell {
...
+      val disperser = for {
+        disperser <- instrumentDisperserO.getOption(p.s)
+      } yield nameMapper.getOrElse(disperser, disperser)
+      val centralWavelength = instrumentDisperserLambdaO.getOption(p.s)
+
+      // Formatter
+      val displayedText = (disperser, centralWavelength) match {
+        case (Some(d), Some(w)) => f"$d @ $w%.0f nm"
+        case _                  => "Unknown"
+      }
```
sraaphorst: Actually, I think now that I'm looking over this that it might need fixing for `Mirror`, since that has no central wavelength and thus `Unknown` will be returned.

I tried to do this by lifting a function `(String, Double) => String` over the `Option`s in a number of ways, e.g.:

`val displayedText = (disperser |@| centralWavelength)((d: String, w: Double) => f"$d @ $w%.0f nm").getOrElse("Unknown")`

However, they all complained about implicits and I couldn't get this to work without importing all of scalaz. This, however, caused issues with operators imported from:

`import japgolly.scalajs.react.vdom.html_<^._`

I'll fix this in another commit for mirror.

sraaphorst:
I am adding a screenshot of the code fixed for mirror, which has no central wavelength defined.
screen shot 2018-02-21 at 1 19 51 pm

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
```diff
@@ -158,4 +158,30 @@ object enumerations {
       "CUSTOM_MASK" -> "Custom Mask"
     )
   }
+
+  object disperser {
+    val GmosSDisperser: Map[String, String] = Map(
+      "MIRROR"      -> "Mirror",
+      "B1200_G5321" -> "B1200",
```
cquiroz: Do we display the right-hand side right? Science wanted the names without the gemini code

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
```diff
@@ -158,4 +158,30 @@ object enumerations {
...
+      "R831_G5322"  -> "R831",
+      "B600_G5323"  -> "B600",
+      "R600_G5324"  -> "R600",
+      "R400_G5325"  -> "R400",
+      "R150_G5326"  -> "R150"
+    )
+
+    // TODO: Do we still need to include the obsolete filters here?
```
cquiroz: I don't think we need them as you wouldn't load sequneces with old filters

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -165,28 +169,36 @@ object StepsTable {
 
     // Columns for the table
     private def columns(p: Props, s: Size): List[Table.ColumnArg] = {
-      val (offsetVisible, exposureVisible, fpuVisible, filterVisible, objectSize) = s.width match {
-        case w if w < PhoneCut      => (false, false, false, false, SSize.Tiny)
-        case w if w < LargePhoneCut => (false, true, false, false, SSize.Small)
-        case _                      => (displayOffsets(p), true, true, true, SSize.Small)
+      val (offsetVisible, exposureVisible, disperserVisible, fpuVisible, filterVisible, objectSize) = s.width match {
+        case w if w < PhoneCut      => (false, false, false, false, false, SSize.Tiny)
+        case w if w < LargePhoneCut => (false, true, false, false, false, SSize.Small)
```
cquiroz: Did you test on a narrower screen? This part of the code will hide columns when we can't display all of them, it would be good to verify how it looks on smaller screen (Responsive mode)

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -165,28 +169,36 @@ object StepsTable {
...
       }
       val offsetColumn =
         p.offsetsDisplay match {
           case OffsetsDisplay.DisplayOffsets(x) =>
             Column(Column.props(ColWidths.OffsetWidthBase + x, "offset", label = "Offset", flexShrink = 0, disableSort = true, cellRenderer = stepStatusRenderer(p.offsetsDisplay))).some
           case _ => None
         }
-        List(
+
+        val hasDisperser = p.steps.exists(s => s.instrument === Instrument.GmosS || s.instrument === Instrument.GmosN)
```
cquiroz: Yeah this isn't very nice. I thin we should make this a property of the instrument, on the `Model` class and do the filtering based on that

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
sraaphorst: Yes, we only display the right hand side names. We look up those names with the left hand side and translate to the right hand side.

modules/edu.gemini.seqexec.model/src/main/scala/edu/gemini/seqexec/model/enumerations.scala
sraaphorst: That's what I figured. I didn't include them since then the mapping wouldn't be injective.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
sraaphorst: Here's a screenshot. I'm assuming that this is what we want, based on the other parameters that are hidden when real estate is small?
screen shot 2018-02-21 at 4 27 55 pm

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
sraaphorst: Yes, this definitely is not the way to do it. I didn't want to make any substantial changes until I confirmed it was working with this easy hack, and until I talked to you to see what you thought was the best way moving forward.

When you say "do the filtering based on that", what do you mean?  I'm not sure about this list of columns. Will they be substantially different as we continue to add instruments?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
cquiroz: Thanks, that seems correct

sraaphorst: Oh, that was my other question, though: is this the position that we want the disperser column in in the table? I don't know.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
cquiroz: Maybe the correct word isn't `filtering`. I think instruments can have a `hasDisperser` property and we include the disperser column based on that

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
cquiroz: Can you ask @andrewwstephens about that?

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
sraaphorst: Sure thing! Will do.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/steps/StepsTable.scala
andrewwstephens: The order looks good to me.

PR #483, opened 2018-02-22
SEQNG-550 Don't try to retrieve guide options if not used.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/tcs/Tcs.scala
```diff
@@ -155,12 +155,14 @@ object Tcs {
...
-      build(buildPwfs1Config,  TELESCOPE_KEY / GUIDE_WITH_PWFS1_PROP, config),
-      build(buildPwfs2Config,  TELESCOPE_KEY / GUIDE_WITH_PWFS2_PROP, config),
-      build(buildOiwfsConfig,  TELESCOPE_KEY / GUIDE_WITH_OIWFS_PROP, config),
-      build(buildOffsetConfig, TELESCOPE_KEY / P_OFFSET_PROP, TELESCOPE_KEY / Q_OFFSET_PROP, config)
-    ).suml.apply(s0)
+      subs.contains(Subsystem.P1WFS).option(build(buildPwfs1Config, TELESCOPE_KEY / GUIDE_WITH_PWFS1_PROP, config)),
+      subs.contains(Subsystem.P2WFS).option(build(buildPwfs2Config, TELESCOPE_KEY / GUIDE_WITH_PWFS2_PROP, config)),
+      subs.contains(Subsystem.OIWFS).option(build(buildOiwfsConfig, TELESCOPE_KEY / GUIDE_WITH_OIWFS_PROP, config)),
+      subs.contains(Subsystem.Mount).option(build(buildOffsetConfig, TELESCOPE_KEY / P_OFFSET_PROP, TELESCOPE_KEY / Q_OFFSET_PROP, config))
+    ).collect{case Some(x) => x}.suml.apply(s0)
```
sraaphorst: Couldn't you just do `.flatten` here instead of `.collect`?
I think it removes all `None` entries and unpacks the `Some(x)` to `x`.

sraaphorst:
```
scala> List(5.some, None, 3.some, None).flatten
res14: List[Int] = List(5, 3)
```

cquiroz: Wart remover doesn't let us use `flatten` for some reason I don't fully understand 😄

jluhrs: Wartremover didn't let me.

sraaphorst:
That is weird... there is no mention of it on the wartremover webpage in their list of "no-nos."
Good to know since I will be held to its standards now. 😸

cquiroz:
i'm not that happy about wartremover high standards

swalker2m:
I've got to defend this wart-removal.   `flatten` is supposed to flatten nested `F[F[A]]` into `F[A]`.  When you have a `List[Option[A]]` `flatten` only works because of an implicit conversion.

sraaphorst:
Does wart-removal always complain about the use of implicit conversions? I can understand why it's good to be aware that such things are being used, but they are supposed to be convenient, and having to write `collect { Some(x) => x }` is a bit tedious.

Would wart-remover not complain if I wrote:
```
List(1.some, None, 2.some, None).map(_.toList).flatten
```

That being said, being a masochist and reading "Effective Modern C++" right now for fun, with the many problems that come up from the compiler doing implicit conversions and the myriad nonsense you have to memorize to make sure you are getting the results you expect, I can see why this is justified.

Example in C++:

```
class MyClass {
public:
  MyClass(int, bool);  // ctor1
  MyClass(std::initializer_list<int>); // ctor2
}
...
MyClass { 5, true };
```

C++14 compilers will actually implicitly convert the boolean `true` to `1` and attempt to use `ctor2` rather than `ctor1`, despite `ctor1` being a perfect match for the parameters.

swalker2m:
If you explicitly convert the `Option` to a `List` it shouldn't complain.  It's clear what is going on when you do it that way.

PR #486, opened 2018-02-23
SEQNG-413A: URL markers for step id

cquiroz:
This is the first part of a change to get the urls to contain the step. Once I started I noted several small defects that could be improved and they are contained on this PR:

* Display the routes on the title

screenshot 2018-02-23 17 47 16

* Have proper links displayed when hovering on a table
monosnap 2018-02-23 17-52-23

* Some small improvements on what's loaded when a sequence is running or the page is just launched with elements on the queue

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
@@ -21,7 +21,6 @@ import org.scalajs.dom.html.TableRow
 import scalacss.ScalaCssReact._
 import scalaz.syntax.show._
 import scalaz.syntax.equal._
-import scalaz.syntax.std.option._
 
 object QueueTableBody {
   type SequencesModel = ModelProxy[StatusAndLoadedSequencesFocus]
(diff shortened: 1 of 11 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
```diff
@@ -51,7 +50,7 @@ object QueueTableBody {
 
   def showSequence(p: Props, s: SequenceInQueue): Callback =
     // Request to display the selected sequence
-    p.sequences.dispatchCB(NavigateTo(InstrumentPage(s.instrument, s.id.some))) >> p.sequences.dispatchCB(SelectIdToDisplay(s.id))
+    p.sequences.dispatchCB(NavigateTo(SequencePage(s.instrument, s.id, 0))) >> p.sequences.dispatchCB(SelectIdToDisplay(s.id))
```
jluhrs: I am guessing the `0` here means the first step, which works only because we are using the step position as its unique id. But at some point we should change it to use the `Step` id type used by the ODB, and also use the `Location` concept (introduced in gem ?). Would this be a good time ?

cquiroz: SEQNG-413 requires including the step in the url, that's what this comes from. At the moment this is unused so I leave it at 0.

I'm not aware of `Location` in gem

jluhrs: That's something @swalker2m mentioned.

PR #488, opened 2018-02-27
SEQNG-525: Fix border for tables shorter than full heigh

cquiroz:
These are some improvements to the way tables are displayed. For example, when a table was too short the borders didn't extend all the way down and we'd get double borders sometimes

The difference is subtle but enough for me to notice it


cquiroz:
There are several tiny changes. This is the most important

before:
before-fix

after:
after-fix

sraaphorst:
Thanks for humouring me!
Looks great. I can see the subtle differences and they are definitely an improvement.

PR #489, opened 2018-03-01
SEQNG-554 Moved SequenceEnd notification to ODB out of last step.

jluhrs:
This solves the issue of the last step not being marked as completed if exposure is stopped, and also ensures the notification is sent, even if the last step is skipped.

### DIFF
--- a/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
+++ b/modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
@@ -10,7 +10,7 @@ import edu.gemini.epics.acm.CaService
 import edu.gemini.pot.sp.SPObservationID
 import edu.gemini.spModel.core.Site
 import edu.gemini.seqexec.engine
-import edu.gemini.seqexec.engine.{Action, Engine, Event, EventSystem, Executed, Failed, Sequence}
+import edu.gemini.seqexec.engine.{Step => _, _}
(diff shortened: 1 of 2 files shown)


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -10,7 +10,7 @@ import edu.gemini.epics.acm.CaService
 import edu.gemini.pot.sp.SPObservationID
 import edu.gemini.spModel.core.Site
 import edu.gemini.seqexec.engine
-import edu.gemini.seqexec.engine.{Action, Engine, Event, EventSystem, Executed, Failed, Sequence}
+import edu.gemini.seqexec.engine.{Step => _, _}
```
sraaphorst: Did not know about `import ... A => _`
I suppose that means ignore `Step` in the package import, because you're using `Step` as defined by another import?

jluhrs: That was Idea's magic. I thing it reduced all the imports from `engine` to all minus Step.

sraaphorst: Interesting. Idea does some very funny things sometimes.

sraaphorst:
I wonder if this has anything to do with me merging my changes after this was submitted?

git should be able to handle that just fine, but who knows? If this doesn't work, do a pull and a rebase?

PR #490, opened 2018-03-07
SEQNG-502: Convert the queue table to use react-virtualized

cquiroz:
This PR will convert the queue table to `react-virtualized`. This will unify the look and feel also keep the headers always in place

seqexec - gs-2018a-q-1-58 2018-03-07 15-45-24

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
@@ -47,12 +47,12 @@ object QueueTableBody {
     var targetName: Option[String]
     var name: String
     var active: Boolean
-    var runningStep: Option[(Int, Int)]
+    var runningStep: Option[RunningStep]
   }
   // scalastyle:on
   object QueueRow {
     @SuppressWarnings(Array("org.wartremover.warts.AsInstanceOf"))
-    def apply(obsId: SequenceId, status: SequenceState, instrument: Instrument, targetName: Option[String], name: String, active: Boolean, runningStep: Option[(Int, Int)]): QueueRow = {
+    def apply(obsId: SequenceId, status: SequenceState, instrument: Instrument, targetName: Option[String], name: String, active: Boolean, runningStep: Option[RunningStep]): QueueRow = {
       val p = (new js.Object).asInstanceOf[QueueRow]
       p.obsId = obsId
       p.status = status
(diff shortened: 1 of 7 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
```diff
@@ -3,152 +3,192 @@
...
+  // ScalaJS defined trait
+  // scalastyle:off
+  trait QueueRow extends js.Object {
+    var obsId: SequenceId
+    var status: SequenceState
+    var instrument: Instrument
+    var targetName: Option[String]
+    var name: String
+    var active: Boolean
+    var runningStep: Option[(Int, Int)]
```
sraaphorst: I'm a bit confused about what the type of `runningStep` indicates. Could you explain what the significance of the `Option[(Int,Int)]` is?

cquiroz: Yeah, I should probably use a case class for that. It is basically `runningStep/totalSteps`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
```diff
@@ -3,152 +3,192 @@
...
+    var instrument: Instrument
+    var targetName: Option[String]
+    var name: String
+    var active: Boolean
+    var runningStep: Option[RunningStep]
+  }
+  // scalastyle:on
+  object QueueRow {
+    @SuppressWarnings(Array("org.wartremover.warts.AsInstanceOf"))
+    def apply(obsId: SequenceId, status: SequenceState, instrument: Instrument, targetName: Option[String], name: String, active: Boolean, runningStep: Option[RunningStep]): QueueRow = {
```
sraaphorst: I'm just curious about this `targetName: Option[String]` in the new table... is there a reason why we'd want to show `Some target` and not the actual target name?

(Sorry... I'm having a little trouble wrapping my head around this PR.)

cquiroz: the `targetName` is optional because it's obscured to non-authenticated users. For those users we remove it completely not only from the UI but also from the data model, hence the `Option`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
```diff
@@ -162,7 +202,6 @@ object QueueTableSection {
     .stateless
     .render_P(p =>
       <.div(
-        ^.cls := "ui segment scroll pane",
```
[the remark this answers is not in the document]
cquiroz: they are orthogonal. `react-virtualized` removes the need for `semantic-ui` tables, but we still use a lot of other components from `semantic-ui`

PR #493, opened 2018-03-14
SEQNG-552 Set EPICS IO timeout.

jluhrs:
Made the EPICS IO configurable, using the last feature of `epics-service`.

### DIFF
--- a/app/seqexec-server-gn/src/main/resources/app.conf
+++ b/app/seqexec-server-gn/src/main/resources/app.conf
@@ -39,6 +39,7 @@ seqexec-engine {
     odbQueuePollingInterval = "3 seconds"
     tops = "tcs=tcs:, ao=ao:, gm=gm:, gc=gc:, gws=ws:, m2=m2:, oiwfs=oiwfs:, ag=ag:, f2=f2:"
     epics_ca_addr_list = "10.2.2.255 10.2.10.21 10.2.126.101"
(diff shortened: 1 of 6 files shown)


app/seqexec-server-gn/src/main/resources/app.conf
```diff
@@ -39,6 +39,7 @@ seqexec-engine {
     odbQueuePollingInterval = "3 seconds"
     tops = "tcs=tcs:, ao=ao:, gm=gm:, gc=gc:, gws=ws:, m2=m2:, oiwfs=oiwfs:, ag=ag:, f2=f2:"
     epics_ca_addr_list = "10.2.2.255 10.2.10.21 10.2.126.101"
+    ioTimeout = "5 seconds"
```
cquiroz: Where the 5 second come from? Or is it just to start testing?

jluhrs: It's just to start testing.

modules/edu.gemini.epics.acm/src/main/java/edu/gemini/epics/acm/CaService.java
```diff
@@ -5,6 +5,7 @@
```
sraaphorst: Just out of my own personal random curiosity, why are we still using Java code here?

PR #494, opened 2018-03-14
SEQNG-549 Added FITS keyword SEQEXVER.

jluhrs:


cquiroz:
The PR is fine but let's wait a bit to discuss the version formatting? @arturog8m what do you think?

jluhrs:
I think that this PR and the definition of a format for the version are two separate issues. The later should not prevent the merge of this PR.

cquiroz:
They are separate issues but with this PR the version format will be forever on FITS files and I think we don't want to change the format later on. Maybe it is a non-issue, i don't know

PR #495, opened 2018-03-15
SEQNG-552 Fixed bug in setting EPICS IO timeout.

jluhrs:


modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
```diff
@@ -531,10 +531,10 @@ object SeqexecEngine {
     // the configuration file or from the environment
     val caInit   = caAddrList.map(a => Task.delay(CaService.setAddressList(a))).getOrElse {
       Task.delay(Option(System.getenv("EPICS_CA_ADDR_LIST"))).flatMap {
-        case Some(_) => Task.delay(CaService.setIOTimeout(java.time.Duration.ofMillis(ioTimeout.toMillis)))
```
jluhrs: Wrong place to put this, because here it is only executed if caAddrList is not set.

jluhrs:
The initialization of the single CaService relies on side effects for its parameters. That should be changed. The initialization should be done explicitly, and pass the parameters there.

modules/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/SeqexecEngine.scala
cquiroz: right, i should have seen this too

PR #496, opened 2018-03-15
SEQNG-356: Performance issues on Firefox

cquiroz:
While doing SEQNG-413 i was hit by some performance issues as the screen was updated too often. This PR fixes that by improving the equality detection. diode wasn't detecting equality correctly for tuples and that led to the extra screen updates.

Also added logging of some actions to simplify debugging and an endpoint to submit log events for testing purposes


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/circuit.scala
```diff
@@ -50,20 +55,59 @@ object circuit {
...
+    implicit val eq: Equal[StepsTableFocus] = Equal.equalA
+  }
+  final case class StepsTableAndStatusFocus(status: ClientStatus, stepsTable: Option[StepsTableFocus]) extends UseValueEq
+  final case class ControlModel(id: SequenceId, isPartiallyExecuted: Boolean, nextStepToRun: Option[Int], status: SequenceState) extends UseValueEq
+  final case class SequenceControlFocus(isLogged: Boolean, isConnected: Boolean, control: Option[ControlModel]) extends UseValueEq
+
+  /**
+   * Diode processor to log some of the action to aid in debugging
+   */
+  final class LoggingProcessor[M <: AnyRef] extends ActionProcessor[M] {
```
sraaphorst: I'm just curious why you do `M <: AnyRef` here. I don't think I've ever seen anyone do `M <: AnyVal` or `M <: AnyRef` before. Is it somehow important?

sraaphorst:
Looks fine to me.

I'm surprised that Diode isn't checking tuple equality correctly. It seems like something you just come to expect.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/circuit.scala
cquiroz: That is the signature required for `ActionProcessor`. `Actions` in diode can be anything though we normally use a marker interface `Action`. So ideally this should be `LoggingProcessor[Action]` but it won't be acceptable for the call `addActionProcessor`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/handlers.scala
```diff
@@ -396,16 +395,16 @@ object handlers {
           case buffer: ArrayBuffer =>
             val byteBuffer = TypedArrayBuffer.wrap(buffer)
             \/.fromTryCatchNonFatal(Unpickle[SeqexecEvent].fromBytes(byteBuffer)) match {
-              case \/-(event) => logger.info(s"Decoding event: ${event.getClass}"); SeqexecCircuit.dispatch(ServerMessage(event))
-              case -\/(t)     => logger.warning(s"Error decoding event ${t.getMessage}")
+              case \/-(event: ServerLogMessage) => SeqexecCircuit.dispatch(ServerMessage(event))
+              case \/-(event)                   => logger.info(s"Decoding event: ${event.getClass}"); SeqexecCircuit.dispatch(ServerMessage(event))
```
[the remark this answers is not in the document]
cquiroz: Fixed

cquiroz:
I was surprised too, the documentation states that using diode's `zip` method it should do the equality correctly but it doesn't, thus i replaced all the tuples by case classes

i added the log testing method as you can see that when adding a `log` entry and the steps table would be updated.

sraaphorst:
Thanks for the explanation, and I was only teasing about the semicolon. It's just always weird to see one in Scala code 👍 .

PR #497, opened 2018-03-16
SEQNG-569: Don't broadcast SequenceRefreshed

cquiroz:
We have discussed for a while to have the ability to send messages directly to en client instead of broadcasting. This PR adds this ability:

1. At startup the client connects and gets an id (UUID)
2. Some requests and responses carry this ID and will be sent to only that client
3. In particular as soon as a client connects it request a state update which is sent on a `SequenceRefreshed` event.
4. With this PR this message will be sent only to that client instead of being broadcasted to all of them

This fixes SENG-569 and SEQNG-371


modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/SharedModelArbitraries.scala
```diff
@@ -26,6 +26,12 @@ object SharedModelArbitraries {
     } yield SequencesQueue(Conditions.default, Some(Operator("operator")), b)
   }
 
+  implicit val clientIdArb: Arbitrary[ClientID] = Arbitrary {
+    for {
+      i <- Gen.uuid
+    } yield i
+  }
+
```
sraaphorst: Can't you just do:
`implicit val clientIdArb: Arbitrary[ClientID] = Arbitrary(Gen.uuid)`
?

modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/SharedModelArbitraries.scala
```diff
@@ -26,6 +26,12 @@ object SharedModelArbitraries {
...
   implicit val instArb: Arbitrary[Instant] = Arbitrary {
```
sraaphorst: I think you could do the same thing here:
`implicit val instArb = Arbitrary(Gen.choose(0L, Long.MaxValue).map(Instant.ofEpochMilli(_)))`

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -116,15 +119,24 @@ class SeqexecUIApiRoutes(devMode: Boolean, auth: AuthenticationService, events:
...
-        WS(
-          Exchange(
-            Process.emit(Binary(trimmedArray(ConnectionOpenEvent(user.toOption)))) ++
-              (pingProcess merge engineOutput.subscribe.map(anonymizeF).filter(filterOutNull).map(v => Binary(trimmedArray(v)))),
-            scalaz.stream.Process.empty
-          )
-        )
+        // Create a client specific process
+        for {
+          clientId <- Task.delay(UUID.randomUUID())
```
sraaphorst: I'm curious why you wrapped the `UUID.randomUUID()` call in a `Task.delay`.

(I'm sure there's a good reason for it. I'm just not sure why you want to delay the creation of the random UUID.)

cquiroz: `Task` wraps the call to `randomUUID` making it pure and allowing me to return a pure `Task` for http4s to reuse

modules/edu.gemini.seqexec.model/src/test/scala/edu/gemini/seqexec/model/SharedModelArbitraries.scala
cquiroz: No reason, I'll change it as you suggest

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.server/src/main/scala/edu/gemini/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
sraaphorst: 👍 
I figured it was something like that, but I wasn't sure.

PR #498, opened 2018-03-20
Updated epics libraries

cquiroz:


sraaphorst:
Seems there were some errors, mostly around channels not being connected?

cquiroz:
Yeah, its failing locally too. I didn't expect that so I may revert it

cquiroz:
after some more testing, it's clear the new epics libraries aren't compatible with the acm code.

PR #499, opened 2018-03-20
SEQNG-212: Display an indication on the UI when there are resource conflicts

cquiroz:
When there is a resource conflict (an attempt to run a sequence no an instrument in use) a notification is displayed. However, this is displayed to all the clients. This fixes that and only the user requesting running the sequence will be notified

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/toolbars/SequenceDefaultToolbars.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/toolbars/SequenceDefaultToolbars.scala
@@ -84,7 +84,7 @@ object SequenceControl {
           val runContinueButton = s"${isPartiallyExecuted ? "Continue" | "Run"} from step $nextStepToRun"
           List(
             // Sync button
-            controlButton(IconRefresh, "purple", $.runState(requestSync(id)), (!allowedToExecute || !s.canSync) && !inConflict, s"Sync sequence", s"Sync $inConflict")
+            controlButton(IconRefresh, "purple", $.runState(requestSync(id)), (!allowedToExecute || !s.canSync) && !inConflict, "Sync sequence", "Sync")
               .when(status.isIdle || status.isError),
             // Run button
             controlButton(IconPlay, "blue", $.runState(requestRun(id)), (!allowedToExecute || !s.canRun) && !inConflict, runContinueTooltip, runContinueButton)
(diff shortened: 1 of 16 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/toolbars/SequenceDefaultToolbars.scala
```diff
@@ -78,16 +78,16 @@ object SequenceControl {
...
         control.whenDefined { m =>
-          val ControlModel(id, isPartiallyExecuted, nextStep, status) = m
+          val ControlModel(id, isPartiallyExecuted, nextStep, status, inConflict) = m
           val nextStepToRun = nextStep.getOrElse(0) + 1
           val runContinueTooltip = s"${isPartiallyExecuted ? "Continue" | "Run"} the sequence from the step $nextStepToRun"
           val runContinueButton = s"${isPartiallyExecuted ? "Continue" | "Run"} from step $nextStepToRun"
           List(
             // Sync button
-            controlButton(IconRefresh, "purple", $.runState(requestSync(id)), !allowedToExecute || !s.canSync, "Sync sequence", "Sync")
+            controlButton(IconRefresh, "purple", $.runState(requestSync(id)), (!allowedToExecute || !s.canSync) && !inConflict, s"Sync sequence", s"Sync $inConflict")
```
sraaphorst: I guess you don't need the string interpolation here for `"Sync sequence"` since you are not making any substitutions.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/handlers.scala
```diff
@@ -36,6 +36,7 @@ import scalaz._
 import Scalaz._
 
 object handlers {
+  private val VoidEffect = Effect(Future(NoAction: Action))
 
   trait Handlers {
     implicit def pfMonoid[A, B]: Monoid[PartialFunction[A, B]] = new Monoid[PartialFunction[A, B]] {
```
sraaphorst: It took me a minute to convince myself that this was a `Monoid` as I wasn't convinced that it was associative, but I think I convinced myself that it is.

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/sequence/toolbars/SequenceDefaultToolbars.scala
cquiroz: That's right. I had it for debuggin but will remove it

modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/handlers.scala
cquiroz: I plan to add tests for these once we move to cats. that would verify that monoid laws are upheld

sraaphorst: 👍

PR #500, opened 2018-03-21
Use sbt-dynver to produce the version

cquiroz:
This PR changes the way to calculate the version in such a way that the hash is always shown. The format is
`<last-tag>+<commits-after-last-tag>-<hash>|-<date>-<time>-UNCOMMITED|`

The part between `|` is only used if there are uncommitted changes

There are 4 scenarios:
* We are on a tag without Uncommitted changes, in that case, the version is:
 **0.8.7+0-468910be**

* We are on a tag with uncommmited changes:
 **0.8.7+0-468910be-20180321-1931-UNCOMMITED**

* We have commits after the last tag:
 **0.8.6+17-468910be**

* Commits after the last tag and uncommited changes
 **0.8.7+0-468910be-20180321-1931-UNCOMMITED**

@tpolecat, @swalker2m , @arturog8m  Does this seem right?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -28,14 +28,20 @@ cancelable in Global := true
 
 // Settings to use git to define the version of the project
 def versionFmt(out: sbtdynver.GitDescribeOutput): String = {
-  val dirtySuffix = out.dirtySuffix.dropPlus.mkString("-", "-UNCOMMITED")
-  out.ref.dropV.value + out.commitSuffix.mkString("+", "-", "") + dirtySuffix
+  val dirtySuffix = if (out.dirtySuffix.mkString("", "").nonEmpty) {
+    "-UNCOMMITED"
+  } else {
+    ""
+  }
+  s"-${out.commitSuffix.sha}$dirtySuffix"
 }
 
 def fallbackVersion(d: java.util.Date): String = s"HEAD-${sbtdynver.DynVer timestamp d}"
 
+val dateFormatter = java.time.format.DateTimeFormatter.BASIC_ISO_DATE
+
 inThisBuild(List(
-  version := dynverGitDescribeOutput.value.mkVersion(versionFmt, fallbackVersion(dynverCurrentDate.value)),
+  version := dateFormatter.format(dynverCurrentDate.value.toInstant().atZone(java.time.ZoneId.systemDefault()).toLocalDate()) + dynverGitDescribeOutput.value.mkVersion(versionFmt, fallbackVersion(dynverCurrentDate.value)),
   dynver := {
      val d = new java.util.Date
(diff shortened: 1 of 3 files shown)


tpolecat:
This seems reasonable, although users will have to report the whole string anyway so I'm not sure what it buys us over the hash alone. If it makes people feel better I'm fine with it and will change gem to use this scheme.

Presumably a rule somewhere will prevent us from deploying an `UNCOMMITTED` version, so users will only see versions like `0.8.7+14-468910be`.

andrewwstephens:
Will we make releases without incrementing the first three numbers in the version ID?

cquiroz:
We shouldn't but there is at the moment no mechanics to prevent it.

@tpolecat Do you know if  `sbt-release` could automate the process to ensure, e.g. you couldn't release with uncommited files and ensure a tag is always made?

tpolecat:
If you want to automate the release then CI will have to open a PR with the updated patch version, or have permission to push to master. It's kind of an automation headache.

If we're releasing by hand then yes sbt-release can do this (not out of the box but it would be easy to set up).

cquiroz:
at least on the seqexec side I expect we'll do manual releases for the foreseeable future

jluhrs:
I have used GitHub releases for Seqexec.

cquiroz:
It was discussed in the last Seqexec meeting that an alternative for the version would be just to use the date and hash as proposed by @andrewwstephens 

* For normal releases it would have the format:
**20180403-3c0a1b06**

* If you try to release with uncommited changes you get:
**20180403-e595fca7-UNCOMMITED**

This should be fully automatic and without needing tags. Also, it gives a clear ordering and tells the user how long ago the release was made

The only question is what happens if you do more than one release the same day but even that would get different hashes

I thought one issue is if you want to depend on a module in this repo but I guess we'll be merging this with gem in one big monorepo

@tpolecat, @swalker2m, @arturog8m, @jluhrs, @andrewwstephens Opinions?

tpolecat:
This seems fine to me. If there's more than one release on the same day you have to ask git for the timestamp of the hash. Note that it's still not a total ordering (it's just an illusion to make users feel better) so *we* should never depend on it.

swalker2m:
Seems fine to me too.  What's the timezone used for coming up with the date?

tpolecat:
> What's the timezone used for coming up with the date?

Good point. Should be UTC I guess?

cquiroz:
Good point, otherwise we are dependent on the machine we are building on for TZ

swalker2m:
Yeah, it doesn't seem like it should matter where the release is made.

cquiroz:
i guess this is settled, I'll merge

PR #501, opened 2018-03-22
SEQNG-559: Sound verification

cquiroz:
This PR will let users check if sounds are working on their workstations. This is a bit convoluted as we need to ensure that the sound is triggered the same way as when a sequence runs.

Instead of adding a button to the UI I added a route that can be used. like
<hostname>/soundtest

after loading the regular UI this will emit the `Sequence completed` sound

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/handlers.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/handlers.scala
@@ -85,7 +85,7 @@ object handlers {
       case SyncToRunning(s) =>
         // We'll select the sequence currently running and show the correct url
         value match {
-          case Root | InstrumentPage(_)        =>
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/model/handlers.scala
```diff
@@ -503,6 +503,12 @@ object handlers {
       case _                        => false
     }
 
+    val soundCheck: PartialFunction[Any, ActionResult[M]] = {
```
sraaphorst: It looks so weird to see a `PartialFunction` with domain `Any`. Do you have to do this?
Could it be `Action`?

cquiroz: I've tried that but it doesn't seem possible as the `handle` function in `ActionHandler` is defined as:
```
  protected def handle: PartialFunction[Any, ActionResult[M]]
```

and we need to override it

PR #502, opened 2018-04-03
Lib updates

cquiroz:
Some library updates, most notably `boopickle`


cquiroz:
NIce, 100k less js with these updates

PR #504, opened 2018-04-05
Small simplification to the engine code

cquiroz:
This is a small simplification to `mapEvalState` suggested by @tpolecat. It may help to simplify the fs2 port

### DIFF
--- a/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
+++ b/modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
@@ -342,13 +342,13 @@ class Engine[D: ActionMetadataGenerator, U](implicit ev: ActionMetadataGenerator
   // Kudos to @tpolecat
   /** Traverse a process with a stateful computation. */
   private def mapEvalState[A, S, B](
-                                     fs: Process[Task, A], s0: S, f: A => StateT[Task, S, (B, Option[Process[Task, A]])]
+                                     fs: Process[Task, A], s0: S, f: (A, S) => Task[(S, B, Process[Task, A])]
                                    ): Process[Task, B] = {
     def go(fi: Process[Task, A], si: S): Process[Task, B] = {
       Process.eval(fi.unconsOption).flatMap {
         case None => Process.halt
-        case Some((h, t)) => Process.eval(f(h).run(si)).flatMap {
-          case (s, (b, p)) => Process.emit(b) ++ go(p.map(_ merge t).getOrElse(t), s)
+        case Some((h, t)) => Process.eval(f(h, si)).flatMap {
+          case (s, b, p) => Process.emit(b) ++ go(p merge t, s)
         }
       }
     }
@@ -356,11 +356,14 @@ class Engine[D: ActionMetadataGenerator, U](implicit ev: ActionMetadataGenerator
     go(fs, s0)
   }
 
-  private def runE(ev: EventType): HandleP[(EventType, StateType)] =
-    run(ev).map((ev, _))
+  private def runE(ev: EventType, s: StateType): Task[(StateType, (EventType, StateType), Process[Task, EventType])] =
+    run(ev).run.run(s).map {
+      case (s, (si, p)) =>
+        (s, (ev, si), p.getOrElse(Process.empty))
+    }
 
   def process(input: Process[Task, EventType])(qs: StateType): Process[Task, (EventType, StateType)] =
-    mapEvalState[EventType, StateType, (EventType, StateType)](input, qs, (e: EventType) => runE(e).run)
+    mapEvalState[EventType, StateType, (EventType, StateType)](input, qs, runE)
 
   // Functions for type bureaucracy
 
@@ -410,6 +413,7 @@ object Engine {
 
     def empty[D](userData: D): State[D] = State(userData, Map.empty)
 
+    implicit def equal[D]: Equal[State[D]] = Equal.equalA
   }
 
   abstract class Types {


modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Engine.scala
```diff
@@ -410,6 +413,7 @@ object Engine {
 
     def empty[D](userData: D): State[D] = State(userData, Map.empty)
 
+    implicit def equal[D]: Equal[State[D]] = Equal.equalA
```
tpolecat: I think I would require `Equal[D]` here and contramap from `Equal[(D, Map[Sequence.Id, Sequence.State])]` otherwise it's not going to be lawful in general.

tpolecat:
```scala
implicit def stateEqual[D: Equal]: Equal[State[D]] =
  Equal[(D, Map[Sequence.Id, Sequence.State])]
    .contramap(s => (s.userData, s. sequences))
```

cquiroz: ok, I couldn't do that as i don't have all of them defined. I'll add those.
On the cats port I'm doing `discipline` testing. Could this problem have been detected that way?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -256,6 +256,12 @@ object Sequence {
       Sequence.Zipper.zipper(q).map(Zipper(_, SequenceState.Idle))
         .getOrElse(Final(Sequence.empty(q.id, q.metadata), SequenceState.Idle))
 
+    implicit val eq: Equal[State] = Equal.equal {
```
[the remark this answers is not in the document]
tpolecat: Doing it pairwise is an ok way to do it, but I would say `Equal[Either[zipper fields, final fields]].contramap { ...` where the *fields* need equal instances. You would not normally define `Equal` instances for `Zipper` and `Final` cases because we treat them as data constructors, not data types.

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
```diff
@@ -342,6 +348,10 @@ object Sequence {
 
     }
 
+    object Zipper {
+      implicit val eq: Equal[Zipper] = Equal.equalA
```
tpolecat: I assume this is just a stub. You should't use `.equalA` unless you have a type that you know for sure has a reasonable universal equality.

cquiroz: I assume `case classes` have a good universal equality

cquiroz: Is that a fair assumption? if a case class is composed of case classes it should have a correct `equals` right?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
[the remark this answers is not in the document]
cquiroz: How do you go about defining equal for more complex ADTs. like

```
    sealed trait SequenceState
    case object Completed         extends SequenceState
    case object Idle              extends SequenceState
    case object Stopped           extends SequenceState
    final case class Running(userStop: Boolean, internalStop: Boolean) extends SequenceState
```

Nested `Either`?

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
tpolecat: It's only true if the fields have reasonable equality; i.e., `Int`, `String`, stuff like that; or are products or sums thereof. Eventually it all gets down to universal equality on primitive types.

cquiroz:
Well, it turns out that if I go deep into the model, I'll reach a `Kleisli` with the actual action done by the seqexec inside. I believe I can't make an `Eq` out of that. I don't need the equal instance at this stage in any case

modules/edu.gemini.seqexec.engine/src/main/scala/edu/gemini/seqexec/engine/Sequence.scala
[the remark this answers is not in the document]
tpolecat: In that case you know that `equalA` will work because the only data members are `Boolean`s.

But yeah you could mechanically write it like

```scala
Equal[Either[Unit, Either[Unit, Either[Unit, (Boolean, Boolean)]]]].contramap {
  case Completed     => Left(())
  case Idle          => Right(Left(())
  case Stopped       => Right(Right(Left(()))
  case Running(a, b) => Right(Right(Right(a, b)))
}
```

Which you could simplify to `Option` since `Option[A] ~ Either[Unit, A]

```scala
Equal[Option[Option[Option[(Boolean, Boolean)]]]].contramap {
  case Completed     => None
  case Idle          => Some(None)
  case Stopped       => Some(Some(None))
  case Running(a, b) => Some(Some(Some(a, b)))
}
```

Or you could use a tuple with  a tag and options of the fields, like 

```scala
Equal[(Int, Option[(Boolea, Boolean)].contramap {
  case Completed     => (1, None)
  case Idle          => (2, None)
  case Stopped       => (3, None)
  case Running(a, b) => (4, Some((a, b)))
}
```

I'm not sure what the best strategy is in general … [kittens](https://github.com/milessabin/kittens) can derive these mechanically but I haven't tried it.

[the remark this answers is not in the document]
cquiroz: I have to say I'm starting to miss the benefit of using stuff like `Eq` if for a very simple ADT you need such a complex construct.
Also, I've read cats documentation and many examples and this is the first time I see the need for `contramap`.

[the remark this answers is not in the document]
tpolecat: Well you can also write it pairwise, which might be the best approach here. You don't *need* to use `contramap`, but you can. Some constructs like `Order` benefit more from this.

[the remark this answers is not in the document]
cquiroz: I guess it would be good being able to create `Equal` from the components `Equal`. i think in `Order` you can do something like that.

PR #505, opened 2018-04-07
SEQNG-583: Obs queue column widths sometimes too narrow

cquiroz:
The width of the columns on the obs queue was fixed and sometimes that produced text to be truncated. This PR adds a simple calculation of the text length to better estimate the columns' width

It also fixes some styles issues that sometimes made the table misaligned

Before:
seqexec - gs-engabc-2018a-q-1-85 2018-04-06 11-13-33

After:
seqexec - gs-engabc-2018a-q-1-72 2018-04-07 08-15-51

### DIFF
--- a/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
+++ b/modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
@@ -12,11 +12,13 @@ import edu.gemini.seqexec.web.client.model.Pages._
 import edu.gemini.seqexec.web.client.ModelOps._
 import edu.gemini.seqexec.web.client.services.HtmlConstants.iconEmpty
 import edu.gemini.seqexec.web.client.semanticui.elements.icon.Icon.{IconAttention, IconCheckmark, IconCircleNotched, IconSelectedRadio}
+import edu.gemini.web.client.utils._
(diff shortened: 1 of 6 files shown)


modules/edu.gemini.seqexec.web/edu.gemini.seqexec.web.client/src/main/scala/edu/gemini/seqexec/web/client/components/QueueArea.scala
```diff
@@ -126,32 +130,49 @@ object QueueTableBody {
...
-      case w if w < PhoneCut      => (0, Nil)
-      case w if w < LargePhoneCut => (TargetNameColumnWidth, List(targetColumn))
-      case _                      => (2 * TargetNameColumnWidth, List(targetColumn, nameColumn))
+
+    // Calculate the column's width
+    val (obsColumnWidth, statusColumnWidth, instrumentColumnWidth, nameColumnWidth, targetNameColumnWidth) = p.sequences().sequences.map {
+      case SequenceInQueue(id, st, i, _, n, t, r) =>
+        (tableTextWidth(id), tableTextWidth(statusText(st, r)), tableTextWidth(i.shows), tableTextWidth(n), tableTextWidth(t.getOrElse("")))
+    }.foldLeft((ObsIdColumnWidth, StateColumnWidth, InstrumentColumnWidth, NameColumnWidth, TargetNameColumnWidth)) {
+      case ((o, s, i, n, t), (co, cs, ci, cn, ct)) =>
```
sraaphorst: I admit that I find this part (with so many variables and the `foldLeft` a little bit confusing, but the end result looks very good, so I'll give this a 👍 .

PR #507, opened 2018-04-16
Use sbt-updates at startup

cquiroz:
As suggested by @tpolecat this change will do an updates check on sbt startup or reload

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,6 @@ install:
 - nvm use 8.7.0
 - npm install jsdom
 script:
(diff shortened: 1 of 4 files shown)


tpolecat:
In gem we have

```scala
// check for library updates whenever the project is [re]load
onLoad in Global := { s =>
  if (sys.props.contains("gem.skipDependencyUpdates")) s
  else "dependencyUpdates" :: s
}
```

So CI can pass `-Dgem.skipDependencyUpdates` to sbt which speeds up the build a little bit. We're not doing this for some reason.

cquiroz:
Yeah. i should add that too

PR #509, opened 2018-04-20
Switch to LBS

cquiroz:
As with gem this PR remove using Typelevel Scala and returns to the base Lightbend scala.
I'm also bumping to the latest scala version

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,9 +78,7 @@ lazy val edu_gemini_web_client_facades = project
       // By necessity facades will have unused params
       "-Ywarn-unused:params"
     ))),
-    libraryDependencies ++= Seq(ScalaJSDom.value, JQuery.value),
-    // And add a custom one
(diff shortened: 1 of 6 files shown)


cquiroz:
@tpolecat After learning a bit about `-Xcheckinit` I realized it had the potential to reduce the js size and in fact I'm getting around 8% reduction. Do you see a drawback on disabling it for scala.js?

tpolecat:
Wow that's a big decrease. It's kind of up to you. 

The effect when it's off is that the `UninitializedFieldError` you were seeing with that `Eq` instance wouldn't have happened. The `Eq` instance would have simply been `null` and you wouldn't know it until someone tried to compare two `Event` instances. Assuming test code exercises things pretty well you're probably safe turning it off for JS. I think it makes sense to leave it on for JVM.

cquiroz:
Yeah, I'd certainly leave it on the JVM side

PR #512, opened 2018-04-27
Updated epics libraries

cquiroz:
This PR contains the epics libraries with the latest versions of jca and caj. @jluhrs you can use these for testing of the epics issues but the unit tests are failing.

### DIFF
--- a/modules/edu.gemini.epics.acm/src/test/java/edu/gemini/epics/acm/test/CaStatusAcceptorTest.java
+++ b/modules/edu.gemini.epics.acm/src/test/java/edu/gemini/epics/acm/test/CaStatusAcceptorTest.java
@@ -68,7 +68,11 @@ public final class CaStatusAcceptorTest {
             caService.unbind();
             caService = null;
         }
-        simulator.stop();
+
+        if (simulator != null) {
+            simulator.stop();
+            simulator = null;
+        }
     }
 
     @Test
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -112,10 +112,10 @@ object Settings {
 
     val apacheXMLRPC      = "3.1.3"
     val opencsv           = "2.1"
-    val epicsService      = "1.0.1"
-    val gmpCommandRecords = "0.7.1"
-    val giapiJmsUtil      = "0.5.1"
-    val giapiJmsProvider  = "1.6.1"
+    val epicsService      = "1.0.2"
+    val gmpCommandRecords = "0.7.2"
+    val giapiJmsUtil      = "0.5.2"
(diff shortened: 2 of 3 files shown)


project/Settings.scala
```diff
@@ -111,9 +111,9 @@ object Settings {
 
     val apacheXMLRPC      = "3.1.3"
     val opencsv           = "2.1"
-    val epicsService      = "1.0.0"
-    val gmpCommandRecords = "0.7.0"
-    val guava             = "24.1-jre"
+    val epicsService      = "0.19"
+    val gmpCommandRecords = "0.6.6"
+    val guava             = "16.0.1"
```
jluhrs: If they use the latest versions, how is it that the numbers go down ?

cquiroz: a more detailed explanation is worthwhile. At the moment we have 2 branches for GMP. the 14.x line based on java 1.6 and the 15.x line base on java 1.8. We use the 15.x line on master at the moment towards supporting GIAPi. However I did my tests with the updated epics library on the 14.x line. Thus these version numbers are newer for the 14.x line.

Now that I think about it, it may be worthwhile to make the same change on 15.x and test, at least we'd be sure this is the only change, WDYT?

cquiroz:
The tests are passing in travis but they fail for me locally 😢

project/Settings.scala
jluhrs: I am a bit confused. Are we pulling epicsService from GMP 14.x ?

cquiroz:
I've updated the libraries to be compatible with the latest ones in `develop` and at the same time use the latest caj/jca

@jluhrs Can you checkout this branch and review the acm tests? They are failing for me locally

cquiroz:
I've updated the libraries to be compatible with the latest ones in `develop` and at the same time use the latest caj/jca

@jluhrs Can you checkout this branch and review the acm tests? They are failing for me locally

cquiroz:
@jluhrs Your fix looks fine. let's wait for travis

modules/edu.gemini.epics.acm/src/test/java/edu/gemini/epics/acm/test/CaStatusAcceptorTest.java
```diff
@@ -68,7 +68,11 @@ public static void tearDown() {
             caService.unbind();
             caService = null;
         }
-        simulator.stop();
+
+        if (simulator != null) {
+            simulator.stop();
+            simulator = null;
+        }
```
jluhrs: It worked before, I changed it only to make it "proper".

PR #517, opened 2018-05-03
SEQNG-586 Modified exceptions test, removed unnecessary cases.

jluhrs:
Cats.IO does not catch fatal VM errors, so the test that verified that those errors were not caught became unnecessary.


modules/edu.gemini.seqexec.engine/src/test/scala/edu/gemini/seqexec/engine/packageSpec.scala
```diff
@@ -279,17 +279,10 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
         )
       ) ) ) )
     )
-
-    intercept[RuntimeException](
-      runToCompletion(s0(new RuntimeException))
+    
+    assertThrows[java.lang.RuntimeException](
```
jluhrs: `assertThrows` if better than `intercept` in this case.

PR #519, opened 2018-05-04
SEQNG-252: GIAPI status getter library

cquiroz:
This is a small library to support GIAPI instruments. At the moment this only supports the `get` status item operation but more will be added as needed.

This is my first attempt at tagless final style. @tpolecat could you check if it is ok?

The GPI Example has been tested against GPI to verify it is working

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -33,7 +33,7 @@ onLoad in Global := { s =>
 }
 
 // Uncomment for local gmp testing
-// resolvers in ThisBuild += "Local Maven Repository" at "file://"+Path.userHome.absolutePath+"/Projects/maven-repo/releases"
+resolvers in ThisBuild += "Local Maven Repository" at "file://"+Path.userHome.absolutePath+"/.m2/repository"
 
 // Settings to use git to define the version of the project
 def versionFmt(out: sbtdynver.GitDescribeOutput): String = {
@@ -64,7 +64,7 @@ lazy val edu_gemini_giapi_client = project
   .enablePlugins(GitBranchPrompt)
   .settings(commonSettings: _*)
   .settings(
-    libraryDependencies += CatsEffect.value
+    libraryDependencies ++= Seq(CatsEffect.value, GiapiJmsUtil, GiapiJmsProvider) ++ Logging
   )
 
 lazy val edu_gemini_web_server_common = project
--- a/build.sbt
+++ b/build.sbt
@@ -33,7 +33,7 @@ onLoad in Global := { s =>
 }
 
 // Uncomment for local gmp testing
-resolvers in ThisBuild += "Local Maven Repository" at "file://"+Path.userHome.absolutePath+"/.m2/repository"
+//resolvers in ThisBuild += "Local Maven Repository" at "file://"+Path.userHome.absolutePath+"/.m2/repository"
 
 // Settings to use git to define the version of the project
 def versionFmt(out: sbtdynver.GitDescribeOutput): String = {
--- a/build.sbt
+++ b/build.sbt
@@ -64,7 +64,7 @@ lazy val edu_gemini_giapi_client = project
   .enablePlugins(GitBranchPrompt)
   .settings(commonSettings: _*)
   .settings(
-    libraryDependencies ++= Seq(CatsEffect.value, GiapiJmsUtil, GiapiJmsProvider) ++ Logging
+    libraryDependencies ++= Seq(CatsEffect.value, Fs2, GiapiJmsUtil, GiapiJmsProvider) ++ Logging
   )
 
 lazy val edu_gemini_web_server_common = project
--- a/build.sbt
+++ b/build.sbt
@@ -58,6 +58,15 @@ enablePlugins(GitBranchPrompt)
 //////////////
 // Projects
 //////////////
+lazy val edu_gemini_giapi_client = project
+  .in(file("modules/edu.gemini.giapi.client"))
+  .enablePlugins(AutomateHeaderPlugin)
+  .enablePlugins(GitBranchPrompt)
+  .settings(commonSettings: _*)
+  .settings(
+    libraryDependencies += CatsEffect.value
+  )
+
 lazy val edu_gemini_web_server_common = project
   .in(file("modules/edu.gemini.web.server.common"))
   .enablePlugins(AutomateHeaderPlugin)
--- a/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
+++ b/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
@@ -3,7 +3,6 @@
 
 package edu.gemini.giapi.client.gpi
 
-import cats.Monad
 import cats.effect.IO
 import edu.gemini.giapi.client.Giapi
 import fs2.Stream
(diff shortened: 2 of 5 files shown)


modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -0,0 +1,47 @@
...
+  def aoDarkLevel: F[Option[Double]] =
+    giapi.get[Double]("gpi:ao:darkLevel")
+
+  // add more items...
+}
+
+object GPIExample extends App {
+
+  private val gpi =
+    Stream.bracket(Giapi.giapiConnection[IO]("failover:(tcp://127.0.0.1:61616)").connect)(
```
cquiroz: I'm not too happy about this. I'm using `bracket` to close the connection at the end but feels cumbersome

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -0,0 +1,98 @@
...
+/**
+  * Algebra to interact with a GIAPI instrument
+  * @tparam F Effect Type
+  */
+trait Giapi[F[_]] {
+
+  /**
+    * Returns a value for the status item. If not found a `None` is returned
+    */
+  def get[A](statusItem: String): F[Option[A]]
```
cquiroz: GIAPI status items are limited to a few types `String`, `Int`, etc. Is there a way I can express that on `A`?

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+    * Close the connection
+    */
+  def close: F[Unit]
```
cquiroz: I think this should be in `GiapiConnection` but if I put it there, I can't get `Stream.bracket` to access `close`

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -0,0 +1,47 @@
...
+            h <- client.heartbeat
+            f <- client.fpmMask
+            o <- client.aoDarkLevel
+          } yield (h, f, o)
+        Stream.eval(r.map(println)) // scalastyle:ignore
+      },
+      _.close
+    )
+
+  gpi.compile.drain.unsafeRunSync()
```
cquiroz: For the record, connected to GPI. this prints
`(Some(5200542),Some(FPM_H),Some(-23.0))`

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -0,0 +1,107 @@
...
+    * Algebra to interact with a GIAPI instrument
+    *
+    * @tparam F Effect Type
+    */
+  trait Giapi[F[_]] {
+
+    /**
+      * Returns a value for the status item. If not found or there is an error, e.g. on types the exception is returned
+      */
+    def get[A: ItemGetter](statusItem: String): F[Option[A]]
```
cquiroz: @tpolecat I changed this with a typeclass and it works nicely. I'm wondering if I should return `F[Either[Throwable, A]]` rather than eating the exception and converting it to `None`. Any opinions?

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+      * Close the connection
+      */
+    def close: F[Unit]
```
cquiroz: I think this should be in GiapiConnection but if I put it there, I can't get Stream.bracket to access close

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
tpolecat: This is a common pattern and I think it's ok for now. Something better may come along but this is what people ar doing.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import cats.Monad
+import cats.effect.IO
+import edu.gemini.giapi.client.Giapi
+import fs2.Stream
+
+/**
+  * Client for GPI
+  */
+class GPIClient[F[_]: Monad](giapi: Giapi[F]) {
```
tpolecat: I would move the `Monad` constraint to the individual methods that require it. I don't think any need it so far.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
tpolecat: The "right" way to do this in MTL style is

```scala
def get[A: ItemGetter](statusItem: String)(
  implicit ev: ApplicativeError[F, String]
): F[A]
```

which would require that `F` provide a way to return an error as a string. A concrete `F` that would allow this is `EitherT[IO, String, ?]` which as you know is just a wrapper for `IO[Either[String, ?]]`, which is what you were thinking in the first place.  The advantage here is that you don't have to talk about the error case in your business logic unless you really want to, in which case you can use `.attempt` to turn an `F[A]` into an `F[Either[String, A]]`.

build.sbt
```diff
@@ -33,7 +33,7 @@ onLoad in Global := { s =>
 }
 
 // Uncomment for local gmp testing
-// resolvers in ThisBuild += "Local Maven Repository" at "file://"+Path.userHome.absolutePath+"/Projects/maven-repo/releases"
+//resolvers in ThisBuild += "Local Maven Repository" at "file://"+Path.userHome.absolutePath+"/.m2/repository"
```
jluhrs: This will not work in my Linux machine :(

cquiroz: well, it is commented and it is mostly for me. Why wouldn't it work in Linux anyway?

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
cquiroz: 👍

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
cquiroz: If I try to do this, I end up with errors defining my `GPIClient` in terms of `F` as I can't produce a generic `ApplicativeError[F, E]` or even if I do it with `String` there isn't a generic `ApplicativeError[F, String]`

Am i missing something?

tpolecat: > I can't produce a generic ApplicativeError[F, E] or even if I do it with String there isn't a generic ApplicativeError[F, String]

I'm not sure what you mean here. `ev` will be available if `F` can treat `String` as an error, like `Either[String, ?]` for instance. So you demand it anywhere you need to raise or trap an error. I'm just waking up but I'll be around in a couple hours if you want to chat.

PR #520, opened 2018-05-09
SEQNG-250: Support sending commands through GIAPI

cquiroz:
This is the next step on GIAPI support with the ability to send commands.

### DIFF
--- a/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/commands.scala
+++ b/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/commands.scala
@@ -21,16 +21,14 @@ package commands {
 }
 
 package object commands {
-
-  type CommandOperationResult[F[_]] = F[CommandResult]
+  val DataLabelCfg = "DATA_LABEL"
 
   implicit val responseEq: Eq[Response] = Eq.instance {
     case (a, b) => a.name === b.name
   }
 
-  def sendCommand[F[_]: Async](
-      commandsClient: CommandSenderClient,
-      command: Command): F[CommandResult] =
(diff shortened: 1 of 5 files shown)


modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/commands.scala
```diff
@@ -0,0 +1,56 @@
...
+        command,
+        (hr: HandlerResponse, _: Command) => {
+          if (hr.getResponse === Response.ERROR || hr.getResponse === Response.NOANSWER) {
+            cb(Right(Error(hr.getResponse, hr.getMessage)))
+          } else {
+            cb(Right(Completed(hr.getResponse)))
+          }
+          ()
+        },
+        2000
```
jluhrs: Magic number ?

cquiroz: Good point, this should be configurable

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -3,13 +3,19 @@
```
jluhrs: This code seems to have the same function for GPI that GmosEpics.scala. Shouldn't it be in seqexec/server/gpi ?

cquiroz: Eventually I"ll move it there

PR #522, opened 2018-05-10
Update scala version and increase paralellism flags

cquiroz:
Update scala version and added a flag to improve compilation performance

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -1,4 +1,5 @@
 import sbt._
+import java.lang.{Runtime => JRuntime}
 
 /**
  * Application settings and dependencies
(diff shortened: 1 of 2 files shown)


project/Settings.scala
```diff
@@ -54,13 +55,14 @@ object Settings {
       "-Ywarn-unused:params",              // Warn if a value parameter is unused.
       "-Ywarn-unused:patvars",             // Warn if a variable bound in a pattern is unused.
       "-Ywarn-unused:privates",            // Warn if a private member is unused.
-      "-Ywarn-value-discard"               // Warn when non-Unit expression results are unused.
+      "-Ywarn-value-discard",              // Warn when non-Unit expression results are unused.
+      "-Ybackend-parallelism", JRuntime.getRuntime.availableProcessors.toString // Run some tasks in parallel
```
cquiroz: @tpolecat Have you seen this one? It comes on `2.12.5`. I've read claims of 3-5% improvement

tpolecat: I have seen this but have't had a chance to try it. If this includes code generation (which does file IO) then we can use some multiple of `availableProcessors` … I'll ask around.

tpolecat: I talked to Seth and he didn't know of a heuristic for this. You might try it with `availableProcessors` and then `availableProcessors * 2`, etc., and see if there's a difference. Build times are really jittery so it may not be noticeable.

cquiroz: I don't expect a huge difference but I'll take anything that speeds up compilation even by a little bit

project/Settings.scala
```diff
@@ -54,13 +55,14 @@ object Settings {
...
     )
   }
 
   /** Library versions */
   object LibraryVersions {
-    val scalaVersion = "2.12.5"
+    val scalaVersion = "2.12.6"
```
tpolecat: You also need to update `.travis.yml` … easy to forget.

cquiroz: True !

PR #523, opened 2018-05-11
SEQNG-252: Add operation to stream GIAPI status items

cquiroz:
This is an extension to the library to read status items. This allows us to stream items as they are produced.

### DIFF
--- a/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
+++ b/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
@@ -4,23 +4,25 @@
 package edu.gemini.giapi.client.gpi
 
 import cats.Show
-import cats.instances.string._
+import cats.effect.Effect
 import cats.instances.double._
 import cats.instances.int._
 import cats.syntax.show._
 import edu.gemini.aspen.giapi.commands.{Activity, Command, DefaultConfiguration, SequenceCommand}
-import edu.gemini.giapi.client.{Giapi, commands}
 import edu.gemini.giapi.client.commands.CommandResult
+import edu.gemini.giapi.client.{Giapi, commands}
 import fs2.Stream
 import mouse.boolean._
+import scala.concurrent.ExecutionContext
 
 /**
   * Client for GPI
   */
-// Status items
-class GPIClient[F[_]](giapi: Giapi[F]) {
+class GPIClient[F[_]: Effect](giapi: Giapi[F], ec: ExecutionContext) {
 
-  // Some items, more will be added as needed
+  ///////////////
+  // Status items
+  ///////////////
   def heartbeat: F[Int] =
     giapi.get[Int]("gpi:heartbeat")
 
@@ -30,9 +32,15 @@ class GPIClient[F[_]](giapi: Giapi[F]) {
   def aoDarkLevel: F[Float] =
     giapi.get[Float]("gpi:ao:darkLevel")
 
-  // add more items...
+  /////////////////////
+  // Streaming statuses
+  /////////////////////
+  def heartbeatS: F[Stream[F, Int]] =
+    giapi.stream[Int]("gpi:heartbeat", ec)
 
-  // Commands
+  ///////////////////
+  // General commands
+  ///////////////////
   def test: F[CommandResult] =
     giapi.command(new Command(SequenceCommand.TEST, Activity.PRESET_START))
 
@@ -153,18 +161,20 @@ object GPIExample extends App {
 
   import cats.effect.IO
   import scala.concurrent.duration._
+  import cats.instances.string._
 
   private val gpiStatus =
     Stream.bracket(
       Giapi.giapiConnection[IO]("failover:(tcp://127.0.0.1:61616)", 2000.millis).connect)(
       giapi => {
-        val client = new GPIClient[IO](giapi)
+        val client = new GPIClient[IO](giapi, scala.concurrent.ExecutionContext.Implicits.global)
         val r =
           for {
-            h <- client.heartbeat
-            f <- client.fpmMask
-            o <- client.aoDarkLevel
-          } yield (h, f, o)
+            hs <- client.heartbeatS.flatMap(_.take(3).compile.toVector)
+            h  <- client.heartbeat
+            f  <- client.fpmMask
+            o  <- client.aoDarkLevel
+          } yield (hs, h, f, o)
         Stream.eval(r.map(println)) // scalastyle:ignore
       },
       _.close
@@ -174,7 +184,7 @@ object GPIExample extends App {
     Stream.bracket(
       Giapi.giapiConnection[IO]("failover:(tcp://127.0.0.1:61616)", 2000.millis).connect)(
       giapi => {
-        val client = new GPIClient[IO](giapi)
+        val client = new GPIClient[IO](giapi, scala.concurrent.ExecutionContext.Implicits.global)
         val r =
           for {
             _ <- client.calExitShutter(true) // Open the shutter
--- a/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
+++ b/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
@@ -13,6 +13,7 @@ import edu.gemini.giapi.client.commands.CommandResult
 import edu.gemini.giapi.client.{Giapi, commands}
 import fs2.Stream
 import mouse.boolean._
+
 import scala.concurrent.ExecutionContext
 
 /**
--- a/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
+++ b/modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
@@ -16,39 +16,20 @@ import edu.gemini.jms.activemq.provider.ActiveMQJmsProvider
 import fs2.{Stream, async}
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.Duration
+import shapeless.Typeable._
 
 package object client {
 
   /**
     * Allowed types according to GIAPI
     */
-  implicit val strItemGetter: ItemGetter[String] = new ItemGetter[String] {
-    def value(p: Any): Option[String] = p match {
-      case x: String => Some(x)
-      case _         => None
-    }
-  }
+  implicit val strItemGetter: ItemGetter[String] = new ItemGetter[String] {}
 
-  implicit val doubleItemGetter: ItemGetter[Double] = new ItemGetter[Double] {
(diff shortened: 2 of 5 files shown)


modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -153,18 +161,20 @@ object GPIExample extends App {
...
           for {
-            h <- client.heartbeat
-            f <- client.fpmMask
-            o <- client.aoDarkLevel
-          } yield (h, f, o)
+            hs <- client.heartbeatS.flatMap(_.take(3).compile.toVector)
+            h  <- client.heartbeat
+            f  <- client.fpmMask
+            o  <- client.aoDarkLevel
+          } yield (hs, h, f, o)
```
cquiroz: On running this against GPI it produces:
`(Vector(5738574, 5738575, 5738576),5738577,FPM_H,-8.0)`

Which corresponds to three heartbeat items, then the other variables

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -153,18 +161,20 @@ object GPIExample extends App {
...
   import cats.effect.IO
   import scala.concurrent.duration._
+  import cats.instances.string._
 
   private val gpiStatus =
     Stream.bracket(
       Giapi.giapiConnection[IO]("failover:(tcp://127.0.0.1:61616)", 2000.millis).connect)(
       giapi => {
-        val client = new GPIClient[IO](giapi)
+        val client = new GPIClient[IO](giapi, scala.concurrent.ExecutionContext.Implicits.global)
```
jluhrs: Shouldn't this be `ec` ?

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -174,7 +184,7 @@ object GPIExample extends App {
...
```
jluhrs: Same here, shouldn't it use the `ExecutionContext` that was passed as a constructor parameter?

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
cquiroz: There is no ec in this context. you need to pass an `ExecutionContext` and I'm just passing `global` directly

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -7,31 +7,68 @@ import cats.Applicative
...
+      case _      => None
+    }
+  }
+
+  implicit val floatItemGetter: ItemGetter[Float] = new ItemGetter[Float] {
+    def value(p: Any): Option[Float] = p match {
+      case x: Float => Some(x)
+      case _        => None
+    }
+  }
```
jluhrs: All this similar looking implementations of `value` could be factored out into a class.

cquiroz: I think the matching of `case x: A` wouldn't work. You need tho concrete class there, otherwise I could have done this directly on the method `stream`

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -75,25 +117,66 @@ package client {
...
+                                       ec: ExecutionContext): F[Stream[F, A]] =
+      Sync[F].delay {
+        implicit val exc: ExecutionContext = ec
+        def statusHandler(q: async.mutable.Queue[F, A]) = new StatusHandler {
+
+          override def update[B](item: StatusItem[B]): Unit =
+          // Check the item name and attempt convert it to A
+            if (item.getName === statusItem) {
+              ItemGetter[A].value(item.getValue).foreach { a =>
+                async.unsafeRunAsync(q.enqueue1(a))(_ => IO.unit)
```
jluhrs: I am always suspicious about the use of `unsafeRunAsync` in the middle of a computation. The rule of thumb is to have only one call to `unsafeRunAsync` at the end of the world. 
The usual case is that you somehow are doing something like `IO(async.unsafeRunAsync(IO(something))`, that could be just `IO(something)`. But, in this code I don't see how to get rid of it.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
```diff
@@ -4,23 +4,25 @@
...
 import fs2.Stream
 import mouse.boolean._
+import scala.concurrent.ExecutionContext
 
 /**
   * Client for GPI
   */
-// Status items
-class GPIClient[F[_]](giapi: Giapi[F]) {
+class GPIClient[F[_]: Effect](giapi: Giapi[F], ec: ExecutionContext) {
```
cquiroz: @tpolecat I'm not that happy having `Effect` here when I use it only for one method but I don't know how to do it otherwise as i need `F` in `Giapi[F]`

tpolecat: Which method needs it? In general you can say `def foo(args)(implicit ev: Effect[F])` to ensure that `F` *also* has `Effect` in addition to whatever else you know. This may mean that some higher-level methods will also need to add the constraint.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
tpolecat: We should keep an eye on this. We may end up needing a different EC here and elsewhere.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -7,31 +7,68 @@ import cats.Applicative
...
   /**
     * Typeclass to present as evidence when calling `Giapi.get`
     */
-  sealed trait ItemGetter[A]
+  sealed trait ItemGetter[A] {
+
+    /**
+      * Attempt to convert any value to A as sent by StatusHandler
+      */
+    def value(p: Any): Option[A]
```
tpolecat: You can implement this, actually.

```scala
def value(p: Any)(implicit ev: shapeless.Typeable[A]): Option[A] =
  ev.cast(p)
```

Or capture as a class constraint `abstract class ItemGetter[A: Typeable] {` which might work better.

`Typeable` instances aren't available for all types (arrays in particular) but if they are available it's safe.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
```diff
@@ -75,25 +117,66 @@ package client {
...
+    private final case class StatusStreamer(aggregate: StatusHandlerAggregate, ss: StatusService)
+
+    private def statusGetter[F[_]: Sync](c: ActiveMQJmsProvider): F[StatusGetter] = Sync[F].delay {
+      val sg = new StatusGetter("client")
+      sg.startJms(c)
+      sg
+    }
+
+    private def commandSenderClient[F[_]: Applicative](c: ActiveMQJmsProvider): F[CommandSenderClient] =
+      Applicative[F].pure { new CommandSenderClient(c) }
```
tpolecat: Does `CommandSenderClient` have any internal state? If so it needs to be in `delay`. Note that you can also say `(new CommandSenderClient(c)).pure[F]` here.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
tpolecat: I don't think you have any choice here because the API requires you to side-effect.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
tpolecat: You can implement this generically but it requires some trickery. See below.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
cquiroz: `stream` uses `Effect` as required my `fs2`

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/gpi/GPIClient.scala
cquiroz: we might. note this is an example app

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
cquiroz: Nice, i'll try that

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
cquiroz: It keeps a reference to `c` but dosen't do anything on the constructor

tpolecat: Ok that's fine as-is.

modules/edu.gemini.giapi.client/src/main/scala/edu/gemini/giapi/client/package.scala
cquiroz: Can you expand on what exactly `Typeable` does. i'm always a bit hestitant about shapeless' magic

tpolecat: For your use case here it's doing more or less what you were doing. For primitive types you can look at the runtime value and figure out the type. The source is [here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/typeable.scala) … there's some complexity in dealing with parameterized types but we're not doing that here as far as I can tell.

tpolecat: We could define our own mini-typeable but Shapeless is there and it works so I think we might as well use it.

PR #525, opened 2018-05-18
SEQNG-584: Rework of the build system for the frontend

cquiroz:
Sorry about the size an complexity of this PR but it is hard to simplify it.
This is a full redo of the build system for the frontend I"ve been working on for quite a while.

The main objective of this change is to improve development speed as scala.js has a fairly slow cycle which hinders progress on UI tasks. On the other side of the fence, javascript tools like webpack are all about development speed.

Using [scalajs-bundler](https://github.com/scalacenter/scalajs-bundler) we can marry `sbt` and `webpack`. The process is fairly complex given how both tools are very complex on their own. that is one of the reasons this work has taken so long. As a result of this work I had to push many [changes](https://github.com/scalacenter/scalajs-bundler/pulls?q=is%3Apr+is%3Aclosed+author%3Acquiroz) to `scalajs-bundler`, I think that gives us more visibility to the internals and ability to influence the evolution of the tool

The speed benefits for development are centered around 2 tools:
* Hot Module Replacement (HMR): This tool lets us launch a development server that will auto refresh when the code changes. It not only automotase the refresh but also keeps the frontend state unlike a manual refresh. This is a huge time saver, with this tool we see about 30-40 seconds from the time a code change is saved until it is on the browser. This compraed with about 1 min and the need for a browser reload.
* Remove ScalaCSS. We used to do css with scalacss and while a nice tool and very safe in terms of typesafety it led to requiring a full recompliation for even tiny style changes. In this PR we ditch `scalacss` and use plain `less` instead. Less was selected as `SemanticUI` uses less. Style changes are now basically instantaneous. After a change is saved you see them on the spot on the browser.

Given how arcane this whole thing is it maybe difficult to follow but feel free to ask questions about the overall design

/cc @swalker2m @tpolecat I plan to use exactly this style for gem too

There are a few open tasks which have been added to jira for implementation later on


modules/edu.gemini.web.client.facades/src/main/scala/edu/gemini/web/client/style/package.scala
```diff
@@ -0,0 +1,33 @@
...
+
+    implicit val eq: Eq[GStyle] = Eq.by(_.htmlClass)
+
+    implicit val monoid: Monoid[GStyle] = new Monoid[GStyle] {
+      override def empty = Zero
+      def combine(x: GStyle, y: GStyle): GStyle = (x, y) match {
+        case (x, y) if x === Zero && y === Zero => Zero
+        case (x, y) if x =!= Zero && y === Zero => x
+        case (x, y) if x === Zero && y =!= Zero => y
+        case (x, y)                             => GStyle(s"$x $y")
```
tpolecat: You don't have a `toString` defined … are you sure you don't want `s"${x.htmlClass} ${y.htmlClass)"` here?

This is kind of complicated … Is there a reason this isn't encoded as `GStyle(styles: List[String])`? If you do it that way you can get a monoid via `Monoid[List[String]].imap(...)(...)` and can get `htmlClass` via `styles.intercalate(" ")`.

PR #530, opened 2018-05-28
Rename packages

### LINKED ISSUE #514: package names
The gem project uses package names without the top level names, e.g. `gem.ui` rather than what the seqexec does using the traditional version `edu.gemini.seqexec`

Should we agree on having the same format or should we let this to continue as is? I'm looking forward to when we unite the repos (Which should be very soon)

any strong opinions? /cc @swalker2m @tpolecat @jdnavarro @sraaphorst @jluhrs

cquiroz:
Mechanical renaming of packages according to the discussion on #514 

Fixes #514


sraaphorst:
Looks like this one has pissed off the mighty Travis.

cquiroz:
After fighting sbt and travis now it seems to be stable, sorry for the fuss

build.sbt
```diff
@@ -292,23 +303,23 @@ lazy val edu_gemini_seqexec_model = crossProject(JVMPlatform, JSPlatform)
...
   .settings(commonSettings: _*)
   .settings(
     addCompilerPlugin(Plugins.kindProjectorPlugin),
     addCompilerPlugin(Plugins.paradisePlugin),
     libraryDependencies ++= Seq(Fs2, CatsEffect.value, Log4s) ++ Monocle.value
   )
 
-lazy val edu_gemini_epics_acm = project
+lazy val acm = project
   .in(file("modules/edu.gemini.epics.acm"))
```
cquiroz: If I change this to `.in(file("modules/acm"))` sbt doesn't know what java files to compile and it fails

PR #532, opened 2018-06-04
SEQNG-602 Fixed unit tests broken by port to FS2.

jluhrs:
This PR fixes two unit tests for seqexec.engine.

### DIFF
--- a/modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
+++ b/modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
@@ -15,7 +15,6 @@ import seqexec.model.{ActionType, UserDetails}
 import fs2.async
 import fs2.async.mutable.Semaphore
 import fs2.Stream
-
 import cats.data.Kleisli
 import cats.implicits._
 import cats.effect._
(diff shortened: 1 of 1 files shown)


modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
```diff
@@ -218,43 +219,42 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
...
-              fileId = None,
-              config = config,
-              resources = Set(GmosS),
-              executions = List(
-                List(fromIO(ActionType.Configure(TCS),
-                  IO.apply {
-                    startedFlag.decrement
-                    finishFlag.increment
-                    Result.OK(Result.Configured(TCS))
-                  }))
```
jluhrs: This yield was producing a Stream[IO, IO[List[Unit]], instead of a Stream[IO, List[Unit]]. The code inside was never evaluated.

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
```diff
@@ -178,15 +177,17 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
     )
     val p = Stream.eval(IO.pure(Event.start(seqId, user, UUID.randomUUID())))
 
-    executionEngine.process(p)(s0).take(1).compile.last.unsafeRunSync.map(_._2)
+    //take(3): Start, Executing, Paused
+    executionEngine.process(p)(s0).take(3).compile.last.unsafeRunSync.map(_._2)
```
jluhrs: Previous to the merge, this would finish naturally after processing the single value in the input Stream and the other Streams generated internally. Because now there is a queue inside, it never terminates, so it is necessary to predict how many events will take to reach the result.

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
```diff
@@ -178,15 +177,17 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
...
   }
 
   "sequence state" should "stay as running when action pauses itself" in {
     assert(actionPause.exists(s => Sequence.State.isRunning(s.sequences(seqId))))
   }
 
-  ignore should "change to Paused if output is Paused" in {
-    assert(actionPause.exists(_.sequences(seqId).current.execution.forall(Action.paused)))
+  "engine" should "change action state to Paused if output is Paused" in {
+    val r = actionPause
```
cquiroz: This look the same as before. Why was it ignored?

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
cquiroz: 👍

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
jluhrs: It didn't work, because `actionPause` was not running the engine to the point were it would actually process the `Action`.

PR #535, opened 2018-06-12
WIP - reasonable json support

tpolecat:
This adds Json encoders for model types, avoiding defining new data transformations. With very few exceptions the codecs are derived from existing optics, and via Circe semi-automatic derivation for case classes and ADTs.

Semi-automatic derivation requires an explicit `deriveEncoder` and `deriveDecoder` for case classes and ADTs, including those nested inside other ADTs you might be deriving codecs for. This is a bit of extra work but compilation is *much* faster as a result.

In order to make the optics work for `RadialVelocity` I changed the light-speed sanity check into a proper `Prism`.

Remaining:
- [x] cross-compile to scala-js
- [x] codec definitions up to `Program`
- [x] ensure all new optics have law checks
- [x] law checks for all codecs
- [x] ensure `gem.json.instances.all` includes everything!
- [x] ~~restore sbt to 1.1.6~~ *we decided to leave this at 1.1.4 for now*
- [x] add something to the client to pull in the `Program` serializer, to check the size


cquiroz:
as a general comment, automatic derivation sometimes produces a large amount of classes which are very bad or scala.js. Possible `semi-automatic` solves that but it is something to keep an eye on

cquiroz:
json should be crossCompiled to scala.js if we want to use it on the client

tpolecat:
Ah good point. I'll change the setup.

modules/core/shared/src/main/scala/gem/Program.scala
```diff
@@ -4,13 +4,13 @@
 package gem
 
 import gem.math.Index
-import scala.collection.immutable.TreeMap
+import scala.collection.immutable.SortedMap
```
tpolecat: I have changed a few `TreeMap/Set`s to `SortedMap/Set` to avoid writing codecs for `TreeMap/Set` that we'll just end up throwing away.

modules/core/shared/src/main/scala/gem/config/GcalConfig.scala
```diff
@@ -24,27 +25,18 @@ final case class GcalConfig(lamp: GcalLamp, filter: GcalFilter, diffuser: GcalDi
 }
 
 object GcalConfig {
-  // We make this a sealed abstract case class in order to force usage of the
-  // companion object constructor.  The OneAnd head is guaranteed to always be
-  // the minimum GcalArc in the group.
-  sealed abstract case class GcalArcs(arcs: OneAnd[Set, GcalArc]) {
+
+  final case class GcalArcs(arcs: NonEmptySet[GcalArc]) {
```
tpolecat: Cats now has `SortedSet` so I git rid of `OneAnd` here.

modules/json/src/main/scala/gem/json/instances/SortedMap.scala
```diff
@@ -0,0 +1,21 @@
...
+package gem.json.instances
+
+import cats.Order
+import scala.collection.immutable.SortedMap
+import io.circe.{ Decoder, Encoder, KeyDecoder, KeyEncoder }
+
+trait SortedMapJson {
+
+  implicit def SortedMapEncoder[K: KeyEncoder, V: Encoder]: Encoder[SortedMap[K, V]] =
+    Encoder[Map[K, V]].contramap(identity)
```
tpolecat: Encoding a map with Circe requires a `KeyEncoder` that explains how to turn the key into a `String`. Decoding requires a `KeyDecoder`. By defining this for `Index` we can serialize index-keyed maps.

modules/json/src/main/scala/gem/json/instances/package.scala
```diff
@@ -0,0 +1,22 @@
...
+       with CoordinatesJson
+       with DynamicConfigJson
+       with EnumeratedJson
+       with EphemerisJson
+       with OffsetJson
+       with ProgramIdJson
+       with ProperMotionJson
+       with TargetJson
+       with TimeJson
+       with WavelengthJson
```
tpolecat: Oops need to make sure everything is here.

cquiroz:
Perhaps we should add some bits to the client code to read/write the json bits thus checking how big an increase the codecs produce. I'm not too worried but it iwould be good to check it

modules/core/shared/src/main/scala/gem/config/GcalConfig.scala
```diff
@@ -24,27 +25,18 @@ final case class GcalConfig(lamp: GcalLamp, filter: GcalFilter, diffuser: GcalDi
...
-      */
-    @SuppressWarnings(Array("org.wartremover.warts.OptionPartial"))
-    def apply(arc0: GcalArc, arcs: List[GcalArc]): GcalArcs = {
-      (arc0 :: arcs).toSet.toList.sorted(Order[GcalArc].toOrdering) match {
-        case h :: t => new GcalArcs(OneAnd(h, t.toSet)) {}
-        case Nil    => sys.error("unpossible")
-      }
-    }
+    def of(arc0: GcalArc, arcs: GcalArc*): GcalArcs =
+      GcalArcs(NonEmptySet.of(arc0, arcs: _*))
```
swalker2m: Presumably if `arc0` is also a member of `arcs` `NonEmptySet` will do the right thing?

modules/core/shared/src/main/scala/gem/math/Angle.scala
```diff
@@ -260,6 +260,24 @@ trait AngleOptics extends OpticsHelpers { this: Angle.type =>
     )
   }
 
+  // Exact signed angles, scaled by moving the decimal point `scale` digits to the left
+  private def signedMicroarcsecondsScaled(scale: Int): SplitMono[Angle, BigDecimal] =
+    signedMicroarcseconds.imapB(_.underlying.movePointRight(scale).longValue, n => new java.math.BigDecimal(n).movePointLeft(scale))
```
swalker2m: `imap` I suppose is invariant map.  What's `imapB`?

modules/json/src/main/scala/gem/json/instances/StaticConfigJson.scala
```diff
@@ -0,0 +1,87 @@
...
+  implicit val GmosShuffleOffsetEncoder: Encoder[GmosConfig.GmosShuffleOffset] = Encoder[Int].contramap(_.detectorRows)
+  implicit val GmosShuffleOffsetDecoder: Decoder[GmosConfig.GmosShuffleOffset] = Decoder[Int].map(GmosConfig.GmosShuffleOffset.unsafeFromRowCount)
+
+  implicit val GmosShuffleCyclesEncoder: Encoder[GmosConfig.GmosShuffleCycles] = Encoder[Int].contramap(_.toInt)
+  implicit val GmosShuffleCyclesDecoder: Decoder[GmosConfig.GmosShuffleCycles] = Decoder[Int].map(GmosConfig.GmosShuffleCycles.unsafeFromCycleCount)
+
+  implicit val GmosNodAndShuffleEncoder: Encoder[GmosConfig.GmosNodAndShuffle] = deriveEncoder
+  implicit val GmosNodAndShuffleDecoder: Decoder[GmosConfig.GmosNodAndShuffle] = deriveDecoder
+
+  implicit val GmosCustomRoiEntryEncoder: Encoder[GmosConfig.GmosCustomRoiEntry] = Encoder[(Short, Short, Short, Short)].contramap(e => (e.xMin, e.yMin, e.xRange, e.yRange))
```
swalker2m: When can something be waived away with `deriveEncoder` vs when you actually have to do something?  I suppose in this case `GmosCustomRoiEntry` isn't a normal `case class` but rather a `sealed abstract case class` with controlled construction so it cannot be easily derived.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.1.6
+sbt.version=1.1.4
```
swalker2m: We went back?

modules/core/shared/src/main/scala/gem/config/GcalConfig.scala
tpolecat: Yes, it's basically a newtype over `SortedSet` with constructors guaranteeing it can't be empty.

modules/core/shared/src/main/scala/gem/math/Angle.scala
tpolecat: Consider 

```scala
type Endo[A] = A => A
```

If we want an `Endo[B]` we would need to supply a `B => A` for the input on the left and `A => B` for the output on the right. `imap[B](f: A => B)(g: B => A): Endo[B]` does this. Types like `Endo` are called invariant functors because `A` appears in both covariant and contravariant positions.

Now consider

```scala
type Iso[A, B] = (A => B, B => A)
```

If we want an `Iso[A, C]` we need `B => C` on the right of the first function, and `C => B` on the left of the second function, as with `Endo`. But if we want `Iso[C, B]` we need the same kind of thing, with `C => A` and `A => C`, also as with `Endo`. So `Iso` is an invariant functor in two different ways.

`Iso` and other bidirectional optics have this 2-headed form so you have `imapA` for the left-side invariant functor, and `imapB` for the right-side one.

Does this make sense?

project/build.properties
cquiroz: weird. in any case the sbt updates seems to alternatively break and fix triggered compilation

modules/json/src/main/scala/gem/json/instances/StaticConfigJson.scala
tpolecat: You can use `deriveEncoder` for case classes whose fields already have encoders (so base types; and `Option`s, `List`s, etc., thereof) and for sealed traits whose cases meet the same criteria. But it *won't* try to derive encoders for fields. This prevents the compile-time goose chase that leads to exponential compile times but requires that you manufacture more things explicitly.

You are correct that the issue with `GmosCustomRoiEntry` is that it's not a normal case class so Circe doesn't know how to construct one and take it apart again. So we derive the codec via the relationship with `(Short, Short, Short, Short)` (which we know about but Circe doesn't). I think I'm going to add an optic to move this logic out, since we'll need it again for the doobie `Composite` (and boopickle codec if we go down that path).

project/build.properties
tpolecat: My editor integration seems to work better with 1.1.4 but I'll put it back before merging.

tpolecat:
> Getting the encoders from the optics is a pretty cool idea.

We reuse the same optics for the doobie row/column serializers, so yeah I think it's a big win. If we change the encoding it propagates to all the external representations.

tpolecat:
> Perhaps we should add some bits to the client code to read/write the json bits thus checking how big an increase the codecs produce. I'm not too worried but it iwould be good to check it

@cquiroz good idea. I will add something to exercise `Program` serializer which will pull everything else along with it.

modules/core/shared/src/main/scala/gem/math/Angle.scala
swalker2m: Yes very much.  Thanks for the detailed explanation.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -291,4 +292,7 @@ object DynamicConfig {
     )
   }
 
+  implicit val EqDynamicConfig: Eq[DynamicConfig] =
+    Eq.fromUniversalEquals // TODO: ensure this is ok
```
cquiroz: Good question

project/build.properties
cquiroz: Perhaps we should keep 1.1.4. Triggered execution is being very slow on 1.1.6 though in theory it was fixed yet again on 1.1.5 and 1.1.6

tpolecat: Fine with me. I'll remove that todo item.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
tpolecat: I think it's fine because it's an ADT that eventually goes down to ground types, but it's bad practice to assume this because it can become wrong in the future. I think we can come back later to review model equality. Opened #541

build.sbt
```diff
@@ -80,17 +80,17 @@ addCommandAlias("gemctl", "ctl/runMain gem.ctl.main")//
...
+// shellPrompt in ThisBuild := { state =>
+//   if (version.value != imageManifest.formatVersion) {
+//     import scala.Console.{ RED, RESET }
+//     print(RED)
+//     println(s"Computed version doesn't match the filesystem anymore.")
+//     println(s"Please `reload` to get back in sync.")
+//     print(RESET)
+//   }
+//   "> "
+// }
```
tpolecat: I think this is unnecessary and it's bothering me so I turned it off for now. Will revisit soon for the continuous deployment task.

tpolecat:
Ok, this is done and rebased. Ready for one last look. It's a lot but it's mostly mechanical and everything has a test so I think we're good.

cquiroz:
We could remove the fullOpt stages. They take a lot of time and could require too much memory

cquiroz:
Looking at the build times it seems we were just below 30m and with this PR we go over it. It may be better to remove `ui/fullOptJS` and `seqexec_web_client/fullOptJS::webpack`

modules/core/shared/src/main/scala/gem/instances/TimeInstances.scala
```diff
@@ -26,5 +27,8 @@ trait TimeInstances {
   implicit val LocalDateOrder:     Order[LocalDate]     = naturalOrder(_ isBefore _)
   implicit val LocalDateTimeOrder: Order[LocalDateTime] = naturalOrder(_ isBefore _)
 
+  implicit val DurationEq: Eq[Duration] =
+    Eq.by(d => (d.getSeconds, d.getNano))
```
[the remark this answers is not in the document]
tpolecat: Duration is a pair. Seconds plus a fractional second in nanos.

modules/core/shared/src/main/scala/gem/optics/Format.scala
```diff
@@ -47,11 +47,11 @@ final case class Format[A, B](getOption: A => Option[B], reverseGet: B => A) {
     composeIso(f)
 
   /** Format is an invariant functor over A. */
```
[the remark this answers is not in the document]
tpolecat: Ah, no. I'll fix those. Thanks.

modules/json/shared/src/main/scala/gem/json/syntax/FormatJsonSyntax.scala
```diff
@@ -0,0 +1,17 @@
...
+
+package gem.json.syntax
+
+import gem.optics._
+import io.circe._
+
+trait FormatJsonSyntax {
+  implicit class JsonFormatOps[A: Encoder: Decoder, B](p: Format[A, B]) {
+    def toEncoder: Encoder[B] = Encoder[A].contramap(p.reverseGet)
+    def toDecoder: Decoder[B] = Decoder[A].emap { a =>
```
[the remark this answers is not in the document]
tpolecat: It's a Circe-specific thing. `Decoder[A] => Either[String, B] => Decoder[B]` where the string is turned into a failure and `B` into success. You can also do it in terms of `flatMap` but this saves some boilerplate.

tpolecat:
> Looking at the build times it seems we were just below 30m and with this PR we go over it. It may be better to remove ui/fullOptJS and seqexec_web_client/fullOptJS::webpack

I'll make that change and see if it helps, but I'm concerned that the project is just too big.

tpolecat:
Ok I really have no idea what's going on. I moved the build steps to a shell script and it appears that the webpack task starts running in the middle of the tests, which screws things up. So, suggestions are welcome.

tpolecat:
I encourage you to look at the log for the latest build failure. I really don't understand it.

cquiroz:
There is a lot going on there. I wonder if we could split this into stages. I'll run some experiments

build.sbt
```diff
@@ -178,14 +179,19 @@ lazy val db = project
     """.stripMargin.trim
   )
 
-lazy val json = project
+lazy val json = crossProject(JVMPlatform, JSPlatform)
+  .crossType(CrossType.Full)
```
cquiroz: I think this project doesn't need to be full. It can be `CrossType.Pure` as there is no js specific code. We can change this in another PR

cquiroz:
I did a lot of experiments and the only way to make it work was to disable tests on the js side in a couple of projects. I sent you a PR with the changes

This is not the end of this though, We really want to run those tests and the project will keep growing. I 
I ran this locally and it just takes 12min, we maybe better finding an alternative to the build. If you need this jenkins server for docker perhaps we should use it for local build which should be much faster.

We should be able to use a reverse proxy to give it access form github.

In the meantime let's keep those tests disable

modules/json/shared/src/test/scala/gem/json/instances/AsterismJsonSpec.scala
```diff
@@ -0,0 +1,18 @@
...
+import gem.Asterism
+import gem.arb._
+import io.circe.testing.CodecTests
+import io.circe.testing.instances._
+
+final class AsterismJsonSpec extends CatsSuite {
+  import ArbAsterism._
+  import asterism._
+
+  checkAll("Asterism", CodecTests[Asterism].unserializableCodec)
```
cquiroz: Why are tests done on `unserializableCodec` rather than `codec`?

tpolecat: If you're working with Spark or Akka everything has to be serializable so circe checks instances for serializability by default. But we don't care about that.

cquiroz:
Tests passing!!

cquiroz:
ouch, `ui-fastopt` grew a lot

tpolecat:
> ouch, ui-fastopt grew a lot

Yeah it's pulling in circe and all the serializers now.

cquiroz:
Yeah that must be

cquiroz:
On the other hand we may use boopickle for scala.js and leave `json` for non scala clients, in that case this is a non-issue.

Boopickle adds its own amount of code of course

tpolecat:
What we really care about is full-opt though right? Once you strip it down to what's actually in the code path it should be a lot smaller. I hope.

cquiroz:
yes, that is true. We had to take full-opt from the build as you know so it is harder to compare

cquiroz:
Feel free to merge. in ocs3 we are not requiring approvals after updates. maybe we should, it is up for discussion

tpolecat:
I have a few little things to change first.

tpolecat:
Ok it finally passed. I'm clicking merge before it breaks again.

PR #536, opened 2018-06-13
Updated copyright year to 2018

cquiroz:
It touches every file but it only changes the copyright year


cquiroz:
Maybe we should. In any case `sbt-header` requires a year 🤷‍♂️

tpolecat:
We could automate it so the first lucky person to compile every year gets a 600+ file diff.

PR #537, opened 2018-06-13
SEQNG-604 Store a map of selected items on the engine metadata

cquiroz:
As part of SEQNG-468 we need to keep a list of selected sequences per instrument on the engine. This PR adds that to the engine metadata, as well as methods to update them.

I'm making more extensive use of monocle lenses and auto generation of those lenses. It requires a change in the engine to produce events with a function that takes the current state as an input. That is called `modifyF`


swalker2m:
> Now that it's one codebase we need to figure out a way to get me and @swalker2m up to speed.

Agree.  The seqexec is pretty big and has been under development for a while so I'm quite behind.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -8,30 +8,34 @@ import cats.effect.IO
...
-
-    val operatorL: Lens[EngineMetadata, Option[Operator]] = GenLens[EngineMetadata](_.operator)
+    def selectedML(instrument: Instrument): Lens[EngineMetadata, Option[SequenceId]] = GenLens[EngineMetadata](_.selected) ^|-> at(instrument)
   }
 
   sealed trait SeqEvent
   final case class SetOperator(name: Operator, user: Option[UserDetails]) extends SeqEvent
   final case class SetObserver(id: Sequence.Id, user: Option[UserDetails], name: Observer) extends SeqEvent
   final case class SetConditions(conditions: Conditions, user: Option[UserDetails]) extends SeqEvent
+  final case class SetSelectedSequences(selected: Map[Instrument, SequenceId], user: Option[UserDetails]) extends SeqEvent
```
jluhrs: I think that `SetSelectedSequences` should be `SetSelectedSequence` and have the `SequenceId` of the selected sequence, not the whole `Map`. The instrument can be inferred from the sequence.

cquiroz: That's ok though I'd still pass the instrument to make things easier on the client

PR #547, opened 2018-06-18
SEQNG-623: Step config table back button not working

cquiroz:
While working on GPI I noted the step settings back button wasn't working and there were some small missalignments

This PR fixes that. Here's the button that's now properly working

seqexec - gs-2018b-q-1-17 2018-06-18 11-31-10

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
@@ -54,8 +54,8 @@ object StepConfigTable {
 
   private def columns(p: Props): List[Table.ColumnArg] =
     List(
-      Column(Column.props(p.size.width.toInt / 2, "name", label = "Name")),
-      Column(Column.props(p.size.width.toInt / 2, "value", label = "Value"))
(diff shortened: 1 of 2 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
```diff
@@ -54,8 +54,8 @@ object StepConfigTable {
 
   private def columns(p: Props): List[Table.ColumnArg] =
     List(
-      Column(Column.props(p.size.width.toInt / 2, "name", label = "Name")),
-      Column(Column.props(p.size.width.toInt / 2, "value", label = "Value"))
+      Column(Column.props(p.size.width.toInt / 2, "name", label = "Name", flexShrink = 0, flexGrow = 0, className = SeqexecStyles.paddedStepRow.htmlClass)),
+      Column(Column.props(p.size.width.toInt / 2, "value", label = "Value", flexShrink = 0, flexGrow = 0, className = SeqexecStyles.paddedStepRow.htmlClass))
```
sraaphorst: So is the default value for `flexShrink` and `flexGrow` nonzero, then?

cquiroz: yes, `flexShrink` is 1 which breaks alignment with the styling we are using

cquiroz: I mean `flexShrink` is `1` by default on the upstream library

PR #551, opened 2018-06-20
SENQG-624: Add configuration to connect to GPI

cquiroz:
This a small update adding a configuration parameter to connect to GPI. I also fixed the launcher which was unnecessarily calling `unsafeRunSync`

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -461,7 +461,13 @@ object SeqexecEngine {
   }
 
   def giapiConnection: Kleisli[IO, Config, Giapi[IO]] = Kleisli { cfg: Config =>
-    Giapi.giapiConnectionIO.connect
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -461,7 +461,13 @@ object SeqexecEngine {
   }
 
   def giapiConnection: Kleisli[IO, Config, Giapi[IO]] = Kleisli { cfg: Config =>
-    Giapi.giapiConnectionIO.connect
+    val instSim = cfg.require[Boolean]("seqexec-engine.instSim")
+    val gpiUrl  = cfg.require[String]("seqexec-engine.gpiUrl")
+    if (instSim) {
+      Giapi.giapiConnectionIO.connect
```
cquiroz: We could in principle use an `Id` based simulator as however this means generalizing most of the seqexec-server code from `IO` to `F[_]`. I started that fix but it is a fairly big change

PR #553, opened 2018-06-22
Horizons paging bugfix

swalker2m:
This PR fixes the [Random ephemeris test failure](https://github.com/gemini-hlsw/gem/issues/278) issue.  The problem was in the code that pages a long request into smaller queries that are palatable for horizons.  One of the complications is that a query for just a single element doesn't work.  The previous version of this code would try to shrink the next-to-last query by one element and thereby make the last query request two elements in this case.  I tried to simplify the logic somewhat and just request an extra element instead.   It's still a bit convoluted unfortunately.

### DIFF
--- a/modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisQuery.scala
+++ b/modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisQuery.scala
@@ -13,6 +13,7 @@ import cats.effect.IO
 import cats.implicits._
 
 import fs2.Stream
+import mouse.boolean._
 
 import java.time.{Duration, Instant, Period}
 
@@ -117,11 +118,6 @@ object HorizonsEphemerisQuery {
         HorizonsClient.stream(reqParams).through(EphemerisParser.elements)
     }
 
-
-  // Utility to calculate the integer ceiling of n/d.
-  private def intCeil(n: Long, d: Long): Long =
-    if (n >= 0) (n + d - 1L) / d else n / d
-
   // Utility to round up a Duration up to the next ms.
   private def roundUpMs(d: Duration): Duration =
     (d.getNano % 1000000) match {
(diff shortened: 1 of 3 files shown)


modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisQuery.scala
```diff
@@ -151,42 +147,45 @@ object HorizonsEphemerisQuery {
...
+    start: Instant,
+    end:   Instant,
+    step:  Duration
+  ): List[HorizonsEphemerisQuery] = {
+
+    // Calculates the list of queries with adjusted end time and step size.
+    def calc(end: Instant, step: Duration): List[HorizonsEphemerisQuery] = {
+      val maxPageDuration = step * (MaxElements - 1).toLong
+
+      Stream.unfold(start) { s =>               // page start time
```
cquiroz: Can you tell a bit more of `Stream.unfold`?

modules/ephemeris/src/test/scala/gem/horizons/HorizonsEphemerisQueryPagingPropSpec.scala
```diff
@@ -54,15 +54,27 @@ class HorizonsEphemerisQueryPagingPropSpec extends PropSpec with PropertyChecks
...
     forAll { (t: TestCase) =>
       t.paging.lastOption match {
         case None    => fail
-        case Some(q) => q.endTime should (be >= t.end and be <= (t.end + t.step))
+        case Some(q) => q.endTime should (be >= t.end and be < (t.end + t.step * 2))
       }
     }
   }
 
+  property("explicitly handle case where last page would have one element using MaxElement-sized pages") {
```
cquiroz: nice

modules/ephemeris/src/main/scala/gem/horizons/HorizonsEphemerisQuery.scala
swalker2m: It's a way of generating a `Stream` from a seed value and a function.

```
def unfold[S, O](s: S)(f: S => Option[(O, S)]): Stream[Pure, O]
```

The description in the doc is actually pretty good:

> Creates a stream by successively applying `f` until a `None` is returned, emitting each output `O` and using each output `S` as input to the next invocation of `f`.

So in this case `S` is an `Instant` beginning with the start time for the first horizons query.   Each successive invocation gets the start time for that particular page.  We stop, returning `None`, when the page start time has reached or gone past the end.  If we are not stopping though we produce a `Some` containing a tuple `(HorizonsEphemerisQuery, Instant)` where the query gets that page of ephemeris data and the `Instant` is the next page's start time.

The end result is a `Stream[Pure, HorizonsEphemerisQuery]` because the `Instant`s are just used in generating the result.

modules/ephemeris/src/test/scala/gem/horizons/HorizonsEphemerisQueryPagingPropSpec.scala
swalker2m: This was a randomly generated test case that produced the issue in the first place.  It's unlikely to come up very often randomly so it seemed like a good idea to leave it as a test case permanently.

modules/ephemeris/src/test/scala/gem/horizons/HorizonsEphemerisQuerySpec.scala
```diff
@@ -117,10 +117,11 @@ final class HorizonsEphemerisQuerySpec extends CatsSuite with EphemerisTestSuppo
...
       end,
-      // Here the first query is shortened so that the last query will have at
-      // least two elements.  There's no way to query for just one element.
-      HorizonsEphemerisQuery(titan, GS, start,           end - OneMinute * 2L, Max - 1),
-      HorizonsEphemerisQuery(titan, GS, end - OneMinute, end,                  2      )
+      // Here the first query has Max elements.  The second query would have
+      // just one element but since that won't work with horizons it is extended
+      // to include an additional element.
+      HorizonsEphemerisQuery(titan, GS, start, end - OneMinute, Max),
+      HorizonsEphemerisQuery(titan, GS, end,   end + OneMinute,   2)
```
swalker2m: This highlights the difference between the previous version and this one.  Before the next to last query would shrink by one and the last query grow by one to keep the total number of elements the same.  Now I just keep the next to last query with max elements and simply request two elements for the last query even though this takes us one step beyond the requested end time.

cquiroz: Thanks for the explanations

PR #555, opened 2018-06-22
Observation import and fetch

swalker2m:
This PR ports the [matching `gem` project PR](https://github.com/gemini-hlsw/gem/pull/282) to `ocs3`. It isn't ready to be merged yet 👎 , but I wanted to discuss it a bit.   The ultimate goal will be to provide an API for the seqexec to be able to import observations from the OCS2 ODB and fetch them over to the seqexec.

I'll review this PR with a couple of topics I wanted to introduce.

### DIFF
--- a/modules/service/src/main/scala/gem/Service.scala
+++ b/modules/service/src/main/scala/gem/Service.scala
@@ -28,6 +28,14 @@ final class Service[M[_]: Sync: LiftIO] private (private val xa: Transactor[M], 
     }
 
   /**
+   * Constuct a program that yields the `Observation` with the matching observation id.
+   */
+  def queryObservationById(id: Observation.Id): M[Observation] =
+    log.log(user, s"queryObservationById(${id.format})") {
+      ObservationDao.select(id).transact(xa)
+    }
+
+  /**
    * Construct a program that attempts to change the user's password, yielding `true` on success.
    */
   def changePassword(oldPassword: String, newPassword: String): M[Boolean] =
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/gem/Service.scala
```diff
@@ -27,6 +27,14 @@ final class Service[M[_]: Sync: LiftIO] private (private val xa: Transactor[M],
       ProgramDao.selectBySubstring(s"%$substr%", max).transact(xa)
     }
 
+  /**
+   * Constuct a program that yields the `Observation` with the matching observation id.
+   */
+  def queryObservationById(id: Observation.Id): M[Observation] =
```
swalker2m: I think we discussed in the chatter a bit whether assuming that the matching observation exists is appropriate but I couldn't find it again.  This should probably explicitly return an `Option[Observation]` instead?

modules/web/src/main/scala/Application.scala
```diff
@@ -37,6 +41,21 @@ object Application {
...
+          gs.ocs2.importObservation(h, oid).flatMap {
+            _.fold(msg => InternalServerError(msg), _ => NoContent())
+          }
+        }
+
+      // Select an observation by id.
+//      case GET -> Root / "api" / "fetch" / "obs" / o as gs =>
+//        withObsId(o) { oid =>
+//          gs.queryObservationById(oid).flatMap { obs =>
+//            Ok(obs.asJson)
```
swalker2m: Are we doing these things in JSON?  What would I need to do to get the codec for an `Observation`?  I assume it is derived because there is no explicit codec in the `gem.json` package but it isn't being picked up.

cquiroz: isn't the codec defined in #535 ?

swalker2m: D'oh, you're right.  I keep forgetting that is still pending.

PR #558, opened 2018-06-26
Gpi testing based fixed

cquiroz:
This PR is the result of a couple of days of instrument testing. It includes several somewhat related fixes:

* Support for all the static GPI configuration. each one has been tested with the instrument
* Fix for the gmp url used for testing. This is needed as we only have a flag to enable all or none of the instruments
* Some silencing of activemq logs that are too noisy
* Renamed commands to start the seqexec and a new command to stop it in development
* Update of http4s and fs2. While I'm a bit reluctant to update fs2 the changes seemed very small, they required no code changes and testing with GPI went very smooth on that side


modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPIController.scala
```diff
@@ -48,43 +156,33 @@ final case class GPIController(gpiClient: GPIClient[IO]) {
...
-  implicit val pcEq: Eq[PupilCamera] = Eq.by(_.displayValue)
+  implicit val pcEq: Eq[LegacyPupilCamera] = Eq.by(_.displayValue)
 
   final case class AOFlags(useAo: Boolean,
                            useCal: Boolean,
                            aoOptimize: Boolean,
-                           alignFpm: Boolean)
+                           alignFpm: Boolean,
+                           magH: Double,
+                           magI: Double)
```
swalker2m: Shouldn't we make more specific types for this information?

modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPIController.scala
```diff
@@ -7,29 +7,137 @@ import cats.{Eq, Show}
...
+        (config.shutters.entranceShutter === LegacyShutter.OPEN)
+          .fold(1, 0)
+          .show)
+      .withConfiguration("gpi:selectShutter.calExitShutter", "-1")
+      .withConfiguration("gpi:observationMode.mode",
+                         obsModeLUT.getOrElse(config.mode, ""))
+      .withConfiguration("gpi:selectPupilCamera.deploy",
+                         (config.pc === LegacyPupilCamera.IN)
+                           .fold(1, 0)
+                           .show)
```
swalker2m: These are the things sent to the instrument itself? That is, `1` or `0`?

modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPIController.scala
cquiroz: i agree but I want to wait to have the `maginutde` from gem integrated

modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPIController.scala
cquiroz: yes we send 1 or 0

PR #559, opened 2018-06-27
SEQNG-625 Use gem.Observation.Id on the seqexec

cquiroz:
This started as a simple exercise about replacing the sequence ids used on the seqexec by `gem.Observation.Id`. it ended up being a massive code change

The seqexec used observation ids from the old ocs2 model and also as just plain strings. This changes all that for `gem.Observation.Id`

I'm throwing the PR in case @swalker2m has time to check it but it still needs some amount of testing  before it could be merged


modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/package.scala
```diff
@@ -3,8 +3,48 @@
...
+              case Some(idx) =>
+                state.enc.writeInt(-idx)
+                ()
+              case None =>
+                // We need a marker
+                state.enc.writeInt(0)
+                // Use the prism to get a representatative for the A and encode it
+                PA.pickle(p.reverseGet(obj))
+                state.addIdentityRef(obj)
+            }
```
swalker2m: Is this related to the change to `Observation.Id`?  Either way, it seems like a good idea to create these instances from the optics.

cquiroz:
Checked manually that this changes leads to an increase of the client size from 2.75M to 2.91M. I expected a somewhat large increase and this is probably acceptable. As we integrate more parts of the gem model the increases should be lower as current model code is removed

cquiroz:
@tpolecat I was also hit by the travis limit in this PR. I added similar changes to the build as I suggested for #535

modules/seqexec/engine/src/main/scala/seqexec/engine/Sequence.scala
```diff
@@ -25,9 +25,7 @@ final case class Sequence(
 
 object Sequence {
 
-  type Id = String
-
-  def empty[A](id: Id, m: SequenceMetadata): Sequence = Sequence(id, m, Nil)
+  def empty[A](id: Observation.Id, m: SequenceMetadata): Sequence = Sequence(id, m, Nil)
```
cquiroz: @jluhrs We could make the `Id` a type parameter though we probably would never use anything but `Observation.Id`. WDYT?

modules/seqexec/model/src/main/scala/seqexec/model/Model.scala
```diff
@@ -93,7 +94,6 @@ object Model {
   type StepConfig = Map[SystemName, Parameters]
   implicit val stEq: Eq[StepConfig] = Eq.fromUniversalEquals
   // TODO This should be a richer type
```
cquiroz: I should remove this comment too

modules/seqexec/server/src/main/scala/seqexec/server/ODBProxy.scala
```diff
@@ -5,54 +5,55 @@ package seqexec.server
 
 import cats.data.EitherT
 import cats.effect.IO
+import cats.implicits._
 import edu.gemini.pot.sp.SPObservationID
```
cquiroz: I think this is the only reference left to ocs2 `SPObservationID` as we need to talk to the old odb. Otherwise we use `Observation.Id` everywhere else

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -207,29 +207,27 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
...
   ).getOrElse(st)
 
-  def moveSequenceInQueue(q: EventQueue, queueName: String, seqId: SPObservationID, d: Int): IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
-    Event.modifyState[executeEngine.ConcreteTypes](moveSeq(queueName, seqId.stringValue, d), NullSeqEvent)
+  def moveSequenceInQueue(q: EventQueue, queueName: String, seqId: Observation.Id, d: Int): IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
+    Event.modifyState[executeEngine.ConcreteTypes](moveSeq(queueName, seqId, d), NullSeqEvent)
   ).map(_.asRight)
 
   def notifyODB(i: (executeEngine.EventType, executeEngine.StateType)): IO[(executeEngine.EventType, executeEngine.StateType)] = {
-    def safeGetObsId(ids: String): SeqAction[SPObservationID] = EitherT(IO.apply(new SPObservationID(ids)).attempt.map { _.leftMap(SeqexecFailure.SeqexecException.apply) })
```
cquiroz: i'm happy we are able to remove this

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/SeqexecUI.scala
```diff
@@ -118,6 +121,19 @@ object SeqexecUI {
     case _                            => s"Seqexec - ${site.show}"
   }
 
+  // Prism from url params to config page
+  def configPageP(instrumentNames: Map[String, Instrument]): Prism[(String, String, Int), SequenceConfigPage] = Prism[(String, String, Int), SequenceConfigPage] {
```
cquiroz: it would be nice to simplify these but there is not at the moment a simple way to compose these in `Monocle`

modules/seqexec/engine/src/main/scala/seqexec/engine/Sequence.scala
jluhrs: I cannot think of any case where we would need another type.

cquiroz:
I have fixed this with the PR comments and added some fixes after doing some testing. The seqexec seems to work fine

It should be ready for a final review

modules/seqexec/engine/src/test/scala/seqexec/engine/StepSpec.scala
```diff
@@ -197,7 +196,7 @@ class StepSpec extends FlatSpec {
           (seqId,
             Sequence.State.Zipper(
               Sequence.Zipper(
-                id = "First",
+                id = Observation.Id.unsafeFromString("GS-2018A-Q-3"),
```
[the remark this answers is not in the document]
tpolecat: Yeah just to avoid confusion I would use a normal id here and elsewhere.

```scala
scala> gem.Observation.Id.unsafeFromString("GS-2018A-Q-3") // works, but weird
res10: gem.Observation.Id = Id(Nonstandard(Some(GS),Some(Semester(2018A)),None,Q),Index(3))

scala> gem.Observation.Id.unsafeFromString("GS-2018A-Q-3-123")
res11: gem.Observation.Id = Id(Science(GS,Semester(2018A),Q,Index(3)),Index(123))
```

You could also generate a random one.

```scala
scala> gem.arb.ArbObservation.arbObservationId.arbitrary.sample.get
res12: gem.Observation.Id = Id(Daily(GN,ENG,2017-09-20),Index(97))
```

cquiroz:
I tried to fix all those but obviously I missed some

PR #561, opened 2018-06-29
SEQNG-617: Support custom observing modes

cquiroz:
This PR add supports for setting GPI non standard modes. 
I took the opportunity to improve the giapi API hiding some of the javaness behind a nicer scala API

Here is a video of a sequence executed on GPI and changing the filter as we go

https://monosnap.com/file/WcCtPZ8e8cawtQhllGpai8qm6r3CIg#

### DIFF
--- a/modules/giapi/src/main/scala/giapi/client/commands.scala
+++ b/modules/giapi/src/main/scala/giapi/client/commands.scala
@@ -3,19 +3,47 @@
 
 package giapi.client
 
-import cats.Eq
+import cats._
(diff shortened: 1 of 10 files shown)


arturog8m:
👏 

The video will be so much cooler if you put the GPI ISD on the side to see how the GPI mechanisms are  configuring - I am assuming this was done with the real instrument, right?

cquiroz:
Even better with the ISD running in the browser 😉 
Yes this is with the real instrument

PR #562, opened 2018-07-02
Update to scala.js 0.6.24

cquiroz:
A minor scala.js update got a lot more complicated due to a bug fix on the floating point formatting algorithim


jluhrs:
What was the bug that was fixed?

cquiroz:
The old scala.js was less strict about string formatting of floats so the old string worked fine. The latest version corrects it but it brings out this bug. it is related to leading zeros on floating point printing

PR #565, opened 2018-07-05
SEQNG-612: Refactor the Seqexec to support DHS and GDS

cquiroz:
On the process to support GDS I realized that the base code on the seqexec server needs a refactoring to make it less dhs specific. Most of the keyword related logic assumes the DHS is used with no room for the GDS which behaves differently.

Another aspect is there is an assumption of a single DHS while there are multiple GDS. This is now abstracted away on a `KeywordsClient` that is a property of `InstrumentSystem`

At the end there is still no logic to actually send commands to the GDS but this refactor will let us work with both DHS and GDS

Further changes will be needed in the near future but I wanted to send this PR to make it easier to review

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/Header.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/Header.scala
@@ -4,11 +4,16 @@
 package seqexec.server
 
 import seqexec.model.dhs.ImageFileId
-import seqexec.server.keywords.DhsClient
-import seqexec.server.keywords.DhsClient._
+import seqexec.server.keywords._
 
+/**
+ * Typeclass for types that can send keywords
+ */
 trait HeaderProvider[A] {
+  // Name of the instrument to be sent to the DHS
   def name(a: A): String
+  // Client to send keywords to an appropriate server
+  def keywordsClient(a: A): KeywordsClient
 }
 
 object HeaderProvider {
@@ -16,14 +21,18 @@ object HeaderProvider {
 
   final class HeaderProviderOps[A: HeaderProvider](val a: A) {
     def name: String = HeaderProvider[A].name(a)
+    def keywordsClient: KeywordsClient = HeaderProvider[A].keywordsClient(a)
   }
 
   implicit def ToHeaderProviderOps[A: HeaderProvider](a: A): HeaderProviderOps[A] = new HeaderProviderOps(a)
 }
 
+/**
+ * Header implementations know what headers sent before and after an observation
+ */
 trait Header {
-  def sendBefore[A: HeaderProvider](id: ImageFileId, inst: A): SeqAction[Unit]
-  def sendAfter[A: HeaderProvider](id: ImageFileId, inst: A): SeqAction[Unit]
+  def sendBefore(id: ImageFileId): SeqAction[Unit]
+  def sendAfter(id: ImageFileId): SeqAction[Unit]
 }
 
 object Header {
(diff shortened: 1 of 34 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -21,6 +21,17 @@ trait InstrumentSystem extends System {
 }
 
 object InstrumentSystem {
+
+  implicit val HeaderProvider: HeaderProvider[InstrumentSystem] = new HeaderProvider[InstrumentSystem] {
+    def name(a: InstrumentSystem): String = a match {
+      case i: DhsInstrument => i.dhsInstrumentName
+      case _                => sys.error("Missing instrument")
```
cquiroz: In the near future we'll add a `GDSInstrument`

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -335,6 +339,15 @@ class SeqexecEngine(settings: SeqexecEngine.Settings) {
 
 // Configuration stuff
 object SeqexecEngine {
+  sealed trait GPIKeywords
+
+  object GPIKeywords {
```
cquiroz: I'd like to use ADTs rather than booleans for flags

modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2.scala
```diff
@@ -19,7 +20,7 @@ import cats.data.{EitherT, Reader}
 import cats.effect.IO
 import cats.implicits._
 
-final case class Flamingos2(f2Controller: Flamingos2Controller) extends InstrumentSystem {
+final case class Flamingos2(f2Controller: Flamingos2Controller, dhsClient: DhsClient) extends InstrumentSystem with DhsInstrument {
```
cquiroz: Now instruments have a reference to their keywords client

modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.http4s.client.Client
+
+/**
+  * Gemini Data service client
+  */
+final case class GDSClient[F[_]](client: Client[F]) extends KeywordsClient {
+  /**
+    * Set the keywords for an image
+    */
+  def setKeywords(id: ImageFileId, keywords: KeywordBag, finalFlag: Boolean): SeqAction[Unit] = {
```
cquiroz: TODO: do the actual calls over xmlrpc

modules/seqexec/server/src/main/scala/seqexec/server/Header.scala
```diff
@@ -28,14 +54,14 @@ object Header {
   def buildBoolean(get: SeqAction[Boolean], name: String ): KeywordBag => SeqAction[KeywordBag] = buildKeyword(get, name, BooleanKeyword)
   def buildString(get: SeqAction[String], name: String ): KeywordBag => SeqAction[KeywordBag]   = buildKeyword(get, name, StringKeyword)
 
-  private def bundleKeywords(inst: String, ks: Seq[KeywordBag => SeqAction[KeywordBag]]): SeqAction[KeywordBag] = {
-    val z = SeqAction(KeywordBag(StringKeyword("instrument", inst)))
+  private def bundleKeywords[A: HeaderProvider](inst: A, ks: Seq[KeywordBag => SeqAction[KeywordBag]]): SeqAction[KeywordBag] = {
+    val z = SeqAction(KeywordBag(StringKeyword("instrument", inst.name)))
```
jluhrs: This is only required for DHS. I think it is better if the "instrument" keyword is added inside DhsClientHttp.setKeywords. The instrument string can be passed as a constructor parameter.
If done that way, I think the "instrument" keywords must be the first one, so it should be prepended.

cquiroz: Yes, this should be isolated for DHS instruments only, or do a `SeqAction.void` for `GDSInstrument`

PR #567, opened 2018-07-06
SEQNG-590 Removed EPICS reads from EPICS monitor callback.

jluhrs:
This PR fix the problem with the cascade of "unresponsive channel" messages coming from CAJ. It seems that is not a good idea to call CAJ functions inside a channel monitor callback.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
@@ -93,24 +93,6 @@ public class CaObserveSenderImpl<C extends Enum<C> & CarStateGeneric> implements
 
     };
 
-    private short getMark(ReadOnlyClientEpicsChannel<Short> ch, short def) {
-        if(ch!=null && ch.isValid()) {
-            try {
-                return ch.getFirst();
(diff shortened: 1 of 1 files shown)


modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -588,7 +570,13 @@ public State onTimeout() {
         final short stopMark;
         final short abortMark;
 
-        WaitObserveCompletion(CaCommandMonitorImpl cm, short stopMark, short abortMark) {
+        WaitObserveCompletion(CaCommandMonitorImpl cm) {
+            this.cm = cm;
+            this.stopMark = MRK_IDLE;
+            this.abortMark = MRK_IDLE;
```
jluhrs: Assume initial value for stop.MARK and abort.MARK channels. This still works, because we only care about them when they change while the observation is in progress (i.e. while in state `WaitObserveCompletion`)

cquiroz:
This is great. Do you think it's worth to add a test case for this?

jluhrs:
I wouldn't know what to test.

swalker2m:
Glad you found this. I think it might have taken me a few years.

PR #569, opened 2018-07-09
SEQNG-644:  Webclient does not show logs

cquiroz:
Fix for logs not being sent to the client. It also does some improvements to the logs table ui


modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -143,7 +143,7 @@ object WebServerLauncher extends StreamApp[IO] with LogInitialization {
       asyncAppender.addAppender(appender)
     }
 
-    Option(LoggerFactory.getLogger("edu.gemini.seqexec")).collect {
+    Option(LoggerFactory.getLogger("seqexec")).collect {
```
jluhrs: Doh!

PR #572, opened 2018-07-09
SEQNG-648: Use Site type from core

cquiroz:
Another removal of a duplicate case from the seqexec mini model and use of the core `Site` class instead

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/ModelOps.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/ModelOps.scala
@@ -3,14 +3,21 @@
 
 package seqexec.web.client
 
-import seqexec.model.Model.{Resource, Instrument, SequenceState, SequenceView, Step, StepState, StandardStep}
 import cats.Show
 import cats.implicits._
+import cats.data.NonEmptyList
+import gem.enum.Site
+import seqexec.model.Model.{Resource, Instrument, SequenceState, SequenceView, Step, StepState, StandardStep}
 
 /**
   * Contains useful operations for the seqexec model
   */
 object ModelOps {
+  implicit val siteStateShow: Show[Site] = Show.show[Site] {
+    case Site.GN => "GN"
+    case Site.GS => "GS"
+  }
+
   implicit val sequenceStateShow: Show[SequenceState] = Show.show[SequenceState] {
     case SequenceState.Completed        => "Complete"
     case SequenceState.Running(true, _) => "Pausing..."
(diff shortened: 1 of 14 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/ModelOps.scala
```diff
@@ -3,14 +3,21 @@
...
 import cats.implicits._
+import cats.data.NonEmptyList
+import gem.enum.Site
+import seqexec.model.Model.{Resource, Instrument, SequenceState, SequenceView, Step, StepState, StandardStep}
 
 /**
   * Contains useful operations for the seqexec model
   */
 object ModelOps {
+  implicit val siteStateShow: Show[Site] = Show.show[Site] {
```
cquiroz: @tpolecat Why there is no `Show[Site]` Is it because sometimes you want `GN/GS` and sometimes the full name?

cquiroz: if `Show` is always `toString` what's the point of having them? I guess I can just use `site.shortName` then

tpolecat: `Show` isn't very useful because very few things have a canonical string representation; and those that do often have an associated parser, which `Show` says nothing about. So I just see it as a better `toString`; better in the sense that you have to opt in.

tpolecat: There is a `Display` typeclass in `Enumerated.scala` that we might revisit. There are no instances, I was just trying to sketch out what a typeclass for UI elements might look like.

cquiroz: hmm, I don't see why would it be better than a plain to`toString` if we delegate. I see the point of the parsing and having better ways to display but if `Show` is just a glorified `toString` I'd rather remove them and save us the extra code it requires

tpolecat: It's better in the sense that not everything has a useful `toString` (functions, for example) so `Show` prevents you from trying to print out things that will be unreadable. But the larger principle is that `toString` and `hashCode` and `equals` and other methods defined on `Any` break parametricity and make our types less meaningful, so we should avoid using them.

But the case for `toString` is weak since it's not that useful to begin with. It wouldn't bother me that much to get rid of the instances, but we should wait and chat with @swalker2m about it.

cquiroz:
Ok, I switched out of `Show` instead using `site.shortName`. I can see the point of the discussion and we can revisit with shane and the others.

PR #577, opened 2018-07-10
SEQNG-614: GDS Client

cquiroz:
This is the first part of the GDS client. at the moment is able to send all the keywords for the standard headers to the GDS.

There is still some amount of work to get the initial and final events and the GPI specific keywords but this is the bulk of the work to get GDS support


modules/seqexec/server/src/main/scala/seqexec/server/Header.scala
```diff
@@ -3,33 +3,32 @@
...
 
   final class HeaderProviderOps[A: HeaderProvider](val a: A) {
-    def name: String = HeaderProvider[A].name(a)
-    def keywordsClient: KeywordsClient = HeaderProvider[A].keywordsClient(a)
+    def keywordsClient: KeywordsClient[IO] = HeaderProvider[A].keywordsClient(a)
   }
 
-  implicit def ToHeaderProviderOps[A: HeaderProvider](a: A): HeaderProviderOps[A] = new HeaderProviderOps(a)
+  implicit def ToHeaderProviderOps[A: HeaderProvider](
+      a: A): HeaderProviderOps[A] = new HeaderProviderOps(a)
```
tpolecat: Kind of a weird place for a newline.

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -3,43 +3,36 @@
...
-  final case class StopObserveCmd(self: SeqAction[Unit]) extends AnyVal
+  final case class StopObserveCmd(self: SeqAction[Unit])  extends AnyVal
   final case class AbortObserveCmd(self: SeqAction[Unit]) extends AnyVal
   final case class PauseObserveCmd(self: SeqAction[Unit]) extends AnyVal
   final case class ContinuePausedCmd(self: Time => SeqAction[ObserveCommand.Result]) extends AnyVal
   final case class StopPausedCmd(self: SeqAction[ObserveCommand.Result]) extends AnyVal
   final case class AbortPausedCmd(self: SeqAction[ObserveCommand.Result]) extends AnyVal
+
+  sealed trait ObserveControl
+  object Uncontrollable extends ObserveControl
```
tpolecat: Why not `case object`?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -503,6 +506,7 @@ object SeqexecEngine {
     val gcalKeywords            = cfg.require[Boolean]("seqexec-engine.gcalKeywords")
     val gnirsKeywords           = cfg.require[Boolean]("seqexec-engine.gnirsKeywords")
     val gpiKeywords             = cfg.require[Boolean]("seqexec-engine.gpiKeywords").fold(GPIKeywords.GPIKeywordsGDS, GPIKeywords.GPIKeywordsSimulated)
+    val gpiGDS                  = Uri.unsafeFromString(cfg.require[String]("seqexec-engine.gpiGDS"))
```
tpolecat: I'm surprised there's not already a `URI` decoder that lets us do `.require[URI]` ... should we add one?

modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
```diff
@@ -3,18 +3,100 @@
...
+          </value>
+        </param>
+        <param>
+          <value>
+            <array>
+              <data>
+                {
+                  for {
+                    k <- ks.keywords
+                  } yield <value><string>{s"${k.name},${k.keywordType.gdsType},${k.value}"}</string></value>
```
tpolecat: This can be simplified to `ks.keywords.map { k => ... }`.

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -3,84 +3,157 @@
...
+    def apply[A](k1: Keyword[A]): KeywordBag =
+      KeywordBag(List(internalKeywordConvert(k1)))
+
+    def apply[A, B](k1: Keyword[A], k2: Keyword[B]): KeywordBag =
+      KeywordBag(List(internalKeywordConvert(k1), internalKeywordConvert(k2)))
+
+    def apply[A, B, C](k1: Keyword[A],
+                       k2: Keyword[B],
+                       k3: Keyword[C]): KeywordBag =
+      KeywordBag(
```
tpolecat: Surely we can do better than all this copy/paste. How about

```scala
def apply(ks: Keyword[_]*): KeywordBag = 
  KeywordBag(ks.map(internalKeywordConvert).toList))
```

Or, better, `KeywordBag` is a monoid, which (when defined) would let you implement it as 

```scala
def apply(ks: Keyword[_]*): KeywordBag = 
  ks.toList.foldMap(KeywordBag.apply)
```

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
cquiroz: Right, it was historically like that but we should change it

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
cquiroz: Could be, in the same line as for `Site`

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
cquiroz: `Keyword[_]` scares me 😄

tpolecat: You can write it as `Keyword[A] forSome { type A }` if it makes you feel better ;-)

modules/seqexec/server/src/main/scala/seqexec/server/Header.scala
cquiroz: This was done by `scalafmt`, it has some quirks

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -3,84 +3,157 @@
...
-  object TypeString extends KeywordType("STRING")
-
+  sealed case class KeywordType protected (str: String, gdsType: String)
+  object TypeInt8    extends KeywordType("INT8", "INT")
+  object TypeInt16   extends KeywordType("INT16", "INT")
+  object TypeInt32   extends KeywordType("INT32", "INT")
+  object TypeFloat   extends KeywordType("FLOAT", "DOUBLE")
+  object TypeDouble  extends KeywordType("DOUBLE", "DOUBLE")
+  object TypeBoolean extends KeywordType("BOOLEAN", "BOOLEAN")
+  object TypeString  extends KeywordType("STRING", "STRING")
```
jluhrs: This is another instance where I let a system specific detail slip into a class that was supposed to be generic. The parameter `str` only has meaning for the DHS implementation, as the new `gdsType` only as meaning for the GDS implementation. IMHO, a better way to do this is to remove those values from `KeywordType`, and translate the types to the implementation specific values inside the DHS and GDS code.

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -3,84 +3,157 @@
...
 
   // At the end, I want to just pass a list of keywords to be sent to the DHS. I cannot do this with Keyword[T],
   // because I cannot mix different types in a List. But at the end I only care about the value as a String, so I
   // use an internal representation, and offer a class to the developer (KeywordBag) to create the list from typed
   // keywords.
 
-  private[keywords] final case class InternalKeyword(name: String, keywordType: KeywordType, value: String)
+  private[keywords] final case class InternalKeyword(name: String,
+                                                     keywordType: KeywordType,
+                                                     value: String)
```
jluhrs: The parameter `value` is a `String` because in DHS, values are sent as strings. Does this make sense for GDS?

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
jluhrs: My :+1: for the `Monoid` suggestion.

cquiroz:
I addressed the comments and added a few tests too

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -3,84 +3,126 @@
...
+  final case class StringKeyword(name: String, value: String)
+      extends Keyword[String](name, TypeString, value)
 
   // At the end, I want to just pass a list of keywords to be sent to the DHS. I cannot do this with Keyword[T],
   // because I cannot mix different types in a List. But at the end I only care about the value as a String, so I
   // use an internal representation, and offer a class to the developer (KeywordBag) to create the list from typed
   // keywords.
 
-  private[keywords] final case class InternalKeyword(name: String, keywordType: KeywordType, value: String)
+  private[server] final case class InternalKeyword(name: String,
```
cquiroz: I think we should tighten this and use type for name and value that takes into account, e.g. the length limits of FITS keywords, maybe in the next PR

modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
```diff
@@ -3,18 +3,111 @@
...
+      .expect[Elem](postRequest)(scalaxml.xml)
+      .attemptT
+      .leftMap {
+        case e: Throwable =>
+          SeqexecFailure.GDSException(e, gdsUri): SeqexecFailure
+      }
+      .flatMap(xml => EitherT.fromEither(GDSClient.checkError(xml, gdsUri)))
+  }
+
+  private def keywordType(k: KeywordType): String = k match {
```
tpolecat: Is there any reason these two string mappings aren't defined on `KeywordType` itself? I kind of wonder if we should push this to the database and generate it into `gem.enum`. Fine for now, just something to consider.

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -3,84 +3,126 @@
...
-        internalKeywordConvert(k4)))
-    def apply[A, B, C, D, E](k1: Keyword[A], k2: Keyword[B], k3: Keyword[C], k4: Keyword[D], k5: Keyword[E]): KeywordBag =
-      KeywordBag(List(internalKeywordConvert(k1), internalKeywordConvert(k2), internalKeywordConvert(k3),
-        internalKeywordConvert(k4), internalKeywordConvert(k5)))
-    def apply[A, B, C, D, E, F](k1: Keyword[A], k2: Keyword[B], k3: Keyword[C], k4: Keyword[D], k5: Keyword[E], k6: Keyword[F]): KeywordBag =
-      KeywordBag(List(internalKeywordConvert(k1), internalKeywordConvert(k2), internalKeywordConvert(k3),
-        internalKeywordConvert(k4), internalKeywordConvert(k5), internalKeywordConvert(k6)))
+
+    implicit val eq: Eq[KeywordBag] = Eq.by(_.keywords)
+    implicit val monoid: Monoid[KeywordBag] = new Monoid[KeywordBag] {
```
tpolecat: 👍

modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
cquiroz: Because they are different for the gds and dhs

cquiroz:
I think we could move some of the keyword related classes to core

modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
tpolecat: Right, I see that they're different. But both could be defined on the enum.

cquiroz: It maybe good as an enum though I'd rather not have a db yet for the seqexec, though we will probably have one sooner than later

PR #578, opened 2018-07-10
SEQNG-648: Remove legacy Site class usage

cquiroz:
I missed removing a couple of usages of the old `Site` class

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -8,7 +8,6 @@ import cats.data.{EitherT, Kleisli, NonEmptyList, Reader}
 import cats.effect.IO
 import cats.implicits._
 import edu.gemini.seqexec.odb.{ExecutedDataset, SeqexecSequence}
-import edu.gemini.spModel.core.Site
 import seqexec.engine.Result.{Configured, FileIdAllocated, Observed}
 import seqexec.engine.{Action, ActionMetadata, Event, Result, Sequence, Step, fromIO}
 import seqexec.model.Model.{Instrument, Resource, SequenceMetadata, StepState}
@@ -37,8 +36,8 @@ import edu.gemini.spModel.seqcomp.SeqConfigNames._
 import org.log4s._
 import squants.Time
 import fs2.Stream
-// import org.http4s.client.Client
 import gem.Observation
+import gem.enum.Site
 import mouse.all._
 
 class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
@@ -333,7 +332,8 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     case _                      => TrySeq.fail(Unexpected(s"Instrument $inst not supported."))
   }
 
-  private def calcResources(sys: List[System]): Set[Resource] = sys.map(resourceFromSystem).toSet
+  private def calcResources(sys: List[System]): Set[Resource] =
+    sys.map(resourceFromSystem).toSet
 
   import TcsController.Subsystem._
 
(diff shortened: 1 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -394,7 +394,7 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
   private def commonHeaders(config: Config, tcsSubsystems: List[TcsController.Subsystem], inst: InstrumentSystem)(ctx: ActionMetadata): Header =
     new StandardHeader(
       inst,
-      ObsKeywordReaderImpl(config, site.displayName.replace(' ', '-')),
+      ObsKeywordReaderImpl(config, site.longName.replace(' ', '-')),
```
tpolecat: What's the meaning of changing `' '` to `'-'`? If we decide to change `longName` at some point will it break things? Do we need to add a `tcsValue` or something?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -484,11 +485,7 @@ object SeqexecEngine {
 
   // scalastyle:off
   def seqexecConfiguration(giapi: Giapi[IO]): Kleisli[IO, Config, Settings] = Kleisli { cfg: Config =>
-    val site = cfg.require[String]("seqexec-engine.site") match {
-      case "GS" => Site.GS
-      case "GN" => Site.GN
-      case _    => Site.GS // Let's default to GS
-    }
+    val site                    = Site.fromTag(cfg.require[String]("seqexec-engine.site")).getOrElse(Site.GS)
```
tpolecat: Can you define a typeclass instance here so you can say `cfg.require[Site]` rather than inlining the logic here? The benefit is that you explicitly say that the `tag` value is how we encode sites in the config.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
cquiroz: You're right we should have the actual values somewhere else. Given it is only used here, it could live in the seqexec rather than being a part of `Site`

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
cquiroz: That's a good idea, let's see if I can figure it out

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
tpolecat: I could argue it either way I guess. Lots of things in the model will need TCC values and it wouldn't bother me to just define them in the model and be done with it. But as you say most of the time we don't care and we could define them somewhere else.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -484,11 +485,7 @@ object SeqexecEngine {
 
   // scalastyle:off
   def seqexecConfiguration(giapi: Giapi[IO]): Kleisli[IO, Config, Settings] = Kleisli { cfg: Config =>
-    val site = cfg.require[String]("seqexec-engine.site") match {
-      case "GS" => Site.GS
-      case "GN" => Site.GN
-      case _    => Site.GS // Let's default to GS
-    }
+    val site                    = cfg.require[Site]("seqexec-engine.site")
```
cquiroz: @tpolecat This looks much better. I guess I'd like not to write these as String but `knobs` doesn't let me override that. Anyway this is less boilerplaty and much more explicit

PR #579, opened 2018-07-10
switch from travis to buildkite (#574)

tpolecat:
This switches the build to use [BuildKite](https://buildkite.com/) instead of Travis. This lets us use our own hardware to build stuff.

- The build runs on `sbfocsdev-lv1` which is a VM with 16G of memory and 4 Xeon processors. It can do the full build including running the JS tests in a little over 9 minutes. Right now there is only one machine but we could add more to do parallel builds.
 - The build runs by way of the buildkite agent daemon, run as the `buildkite-agent` user. This daemon polls BuildKite for work, so we don't need to open any ports.
- BuildKite watches GitHub and pushes work to agents to build PRs and merged/pushed branches (just to the `develop-buildkite-temp` branch for now but we'll remove the filter once this is merged) and reports status back to GH just like Travis does.
- Because the build runs on *our* hardware we need to be careful. Tests and the build itself can run arbitrary code, so the build will immediately fail for PRs that come from untrusted repos. Trusted ones are simply hardcoded in the `/etc/buildkite-agent/hooks/environment` script which is run before the build starts. Right now Carlos and I are the only ones who are set up.

TODO:

- [x] invite everyone to the `gemini-hlsw` organization on BuildKite
- [x] document the setup on the wiki
- [ ] get vague nods of approval from everyone

After merging:

- [ ] remove branch filters
- [ ] may need to kick existing PRs


build/buildkite-build.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# The postgres user already exists, so we can go ahead and create the database
+until docker exec $CID psql -U postgres -c 'create database gem'
+do
+  echo "waiting for postgres container..."
+  sleep 0.5
+done
+
+# Set up the schema and run tests
+/usr/local/bin/sbt                                        \
```
cquiroz: Would it be better to run all sbt tasks in one go to benefit from the JVM being warm?

tpolecat:
The Travis CI will never report success because the build definition is gone, so we'll have to turn off the branch protection temporarily to merge this.

After merging we may end up with some hung PRs that need to be closed and re-opened.

tpolecat:
Added https://github.com/gemini-hlsw/ocs3/wiki/Continuous-Integration

tpolecat:
@jdnavarro @jluhrs @arturog8m can I get a 👍 or other reaction from each of you just acknowledging what's happening? This is a reasonably big process change and I don't want to surprise anyone.

swalker2m:
> @jdnavarro @jluhrs @arturog8m can I get a 👍 or other reaction from each of you just to acknowledge you read the description above?

Looks good to me. 👍 Eventually it would probably be good to get a couple of VMs but we can see how it goes I suppose.

arturog8m:
🚢 it 👍

jluhrs:
My vague :+1:

PR #580, opened 2018-07-12
keyword package refactoring

cquiroz:
This PR contains some refactorings about keywords:

* Some classes are moved to the keywords package
* Made implicit conversions explicit
* Move implicits to the `keywords` package
* Split some large classes

Mostly mechanical work


modules/seqexec/server/src/main/scala/seqexec/server/gcal/GcalKeywordReader.scala
```diff
@@ -18,13 +18,13 @@ trait GcalKeywordReader {
 
 object DummyGcalKeywordsReader extends GcalKeywordReader {
 
-  def getDiffuser: SeqAction[Option[String]] = None
+  def getDiffuser: SeqAction[Option[String]] = None.toSeqActionO
```
tpolecat: Oh was this an implicit conversion before? 👍 for explicit syntax.

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -119,10 +137,75 @@ package keywords {
       KeywordBag(ks.toList.map(internalKeywordConvert))
 
   }
+
+  // A simple typeclass to encapsulate default values
+  trait DefaultValue[A] {
```
tpolecat: This is kind of a dangerous typeclass because it's lawless and easily abused. Can we give it a more specific name and restrict its usage?

```scala
sealed abstract class DefaultHeaderValue[A] private {
  ...
}
object DefaultHeaderValue {
  // define every instance here
}
```

cquiroz: Makes sense

modules/seqexec/server/src/main/scala/seqexec/server/gcal/GcalKeywordReader.scala
cquiroz: yes, it was implicit and very confusing for me

PR #581, opened 2018-07-12
smaller build steps

tpolecat:
Messing around with the build script. I'm thinking maybe breaking it up into smaller steps will make it more reliable.


tpolecat:
This increases the build time by like 30 seconds but it seems to be more stable, at least relative to earlier today. I ran it 10 times and it failed twice.

cquiroz:
Seems good, 30s sounds worth more stability

tpolecat:
This also moves the pipeline definition to a YAML file which allows us to specify retries (set to 2 for now).

tpolecat:
@cquiroz can you approve this? I think it's worth doing.

PR #582, opened 2018-07-12
SEQNG-613: GPI Keywords

cquiroz:
This PR does the bulk of the work to get keywords onto GPI images. It has been tested against the instrument and it gets the keywords written except for the following exceptions:

* CRFOLLOW which seems to be not in use. we'll follow up with science
* Keywords from gcal/gws. They need to be modified to work with either the DHS or GDS

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
@@ -34,28 +34,18 @@ final case class GDSClient(client: Client[IO], gdsUri: Uri)
             <string>{id}</string>
           </value>
         </param>
-        <param>
-          <value>
-            <array>
-              <data>
-                {
-                  ks.keywords.map { k =>
-                    <value><string>{s"${k.name},${KeywordType.gdsKeywordType(k.keywordType)},${k.value}"}</string></value>
-                  }
-                }
-              </data>
-            </array>
-          </value>
-        </param>
+        {keywordsParam(ks)}
       </params>
     </methodCall>
 
-  private def handleConnectionError: PartialFunction[Throwable, SeqexecFailure] = {
+  private def handleConnectionError
+    : PartialFunction[Throwable, SeqexecFailure] = {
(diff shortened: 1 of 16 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
```diff
@@ -33,22 +34,19 @@ final case class GDSClient(client: Client[IO], gdsUri: Uri)
...
-              </data>
-            </array>
-          </value>
-        </param>
+        {keywordsParam(ks)}
       </params>
     </methodCall>
 
+  private def handleConnectionError
+    : PartialFunction[Throwable, SeqexecFailure] = {
```
tpolecat: Why is this a `PartialFunction`?

cquiroz: We pass it to `EitherT.leftMap` which takes that signature

tpolecat: It does? Weird.

tpolecat: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/EitherT.scala#L109

?

cquiroz: Oh, i got confused with the transformers i guess when trying to disentangle the monad transformers stack

tpolecat: I would 👍 a PR to remove all the type aliases ;-)

cquiroz: yeah, maybe I can come up with some magic sed command to do it

cquiroz: Fixed the signature

PR #584, opened 2018-07-13
SEQNG-654: Support CRFOLLOW

cquiroz:
This adds support for the CRFOLLOW keyword in Gpi and also fixes reading the corresponding epics channels from the TCS

### DIFF
--- a/modules/seqexec/server/src/main/resources/Tcs.xml
+++ b/modules/seqexec/server/src/main/resources/Tcs.xml
@@ -1011,6 +1011,11 @@
             <type>INTEGER</type>
             <description>CASS Rotator Follow</description>
         </attribute>
+        <attribute name="parangle">
+            <channel>parAngle</channel>
+            <type>DOUBLE</type>
+            <description>Parallactic Angle</description>
+        </attribute>
         <attribute name="decson">
             <channel>chopControl.VALB</channel>
             <type>INTEGER</type>
--- a/modules/seqexec/server/src/main/resources/Tcs.xml
+++ b/modules/seqexec/server/src/main/resources/Tcs.xml
@@ -1007,12 +1007,12 @@
     <Status name="tcsstate">
         <top>tcs</top>
         <attribute name="crfollow">
-            <channel>crFollow</channel>
+            <channel>crFollow.VAL</channel>
             <type>INTEGER</type>
             <description>CASS Rotator Follow</description>
         </attribute>
         <attribute name="parangle">
-            <channel>parAngle</channel>
+            <channel>sad:parAngle.VAL</channel>
             <type>DOUBLE</type>
             <description>Parallactic Angle</description>
         </attribute>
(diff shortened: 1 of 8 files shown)


modules/seqexec/server/src/main/resources/Tcs.xml
```diff
@@ -1006,6 +1006,16 @@
     </Status>
     <Status name="tcsstate">
         <top>tcs</top>
+        <attribute name="crfollow">
+            <channel>crFollow</channel>
```
jluhrs: crFollow.VAR, to be precise.

modules/seqexec/server/src/main/resources/Tcs.xml
```diff
@@ -1006,6 +1006,16 @@
...
+            <type>INTEGER</type>
+            <description>CASS Rotator Follow</description>
+        </attribute>
+        <attribute name="parangle">
+            <channel>parAngle</channel>
```
jluhrs: Its "sad:parAngle.VAL". Why this has the sad: prefix and the crfollow doesn't? No idea.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -104,6 +104,28 @@ object Tcs {
...
+        case 0  => Off.some
+        case 1  => On.some
+        case _  => none
+      } {
+        case Off => 0
+        case On  => 1
+      }
+
+  }
+
```
jluhrs: I am not sure this class is the right place for this piece of code.

cquiroz: Do you mean the whole CRFollow or just the prism. The Prism is used to decode the epics values

modules/seqexec/server/src/main/resources/Tcs.xml
cquiroz: [software@sbfgpidev1 ~]$ caget tcs:sad:parAngle.VAL
tcs:sad:parAngle.VAL           -41.472

modules/seqexec/server/src/main/resources/Tcs.xml
cquiroz: [software@sbfgpidev1 ~]$ caget tcs:crFollow.VAL
tcs:crFollow.VAL               0

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -104,6 +104,28 @@ object Tcs {
...
+  object CRFollow {
+    case object On extends CRFollow
+    case object Off extends CRFollow
+
+    def keywordValue(cr: CRFollow): String = cr match {
+      case On  => "yes"
+      case Off => "no"
+    }
+
+    def fromInt: Prism[Int, CRFollow] =
```
tpolecat: Can you add a law test for this prism?

PR #585, opened 2018-07-13
Auto-deploy to sbfocstest-lv1 for `develop` merges

tpolecat:
This is work toward getting continuous deployment working. It fixes the deploy stuff to use the versioning scheme @cquiroz came up with, so we're now consistent there; and it adds a step to create a docker image for `develop` merge builds (i.e., not for PRs or other branches) and deploys it to the test machine `sbfocstest-lv1`, HTTP REST on 9090 and telnet on 9091. Each deployment destroys the previous one; there is no data migration yet.

This also splits the log up into sections and adds wee emojis to delight and entertain.

image

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,9 +8,6 @@ import NativePackagerHelper._
 import sbtcrossproject.{crossProject, CrossType}
 import com.typesafe.sbt.packager.docker._
 
-// our version is determined by the current git state (see project/ImageManifest.scala)
-def imageManifest = ImageManifest.current("postgres:9.6.0").unsafeRunSync
-
 name := Settings.Definitions.name
 
 organization in Global := "edu.gemini.ocs"
@@ -287,9 +284,9 @@ lazy val ctl = project
       CatsFree.value,
       Decline
     ),
-    TaskKey[Unit]("deployTest") := (runMain in Compile).toTask {
-      s" gem.ctl.main --no-ansi --host sbfocstest-lv1.cl.gemini.edu deploy-test ${imageManifest.formatVersion}"
-    } .value,
+    // TaskKey[Unit]("deployTest") := (runMain in Compile).toTask {
+    //   s" gem.ctl.main --no-ansi --host sbfocstest-lv1.cl.gemini.edu deploy-test ${version.value}"
+    // } .value,
     fork in run := true
   )
 
@@ -305,7 +302,7 @@ lazy val main = project
     dockerBaseImage       := "openjdk:8u141",
     dockerExposedPorts    := List(9090, 9091),
     dockerRepository      := Some("sbfocsdev-lv1.cl.gemini.edu"),
-    dockerLabels          := imageManifest.labels,
+    dockerLabels          := ImageManifest.current("postgres:9.6.0", version.value).unsafeRunSync.labels,
 
     // Install nc before changing the user
     dockerCommands       ++= dockerCommands.value.flatMap {
--- a/build.sbt
+++ b/build.sbt
@@ -284,9 +284,11 @@ lazy val ctl = project
       CatsFree.value,
       Decline
     ),
-    // TaskKey[Unit]("deployTest") := (runMain in Compile).toTask {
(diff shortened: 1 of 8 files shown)


tpolecat:
note to self: squash-merge this

modules/ctl/src/main/scala/free/ctl.scala
```diff
@@ -53,7 +53,8 @@ object ctl {
         case None    =>
           o.lines.traverse(error(_))      *>
           error(s"exited (${o.exitCode})") *>
-          exit[A](o.exitCode)
+          error(s"Requirement failed. Exiting.") *>
+          exit[A](-1)
```
tpolecat: This fails the deployment if the output of a shell command doesn't match what was expected. Like, if we expected one line of output but we got two. Before it only failed if the command itself failed, which wasn't good enough. The point here is that ultimately `runMain` will exits abnormally in this case, signaling a failed build.

modules/ctl/src/main/scala/hi/deploy.scala
```diff
@@ -11,13 +11,14 @@ object Deploy {
...
-      findRunningContainersWithLabel("gem.version").flatMap {
-        case Nil => info("None found, nothing to do!")
-        case cs  => cs.traverse { c =>
-          info(s"Stopping and removing ${c.hash}.") *> destroyContainer(c)
-        }.void
+      currentDeployment.flatMap {
+        case None => info("None found, nothing to do!")
+        case Some(Deployment(g, p)) =>
+          containerImage(g).flatMap { oldGemImage =>
+            destroyContainer(g, !(oldGemImage like newGemImage)) *> destroyContainer(p, false)
```
tpolecat: When we shut down the existing deployment (containers `g` for Gem and `p` for Postgres) we also destroy `g`'s *image* if we're deploying a new version. Otherwise we just shut it down and the existing image will be re-used.

project/ImageManifest.scala
```diff
@@ -8,53 +8,20 @@ import java.time._
...
-  def formatVersion: String =
-    if (unstable) f"$j2000%8.4f-$commit-UNSTABLE"
-    else          f"$j2000%8.4f-$commit"
-
   /** Docker labels for this ImageManifest, in a format sbt-native-packager likes. */
   def labels: Map[String, String] =
     Map(
       Keys.Commit   -> commit,
-      Keys.History  -> history.intercalate(","),
+      // Keys.History  -> history.intercalate(","),
```
tpolecat: It turns out there is an internal limit to the size of image labels, so the strategy of listing all the commits (in order to determine upgrade compatibility) isn't going to work. It's off for now, pending another solution.

tpolecat:
Ok this is ready for review. The changes are uniformly fiddly and I have tried to explain them in comments above.

cquiroz:
I like the emojis

build.sbt
```diff
@@ -287,8 +284,10 @@ lazy val ctl = project
       CatsFree.value,
       Decline
     ),
-    TaskKey[Unit]("deployTest") := (runMain in Compile).toTask {
-      s" gem.ctl.main --no-ansi --host sbfocstest-lv1.cl.gemini.edu deploy-test ${imageManifest.formatVersion}"
+    TaskKey[Unit]("deployTest") := Def.taskDyn {
+      (runMain in Compile).toTask {
+        s" gem.ctl.main --verbose --host sbfocstest-lv1.cl.gemini.edu deploy-test ${version.value}"
```
cquiroz: Does this have any effect if e.g. we are not in VPN and this tries to connect to `sbfoctest-lv1`?

swalker2m: I think it is only called automatically for a merge to develop?

tpolecat: Yeah it needs to be able to see `sbfocstest-lv1.cl.gemini.edu` and also needs to be able to `ssh` there without a password prompt. Otherwise it will fail and do nothing.

tpolecat: But *you* can say `deployTest` (or, more generally, `runMain gem.ctl.main <options>`) at the sbt prompt to deploy to the test machine (or any other machine that has docker set up).

PR #586, opened 2018-07-13
SEQNG-650 Refactor of configuration parameters

jluhrs:


jluhrs:
I found some issues related to the initialization of `SeqexecEngine`, that will be fixed in another task.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -352,26 +349,28 @@ object SeqexecEngine extends SeqexecConfiguration {
...
+                            gsaoiControl: ControlStrategy,
+                            gwsControl: ControlStrategy,
+                            nifsControl: ControlStrategy,
+                            niriControl: ControlStrategy,
+                            tcsControl: ControlStrategy,
+                            odbNotifications: Boolean,
+                            instForceError: Boolean,
+                            failAt: Int,
+                            odbQueuePollingInterval: Duration,
+	                    gpiGiapi: Giapi[IO],
```
cquiroz: I bet your hand itch to align these

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -106,4 +124,17 @@ package object server {
     }
   }
 
+  implicit class ControlStrategyOps(v: ControlStrategy) {
+    val connect: Boolean = v match {
+      case Simulated => false
+      case _         => true
+    }
+    // If connected, then use real values for keywords
+    val realKeywords: Boolean = connect
```
cquiroz: Wouldn't this forbid me to send things to the GDS without connecting to the instruments. Its maybe a fringe case but I've been doing that during development

jluhrs: No. `realKeywords` means the keyword values are real, i.e. they are read from the systems, wich implies that you connected to the system.
For GPI, I added a separate parameter for GDS. The meaning is the same: if `gpiGDS` is `full`, keywords will be sent to the real GDS

cquiroz: I see

PR #587, opened 2018-07-13
SEQNG-643: GHOST Model

cquiroz:
Read the GHOST model from the ocs

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/properties.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/properties.scala
@@ -4,7 +4,7 @@
 package seqexec.model
 
 import seqexec.model.Model.Instrument
-import seqexec.model.Model.Instrument.{F2, GNIRS, GmosN, GmosS, GPI}
+import seqexec.model.Model.Instrument._
 
 object properties {
   sealed trait InstrumentProperties
@@ -15,6 +15,7 @@ object properties {
     GmosS -> Set(Disperser),
     GmosN -> Set(Disperser),
     GNIRS -> Set(Disperser),
-    GPI   -> Set.empty
+    GPI   -> Set.empty,
+    GHOST -> Set.empty
   )
 }
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -20,6 +20,7 @@ import seqexec.server.InstrumentSystem._
 import seqexec.server.flamingos2.{Flamingos2, Flamingos2Controller, Flamingos2Header}
 import seqexec.server.keywords._
 import seqexec.server.gpi.{GPI, GPIController, GPIHeader}
+import seqexec.server.ghost.{GHOST, GHOSTController}
 import seqexec.server.gcal._
 import seqexec.server.gmos.{GmosController, GmosHeader, GmosNorth, GmosSouth}
 import seqexec.server.gws.{DummyGwsKeywordsReader, GwsHeader, GwsKeywordsReaderImpl}
@@ -32,6 +33,7 @@ import edu.gemini.spModel.gemini.altair.AltairConstants
 import edu.gemini.spModel.obscomp.InstConstants._
 import edu.gemini.spModel.seqcomp.SeqConfigNames._
 import org.log4s._
+import org.http4s.Uri._
 import squants.Time
 import fs2.Stream
 import gem.Observation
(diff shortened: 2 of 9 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -456,6 +459,7 @@ object SeqTranslate {
       case GmosNorth.name  => TrySeq(Model.Instrument.GmosN)
       case Gnirs.name      => TrySeq(Model.Instrument.GNIRS)
       case GPI.name        => TrySeq(Model.Instrument.GPI)
+//      case GHOST.name      => TrySeq(Model.Instrument.GHOST)
```
jluhrs: Why are these lines commented?

cquiroz:
This could be reviewed now. Please note this is really a proof of concept, there is still a lot to do, in particular I think IFUs should be split into another class and we need to implement configuration and the GIAPI client

But we can load a sequence:
seqexec - gs-2018b-q-0-1 2018-07-17 14-38-12

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
```diff
@@ -0,0 +1,151 @@
...
+  override val gdsClient: GDSClient = controller.gdsClient
+
+  override val resource: Resource = Instrument.GHOST
+
+  override val sfName: String = "GHOST"
+
+  override val contributorName: String = "ghost"
+
+  override val observeControl: InstrumentSystem.ObserveControl =
+    InstrumentSystem.Uncontrollable
```
swalker2m: Another teaching moment for me.  What is this?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
```diff
@@ -0,0 +1,151 @@
...
+//
+//  private def dmsAngle(config: Config, name: String): Either[ExtractFailure, Option[Angle]] =
+//    for {
+//      a <- config.extract(INSTRUMENT_KEY / name).as[String]
+//        .map(Angle.fromStringSignedDMS.getOption)
+//    } yield a
+//
+//
+//  private def targetName(config: Config, name: String): Either[ExtractFailure, Option[String]] =
+//    config.extract(INSTRUMENT_KEY / name).as[Option[String]]
```
swalker2m: Are you planning on coming back to all of the commented out stuff?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
```diff
@@ -0,0 +1,151 @@
...
+      config: Config): SeqActionF[F, GHOSTConfig] =
+    EitherT(
+      Sync[F].delay(
+        (for {
+          baseRAHMS          <- config
+                                 .extract(INSTRUMENT_KEY / Ghost.BaseRAHMS)
+                                 .as[Option[HourAngle]]
+          baseDecDMS         <- config
+                                 .extract(INSTRUMENT_KEY / Ghost.BaseDecDMS)
+                                 .as[Option[Angle]]
```
swalker2m: I suppose you could add an `extractAs[A](...)` and make these 1 liners.

modules/seqexec/model/src/main/scala/seqexec/model/properties.scala
```diff
@@ -15,6 +15,7 @@ object properties {
     GmosS -> Set(Disperser),
     GmosN -> Set(Disperser),
     GNIRS -> Set(Disperser),
-    GPI   -> Set.empty
+    GPI   -> Set.empty,
+    GHOST -> Set.empty
```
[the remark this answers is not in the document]
cquiroz: It is a UI side property, it tells what extra columns to display on the steps table

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
cquiroz: I suppose @sraaphorst will take those when the time comes

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
cquiroz: Right, that's a good idea, though we won't need these with the new model

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -342,6 +346,7 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
     case Model.Instrument.NIFS  => true
     case Model.Instrument.NIRI  => true
     case Model.Instrument.GPI   => true
+    case Model.Instrument.GHOST => false
```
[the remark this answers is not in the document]
cquiroz: @jluhrs How is this used in practice

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
swalker2m: Good point.  I only mention it because of the scala check method length test.

cquiroz: yes, we keep hitting that limit in these methods. I'd rather remove the 50 lines limit check in scalastyle

tpolecat: Yeah the boilerplate could be reduced here, and you could use `mapN` instead of a for-comprehension if you wanted to (not a big deal), I would rather keep the 50-line limit and disable it for cases where you have a good reason to have a long method. It's a good stylistic check.

cquiroz: Actually, it was very easy to add and looks a lot more compact. thanks for the idea

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
cquiroz: The naming is probably scary, it is used to indicate if you can interrupt observations

cquiroz:
The 50 lines limit is probably ok for the kind of basic model code we have but for more complex cases, where you necessarily need to e.g. read N channels, it just becomes an unreal requirement.

At the end all the code inside the long method gets unchecked on style or we break it up artificially

cquiroz:
and the 50 lines is arbitrary, why not 10 or 200?

tpolecat:
I think you can turn off just that one warning. But if it's really an issue we can change it. I'd like to switch to scalafix at some point anyway.

cquiroz:
yes, you can, and we do it for long methods.

I guess scalafix does more syntactic checks than style ones right?

tpolecat:
10 sounds good.

tpolecat:
Scalafix has full syntactic and semantic information (i.e., types) so it can do much better checks than scalastyle + wartremover.

cquiroz:
it is probably good to keep discussing, but I'll merge this for now

PR #588, opened 2018-07-13
SEQNG-658; Strange display of breakpoints

cquiroz:
Tiny css fix that's breaking the the UI with breakpoints for logged out users


cquiroz:
Tested in FF, Chrome, Safari and iOS

PR #589, opened 2018-07-17
SEQNG-653: Send GCAL and GWS keywords to either GDS or DHS as needed

cquiroz:
This PR extends GDS support to Gcal and gws keywords. This has been tested with GPI

As a part of this PR I was able to bring several simplifications to the way using either GDS or DHS is done. Thus `HeaderProvider` is gone ad Gcal doesn't take a dhs client anymore


modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPIHeader.scala
```diff
@@ -4,26 +4,26 @@
...
     new Header {
       override def sendBefore(obsId: Observation.Id,
                               id: ImageFileId): SeqAction[Unit] = {
-        val ks = bundleKeywords(
-          inst,
+        val ks = inst.keywordsClient.bundleKeywords(
           List(
             buildDouble(tcsKeywordsReader.getParallacticAngle
-                          .map(_.map(_.toDoubleDegrees))
+                          .map(_.map(_.toSignedDoubleDegrees))
```
cquiroz: This returns a double with many digits of precision, like -22.291460009476502. We are going to leave it this way as the same amount of digits is used for other values.

This is up for review though

modules/seqexec/server/src/main/scala/seqexec/server/keywords/StandardHeader.scala
```diff
@@ -102,11 +102,11 @@ class StandardHeader[A: HeaderProvider](
     buildString(obsReader.getObsType, "OBSTYPE"),
     buildString(obsReader.getObsClass, "OBSCLASS"),
     buildString(obsReader.getGemPrgId, "GEMPRGID"),
-    buildString(obsReader.getObsId, "obsid"),
+    buildString(obsReader.getObsId, "OBSID"),
     buildString(obsReader.getDataLabel, "DATALAB"),
     buildString(stateReader.getObserverName, "OBSERVER"),
     buildString(obsReader.getObservatory, "OBSERVAT"),
-    buildString(obsReader.getTelescope, "telescope"),
+    buildString(obsReader.getTelescope, "TELESCOP"),
```
cquiroz: `telescope` is not a valid keyword name. I'll do a refactoring of the keywords code to make this impossible

jluhrs: `telescope` is valid in the DHS.

cquiroz: Would it work regardless with the keyword in capitals? Let me review why GPI uses this:

`TELESCOP= 'Gemini-South'       / Name of telescope (Gemini-North|Gemini-South)`

modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPIHeader.scala
tpolecat: What is your concern re: very precise angles? Are you concerned that the TCS won't be able to parse them or something?

cquiroz: it is a "change" in the sense that FITS files prevously only had 3 decimals. I'm not sure if there is any effect downstream

cquiroz:
This now contains the changes for SEQNG-659 and SEQNG-653

jluhrs:
I tested the PR against the DHS. I run Object and GCAL observations for GMOS, GNIRS and FLAMINGOS-2. They worked, and the headers look ok.

PR #590, opened 2018-07-17
SEQNG-651 Fixed function smartSetDoubleParam.

jluhrs:
The reported bug was that Seqexec was not setting the GNIRS parameter `centralWavelength`. The cause was a bug in the function that compares the current value to the requested value.
I also change the comparison criteria to use a relative threshold. That makes more sense for floating point numbers.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
@@ -187,7 +187,7 @@ object GnirsControllerEpics extends GnirsController {
   private def setOtherCCParams(config: Other): List[SeqAction[Unit]] = {
     val open = "Open"
     val focus = "best focus"
-    val wavelengthTolerance = 0.0005
+    val wavelengthTolerance = 0.0001
     val refocusParams = setAcquisitionMirror(config.mode) ++
       smartSetParam(encode(config.filter1), epicsSys.filter1.map(removePartName), ccCmd.setFilter1(encode(config.filter1))) ++
       setFilter2(config.filter2, config.wavel) ++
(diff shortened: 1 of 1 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -187,7 +187,7 @@ object GnirsControllerEpics extends GnirsController {
   private def setOtherCCParams(config: Other): List[SeqAction[Unit]] = {
     val open = "Open"
     val focus = "best focus"
-    val wavelengthTolerance = 0.0005
+    val wavelengthTolerance = 0.0001
```
cquiroz: Where is this used?

tpolecat: If you're using a double and rounding then you might want to use `gem.math.Wavelength` which is exact.

jluhrs: @cquiroz It is used in only one place, in the call to `smartSetDoubleParam` at the end of the same function.

jluhrs: @tpolecat `wavelengthTolerance` is not a wavelength, it is a dimensionless quantity, the relative threshold to decide if the wavelength parameter needs to be applied.

swalker2m: There is a dimension there though, if I understand it correctly.  You compare one wavelength to another to see if it differs sufficiently to change?  If they were `Wavelength`s then we could state exactly how many angstroms different that they need to be.

jluhrs: @swalker2m It is relative, like in `(current - requested)/requested > threshold`. Therefore, it has no units.

PR #592, opened 2018-07-18
SEQNG-659: KeywordName enum

cquiroz:
I'm a bit unhappy that we have keyword names as String while there are restrictions on what strings can be (max 8 chars, etc.)

This PR changes that creating an enum for it and replacing the keyword usage for fixed values.

This also fixes the issue of the DHS accepting non conforming keywords

I think this is a change big enough to warrant some experimentation with the instruments to ensure I haven't broken something


modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
```diff
@@ -28,6 +29,12 @@ import cats.implicits._
 class DhsClientHttp(val baseURI: String) extends DhsClient {
   import DhsClientHttp._
 
+  private val KeywordsLUT: Map[KeywordName, String] = Map(
```
cquiroz: This will adapt a few keyword names that must be sent especially formatted to the DHS

modules/core/shared/src/test/scala/gem/arb/ArbKeywordName.scala
```diff
@@ -0,0 +1,21 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+package arb
+
+import gem.enum.KeywordName
+import org.scalacheck._
+import org.scalacheck.Cogen._
+
+trait ArbKeywordName {
```
tpolecat: This is unnecessary; you get these for free via `ArbEnumerated`.

modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
tpolecat: An option here would be to add a `dhsName` property to `KeywordName` and push this mapping down into the database. No strong opinion.

tpolecat:
👍

tpolecat:
It's generated on demand. We check them in because they rarely change and it's nice to have the source for reference.

modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
swalker2m: It could be nullable I suppose so you'd only need to add the exceptions.

cquiroz: it should be nullable in that case, only 3 of the nearly 1K keywords need it

swalker2m:
> > If I understand correctly, KeywordsName.scala is generated from the DB during compilation. Why is then included in the PR?

> It's generated on demand. We check them in because they rarely change and it's nice to have the source for reference.

It seemed a little odd at first when we started checking them in, but in practice it has been much nicer that way.   Before I was always regenerating enums and wondering if I had the latest version.  Now I never have to think about it unless I add a new enum or update an existing one. To summarize some of the points that have been raised elsewhere

* The code is available immediately when you check out the project so you don't need to set up a database to do the initial compilation.

* You can open the project in an editor and have all the source code available from the beginning.

* When you update an enum (via a SQL migration), the changes to the generated enum appear in the PR making it easier to see the impact.

modules/sql/src/main/resources/db/migration/V056__Fits_KeywordNames.sql
```diff
@@ -0,0 +1,663 @@
...
+
+CREATE TABLE e_fits_keyword_names (
+    id   identifier       PRIMARY KEY,
+    name VARCHAR(8)       NOT NULL CHECK (name  ~     '[A-Z0-9_-]{1,8}')
+);
+
+ALTER TABLE e_fits_keyword_names OWNER TO postgres;
+
+COPY e_fits_keyword_names (id, name) FROM stdin;
+INSTRUMENT	INSTRUME
```
jluhrs: Is it possible to have to fill the second column only when the keyword is different from the id ?

cquiroz: I think you’d need to add a NULL at the end of every other row

jluhrs:
It still feels weird to me, because it breaks one of the golden rules I learnt when I was just starting working in software. Of course, on those times the rule to not archive generated files applied to object files and executables. But I recognizes this is just my personal dogmatic position, and I can let it pass.
The one related issue that really bothers me is the reliance on human action to keep the files synchronized. Can this be automatized? I don't trust humans, specially myself. I know that if I make a change, I will forget to run the command to update the scala file, and I will waste half a day trying to figure out why it doesn't work.
My proposal is this: to modify `seqexec_server` project in build.sbt, and add a task to its `sourceGenerators` that would run the proper command on compile, to generate  KeywordsName.scala when needed.

jluhrs:
Ok, so generating KeywordsName.scala is not that simple as I thought. Given that, I'll settle for the suggestion of adding a build check, to be done in another Jira task.

modules/sql/src/main/resources/db/migration/V056__Fits_KeywordNames.sql
```diff
@@ -0,0 +1,663 @@
...
+SFRT2	SFRT2
+SFTILT	SFTILT
+SFLINEAR	SFLINEAR
+AOFOLD	AOFOLD
+PWFS1_ST	PWFS1_ST
+PWFS2_ST	PWFS2_ST
+OIWFS_ST	OIWFS_ST
+AOWFS_ST	AOWFS_ST
+SCIBAND	SCIBAND
+NUMREQTW	NUMREQT
```
jluhrs: Missing a W, it should be NUMREQTW

cquiroz: Great find

cquiroz:
Closing in favor of #589

PR #595, opened 2018-07-19
Fail the build if genEnums produces unchecked code

cquiroz:
Don't merge

### DIFF
--- a/build/buildkite-build.sh
+++ b/build/buildkite-build.sh
@@ -70,11 +70,13 @@ echo "--- :scala: Running tests"
   test
 
 # Check git status. if genEnums generated something mismatching, this should fail
-echo "--- :git: check code generation is current"
-if [ -n "$(git status --untracked-files=no --porcelain)" ]; then
-  # Uncommitted changes in tracked files
-  echo "--- :shrug: Enum generation produced unmatching files"
-  exit 1
+if [ "$BUILDKITE" = "true" ]; then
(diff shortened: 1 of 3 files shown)


tpolecat:
Slight confusion … the check needs to happen after you do `genEnums` right?

build/buildkite-build.sh
```diff
@@ -59,15 +59,24 @@ do
...
   -Docs3.databaseUrl=jdbc:postgresql://$HOST_AND_PORT/gem \
   sql/flywayMigrate                                       \
+  genEnums                                                \
   test
 
+# Check git status. if genEnums generated something mismatching, this should fail
+echo "--- :git: check code generation is current"
+if [ -n "$(git status --untracked-files=no --porcelain)" ]; then
+  # Uncommitted changes in tracked files
+  echo "--- :shrug: Enum generation produced unmatching files"
```
swalker2m: I'm not sure why we need this, but I guess it doesn't hurt anything.  Won't it skip new enums though since they will initially be untracked.

Is it necessarily only changes produced by `genEnums`?  It looks like you can't continue with the build until you commit things you've modified in general.

cquiroz: right, we could remove the `untracked` check

tpolecat: `if [ "$BUILDKITE" = "true" ]; then` ...

PR #596, opened 2018-07-20
Upgrade Deployment Pipeline

tpolecat:
This adds an upgrade pipeline to deploy a new version (which must have been built previously from `develop` on the main pipeline) and migrates data over. This is a manual process. You click "new build" in the BuildKite web UI and then it prompts for a few details before doing its thing.

image

This PR also adds the other side of the "write github history to a file" PR from yesterday, which is necessary for upgrades. It reads the history from the version you're trying to deploy, to verify that it understands the [older] schema used by the currently deployed version.

❗️This really really needs to be squashed. Lots of experiments here.


build/upgrade/step-2.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# This dynamically generates a pipeline containing a click to unblock step with
+# custom fields.
```
tpolecat: Oops, this comment is from an earlier copy/paste. I'll remove it.

cquiroz:
That looks very neat. Some question:
Can the version number be just left empty and using the one generated on `build.sbt`?
Will it run a full build (tests, etc.) and then deploy?

tpolecat:
> Can the version number be just left empty and using the one generated on build.sbt?

Conceptually all this task does is deploy existing images … it uses the latest version of the `ctl` project to do this, which could really be its own repo. So it happens to know the current version on `develop` `HEAD` but I'm hesitant to use it. It also seems a little dangerous to me … for production deployments I think we probably want to be more deliberate and select a specific version. Also I think we may end up wanting a staging environment for pre-deploy validation, which will lag behind the test environment. Anyway it's not really clear what the workflow should be so I'm trying to keep it simple for now.

> Will it run a full build (tests, etc.) and then deploy?

No, this doesn't build at all. It deploys a version that has already been built, verified, and imaged by the main build pipeline.

PR #597, opened 2018-07-20
SEQNG-662: Reduce semantic ui css

cquiroz:
Semantic UI Css is fairly big but it can be reduced by selecting a subset of the widgets

This PR does that reducing the css size from 545Kb to 408Kb


modules/seqexec/web/client/src/main/resources/theme/semantic.less
```diff
@@ -0,0 +1,156 @@
...
+& {
+  @import "~semantic-ui-less/definitions/elements/button";
+}
+& {
+  @import "~semantic-ui-less/definitions/elements/container";
+}
+& {
+  @import "~semantic-ui-less/definitions/elements/divider";
+}
+// & {
```
cquiroz: Commented widgets aren't used but i'll keep them here in case we need them later on

PR #598, opened 2018-07-20
SEQNG-664: Parallelize keyword collection

cquiroz:
Looking at the code I can tell we can run some of the keyword related tasks in parallel. After a serious fight with cats implicits I managed to get it working

I tested these against GPI without any issues. I hope the DHS can also accept this

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecFailure.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecFailure.scala
@@ -3,12 +3,14 @@
 
 package seqexec.server
 
+import cats.Semigroup
 import edu.gemini.seqexec.odb.SeqFailure
 import org.http4s.Uri
 
 sealed trait SeqexecFailure
 
 object SeqexecFailure {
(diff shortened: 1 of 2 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecFailure.scala
```diff
@@ -44,6 +46,8 @@ object SeqexecFailure {
   /** XMLRPC error while communicating with the GDS */
   final case class GDSXMLError(msg: String, url: Uri) extends SeqexecFailure
 
+  final case class SeqexecCombinedFailure(a: SeqexecFailure, b: SeqexecFailure) extends SeqexecFailure
```
tpolecat: Another solution in cases like this is to use `NonEmptyList[SeqexecFailure]` which gives you a free semigroup, so you don't need the reified constructor for `|+|`. Just FYI. But this is fine.

jluhrs:
Unfortunately, Science requested to preserve the order of the keywords. DHS puts the keywords in the order it receives them. That means we cannot send the different groups in parallel :(

jluhrs:
We could collect the keywords in parallel though.

tpolecat:
> Science requested to preserve the order of the keywords

Weird. Any idea why?

cquiroz:
Would it make sense to make it "backend" optional?

PR #601, opened 2018-07-23
dependency updates

tpolecat:
This updates a few things

- cats 1.2.0
- flyway 5.1.4
- decline 0.4.2
- wartremover 2.3.1

### DIFF
--- a/modules/core/shared/src/test/scala/gem/util/LocationSpec.scala
+++ b/modules/core/shared/src/test/scala/gem/util/LocationSpec.scala
@@ -9,7 +9,7 @@ import cats.kernel.laws.discipline._
 import cats.kernel.Comparison.{ GreaterThan => GT, LessThan => LT, EqualTo => EQ }
 import cats.tests.CatsSuite
 
-@SuppressWarnings(Array("org.wartremover.warts.Equals", "org.wartremover.warts.NonUnitStatements"))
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/test/scala/gem/util/LocationSpec.scala
```diff
@@ -9,7 +9,7 @@ import cats.kernel.laws.discipline._
 import cats.kernel.Comparison.{ GreaterThan => GT, LessThan => LT, EqualTo => EQ }
 import cats.tests.CatsSuite
 
-@SuppressWarnings(Array("org.wartremover.warts.Equals", "org.wartremover.warts.NonUnitStatements"))
+@SuppressWarnings(Array("org.wartremover.warts.Equals", "org.wartremover.warts.NonUnitStatements", "org.wartremover.warts.TraversableOps"))
```
tpolecat: This test uses `.min` and `.max` which are unsafe in general but ok here so I just disabled the [new] warning.

cquiroz:
in a related note. There are some eviction warnings about pre 1.0 cats usage in `knobs`. Since it seems those verizon libs aren't updated anymore. Should we keep a fork?

tpolecat:
@cquiroz are you doing anything that depends on knob specifically? I think PureConfig is kind of the dominant config library at this point.

cquiroz:
We started using it at some moment a while time ago and it has just stuck. It does the job just fibe so I don’t see the need to redo it with another liv unless there is a very good reason

tpolecat:
I think knobs being abandoned is probably a good enough reason. I don't want to maintain a fork.

PR #602, opened 2018-07-24
SEQNG-661: Resizable config table and preserve state

cquiroz:
This is a fairly large PR which makes the basis to make all tables columns-resizable and to remove a family of quirks related to state being lost when the core model changes. The changes are fairly big so I'm testing them on the step config table before tackling the Steps table where the bugs are more evident

This PR contains:
* Made the Step Config table columns resizable
* Factor out some common code for tables with resizable columns
* Preserve the resize and scroll position across model changes. This can be observed on the image below where the column sizes and scroll position is kept when the step changes:

screencast 2018-07-24 10-38-30

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
@@ -32,7 +32,10 @@ object StepConfigTable {
 
   type Backend = RenderScope[Props, TableState[TableColumn], Unit]
 
-  final case class Props(step: Step, size: Size, startState: TableState[TableColumn]) {
+  final case class Props(step: Step,
+                         size: Size,
+                         startState: TableState[TableColumn]) {
+
     val settingsList: List[(SystemName, String, String)] =
       step.config.toList.flatMap {
         case (s, c) =>
@@ -44,7 +47,9 @@ object StepConfigTable {
     val rowCount: Int = settingsList.size
 
     def rowGetter(idx: Int): SettingsRow =
-      settingsList.lift(idx).fold(SettingsRow.Zero)(Function.tupled(SettingsRow.apply))
+      settingsList
+        .lift(idx)
+        .fold(SettingsRow.Zero)(Function.tupled(SettingsRow.apply))
   }
 
   // ScalaJS defined trait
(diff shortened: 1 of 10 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
```diff
@@ -25,7 +47,9 @@ object StepConfigTable {
     val rowCount: Int = settingsList.size
 
     def rowGetter(idx: Int): SettingsRow =
-      settingsList.lift(idx).fold(SettingsRow.Zero)(Function.tupled(SettingsRow.apply))
+      settingsList
+        .lift(idx)
+        .fold(SettingsRow.Zero)(Function.tupled(SettingsRow.apply))
```
tpolecat: `SettingsRow` is mutable … does it really make sense to have a global mutable `Zero` value?

modules/shared/web/client/src/main/scala/web/client/table/package.scala
```diff
@@ -0,0 +1,146 @@
...
+package table {
+
+  sealed trait UserModified extends Product with Serializable
+  case object IsModified  extends UserModified
+  case object NotModified extends UserModified
+
+  object UserModified {
+    implicit val eq: Eq[UserModified] = Eq.fromUniversalEquals
+
+    def fromBool(b: Boolean): UserModified = b.fold(IsModified, NotModified)
```
tpolecat: I prefer `if (b) IsModified else NotModified` because it's so easy to swap the arguments to `fold`.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
cquiroz: This is a good question. `SettingsRow` has vars inside as it sits at the border with `js`. it can't be a `val` an interoperate properly. However, the way we use it is effectively immutable.
Yes the compiler cannot prevent in this case somebody changing `Zero`. I suppose it could be constructed at the call site if you think that's better

tpolecat: Oh weird, everything has to be a var? Crazy. I think I would feel better if there were a factory method that returned a new empty instance.

cquiroz: you can make a `js trait` with `vals` but it could be useless as you cannot have a constructor nor create new instances as `new Trait {`.

I'll change the Zero though

PR #605, opened 2018-07-25
SEQNG-633A: Better error handling reading status items

cquiroz:
Trying to make the giapi connection more robust this PR adds more checks and a method to optionally get a value

More important a realistic test is built that will create a mini gmp to test e.g. that timeouts are being accepted

### DIFF
--- a/modules/giapi/src/main/scala/giapi/client/package.scala
+++ b/modules/giapi/src/main/scala/giapi/client/package.scala
@@ -36,6 +36,8 @@ package client {
 
   import giapi.client.commands.CommandResult
 
+  final case class GiapiException(str: String) extends RuntimeException(str)
+
   /**
     * Typeclass to present as evidence when calling `Giapi.get`
     */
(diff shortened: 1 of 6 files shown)


modules/giapi/src/main/scala/giapi/client/package.scala
```diff
@@ -156,18 +163,29 @@ package client {
       * @param url Url to connect to
       * @tparam F Effect type
       */
+    // scalastyle:off
```
cquiroz: This is a case where the IMHO the line limit check in scala style is broken. The method includes the creation of a trait instance which is long but composed of small methods

modules/giapi/src/main/scala/giapi/client/package.scala
```diff
@@ -156,18 +163,29 @@ package client {
...
+            @SuppressWarnings(Array("org.wartremover.warts.Throw"))
             override def get[A: ItemGetter](statusItem: String): F[A] =
               Sync[F].delay {
                 val item = sg.getStatusItem[A](statusItem)
-                // Note item.getValue can throw if e.g. the item is unknown
-                item.getValue
+                // Note item.getValue can throw if e.g. the item is unknown, let's avoid NPEs though
+                Option(item) match {
+                  case Some(i) => i.getValue
+                  case None    => throw GiapiException(s"Status item $statusItem not found")
```
jluhrs: Is an exception really necessary here?

cquiroz: we discussed this over Stride and that was the conclusion. If no item is found you'll get an NPE which is maybe worse

cquiroz: Sorry I merged but it is a good idea. I'll add it next

PR #606, opened 2018-07-26
Observation import and fetch service API

swalker2m:
I'm reopening this PR in the new buildkite world now that the JSON codec for observations is working again. It provides a minimal API for fetching observations from the database, looking ahead to converting the seqexec.

I believe this signature in the `Service` API:

```scala
def queryObservationById(id: Observation.Id): M[Observation]
```

should probably be changed to handle the presumably common case in which the indicated observation doesn't exist.  I wanted to get feedback first though since we should have a common way of handling this and we may want to update the DAO API accordingly.  At the moment, everything assumes the named observation exists.

### DIFF
--- a/modules/service/src/main/scala/gem/Service.scala
+++ b/modules/service/src/main/scala/gem/Service.scala
@@ -95,7 +95,7 @@ final class Service[M[_]: Sync: LiftIO] private (private val xa: Transactor[M], 
      * existing program with the same id.
      */
     def importProgram(host: String, pid: Program.Id): M[Either[String, Unit]] =
-      log.log(user, s"ocs2.importProgram($host, $pid") {
+      log.log(user, s"ocs2.importProgram($host, $pid)") {
         OdbClient.fetchProgram[IO](host, pid).liftIO[M].flatMap { _.traverse { case (p, ds) =>
           Importer.importProgram(p, ds).transact(xa)
         }}
--- a/modules/service/src/main/scala/gem/Service.scala
+++ b/modules/service/src/main/scala/gem/Service.scala
@@ -28,6 +28,14 @@ final class Service[M[_]: Sync: LiftIO] private (private val xa: Transactor[M], 
     }
 
   /**
+   * Constuct a program that yields the `Observation` with the matching observation id.
+   */
+  def queryObservationById(id: Observation.Id): M[Observation] =
+    log.log(user, s"queryObservationById(${id.format})") {
+      ObservationDao.select(id).transact(xa)
+    }
+
+  /**
    * Construct a program that attempts to change the user's password, yielding `true` on success.
    */
   def changePassword(oldPassword: String, newPassword: String): M[Boolean] =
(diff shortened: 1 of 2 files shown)


modules/web/src/main/scala/Application.scala
```diff
@@ -38,6 +42,21 @@ object Application {
         val limit   = n.getOrElse(100)
         gs.queryProgramsByName(pattern, limit).flatMap(ps => Ok(ps.map(p => (p._1, p._2)).asJson ))
 
+      // Import an observation from the legacy ODB.
+      case GET -> Root / "api" / "import" / "obs" / o :? Host(h) as gs =>
```
cquiroz: i wonder if these endpoints should be tested

tpolecat: Yeah it would be good to add tests … maybe in another pr?

tpolecat: I also think we need to look into the Rho DSL for defining endpoints because it generates swagger doc. I'm concerned that we don't have a good way to document the API.

cquiroz: GraphQL maybe a better approach and it support generated documentation by default

tpolecat:
@swalker2m I wonder if we need a convention like `fetchFoo(id): M[Foo]` vs. `queryFoo(id): M[Option[Foo]]` or something like that. I don't know what the common case is, but it seems to me that the only way to get an id is from a valid instance, and the case of deleting things is relatively infrequent so we may want both. That is, it might be equally reasonable to treat a bogus id as an error, or as a reasonable possibility, depending on the situation.

cquiroz:
It is kind of irrelevant but in many cases you see `F[_]` rather than `M[_]`. ANy reason to prefer `M`?

tpolecat:
@cquiroz I default to `F[_]` and only use `M[_]` if I know it's a monad. This is a weird convention and we should probably just use `F[_]` everywhere.

modules/service/src/main/scala/gem/Service.scala
```diff
@@ -67,23 +77,25 @@ final class Service[M[_]: Sync: LiftIO] private (private val xa: Transactor[M],
...
-      * existing observation with the same id.
-      */
+    /**
+     * Constructs a program that fetches the corresponding observation from the
+     * indicated ODB and then stores it in the database, replacing any
+     * existing observation with the same id.
+     */
     def importObservation(host: String, oid: Observation.Id): M[Either[String, Unit]] =
       log.log(user, s"ocs2.importObservation($host, $oid)") {
         OdbClient.fetchObservation[IO](host, oid).liftIO[M].flatMap { _.traverse { case (o, ds) =>
```
tpolecat: Seems like there should be a combinator for `.flatMap(_.traverse` but there's not.

swalker2m:
Thanks, I'll add this for now and then look into making a fetch vs. query distinction and changing the way endpoints are defined.

PR #608, opened 2018-07-27
Code simplifications

cquiroz:
Some minor simplifications:

* Use `extractAs[A]` instead of `extract().as[A]`
* Remove some parameters on the table columns that default correctly upstream

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPI.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPI.scala
@@ -10,6 +10,7 @@ import cats.data.Reader
 import edu.gemini.spModel.config2.Config
 import edu.gemini.spModel.gemini.gpi.Gpi._
 import edu.gemini.spModel.seqcomp.SeqConfigNames._
+import java.lang.{Boolean => JBoolean, Double => JDouble, Integer => JInt}
 import seqexec.model.dhs.ImageFileId
 import seqexec.model.Model.{Instrument, Resource}
 import seqexec.server.ConfigUtilOps._
@@ -55,8 +56,7 @@ final case class GPI[F[_]: Sync](controller: GPIController[F])
 
   override def calcObserveTime(config: Config): Time =
     config
-      .extract(OBSERVE_KEY / EXPOSURE_TIME_PROP)
-      .as[java.lang.Double]
+      .extractAs[JDouble](OBSERVE_KEY / EXPOSURE_TIME_PROP)
       .map(x => Seconds(x.toDouble))
       .getOrElse(Seconds(360))
 }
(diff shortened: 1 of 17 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -117,11 +117,11 @@ object Tcs {
 
   @SuppressWarnings(Array("org.wartremover.warts.Overloading"))
   def build[T, P: ClassTag](f: P => Endo[T], k: ItemKey, config: Config): Endo[T] =
-    config.extract(k).as[P].map(f).foldK
+    config.extractAs[P](k).map(f).foldK
 
   @SuppressWarnings(Array("org.wartremover.warts.Overloading"))
   def build[T, P: ClassTag, Q: ClassTag](f: (P, Q) => Endo[T], k1: ItemKey, k2: ItemKey, config: Config): Endo[T] =
-    config.extract(k1).as[P].product(config.extract(k2).as[Q]).map(f.tupled).foldK
+    config.extractAs[P](k1).product(config.extractAs[Q](k2)).map(f.tupled).foldK
```
tpolecat: This should be the same as `.mapN(f)`

```scala
(config.extractAs[P](k1), config.extractAs[Q](k2)).mapN(f).foldK
```

modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPI.scala
```diff
@@ -66,100 +66,81 @@ object GPI {
...
   private def gpiAoFlags(config: Config): Either[ExtractFailure, AOFlags] =
     for {
-      useAo      <- config.extract(INSTRUMENT_KEY / USE_AO_PROP).as[java.lang.Boolean]
-      useCal     <- config.extract(INSTRUMENT_KEY / USE_CAL_PROP).as[java.lang.Boolean]
+      useAo      <- config.extractAs[JBoolean](INSTRUMENT_KEY / USE_AO_PROP)
+      useCal     <- config.extractAs[JBoolean](INSTRUMENT_KEY / USE_CAL_PROP)
       aoOptimize <- config
-                      .extract(INSTRUMENT_KEY / AO_OPTIMIZE_PROP)
-                      .as[java.lang.Boolean]
+                      .extractAs[JBoolean](INSTRUMENT_KEY / AO_OPTIMIZE_PROP)
```
tpolecat: Not a big deal but you could remove the leftover newlines in this method and in a few other places.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
cquiroz: nice

modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPI.scala
cquiroz: ok, I've been running scalafmt to ensure the width is under 80, I'll check that again

PR #609, opened 2018-07-27
SEQNG-633 Proper error treatment of unhandled commands

cquiroz:
The giapi client isn't properly handling command errors. This PR fixes that and also removes some unused code

### DIFF
--- a/modules/giapi/src/main/scala/giapi/client/commands.scala
+++ b/modules/giapi/src/main/scala/giapi/client/commands.scala
@@ -18,10 +18,8 @@ import scala.collection.JavaConverters._
 import scala.concurrent.duration.Duration
 
 package commands {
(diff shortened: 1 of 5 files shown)


modules/giapi/src/main/scala/giapi/client/commands.scala
```diff
@@ -86,23 +76,23 @@ package object commands {
         command.toGiapi,
         (hr: HandlerResponse, _: GiapiCommand) => {
           if (hr.getResponse === Response.ERROR || hr.getResponse === Response.NOANSWER) {
-            cb(Right(Error(hr.getResponse, hr.getMessage)))
+            cb(Left(CommandResultException(hr.getResponse, hr.getMessage)))
```
cquiroz: Big mistake!

PR #611, opened 2018-07-30
SEQNG-668 Change the background of executed rows

cquiroz:
It was requested to make the executed rows more visible:

Before
seqexec - gs-2018b-q-0-2 2018-07-30 18-14-48

After
seqexec - gs-2018b-q-0-2 2018-07-30 17-30-41

The changes may look extensive but i'm mostly cleaning up variables usage on the css/less

### DIFF
--- a/modules/seqexec/web/client/src/main/resources/less/style.less
+++ b/modules/seqexec/web/client/src/main/resources/less/style.less
@@ -2,19 +2,8 @@
 @import "~semantic-ui-less/themes/default/collections/table.variables";
 @import "~semantic-ui-less/themes/default/elements/segment.variables";
 
-@color_2: #573a08;
-@color_3: black;
-@text_color: rgba(0, 0, 0, 0.95);
-@color_5: #9f3a38;
-@color_7: rgba(0, 0, 0, 0.87);
-@color_8: #2c662d;
-@background_color_3: lightcyan;
-@background_color_4: #fffaf3;
 @background_color_5: #f5f5f5;
 @background_color_6: #f9fafb;
-@background_color_7: rgba(0, 0, 0, 0.05);
-@background_color_8: #fff6f6;
-@background_color_10: #fcfff5;
 @table_row_left_padding: 7px;
 @breakpoint_line_color: #a5673f;
 @gutter_end_color: rgba(
@@ -170,9 +159,9 @@ body {
 }
 
 .queueText() {
-  color: @text_color;
+  color: @selectedTextColor;
   &:hover {
-    color: @text_color;
+    color: @selectedTextColor;
   }
 }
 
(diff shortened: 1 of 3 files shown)


andrewwstephens:
@tpolecat - the feedback was that the font color was too light, making it difficult to read the file number (which the observer needs to check the data quality or run gacq).  I'm fine with the gray background or just making the font color a bit darker.

tpolecat:
I vote for making it darker gray, although that might make it look too much like black. Worth a try, seems like.

cquiroz:
Some alternatives:

On green:
seqexec - gs-2018b-q-0-2 2018-07-30 17-34-03

A bit darker gray;
seqexec - gs-2018b-q-0-2 2018-07-30 23-12-17

darker background
seqexec - gs-2018b-q-0-2 2018-07-30 23-14-01

italics:
seqexec - gs-2018b-q-0-2 2018-07-30 23-17-30

andrewwstephens:
I like "A bit darker gray"...

cquiroz:
This is how it would look at the end
seqexec - gs-2018b-q-0-2 2018-07-31 13-00-05

modules/seqexec/web/client/src/main/resources/less/style.less
```diff
@@ -383,52 +372,52 @@ body {
   .borderTop();
   .borderRight();
   background-color: @white;
```
sraaphorst: I'm curious why you use `@white` here (which I assume is being substituted in) but `lightcyan` (hard-coded?) on L364.

sraaphorst:
If these are successfully completed rows, I think green looks the best, since green has an association with success, whereas grey has an association with disabled.

modules/seqexec/web/client/src/main/resources/less/style.less
cquiroz: `@white` is a less variable and it is relevant if we ever support theming. A dark theme may define `@white` as `black` though that may sound strange.

There is no variable for `lightcyan`, so that woudn't be themable

cquiroz:
I agree that I prefer green

swalker2m:
FWIW, I think I'm on team "a bit darker grey".  They are disabled in a way considering that you cannot modify anything about an executed step.  The upcoming steps may be changed though.

modules/seqexec/web/client/src/main/resources/less/style.less
sraaphorst: Theming would be great, but wouldn't it be detrimental, then, to use any colours that are not less variables and thus not themable?

cquiroz: it would, theming would have two parts, theming at the semantic-ui level where those variables are used. We should also need to theme our own css and there we may need to replace `lightcyan` for a variable too

jluhrs:
I also prefer the green option. With grey we risk that what somebody considers "dark enough to be legible" will be "too dark to be distinguishable" for someone else.

cquiroz:
We agreed on dark grey at the end. Can somebody approve this PR?

PR #612, opened 2018-08-01
SEQNG-666: Make log table columns resizable

cquiroz:
This PR makes the log table columns resizable. I'm also expanding code reuse across all the tables;

Here is how it looks like:
screencast 2018-08-01 11-33-40

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/LogArea.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/LogArea.scala
@@ -3,12 +3,15 @@
 
 package seqexec.web.client.components
 
+import cats.Eq
+import cats.data.NonEmptyList
 import cats.implicits._
 import mouse.all._
 import diode.react.ModelProxy
 import gem.enum.Site
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.component.Scala.Unmounted
+import japgolly.scalajs.react.component.builder.Lifecycle.RenderScope
 import japgolly.scalajs.react.CatsReact._
 import japgolly.scalajs.react.vdom.html_<^._
 import java.time.{Instant, LocalDateTime}
@@ -27,6 +30,7 @@ import seqexec.web.client.model.{GlobalLog, SectionOpen}
 import seqexec.web.client.actions.ToggleLogArea
 import seqexec.web.common.FixedLengthBuffer
 import web.client.style._
+import web.client.table._
 
 /**
   * Area to display a sequence's log
@@ -47,6 +51,17 @@ object CopyLogToClipboard {
   * Area to display a sequence's log
   */
 object LogArea {
+  type Backend = RenderScope[Props, State, Unit]
+
+  sealed trait TableColumn
+  case object TimestampColumn  extends TableColumn
+  case object LevelColumn extends TableColumn
+  case object MsgColumn extends TableColumn
+  case object ClipboardColumn extends TableColumn
+
+  object TableColumn {
+    implicit val eq: Eq[TableColumn] = Eq.fromUniversalEquals
+  }
 
   // Date time formatter
   private val formatter  = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SS")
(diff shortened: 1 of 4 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/LogArea.scala
```diff
@@ -82,66 +109,155 @@ object LogArea {
...
-      }.getOrElse(LogRow.Zero)
+    def rowGetter(s: State)(i: Int): LogRow =
+      reverseLog
+        .filter_(levelFilter(s) _)
+        .lift(i)
+        .map { l =>
+          val localTime = LocalDateTime.ofInstant(l.timestamp, site.timezone)
+          LogRow(formatter.format(localTime), l.timestamp, l.level, l.msg)
+        }
+        .getOrElse(LogRow.Zero)
```
tpolecat: If `LogRow` is a monoid you can use `foldMap` here. Otherwise I think `fold(z)(f)` expresses intent a little better than `.map(f).getOrElse(z)`.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/LogArea.scala
```diff
@@ -82,66 +109,155 @@ object LogArea {
...
+
+    val ZeroTableState: TableState[TableColumn] = TableState[TableColumn](
+      userModified   = NotModified,
+      scrollPosition = 0,
+      columns        = NonEmptyList.of(TimestampColumnMeta,
+                                       LevelColumnMeta,
+                                       MsgColumnMeta,
+                                       ClipboardColumnMeta))
+
+    val Zero: State =
```
tpolecat: When I see a member called `Zero` I expect it to be a zero with respect to some operator (i.e., a monoidal zero) so if it's not a monoid maybe something like `Initial` or `Default` would express the meaning more clearly.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/QueueTable.scala
```diff
@@ -219,8 +219,8 @@ object QueueTableBody {
         // Normalize based on visibility
         State.columns.modify(_.map {
           case c @ ColumnMeta(t, _, _, true, PercentageColumnWidth(_)) =>
-            c.copy(
-              width = PercentageColumnWidth(optimalSizes.getOrElse(t, 0).toDouble / width))
+            PercentageColumnWidth.fromDouble(optimalSizes.getOrElse(t, 0).toDouble / width)
+              .map(w => c.copy(width = w)).getOrElse(c)
```
tpolecat: Same comment as above; I think some kind of fold is usually preferable to `.map(f).getOrElse(z)`.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepConfigTable.scala
```diff
@@ -79,54 +78,47 @@ object StepConfigTable {
...
-          UpdateStepsConfigTableState(s))
-      }
+    userModified   = NotModified,
+    scrollPosition = 0,
+    columns        = NonEmptyList.of(TableColumnMeta, ValueColumnMeta))
+
+  private def colBuilder(b: Backend): ColumnRenderArgs[TableColumn] => Table.ColumnArg = tb => {
+    val state = b.state
+    def updateState(s: TableState[TableColumn]): Callback =
+      b.setState(s) >> SeqexecCircuit.dispatchCB(UpdateStepsConfigTableState(s))
```
tpolecat: If you're using `>>` because you need the RHS to be by-name then it's probably worth a comment. Otherwise prefer `*>`.

modules/shared/web/client/src/main/scala/web/client/table/package.scala
```diff
@@ -74,13 +88,16 @@ package table {
       val result = columns.toList.zipWithIndex.map {
         case (c @ ColumnMeta(_, _, _, true, PercentageColumnWidth(x)), idx) if idx === indexOf     =>
           val nv = x + delta
-          (nv >= 0 && nv <= 1).fold(c.copy(width = PercentageColumnWidth(nv)), c)
+          (nv >= 0 && nv <= 1)
+            .fold(c.copy(width = PercentageColumnWidth(nv)), c)
```
tpolecat: Not a fan of boolean fold because it still typechecks if you write it backwards. I think we have argued about this before ;-)

cquiroz: yeah, maybe i'm carried out by scala.js style that forces you to use it. I mean if you want to submit a PR to scala.js

tpolecat: Oh scala.js uses it? Huh.

cquiroz: I think I'll agree on this, especially as `Option.fold` works the other way around

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/LogArea.scala
cquiroz: `LogRow` doesn't make sense as a `Monoid` but will do the `fold`

PR #613, opened 2018-08-01
SEQNG-633: Add timeout for configuring

cquiroz:
This is a fairly important change and should not be merged until tested with an instrument.

When a giapi command is sent it can reply that something is excuted right away or that it will complete later. At the moment we assume that it will eventually complete but that may not be always true due to instrument code bugs or other unforseen events.

This PR changes the api so each command will take a timeout and fail if the timeout isn't fullfilled

This led to using some advanced `cats-effect` functions which took me a while to grasp.

The PR also improves the testing situation and it even test the particular case of a command never completing

### DIFF
--- a/modules/giapi/src/main/scala/giapi/client/gpi/GPIClient.scala
+++ b/modules/giapi/src/main/scala/giapi/client/gpi/GPIClient.scala
@@ -17,7 +17,7 @@ import scala.concurrent.duration._
   */
 class GPIClient[F[_]](giapi: Giapi[F]) {
   // GPI documentation specify 60 seconds as the max time to move muchanism
-  val defaultCommandTimeout: FiniteDuration = 60.seconds
+  val DefaultCommandTimeout: FiniteDuration = 60.seconds
 
   ///////////////
   // Status items
@@ -43,45 +43,45 @@ class GPIClient[F[_]](giapi: Giapi[F]) {
   def test: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.TEST, Activity.PRESET_START, Configuration.Zero),
-      defaultCommandTimeout)
+      DefaultCommandTimeout)
 
   def init: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.INIT, Activity.PRESET_START, Configuration.Zero),
-      defaultCommandTimeout)
+      DefaultCommandTimeout)
 
   def datum: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.DATUM, Activity.PRESET_START, Configuration.Zero),
-      defaultCommandTimeout)
+      DefaultCommandTimeout)
 
   def park: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.PARK, Activity.PRESET_START, Configuration.Zero),
-      defaultCommandTimeout)
+      DefaultCommandTimeout)
 
   def verify: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.VERIFY,
               Activity.PRESET_START,
-              Configuration.Zero), defaultCommandTimeout)
+              Configuration.Zero), DefaultCommandTimeout)
 
   def endVerify: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.END_VERIFY,
               Activity.PRESET_START,
-              Configuration.Zero), defaultCommandTimeout)
+              Configuration.Zero), DefaultCommandTimeout)
 
   def guide: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.GUIDE, Activity.PRESET_START, Configuration.Zero),
-      defaultCommandTimeout)
+      DefaultCommandTimeout)
 
   def endGuide: F[CommandResult] =
     giapi.command(
       Command(SequenceCommand.END_GUIDE,
               Activity.PRESET_START,
-              Configuration.Zero), defaultCommandTimeout)
+              Configuration.Zero), DefaultCommandTimeout)
 
   def observe[A: Show](dataLabel: A, expTime: FiniteDuration): F[CommandResult] =
     giapi.command(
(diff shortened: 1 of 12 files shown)


modules/giapi/src/main/scala/giapi/client/gpi/GPIClient.scala
```diff
@@ -10,12 +10,13 @@ import fs2.Stream
...
 import giapi.client.{Giapi, commands}
 import mouse.boolean._
-import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
 
 /**
   * Client for GPI
   */
-class GPIClient[F[_]](giapi: Giapi[F], ec: ExecutionContext) {
+class GPIClient[F[_]](giapi: Giapi[F]) {
```
cquiroz: now the `ExecutionContent` is passed to the giapi connection

modules/giapi/src/main/scala/giapi/client/package.scala
```diff
@@ -90,12 +90,12 @@ package client {
...
       * This decision may be review in the future
       */
-    def command(command: Command): F[CommandResult]
+    def command(command: Command, timeout: FiniteDuration): F[CommandResult]
 
     /**
       * Returns a stream of values for the status item.
       */
-    def stream[A: ItemGetter](statusItem: String, ec: ExecutionContext): F[Stream[F, A]]
+    def stream[A: ItemGetter](statusItem: String): F[Stream[F, A]]
```
cquiroz: Let's remove the `ExecutionContext` and pass it to the `GiapiConnection` builder

modules/giapi/src/main/scala/giapi/client/package.scala
```diff
@@ -157,42 +157,67 @@ package client {
...
       }
 
+    // taken from cats-effect 1.x
+    private def toIOFromRunAsync[F[_], A](f: F[A])(implicit F: Effect[F]): IO[A] =
+      IO.async { cb =>
+        F.runAsync(f)(r => IO(cb(r))).unsafeRunSync()
+      }
+
+    // Implementations of timeout suggested from cats-effect documentation
+    private def timeoutTo[F[_]: Effect, A](fa: F[A], after: FiniteDuration, fallback: F[A])
```
cquiroz: This is taken from `cats-effect` but on `F[_]` rather than `IO`. It maybe better to use `IO` instead and convert to `F[_]` on the caller instead. I can't tell which one is better

modules/giapi/src/main/scala/giapi/client/package.scala
```diff
@@ -157,42 +157,67 @@ package client {
...
               }
 
             def getO[A: ItemGetter](statusItem: String): F[Option[A]] =
               Sync[F].delay {
                 val item = sg.getStatusItem[A](statusItem)
                 Option(item).map(_.getValue)
               }
+            override def command(command: Command, timeOut: FiniteDuration): F[CommandResult] = {
+              val error = CommandResultException.timedOut(timeOut)
+              timeout(commands.sendCommand(cc, command, commandsAckTimeout), timeOut, error)
```
cquiroz: Now a command will timeout if it doesn't complete on the given duration

jluhrs:
Does it handle the case when a GPI declares a command completed after it timed out on the Seqexec side?

cquiroz:
in that case the step will be declared an error. That's why the timeout value should be relatively large

modules/giapi/src/main/scala/giapi/client/package.scala
cquiroz: /cc @tpolecat

modules/giapi/src/main/scala/giapi/client/gpi/GPIClient.scala
tpolecat: Is this used for blocking operations, or is it all nonblocking from the EPICs client point of view? This will affect the kind of EC you want to use here.

modules/giapi/src/main/scala/giapi/client/gpi/GPIClient.scala
```diff
@@ -10,12 +10,13 @@ import fs2.Stream
...
+  val defaultCommandTimeout: FiniteDuration = 2000.milliseconds
```
tpolecat: Can we capitalize this since it's a constant? i.e., `DefaultCommandTimeout`

I would not be entirely opposed to this being a default argument on `Command.apply`, just to remove the repetition. I don't normally like default args but it might make sense here. What do you think?

cquiroz:
I updated the timeouts with some consultations with the GPI team and from real world testing

modules/giapi/src/main/scala/giapi/client/package.scala
tpolecat: So with cats-effect 1.0 you have `F <~> IO` given `Effect[F]` via `fa.toIO` and then `io.liftIO[F]` … so what I might do here is use an implicit class to add a `.toIO` method for any `F: Effect` (`.liftIO` already works) and then delegate the timeout stuff to the existing `IO` combinator. When cats-effect 1.0 comes out we can just delete the implicit class. Does that make sense?

modules/giapi/src/main/scala/giapi/client/package.scala
```diff
@@ -157,42 +157,67 @@ package client {
...
-                // Note item.getValue can throw if e.g. the item is unknown, let's avoid NPEs though
-                Option(item) match {
-                  case Some(i) => i.getValue
-                  case None    => throw GiapiException(s"Status item $statusItem not found")
-                }
+              getO[A](statusItem).flatMap {
+                case Some(a) => a.pure[F]
+                case None =>
+                  Sync[F].raiseError(
+                    new GiapiException(s"Status item $statusItem not found"))
```
tpolecat: FYI you can say `OptionT(getO[A](statusItem)).getOrElseF[A](Sync[F].raiseError(...)).value` if that seems better. What you have here is fine though.

modules/giapi/src/main/scala/giapi/client/gpi/GPIClient.scala
cquiroz: In this case probably we don't want to have a default arg. This value is mostly the same for all GPI commands but it is not necessarily the same for other GIAPI instruments like GHOST

cquiroz:
The EC is used to get a `Timer[IO]` as required by `IO.race` on the timeouts. I can't really tell if we'd ever need something different than scala's default EC but this leaves the option open

cquiroz:
Testing with GPI was quit succesful. Here is a screenshot of a forced timeout. I did this starting an obs and killing the software on the instrument side resposible for sending confirmation of completion

seqexec - gs-2018b-q-0-15 2018-08-02 13-57-52

modules/seqexec/web/server/src/main/resources/app.conf
```diff
@@ -67,6 +67,7 @@ seqexec-engine {
     smartGCalHost = "gsodbtest.gemini.edu"
     # Tmp file for development
     smartGCalDir = "/tmp/smartgcal"
-    gpiUrl = "vm://gpi?marshal=false&broker.persistent=false"
-    gpiGDS = "http://localhost:8888/xmlrpc"
+    #gpiUrl = "vm://gpi?marshal=false&broker.persistent=false"
+    gpiUrl = "failover:(tcp://172.17.107.50:61616)?timeout=4000"
+    gpiGDS = "http://172.17.107.50:8888/xmlrpc"
```
jluhrs: These changes are from the testing with the real instruments. They must be reverted before the commit.

PR #614, opened 2018-08-02
Improve TCS subsystem logging

cquiroz:
This is a small improvement on how tcs logging is done from:
seqexec - gs-2018b-q-0-2 2018-08-02 11-11-56

to
seqexec - gs-2018b-q-0-2 2018-08-02 11-19-23


tpolecat:
We might want to generate this enumeration from the database at some point.

cquiroz:
Do you think that would be needed, these subsystems will probably never ever change.
Or do you think it would be good to do queries?

tpolecat:
Yeah that's what I mean. If we ever start logging stuff to the database by subsystem then it would be nice to have the enum defined there. It's ok for now.

cquiroz:
Maybe we should consider a task of migrating the many enums we have in the seqexec

tpolecat:
I don't think there's any urgency.

cquiroz:
This PR got somhow disconnected from buildkite, I'll close and reopen

PR #616, opened 2018-08-03
Don't use arbitrary automatic derivation

cquiroz:
Some time ago i tried to use `scalacheck-shapeless` to do automatic derivation of `Arbitrary`. While it removes boiler plate it increases substantially compilation time

Here I'm removing it and write the arbitraries explicitly 

It should have some 20-30s of compilation time


sraaphorst:
Using shapeless like that is something that never would have occurred to me. Very interesting that it made writing the arbitraries so easy.

Anyways, this looks good.

PR #617, opened 2018-08-09
extract seqexec enums

tpolecat:
I'm going to be doing some janitorial work on the seqexec codebase, to help readability. This just moves the enumerated types from `Model.scala` and into their own source files, in their own package.

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/enum/Resource.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/enum/Resource.scala
@@ -1,0 +1,68 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.model.enum
+
+import cats.{ Eq, Order, Show }
+import cats.data.NonEmptyList
+import cats.implicits._
+
+/** A Seqexec resource represents any system that can be only used by one single agent. */
+sealed abstract class Resource(val ordinal: Int, val label: String)
+  extends Product with Serializable
+
+object Resource {
+
+  case object P1     extends Resource(1, "P1")
+  case object OI     extends Resource(2, "OI")
+  case object TCS    extends Resource(3, "TCS")
+  case object Gcal   extends Resource(4, "Gcal")
+  case object Gems   extends Resource(5, "Gems")
+  case object Altair extends Resource(6, "Altair")
+
+  // Mount and science fold cannot be controlled independently. Maybe in the future.
+  // For now, I replaced them with TCS
+  //  case object Mount extends Resource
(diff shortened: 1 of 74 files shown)


swalker2m:
This seems straightforward and correct to me, but I'll let Javier and Carlos take a look.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -319,13 +320,13 @@ class SeqTranslate(site: Site, systems: Systems, settings: Settings) {
...
-    case Model.Instrument.GPI   => TrySeq(GPI(systems.gpi))
-    case Model.Instrument.GHOST => TrySeq(GHOST(GHOSTController[IO](GDSClient(GDSClient.alwaysOkClient, uri("http://localhost:8888/xmlrpc"))))) // todo put the controller on systems
+  private def toInstrumentSys(inst: Instrument): TrySeq[InstrumentSystem[IO]] = inst match {
+    case Instrument.F2    => TrySeq(Flamingos2(systems.flamingos2, systems.dhs))
+    case Instrument.GmosS => TrySeq(GmosSouth(systems.gmosSouth, systems.dhs))
+    case Instrument.GmosN => TrySeq(GmosNorth(systems.gmosNorth, systems.dhs))
+    case Instrument.GNIRS => TrySeq(Gnirs(systems.gnirs, systems.dhs))
+    case Instrument.GPI   => TrySeq(GPI(systems.gpi))
+    case Instrument.GHOST => TrySeq(GHOST(GHOSTController[IO](GDSClient(GDSClient.alwaysOkClient, uri("http://localhost:8888/xmlrpc"))))) // todo put the controller on systems
     case _                      => TrySeq.fail(Unexpected(s"Instrument $inst not supported."))
```
cquiroz: We should likely align this one

modules/seqexec/model/src/main/scala/seqexec/model/enum/Resource.scala
```diff
@@ -0,0 +1,68 @@
...
+    Show.show(_.label)
+
+  implicit val ordering: Ordering[Resource] =
+    order.toOrdering
+}
+
+sealed abstract class Instrument(ordinal: Int, label: String)
+  extends Resource(ordinal, label)
+
+object Instrument {
```
cquiroz: We should try soon to replace this with `gem.enum.Instrument`

cquiroz:
Looks good to me.

modules/seqexec/model/src/main/scala/seqexec/model/enum/Resource.scala
tpolecat: Yeah I'm going to tidy up a bit and then I'll circle back and try to replace stuff that exists in core.

PR #619, opened 2018-08-09
remove Model.scala

tpolecat:
This removes `seqexec.model.Model` … all the inner classes are now top-level and the type aliases are in a package object.


modules/seqexec/model/src/main/scala/seqexec/model/package.scala
```diff
@@ -0,0 +1,21 @@
...
+import cats._
+import seqexec.model.enum._
+
+package object model {
+  type ParamName = String
+  type ParamValue = String
+  type Parameters = Map[ParamName, ParamValue]
+  type StepConfig = Map[SystemName, Parameters]
+  implicit val stEq: Eq[StepConfig] = Eq.fromUniversalEquals
+  type StepId = Int
```
cquiroz: Would you switch this for `Index`?

modules/seqexec/model/src/main/scala/seqexec/model/package.scala
```diff
@@ -0,0 +1,21 @@
...
+  type ObservationName = String
+  type TargetName = String
```
cquiroz: Observation and Target name probably don't have a type in gem. are they just strings?

modules/seqexec/model/src/main/scala/seqexec/model/package.scala
tpolecat: `Index` is a positive short, so yeah anything where you start numbering at 1.

modules/seqexec/model/src/main/scala/seqexec/model/package.scala
tpolecat: Yeah they're just strings.

PR #620, opened 2018-08-10
Library/sbt updates

cquiroz:
The new `sbt-crossproject` release lets us simplify the build somewhat. It requires sbt 1.2.1 though

I'm updating dependencies too


build/buildkite-build.sh
```diff
@@ -59,6 +59,9 @@ do
   sleep 1
 done
 
+echo "--- :node: Install jsdom"
+npm install jsdom
```
cquiroz: @tpolecat I have to add this to run some of the tests. I don't know if there is a way to have this installed only once

PR #622, opened 2018-08-13
Asset size checks

cquiroz:
We used to have checks on the js asset sizes on travis. The move to buildkite broke them and this PR restores them by adapting the script setting github statuses

I also added checks for the fully optimized assets

### DIFF
--- a/modules/seqexec/web/client/src/webpack/webpack.parts.js
+++ b/modules/seqexec/web/client/src/webpack/webpack.parts.js
@@ -57,11 +57,19 @@ module.exports.lessLoader = (options = {}) => {
 };
 
 // Extract css to a file, use only in production
-module.exports.extractCSS = ({ devMode, include, exclude, use = [] }) => {
+module.exports.extractCSS = ({
+  devMode,
+  include,
+  exclude,
+  use = [],
+  ci = false
+}) => {
+  const filename = ci ? "[name].css" : "[name].[contenthash].css";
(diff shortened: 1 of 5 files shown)


modules/seqexec/web/client/src/webpack/webpack.parts.js
```diff
@@ -57,11 +57,25 @@ module.exports.lessLoader = (options = {}) => {
...
+    filename: devMode
+      ? "[name].css"
+      : ci
+        ? "[name].css"
+        : "[name].[contenthash].css",
+    chunkFilename: devMode
+      ? "[id].css"
+      : ci
+        ? "[id].css"
+        : "[id].[contenthash].css"
```
sraaphorst: This is rather confusing. I can see what you are doing, but the nested ? : without parentheses makes me feel uncomfortable.

sraaphorst:
Buildkite doesn't seem to like your changes... I'll okay them once you get buildkite working!

modules/seqexec/web/client/src/webpack/webpack.parts.js
cquiroz: This is made by prettier a very standard js beautifier. I'll try to get something more readable in

cquiroz:
yeah, i need to work a bit to get buildkite to run it. It worked fine on my own buildpipe but not on main

tpolecat:
This is great, thanks! 👍

PR #625, opened 2018-08-21
Simplified execution core.

jluhrs:
I moved some sequence details up to package `seqexec.server`, making the code in `seqexec.engine` simpler.
The way parameter updates are passed to the Step execution code was also changed (parameters like observer, operator, observing conditions). Until now, the actual IO for the Action would be created just before running it. In this new code I keep all the information necessary to create the sequence, and refresh the actual tasks when any parameter change. That removes the need to pass the changing parameters through the execution engine, and simplifies the future addition of new parameters.


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -22,15 +21,14 @@ import org.log4s.getLogger
 
 import scala.concurrent.ExecutionContext
 
-class Engine[D: ActionMetadataGenerator, U](implicit ev: ActionMetadataGenerator[D]) {
+class Engine[D, U](stateL: Lens[D, Engine.State]) {
```
jluhrs: I basically turned around the relation between Engine.State and the state used by the upper level.

modules/seqexec/engine/src/main/scala/seqexec/engine/UserEvent.scala
```diff
@@ -7,78 +7,61 @@ import cats.effect.IO
...
 import gem.Observation
-import seqexec.model.{ ClientID, Observer, UserDetails }
+import seqexec.model.ClientID
+import seqexec.model.UserDetails
 
 /**
   * Events generated by the user.
   */
-sealed trait UserEvent[+D <: Engine.Types] {
+sealed trait UserEvent[D<:Engine.Types] {
```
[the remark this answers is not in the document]
cquiroz: no variance is always simpler 😄

modules/seqexec/engine/src/main/scala/seqexec/engine/Event.scala
```diff
@@ -18,20 +19,16 @@ final case class EventSystem(se: SystemEvent) extends Event[Nothing]
 
 object Event {
 
-  def start[D<:Engine.Types](id: Observation.Id, user: UserDetails, clientId: ClientID): Event[D] = EventUser[D](Start(id, user.some, clientId))
+  def start[D<:Engine.Types](id: Observation.Id, user: UserDetails, clientId: ClientID, f: D#StateType => Boolean): Event[D] = EventUser[D](Start[D](id, user.some, clientId, f))
   def pause[D<:Engine.Types](id: Observation.Id, user: UserDetails): Event[D] = EventUser[D](Pause(id, user.some))
   def cancelPause[D<:Engine.Types](id: Observation.Id, user: UserDetails): Event[D] = EventUser[D](CancelPause(id, user.some))
-  def load[D<:Engine.Types](id: Observation.Id, sequence: Sequence): Event[D] = EventUser[D](Load(id, sequence))
```
cquiroz: Why are these gone? `load`, `unload`, `setObserver`

modules/seqexec/server/src/main/scala/seqexec/server/SequenceGen.scala
```diff
@@ -0,0 +1,24 @@
...
+
+
+import cats.Eq
+import cats.implicits._
+import gem.Observation
+import seqexec.model.StepConfig
+import seqexec.model.enum.{Instrument, Resource}
+import seqexec.engine.{Step => EngineStep}
+
+final case class SequenceGen(id: Observation.Id, title: String, instrument: Instrument, steps: List[SequenceGen.Step]) {
```
cquiroz: What's the role of this class?

modules/seqexec/server/src/main/scala/seqexec/server/SequenceGen.scala
```diff
@@ -0,0 +1,24 @@
...
+  val resources: Set[Resource] = steps.foldMap(_.resources)
+}
+
+object SequenceGen {
+
+  implicit val eq: Eq[SequenceGen] = Eq.by(x => (x.id))
```
cquiroz: This doesn't seem right. it only uses the id field for equality

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -16,21 +16,27 @@ import monocle.Lens
...
-  object EngineMetadata {
-    implicit val eq: Eq[EngineMetadata] = Eq.by(x => (x.queues, x.selected, x.conditions, x.operator))
+  object ObserverSequence {
+    implicit val eq: Eq[ObserverSequence] = Eq.by(x => (x.observer, x.seq))
+  }
+  @Lenses
+  final case class EngineState(queues: ExecutionQueues, selected: Map[Instrument, Observation.Id], conditions: Conditions, operator: Option[Operator], sequences: Map[Observation.Id, ObserverSequence], executionState: Engine.State)
+  @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
+  object EngineState {
+    implicit val eq: Eq[EngineState] = Eq.by(x => (x.queues, x.selected, x.conditions, x.operator, x.sequences))//, x.executionState))
```
cquiroz: Why is `executionState` left out? Do we have law tests for these Eq instances?

modules/seqexec/server/src/test/scala/seqexec/server/SeqexecEngineSpec.scala
```diff
@@ -170,221 +175,269 @@ class SeqexecEngineSpec extends FlatSpec with Matchers {
...
-    }).compile.last.unsafeRunSync
+    }).unsafeRunSync
   }
+
   it should "not add sequence id if sequence is running or completed" in {
-    val s0 = Engine.State.empty(EngineMetadata.default)
-      .copy(sequences = Map(seqObsId1 -> Sequence.State.status.set(SequenceState.Running.init)(sequence(seqObsId1)),
-        seqObsId2 -> Sequence.State.status.set(SequenceState.Completed)(sequence(seqObsId2))
-      ) )
+    val s0 = (SeqexecEngine.loadSequenceEndo(seqObsId1, sequence(seqObsId1)) >>>
```
cquiroz: is `>>>` the same as `andThen` right? is that on the core lib or from cats?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -456,7 +467,7 @@ object SeqTranslate {
   }
 
   private def extractInstrument(config: Config): TrySeq[Instrument] = {
-    config.extractAs[String](INSTRUMENT_KEY / INSTRUMENT_NAME_PROP).asTrySeq.flatMap {
+    config.extract(INSTRUMENT_KEY / INSTRUMENT_NAME_PROP).as[String].asTrySeq.flatMap {
```
cquiroz: Why reverting to `extract().as[A]`?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -143,29 +151,27 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
   def seqQueueRefreshStream: Stream[IO, executeEngine.EventType] =
     Scheduler[IO](corePoolSize = 1).flatMap { scheduler =>
       val fd = Duration(settings.odbQueuePollingInterval.toSeconds, TimeUnit.SECONDS)
-      scheduler.fixedDelay[IO](fd).flatMap { _ =>
-        Stream.emit(Event.getState[executeEngine.ConcreteTypes](refreshSequenceList()))
+      scheduler.fixedDelay[IO](fd).evalMap(_ => odbProxy.queuedSequences.value).map { x =>
+        Event.getState[executeEngine.ConcreteTypes](st => x.map(refreshSequenceList(_)(st)).valueOr(r => List(Event.logWarningMsg(SeqexecFailure.explain(r)))).some.filter(_.nonEmpty).map(Stream.emits(_).covary[IO]))
```
cquiroz: Could we reformat the code to make this clearer?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -143,29 +151,27 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
+          toSeqexecEvent(ev, qState)(
+            SequencesQueue(
+              EngineState.selected.get(qState),
+              EngineState.conditions.get(qState),
+              EngineState.operator.get(qState),
+              qState.sequences.values.map(
+                s => qState.executionState.sequences.get(s.seq.id).map(x => viewSequence(s, x.toSequence, x))
+              ).collect{ case Some(x) => x }.toList
             )
-            Stream.eval(updateMetrics[IO](ev, sequences).map(_ => event))
```
cquiroz: Please restore this

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -176,28 +182,32 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
 
-  private def addSeq(name: String, seqId: Observation.Id)(st: executeEngine.StateType): executeEngine.StateType = (
+  private def addSeq(name: String, seqId: Observation.Id)(st: EngineState): EngineState = (
     for {
-      q   <- st.userData.queues.get(name)
-      seq <- st.sequences.get(seqId)
-    } yield if(!q.status(st).isRunning && !seq.status.isRunning && !seq.status.isCompleted && !q.contains(seqId)) st.copy(userData = EngineMetadata.queues.modify(_.updated(name, q :+ seqId))(st.userData)) else st
+      q   <- st.queues.get(name)
+      seq <- st.executionState.sequences.get(seqId)
+    } yield if(!q.status(st).isRunning && !seq.status.isRunning && !seq.status.isCompleted && !q.contains(seqId))
```
cquiroz: Maybe we put these on a `{ ... }` block for readability

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -215,88 +225,43 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
 
-  private def loadEvents(seqId: Observation.Id): SeqAction[List[executeEngine.EventType]] = {
-    val t: EitherT[IO, SeqexecFailure, (List[SeqexecFailure], Option[Sequence])] = for {
-      odbSeq       <- SeqAction.either(odbProxy.read(seqId))
-      progIdString <- SeqAction.either(odbSeq.config.extract(OCS_KEY / InstConstants.PROGRAMID_PROP).as[String].leftMap(ConfigUtilOps.explainExtractError))
-      _            <- SeqAction.either(Either.catchNonFatal(IO.pure(SPProgramID.toProgramID(progIdString))).leftMap(e => SeqexecFailure.SeqexecException(e): SeqexecFailure))
+  private def loadEvents(seqId: Observation.Id): List[executeEngine.EventType] = {
+    val t: Either[SeqexecFailure, (List[SeqexecFailure], Option[SequenceGen])] = for {
+      odbSeq       <- odbProxy.read(seqId)
+      progIdString <- odbSeq.config.extract(OCS_KEY / InstConstants.PROGRAMID_PROP).as[String].leftMap(ConfigUtilOps.explainExtractError)
```
cquiroz: `extractAs[String]...`

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -307,7 +272,7 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
     (e match {
       // TODO Add metrics for more events
       case engine.EventUser(ue) => ue match {
-        case engine.Start(id, _, _) => instrument(id).map(sm.startRunning[F]).getOrElse(Sync[F].unit)
+        case engine.Start(id, _, _, _) => instrument(id).map(sm.startRunning[F]).getOrElse(Sync[F].unit)
```
cquiroz: Alignment?

modules/seqexec/engine/src/main/scala/seqexec/engine/UserEvent.scala
```diff
@@ -7,78 +7,61 @@ import cats.effect.IO
...
-final case class SkipMark(id: Observation.Id, user: Option[UserDetails], step: Step.Id, v: Boolean) extends UserEvent[Nothing]
-//final case class SetOperator(name: Operator, user: Option[UserDetails]) extends UserEvent[Nothing]
-final case class SetObserver(id: Observation.Id, user: Option[UserDetails], name: Observer) extends UserEvent[Nothing]
-//final case class SetConditions(conditions: Conditions, user: Option[UserDetails]) extends UserEvent[Nothing]
-//final case class SetImageQuality(iq: ImageQuality, user: Option[UserDetails]) extends UserEvent[Nothing]
-//final case class SetWaterVapor(wv: WaterVapor, user: Option[UserDetails]) extends UserEvent[Nothing]
-//final case class SetSkyBackground(wv: SkyBackground, user: Option[UserDetails]) extends UserEvent[Nothing]
-//final case class SetCloudCover(cc: CloudCover, user: Option[UserDetails]) extends UserEvent[Nothing]
-final case class Poll(clientId: ClientID) extends UserEvent[Nothing] {
+final case class Start[D<:Engine.Types](id: Observation.Id, user: Option[UserDetails], clientId: ClientID, f: D#StateType => Boolean) extends UserEvent[D]
```
jluhrs: `f` is a user provided check. It is currently used to test resource conflict. I am not sure it was the best mechanism.
Other options I thought of:
1. Use a `GetState` message to inspect the current state, and produce a `Start` event if there was no resource conflict. This is not feasible, because it risk that another event slips in between the processing of the `GetState` and the `Start`. The state check and the sequence status change to "Running" must be done in the same cycle. Another issue is how to report that the sequence was not started because of resource conflicts. Maybe produce a `Busy` event?
2. Remove the event `Start` and make the method `Engine.start` public, with a signature similar to `Engine.load`, that could be used from the function inside a `ModifyState` message. That keeps the state check and the sequence start in the same cycle.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
jluhrs: It does not compile if included.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -16,21 +16,27 @@ import monocle.Lens
...
-  final case class EngineMetadata(queues: ExecutionQueues, selected: Map[Instrument, Observation.Id], conditions: Conditions, operator: Option[Operator])
+  final case class ObserverSequence(observer: Option[Observer], seq: SequenceGen)
   @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
-  object EngineMetadata {
-    implicit val eq: Eq[EngineMetadata] = Eq.by(x => (x.queues, x.selected, x.conditions, x.operator))
+  object ObserverSequence {
+    implicit val eq: Eq[ObserverSequence] = Eq.by(x => (x.observer, x.seq))
+  }
+  @Lenses
+  final case class EngineState(queues: ExecutionQueues, selected: Map[Instrument, Observation.Id], conditions: Conditions, operator: Option[Operator], sequences: Map[Observation.Id, ObserverSequence], executionState: Engine.State)
```
jluhrs: There is a one to one relationship between the elements of `sequences` and `executionState.sequences` that I didn't know how to represent without complicating the structure of executionState.

modules/seqexec/server/src/main/scala/seqexec/server/SequenceGen.scala
jluhrs: That class keeps the "recipe" to produce an `engine.Sequence`, which allows for recreating those sequences when any parameter changes.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: I don't remember making this change.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: I. Hate. You.

jluhrs:
Changes to Operator, Observer and condition changes are propagated to the `engine.Sequence` definition, and are properly set in the FITS header. Resource checks are also preserved.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
tpolecat: > It does not compile if included.

What is your reasoning here? We can't have invalid instances.

modules/seqexec/server/src/test/scala/seqexec/server/SeqexecEngineSpec.scala
jluhrs: Yes, it is the same, and it is from cats.

modules/seqexec/engine/src/main/scala/seqexec/engine/Event.scala
jluhrs: They are implemented differently, and don't require an event anymore.

jluhrs:
Yes, I had to remove them because of the lack of `Eq`, even the ones for manually coded lenses.

cquiroz:
@tpolecat Suggested we could make local `Eq` instances for the tests, to verify at least the manually coded lenses

swalker2m:
This will be the first application delivered under the "ocs3" banner so it's pretty important.  It's also a fairly large codebase and a bit difficult to approach if you haven't been involved in the design from the outset.  We need to find a way to bring the rest of us up to speed on how the code works.  I find it difficult to follow by just looking at the code.

One thing that would help, as mentioned in PR comments, would be to increase documentation in the code.  For example, `Engine` sounds like a pretty important part but there isn't any comment about what it does or what `D` or `U` are for:

```scala
class Engine[D, U](stateL: Lens[D, Engine.State])
```

I see later that

```scala
  override type StateType = D
  override type EventData = U
```

but it isn't clear to the uninitiated what these mean either.

To be fair, you could level the same criticism about a lot of our code and a lot of code that I've contributed as well!  Nevertheless, we should all improve our documentation.

Because it is a large application that we intend to maintain indefinitely going forward, perhaps a brief design document would help as an initial introduction to the codebase?

cquiroz:
@swalker2m I think that is a fair assesment, a design document and a presentation would be useful  in general. this applies to several parts, the UI, the engine etc. A difficult is of course that the application has evolved a lot, such document would be already obsolete if it had been done a few months ago

swalker2m:
> such document would be already obsolete if it had been done a few months ago

Right, that's always an issue.  I remember seeing extensive documentation of one of the Gemini principal systems with class diagrams for every class and interaction diagrams, etc.  At that level the documentation quickly stops matching the code.  A brief markdown document kept with the code that introduces the concepts though might be useful for longer and at least help the rest of us get up to speed, idk.

jluhrs:
@swalker2m  I created Jira task SEQNG-696 to document code design.

PR #629, opened 2018-08-27
SEQNG-691: Increase size of the step pause button

cquiroz:
Some users have requested the breakpoint button to be made bigger. This PR addresses this and also shows a representation of the breakpoint while hovering. This should be user tested to get a better idea if this is enough

screencast 2018-08-27 15-11-52

The PR is somewhat larger than the actual changes as I did some small refactorings here and there

### DIFF
--- a/modules/seqexec/web/client/src/main/resources/less/style.less
+++ b/modules/seqexec/web/client/src/main/resources/less/style.less
@@ -482,6 +482,7 @@ body {
     @breakpoint_line_color 0px,
     @breakpoint_line_color
   );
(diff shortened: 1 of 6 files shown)


jluhrs:
Is it possible to cut the bottom of the previous row and/or the top of the following row when showing the breakpoints, to keep the following rows in the same position?

cquiroz:
What about this?

https://monosnap.com/file/byjzrAGP9eyQ60uc41yWn7LKIf8vm2#

PR #631, opened 2018-08-29
SEQNG-703: Remove use of RefTo in the model

cquiroz:
In diode one of the features is `RefTo` which lets you do references across the model. While teorethically nice in practice they don't work well as we need usually two actions to get the model correct. This is more evident when doing a load sequence which ends up being unnecessarily slow

In this PR I remove this simplifying the updates though by necessity this requires a bit more manual bookkeeping.

In addition I'm using monocle Lenses to generate diode `ModelRW` instances as they are easier to build and compose

The PR is largish but it is mostly test code

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/SequencesQueue.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/SequencesQueue.scala
@@ -6,12 +6,15 @@ package seqexec.model
 import cats._
 import cats.implicits._
 import gem.Observation
+import monocle.Getter
+import monocle.macros.Lenses
 import seqexec.model.enum.Instrument
 
 /**
   * Represents a queue with different levels of details. E.g. it could be a list of Ids
   * Or a list of fully hydrated SequenceViews
   */
+@Lenses
 final case class SequencesQueue[T](
   loaded:     Map[Instrument, Observation.Id],
   conditions: Conditions,
@@ -19,9 +22,12 @@ final case class SequencesQueue[T](
   queue:      List[T]
 )
 
+@SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
 object SequencesQueue {
 
   implicit def equal[T: Eq]: Eq[SequencesQueue[T]] =
-    Eq.by(x => (x.conditions, x.operator, x.queue))
+    Eq.by(x => (x.loaded, x.conditions, x.operator, x.queue))
 
+  def queueItemG[T](predicate: T => Boolean): Getter[SequencesQueue[T], Option[T]] =
+    SequencesQueue.queue composeGetter Getter[List[T], Option[T]] { _.find(predicate) }
 }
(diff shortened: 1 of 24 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
```diff
@@ -37,19 +36,35 @@ final case class SequencesOnDisplay(sequences: Zipper[SequenceTab]) {
   /**
    * List of loaded sequence ids
    */
-  val loadedIds: List[Observation.Id] =
-    sequences.toList.collect {
-      case InstrumentSequenceTab(_, curr, _, _) => curr().map(_.id)
-    }.collect {
-      case Some(x) => x
+  val loadedId: List[Observation.Id] =
```
jluhrs: Not plural anymore?

cquiroz: hmm I did that to facilitate the refactoring, it should be reverted back to plural

modules/seqexec/model/src/main/scala/seqexec/model/SequencesQueue.scala
```diff
@@ -6,22 +6,28 @@ package seqexec.model
...
 import gem.Observation
+import monocle.Getter
+import monocle.macros.Lenses
 import seqexec.model.enum.Instrument
 
 /**
   * Represents a queue with different levels of details. E.g. it could be a list of Ids
   * Or a list of fully hydrated SequenceViews
   */
+@Lenses
```
sraaphorst: Does `@Lenses` generate the lenses automatically?

modules/seqexec/model/src/test/scala/seqexec/model/SeqexecModelArbitraries.scala
```diff
@@ -26,12 +25,6 @@ trait SeqexecModelArbitraries extends ArbObservation {
 
   implicit val clientIdArb: Arbitrary[ClientID] = Arbitrary(Gen.uuid)
 
-  implicit val instArb: Arbitrary[Instant] = Arbitrary {
-    for {
-      i <- Gen.choose(0L, Long.MaxValue)
-    } yield Instant.ofEpochMilli(i)
-  }
-
```
sraaphorst: Why are we getting rid of the arbitrary `Instant`?

modules/seqexec/model/src/main/scala/seqexec/model/SequencesQueue.scala
cquiroz: Yes, it makes life easier though WartRemover complains 😄

modules/seqexec/model/src/test/scala/seqexec/model/SeqexecModelArbitraries.scala
cquiroz: Because there is another one defined on the `core` project. I'm using that instead

PR #632, opened 2018-08-30
SEQNG-706 Changed ModifyState event, propagated result.

jluhrs:
This change takes advantage that the state machine can produce outputs (which until now were discarded). This is now used in the event `ModifyState` to propagate a result from the user function. It can be used for other events, but that will be left for another task.


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -297,44 +298,43 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
...
   @SuppressWarnings(Array("org.wartremover.warts.ImplicitParameter"))
-  private def run(ev: EventType)(implicit ec: ExecutionContext): HandleP[D] = {
-    def handleUserEvent(ue: UserEventType): HandleP[Unit] = ue match {
-      case Start(id, _, clid, userCheck) => Logger.debug(s"Engine: Start requested for sequence $id") *> start(id, clid, userCheck)
-      case Pause(id, _)                  => Logger.debug("Engine: Pause requested") *> pause(id)
-      case CancelPause(id, _)            => Logger.debug("Engine: Pause canceled") *> cancelPause(id)
+  private def run(ev: EventType)(implicit ec: ExecutionContext): HandleP[ResultType] = {
+    def handleUserEvent(ue: UserEventType): HandleP[ResultType] = ue match {
+      case Start(id, _, clid, userCheck) => Logger.debug(s"Engine: Start requested for sequence $id") *> start(id, clid, userCheck) *> pure(UserCommandResponse(ue, EventResult.Ok, None))
+      case Pause(id, _)                  => Logger.debug("Engine: Pause requested") *> pause(id) *> pure(UserCommandResponse(ue, EventResult.Ok, None))
```
cquiroz: Where's `pure` defined?

jluhrs: In the same file, line 372.

PR #634, opened 2018-08-31
SEQNG-606 Don't load sequence if instrument is in use.

jluhrs:
This change prevents to assign a sequence to an instrument if the current one is running.
It also includes some changes to a show a notification to the user. The actual pop-up will be implemented in SEQNG-712


modules/seqexec/model/src/main/scala/seqexec/model/Notification.scala
```diff
@@ -0,0 +1,32 @@
...
+final case class InstrumentInUse(sid: Observation.Id, ins: Instrument) extends Notification
+
+object Notification {
+  def header(n: Notification): String = n match {
+    case ResourceConflict(_)           => "Resource conflict"
+    case InstrumentInUse(_, _) => "Instrument busy"
+  }
+
+  def body(n: Notification): List[String] = n match {
+    case ResourceConflict(sid)              => List(
```
cquiroz: Let's see how this works when displayed on the UI

PR #640, opened 2018-09-05
Fetch vs. Query

swalker2m:
This PR addresses feedback from #606, splitting the observation retrieval options in two: fetch vs. query.  Fetch expects an observation with the indicated ID to exist and raises an error if it doesn't while query returns optional results.

I let this update trickle all the way down to `ObservationDao`, though I suppose another option would be to leave the DAO with just the original exception raising option and handle the query option by dealing with the exception in `Service`.

### DIFF
--- a/modules/db/src/main/scala/gem/dao/ObservationDao.scala
+++ b/modules/db/src/main/scala/gem/dao/ObservationDao.scala
@@ -37,7 +37,7 @@ object ObservationDao {
 
   /**
    * Construct a program to select the specified observation, with the
-   * instrument but not targets nor steps.  Raises an error if the indicated
+   * instrument but not targets nor steps.  Raises an exception if the indicated
    * observation does not exist.
    */
   def fetchFlat(id: Observation.Id): ConnectionIO[(String, Instrument)] =
@@ -52,7 +52,7 @@ object ObservationDao {
 
   /**
    * Construct a program to select the specified observation, with the
-   * targets and instrument type but not steps.  Raises an error if the
+   * targets and instrument type but not steps.  Raises an exception if the
    * indicated observation does not exist.
    */
   def fetchTargets(id: Observation.Id): ConnectionIO[(String, TargetEnvironment)] =
@@ -73,7 +73,7 @@ object ObservationDao {
 
   /**
    * Construct a program to select the specified observation, with static
-   * config but not targets nor steps.  Raises an error if the indicated
+   * config but not targets nor steps.  Raises an exception if the indicated
    * observation does not exist.
    */
   def fetchStatic(id: Observation.Id): ConnectionIO[(String, StaticConfig)] =
@@ -94,7 +94,7 @@ object ObservationDao {
 
   /**
    * Construct a program to select the specified observation, with static
-   * config and steps but not targets.  Raises an error if the indicated
+   * config and steps but not targets.  Raises an exception if the indicated
    * observation does not exist.
    */
   def fetchConfig(id: Observation.Id): ConnectionIO[(String, StaticConfig, TreeMap[Location.Middle, Step])] =
@@ -115,8 +115,8 @@ object ObservationDao {
 
   /**
    * Construct a program to select a fully specified observation, with targets,
-   * static config and steps.  Raises an error if the indicated observation does
-   * not exist.
+   * static config and steps.  Raises an exception if the indicated observation
+   * does not exist.
    */
   def fetch(id: Observation.Id): ConnectionIO[Observation] =
     for {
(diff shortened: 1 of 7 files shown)


modules/web/src/main/scala/Application.scala
```diff
@@ -50,11 +50,17 @@ object Application {
...
-          gs.queryObservationById(oid).flatMap { obs =>
-            Ok(obs.asJson)
+          gs.fetchObservationById(oid).flatMap { obs => Ok(obs.asJson) }
+        }
+
+      // Query an observation by id.
+      case GET -> Root / "api" / "query" / "obs" / o as gs =>
+        withObsId(o) { oid =>
+          gs.queryObservationById(oid).flatMap { o =>
+            o.fold(NotFound())(obs => Ok(obs.asJson))
```
swalker2m: Another request from #606 feedback was test cases for the web service routes.  I'll add that in the next PR.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -35,98 +35,161 @@ object ObservationDao {
...
 
-  /** Construct a program to select the specified observation, with the
-    * targets and instrument type but not steps.
-    */
-  def selectTargets(id: Observation.Id): ConnectionIO[(String, TargetEnvironment)] =
+  /**
+   * Construct a program to query for the specified observation, with the
+   * instrument but not targets nor steps.
+   */
+  def queryFlat(id: Observation.Id): ConnectionIO[Option[(String, Instrument)]] =
```
swalker2m: I suppose this could be `OptionT[ConnectionIO, ...]`.

modules/db/src/main/scala/gem/dao/ObservationDao.scala
```diff
@@ -35,98 +35,161 @@ object ObservationDao {
...
-  def selectStatic(id: Observation.Id): ConnectionIO[(String, StaticConfig)] =
+  /**
+   * Construct a program to query for the specified observation, with the
+   * targets and instrument type but not steps.
+   */
+  def queryTargets(id: Observation.Id): ConnectionIO[Option[(String, TargetEnvironment)]] =
+    for {
+      o <- queryFlat(id)
+      t <- o.as(TargetEnvironmentDao.selectObs(id)).sequence
+    } yield o.map(_._1).product(t)
```
cquiroz: I thought this was `fproduct`

modules/db/src/test/scala/gem/dao/ObservationDaoSpec.scala
```diff
@@ -64,14 +90,39 @@ class ObservationDaoSpec extends PropSpec with PropertyChecks with DaoTest {
     }
   }
 
-  property("ObservationDao should select target-only observations") {
-    val oid = Observation.Id(pid, Index.One)
+  property("ObservationDao should fail fetchStatic if missing") {
```
cquiroz: What's `property`?

modules/service/src/main/scala/gem/Service.scala
```diff
@@ -28,11 +28,21 @@ final class Service[M[_]: Sync: LiftIO] private (private val xa: Transactor[M],
     }
 
   /**
-   * Constuct a program that yields the `Observation` with the matching observation id.
+   * Constuct a program that yields the `Observation` with the matching
+   * observation id, raising an error if none.
    */
-  def queryObservationById(id: Observation.Id): M[Observation] =
+  def fetchObservationById(id: Observation.Id): M[Observation] =
+    log.log(user, s"fetchObservationById(${id.format})") {
```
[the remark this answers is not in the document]
tpolecat: Yeah this logs to a JDK logger and also to the database, times the inner operation, and logs success/failure. We'll need to revisit all of this for the integrated logging/metrics stuff I guess.

swalker2m:
> Why is the build failing?

Unrelated.

```
[info] - Arbitrary IO should be parallel. *** FAILED ***
[info]   HorizonsClientSpec.this.exclusionTest(fs2.Stream.eval[cats.effect.IO, Unit](cats.effect.IO.unit)) was true (HorizonsClientSpec.scala:22)
```

modules/db/src/main/scala/gem/dao/ObservationDao.scala
swalker2m: `fproduct` from `Functor` takes an `F[A]` and a function `A => B` and creates a tuple `F[(A, B)]` while `product` from `Apply` just tuples two values in `F` together: 

```
def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
```

here `F` is `Option` of course.

modules/db/src/test/scala/gem/dao/ObservationDaoSpec.scala
swalker2m: Something from `scalatest` that "register[s] a property-based test with the specified name". I'm not a big fan of any of the test DSLs.  There are so many different ways to do the same thing and I never know which one to use.  I always just end up copying some existing test that seems similar and doing whatever it did.

PR #642, opened 2018-09-06
Application Tests

swalker2m:
Following a suggestion from #606, this PR offers a few basic `gem.web.Application` tests.  For the most part they just check that things are hooked up correctly.  In particular more extensive testing of the program listing query could be provided if we keep it in its current form.


swalker2m:
> do we need to do something to run this on development?

As long as you have a postgres database setup with `sql/flywayMigrate` it should work.  If the DAO tests run then this should too.

PR #643, opened 2018-09-06
SEQNG-715 Fixed code that updates and reloads sequences

jluhrs:
The `update` method in `Sequence.State.Zipper` was trying to do too many things. It was used to refresh the steps `executions` when, for example, the operator name changed, and also to update the whole sequence when reloading it from the ODB. I separated it into two functions, `update` and `reload`.

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -415,7 +415,7 @@ object Engine {
 
   object State {
     def empty: State = State(Map.empty)
-    def atSequence(id: Observation.Id): Optional[Map[Observation.Id, Sequence.State], Sequence.State] = index(id)
+    private def atSequence(id: Observation.Id): Optional[Map[Observation.Id, Sequence.State], Sequence.State] = index(id)
     def sequenceState[D](id: Observation.Id): Optional[State, Sequence.State] = State.sequences ^|-? atSequence(id)
   }
 
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -12,10 +12,9 @@ import seqexec.engine.Result.{PartialVal, PauseContext, RetVal}
 import seqexec.model.{ClientID, SequenceState}
 import fs2.Stream
 import gem.Observation
-import monocle.Lens
+import monocle.{Lens, Optional}
 import monocle.macros.Lenses
-import monocle.function.At.at
-import monocle.function.At.atMap
+import monocle.function.Index.index
 import mouse.boolean._
 import org.log4s.getLogger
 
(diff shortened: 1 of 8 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/Sequence.scala
```diff
@@ -231,12 +237,27 @@ object Sequence {
...
+        .getOrElse(Final(q, SequenceState.Idle))
+
+    /**
+      * Rebuilds the state of a sequence with a new steps definition, but preserving breakpoints and skip marks
+      * The sequence must not be running.
+      * @param q New sequence definition
+      * @param st Old sequence state
+      * @return The new sequence state
+      */
+    def reload(steps: List[Step], st:State): State =
```
jluhrs: Important: `reload` uses the step state from `steps`. In the case of reloading from the ODB, this means that if a step was run with the old Seqexec, that fact will be reflected in the sequence state.

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -407,8 +415,8 @@ object Engine {
 
   object State {
     def empty: State = State(Map.empty)
-    def atSequence(id: Observation.Id): Lens[Map[Observation.Id, Sequence.State], Option[Sequence.State]] = at(id)
-    def sequenceState[D](id: Observation.Id): Lens[State, Option[Sequence.State]] = State.sequences ^|-> atSequence(id)
+    def atSequence(id: Observation.Id): Optional[Map[Observation.Id, Sequence.State], Sequence.State] = index(id)
```
jluhrs: Just taking advantage of monocle.Optional

cquiroz: Did you add tests to these?

modules/seqexec/engine/src/main/scala/seqexec/engine/Sequence.scala
```diff
@@ -192,7 +192,13 @@ object Sequence {
...
     def start(i: Int): State
 
-    def update(sequence: Sequence): State
+    /**
+      * Updates the steps executions.
+      * It preserves the number of steps.
+      * @param stepDefs New executions.
+      * @return Updated state
+      */
+    def update(stepDefs: List[List[Actions]]): State
```
cquiroz: I've always found these a bit confusing with the nested `List` I wonder if there is a better way to express this

jluhrs: I could create type aliases

tpolecat: What's wrong with a list of lists? It's very clear what it means.

cquiroz: the val name `stepDeps` helps somewhat but perhaps another type could moke it clearer.

modules/seqexec/engine/src/main/scala/seqexec/engine/Sequence.scala
cquiroz: Is this called on `Sync`?

jluhrs: Yes

cquiroz:
Thanks for the tests

PR #645, opened 2018-09-07
SEQNG-701: Fixed giapi channel to set the GPI prism

cquiroz:
Tested with the instrument

seqexec - gs-2018b-q-0-22 2018-09-07 10-23-52


cquiroz:
I think that is a valid concern but at the end the instrument will accept those Strings verbatim. The one improvement could be to use constant but that dosen't give you a lot more

PR #647, opened 2018-09-12
SEQNG-720 Implemented basic Calibration Queue functionality.

jluhrs:
This PR implements basic functionality for the Calibration Queue, to be used to build the UI.


modules/seqexec/server/src/main/scala/seqexec/server/ExecutionQueue.scala
```diff
@@ -0,0 +1,20 @@
...
+package seqexec.server
+
+import cats.Eq
+import cats.implicits._
+import gem.Observation
+import monocle.macros.Lenses
+import seqexec.model.SequencesBatch.CommandState
+
+@Lenses
+final case class ExecutionQueue(name: String, cmdState: CommandState, queue: List[Observation.Id])
```
cquiroz: Why not `QueueId` here?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -213,54 +213,43 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
-        (EngineState.queues ^|-? index(name)).modify(moveElement(_, seqId, d))(st)
+      q <- st.queues.get(qid)
+    } yield if(q.status(st) =!= SequencesBatch.State.Running && q.queue.contains(seqId))
+        (EngineState.queues ^|-? index(qid)).modify(_.moveSeq(seqId, d))(st)
       else st
   ).getOrElse(st)
 
-  def moveSequenceInQueue(q: EventQueue, queueName: String, seqId: Observation.Id, d: Int): IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
-    Event.modifyState[executeEngine.ConcreteTypes](moveSeq(queueName, seqId, d) withEvent NullSeqEvent)
+  def moveSequenceInQueue(q: EventQueue, qid: QueueId, seqId: Observation.Id, d: Int): IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
```
cquiroz: Should there be checks  on the range of d?

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -85,6 +88,7 @@ package object server {
     Eq[Int].contramap(_.ordinal())
 
   val CalibrationQueueName: String = "Calibration Queue"
+  val CalibrationQueueId: UUID = UUID.fromString("7156fa7e-48a6-49d1-a267-dbf3bbaa7577")
```
cquiroz: Why the specific UUID?

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -125,15 +129,37 @@ package object server {
...
-      statuses.find(_.isRunning).orElse(statuses.find(_.isError)).orElse(statuses.find(_.isStopped))
-        .orElse(statuses.find(_.isIdle)).getOrElse(SequenceState.Completed)
+    if(d === 0 || idx<0) l
+    else {
+      val (h, t) = l.filterNot(_ === e).splitAt(idx+d)
+      (h :+ e) ++ t
     }
   }
 
+  implicit class ExecutionQueueOps(q: ExecutionQueue) {
```
cquiroz: Can it be a value class? `extends AnyVal`?

modules/seqexec/model/src/main/scala/seqexec/model/SequencesBatch.scala
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.model
+
+import cats.Eq
+
+object SequencesBatch {
+  sealed trait CommandState extends Product
```
tpolecat: `Product with Serializable`? The point of doing this is to avoid weird LUBs among the cases. Why not also for `State` below?

tpolecat: Should this stuff be in the `enum` package since they're enumerated types?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -485,6 +473,31 @@ object SeqexecEngine extends SeqexecConfiguration {
     )
   }
 
+  def nextRunnableObservations(qid: QueueId)(st: EngineState): Set[Observation.Id] = {
+
+    def findRunnables(used: Set[Resource], obs: Map[Observation.Id, Set[Resource]], acc: Set[Observation.Id]): Set[Observation.Id] = {
```
tpolecat: Can you add a comment about what this does? Are you sure it can't be simplified?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -485,6 +473,31 @@ object SeqexecEngine extends SeqexecConfiguration {
...
+        }
+      }
+    }
+
+    // Get sequence ids, resources and states
+    val seqInfos = st.sequences.map { case (id, ObserverSequence(_, seq)) => id -> ((seq.resources, st.executionState.sequences.get(id))) }.collect { case (id, (res, Some(s))) => id -> ((res, s.status)) }
+    // Set of resources used by running sequences
+    val used = seqInfos.collect { case (_, (res, status)) if (status.isRunning) => res }.foldRight(Set[Resource]())(_.union(_))
+    // resources for observations in queue that are not yet run
+    val obs = st.queues.get(qid).map(_.queue.map(id => id -> seqInfos.get(id)).collect {
```
tpolecat: `queue.map(id => id -> seqInfos.get(id))` can be written `queue.fproduct(seqInfos.get)`

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -485,6 +473,31 @@ object SeqexecEngine extends SeqexecConfiguration {
...
+      case (id, Some((res, status))) if (!status.isRunning && !status.isCompleted) => id -> res
+    }.toMap).getOrElse(Map.empty)
```
tpolecat: `.getOrElse(Map.empty)` can be written `.orEmpty`

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -125,15 +129,37 @@ package object server {
       ab.fold(a => Validated.Invalid(NonEmptyList.of(a)), b => Validated.Valid(b))
   }
 
-  implicit class ExecutionQueueOps(q: ExecutionQueue) {
-    def status(st: EngineState): SequenceState = {
-      val statuses: List[SequenceState] = q.map(st.executionState.sequences.get(_).map(_.status).getOrElse(SequenceState.Idle))
+  // This assumes that there is only one instance of e in l
+  private def moveElement[T](l: List[T], e: T, d: Int)(implicit eq: Eq[T]): List[T] = {
```
tpolecat: What does this method do?

modules/seqexec/server/src/main/scala/seqexec/server/ExecutionQueue.scala
jluhrs: It would be redundant. `QueueId` is the key to the map of `ExecutionQueue`s

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
jluhrs: Because it is a predefined queue. I thought it would be easier for the client implementation if the id of the calibration queue was known.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: That is not necessary. The element will be moved to the first or last position if `d` is out of range.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
jluhrs: Yes. I'll change it.

modules/seqexec/model/src/main/scala/seqexec/model/SequencesBatch.scala
jluhrs: To be completely honest, I don't know what `Product` is. I only put it there because otherwise the code would not compile.

jluhrs: @tpolecat You mean, to move it to seqexec.model.enum ?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: I guess so. One obvious optimization I see is that `findRunnables` only needs to check the obs that have not being checked yet.

swalker2m: Looks like it could be done with a fold over the `Map`.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
jluhrs: It moves the element `e` `d` positions up or down inside the list `l`.

tpolecat: Sorry, I mean, can you add a comment explaining what it does?

PR #648, opened 2018-09-14
SEQNG-700 Force AC park if instrument is in port 1.

jluhrs:
This is a small PR that forces the AC pickup mirror to be parked in the AG unit if the instrument in use is in port 1 (bottom port). It only has an effect if the TCS is configured for that step. That means, it will be used for GCAL calibrations or on sky observations, but ignored for Biases or Darks.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
@@ -3,7 +3,7 @@
 
 package seqexec.server.tcs
 
-import seqexec.server.tcs.TcsController._
+import seqexec.server.tcs.TcsController.{HrwfsConfig, _}
 import seqexec.model.enum.Instrument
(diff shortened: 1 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -389,12 +389,21 @@ object TcsControllerEpics extends TcsController {
       case Subsystem.ScienceFold => setScienceFoldPosition(tcs.agc.sfPos)
     }
 
-    subsystems.tail.foldLeft(configSubsystem(subsystems.head))((b, a) => b *> configSubsystem(a)) *>
+    val overrideHRProbe: Boolean = tcs.agc.sfPos.flatMap{
```
cquiroz: I think this could be simplified with `collect`

jluhrs: I would agree, if it weren't for `getInstPort` returning an `Option`.

PR #649, opened 2018-09-25
SEQNG-717: Add a tab for the cal queue

cquiroz:
This PR adds a tab to show the calibration queue:

seqexec - daycal queue 2018-09-25 13-02-01

The PR is fairly big because:
* This is a largish change on the model as it was previously assumed that each tab had a sequence
* Several files containing more than one clasess were split 
* Code has been reformatted using scalafmt

This PR also contains a fix for SEQNG-721 which would make some rows unrendered:
seqexec - gs-2018b-q-0-6 2018-09-11 14-45-17

We should consider whether to put this on a separate branch to avoid shipping it to production before it is ready. it would make merges a lot harder though


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/circuit/SeqexecCircuit.scala
```diff
@@ -95,49 +95,63 @@ object SeqexecCircuit extends Circuit[SeqexecAppRootModel] with ReactConnector[S
...
-      case (s, (log, SequencesOnDisplay(sequences))) => sequences.withFocus.map{
-        case (tab, active) => SequenceTabContentFocus(s.isLogged, tab.instrument, tab.sequence.map(_.id), active, log)
-      }.toNel
+    this.zoomL(SeqexecAppRootModel.logDisplayL)
+
+  val tabsReader: ModelR[SeqexecAppRootModel, TabFocus] = {
+    val getter = SeqexecAppRootModel.uiModel composeGetter (SeqexecUIModel.sequencesOnDisplay composeGetter SequencesOnDisplay.availableTabsG).zip(SeqexecUIModel.defaultObserverG)
+    val constructor = ClientStatus.canOperateG.zip(getter) >>> { p =>
+      val (o, (t, ob)) = p
+      TabFocus(o, t, ob)
```
tpolecat: I prefer `{ case (o, (t, ob)) => ...` in cases like this.

modules/seqexec/web/shared/src/main/scala/seqexec/web/common/Zipper.scala
```diff
@@ -3,57 +3,87 @@
...
 /**
- * Minimal zipper based on scalaz's implementation
- * This is only meant for small collections. performance has not been optimized
- */
+  * Minimal zipper based on scalaz's implementation
+  * This is only meant for small collections. performance has not been optimized
+  */
+@Lenses
 final case class Zipper[A](lefts: List[A], focus: A, rights: List[A]) {
+
```
tpolecat: We should be able to use `cats.collection.Zipper` pretty soon.

cquiroz:
yes I'm using scalafmt on the client side. The formatting is a bit odd but I really want to get out of the business of fiddling with formatting. This is my conf:

```
   1   │ style = default
   2   │
   3   │ maxColumn = 80
   4   │
   5   │ // Vertical alignment is pretty, but leads to bigger diffs
   6   │ align = most
   7   │
   8   │ rewrite.rules = [
   9   │   AvoidInfix
  10   │   RedundantBraces
  11   │   RedundantParens
  12   │   AsciiSortImports
  13   │   PreferCurlyFors
  14   │ ]
  15   │
  16   │ align.tokens.add = [{code = "=>", owner = "Case"}]
  17   │ align.tokens.add = [{code = ":", owner = "Term.Param"}, "=", "shouldBe"]
  18   │ spaces.inImportCurlyBraces = true
```

modules/seqexec/web/shared/src/main/scala/seqexec/web/common/Zipper.scala
cquiroz: ok, this is fairly simple so we may not need more for this use case

cquiroz:
I agree with havig a design document, commenting seems a bit less important as the ui side changes a lot, which is reflected on the size of the PRs

In general UI code is fairly hard to understand just looking at the code. and even worse if you are less familiar with the underlying tech

modules/seqexec/web/client/src/test/scala/seqexec/web/client/model/ModelSpec.scala
```diff
@@ -38,6 +39,11 @@ final class ModelSpec extends CatsSuite with ArbitrariesWebClient {
...
   checkAll("Lens[SequenceTab, Option[Int]]", LensTests(SequenceTab.stepConfigL))
-  checkAll("Lens[SequenceTab, RefTo[Option[SequenceView]]]", LensTests(SequenceTab.currentSequenceL))
-  checkAll("Optional[SequenceTab, SequenceView]", OptionalTests(SequenceTab.completedSequenceO))
+
+  // Property generation errors
+  // checkAll("SeqexecTab.previewTab", PrismTests(SeqexecTab.previewTab))
+  // checkAll("SeqexecTab.instrumentTab", PrismTests(SeqexecTab.instrumentTab))
+  // checkAll("SeqexecTab.sequenceTab", PrismTests(SeqexecTab.sequenceTab))
+  // checkAll("SequencesOn.focusSequence", OptionalTests(SequencesOnDisplay.focusSequence))
+  // checkAll("SequencesOnDisplay.previewTab", TraversalTests(SequencesOnDisplay.previewTab))
```
jluhrs: Why are these tests included, but commented?

cquiroz: because they fail due to a `scalacheck` issue which i hope we can fix at some moment in time

PR #650, opened 2018-09-25
Lib updates

cquiroz:
Some library updates

@tpolecat is the circe update ok?


tpolecat:
It's a breaking update but we'll see.

PR #652, opened 2018-09-26
SEQNG-561: Add calibration queue to the global state for the client

cquiroz:
This updates the client model to carry the calibration queue to the client. I also did some work on tests to make them faster and did some formatting

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
@@ -1,7 +1,7 @@
 // Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
 // For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-package seqexec.server
+package seqexec.model
 
 import cats.Eq
 import cats.implicits._
--- a/modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
@@ -10,11 +10,15 @@ import monocle.macros.Lenses
 import seqexec.model.enum.BatchCommandState
 
 @Lenses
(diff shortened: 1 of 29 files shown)


modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
 // For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-package seqexec.server
+package seqexec.model
```
cquiroz: Note I moved this to model

modules/seqexec/model/src/main/scala/seqexec/model/SequencesQueue.scala
```diff
@@ -16,18 +16,20 @@ import seqexec.model.enum.Instrument
...
 @Lenses
 final case class SequencesQueue[T](
-  loaded:     Map[Instrument, Observation.Id],
-  conditions: Conditions,
-  operator:   Option[Operator],
-  queue:      List[T]
+  loaded:       Map[Instrument, Observation.Id],
+  conditions:   Conditions,
+  operator:     Option[Operator],
+  queues:       Map[QueueId, ExecutionQueue],
```
cquiroz: This is the main change in this PR, we take the queues to the client

modules/seqexec/web/client/src/main/scala/seqexec/web/client/services/SeqexecWebClient.scala
```diff
@@ -33,17 +52,12 @@ object SeqexecWebClient extends ModelBooPicklers {
     Unpickle[A].fromBytes(ab)
   }
 
-  def sync(id: Observation.Id): Future[SequencesQueue[Observation.Id]] =
-    Ajax.get(
+  def sync(id: Observation.Id): Future[String] =
```
cquiroz: This PR showed me we have an issue on the way requests are decoded. I'll work on that in another PR

PR #653, opened 2018-09-26
SEQNG-724: Extracted methods from GPIClient

sraaphorst:
Some methods currently implemented in `GPIClient` will be used by all clients of thhe `GIAPI`, such as the up-and-coming `GHOSTClient`, so this is a refactor to make a trait implementing those common methods.


modules/giapi/src/main/scala/giapi/client/GiapiClient.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def genericApply(configuration: Configuration): F[CommandResult] =
+    giapi.command(
+      Command(
+        SequenceCommand.APPLY,
+        Activity.PRESET_START,
+        configuration
+      ), GiapiClient.DefaultCommandTimeout)
+
+    def observe[A: Show](dataLabel: A, expTime: FiniteDuration): F[CommandResult] =
```
cquiroz: This method is missagligned

sraaphorst: Oooops. Okay, pushing another commit. Sorry!

cquiroz:
We should move all the standard methods `test`, `verify`, etc tothe same trait

sraaphorst:
@cquiroz Excellent. I figured there would be more, but I wasn't sure which ones would be standard.
Tomorrow if you are  around, we can discuss.

jluhrs:
In other words, everything under "General commands" in GPIClient.scala should be moved to GiapiClient.scala.

sraaphorst:
@cquiroz and @jluhrs: I've moved all the common commands from `GPIClient` to `GiapiClient`.

PR #655, opened 2018-09-27
SEQNG-596: Enable minification of js

cquiroz:
With the update to uglyfyJS to 2.x i can restor minification of js to get a nice reduction on the final size


cquiroz:
Nice reduction of :  -813,571 bytes (-34.29%)

PR #657, opened 2018-09-28
Run calibration queue.

jluhrs:
Running the calibration queue means that at any point in time after the user commanded to run the queue, if any sequence in the queue can be run, it must be running.
When the command arrives, every sequence in the queue that can be run (minding resource usage) is started. Then, whenever a sequence finishes and frees resources, the check is repeated.

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -232,7 +232,7 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
     * Main logical thread to handle events and produce output.
     */
   @SuppressWarnings(Array("org.wartremover.warts.ImplicitParameter"))
-  private def run(ev: EventType)(implicit ec: ExecutionContext): HandleType[ResultType] = {
+  private def run(userReact: PartialFunction[SystemEvent, HandleType[Unit]])(ev: EventType)(implicit ec: ExecutionContext): HandleType[ResultType] = {
     def handleUserEvent(ue: UserEventType): HandleType[ResultType] = ue match {
       case Start(id, _, clid, userCheck) => Logger.debug(s"Engine: Start requested for sequence ${id.format}") *> start(id, clid, userCheck) *> pure(UserCommandResponse(ue, EventResult.Ok, None))
       case Pause(id, _)                  => Logger.debug(s"Engine: Pause requested for sequence ${id.format}") *> pause(id) *> pure(UserCommandResponse(ue, EventResult.Ok, None))
@@ -267,7 +267,7 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
 
     (ev match {
       case EventUser(ue)   => handleUserEvent(ue)
-      case EventSystem(se) => handleSystemEvent(se)
+      case EventSystem(se) => handleSystemEvent(se).flatMap(x => userReact.applyOrElse(se, (_:SystemEvent) => unit).map(_ => x))
     })
   }
 
@@ -289,15 +289,15 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
   }
 
   @SuppressWarnings(Array("org.wartremover.warts.AnyVal", "org.wartremover.warts.ImplicitParameter"))
-  private def runE(ev: EventType, s: D)(implicit ec: ExecutionContext): IO[(D, (ResultType, D), Stream[IO, EventType])] =
-    run(ev).run.run(s).map {
+  private def runE(userReact: PartialFunction[SystemEvent, HandleType[Unit]])(ev: EventType, s: D)(implicit ec: ExecutionContext): IO[(D, (ResultType, D), Stream[IO, EventType])] =
+    run(userReact)(ev).run.run(s).map {
       case (si, (r, p)) =>
         (si, (r, si), p.getOrElse(Stream.empty))
     }
 
   @SuppressWarnings(Array("org.wartremover.warts.AnyVal", "org.wartremover.warts.ImplicitParameter"))
-  def process(input: Stream[IO, EventType])(qs: D)(implicit ec: ExecutionContext): Stream[IO, (ResultType, D)] =
-    mapEvalState[EventType, D, (ResultType, D)](input, qs, runE)
+  def process(userReact: PartialFunction[SystemEvent, HandleType[Unit]])(input: Stream[IO, EventType])(qs: D)(implicit ec: ExecutionContext): Stream[IO, (ResultType, D)] =
+    mapEvalState[EventType, D, (ResultType, D)](input, qs, runE(userReact))
 
   //private def userModify(f: D => (D, U)): HandleType[U] = StateT[IO, D, U]( st => IO(f(st)) ).toHandleP
 
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -21,6 +21,7 @@ import org.log4s.getLogger
 import scala.concurrent.ExecutionContext
 
 class Engine[D, U](stateL: Lens[D, Engine.State]) {
+  import Engine._
 
   class ConcreteTypes extends Engine.Types {
     override type StateType = D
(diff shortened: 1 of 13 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/UserEvent.scala
```diff
@@ -33,7 +33,7 @@ final case class GetState[D<:Engine.Types](f: D#StateType => Option[Stream[IO, E
 }
 // Generic event to put a function in the main Process process, which changes the state
 // depending on the current state
-final case class ModifyState[D<:Engine.Types](f: D#StateType => (D#StateType, D#EventData)) extends UserEvent[D] {
+final case class ModifyState[D<:Engine.Types](f: Engine.HandleP[D#StateType, Event[D], D#EventData]) extends UserEvent[D] {
```
jluhrs: Changing `f` to a `HandleP` means that `f` now can produce a feedback `Stream`. This allows, for example, to start a sequence inside `f`.

modules/seqexec/model/src/main/scala/seqexec/model/enum/BatchCommandState.scala
```diff
@@ -4,12 +4,13 @@
...
 
 import cats.Eq
+import seqexec.model.ClientID
 
 sealed trait BatchCommandState extends Product with Serializable
 
 object BatchCommandState {
   case object Idle extends BatchCommandState
-  case object Run extends BatchCommandState
+  final case class Run(clientId: ClientID) extends BatchCommandState
```
jluhrs: `clientId` was needed to start the sequences.

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -29,86 +30,16 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
...
-        }
-
-      // Done
-      HandleP(st)
-    }
-  }
-
-  implicit class HandleToHandleP[A](self: Handle[A]) {
-    def toHandleP: HandleP[A] = HandleP(self.map((_, None)))
-  }
```
jluhrs: Moved `HandleP` to the companion object to make it accessible to other classes. Should it be in its own source file?

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -278,36 +204,36 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
...
-  private def send(ev: EventType): HandleP[Unit] = HandleP.fromStream(Stream(ev))
+  private def send(ev: EventType): HandleType[Unit] = HandleP.fromStream(Stream(ev))
 
   /**
     * Main logical thread to handle events and produce output.
     */
   @SuppressWarnings(Array("org.wartremover.warts.ImplicitParameter"))
-  private def run(ev: EventType)(implicit ec: ExecutionContext): HandleP[ResultType] = {
-    def handleUserEvent(ue: UserEventType): HandleP[ResultType] = ue match {
+  private def run(userReact: PartialFunction[SystemEvent, HandleType[Unit]])(ev: EventType)(implicit ec: ExecutionContext): HandleType[ResultType] = {
```
jluhrs: Now the user of `Engine` can pass a function to react to system events inside the engine loop.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -653,6 +683,7 @@ object SeqexecEngine extends SeqexecConfiguration {
     case LoadSequence(id)              => SequenceLoaded(id, svs)
     case UnloadSequence(id)            => SequenceUnloaded(id, svs)
     case NotifyUser(m, cid)            => UserNotification(m, cid)
+    case StartQueue(_, _)              => NullEvent
```
cquiroz: No client event?

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
cquiroz: probably. our source files tend to be too large

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -29,86 +30,16 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
   type EventType = Event[ConcreteTypes]
   type ResultType = EventResult[ConcreteTypes]
   type UserEventType = UserEvent[ConcreteTypes]
+  type HandleType[A] = HandleP[D, EventType, A]
```
cquiroz: Should we consider not using type aliases? They tend to make the code less clear sometimes

tpolecat: 💯 kill ~~all~~ most type aliases

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -424,4 +350,74 @@ object Engine {
     type EventData
   }
 
+
+  /**
+    * Type constructor where all Seqexec side effect are managed.
+    *
+    * It's named `Handle` after `fs2.Handle` in order to give a hint in a future
```
cquiroz: is there really a `fs2.Handle`?

tpolecat: Not anymore, anyway.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: I leave you the honor.

cquiroz:
Could you a descripton to the PR?

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -341,7 +267,7 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
 
     (ev match {
       case EventUser(ue)   => handleUserEvent(ue)
-      case EventSystem(se) => handleSystemEvent(se)
+      case EventSystem(se) => handleSystemEvent(se).flatMap(x => userReact.applyOrElse(se, (_:SystemEvent) => unit).map(_ => x))
```
cquiroz: How is `userReact` used in practice?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -198,6 +198,9 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
     }
   }
 
+  private[server] def stream(p: Stream[IO, executeEngine.EventType])(s0: EngineState): Stream[IO, (executeEngine.ResultType, EngineState)] = executeEngine.process(iterateQueues)(p)(s0)
```
cquiroz: Could you split this to show the implementation in a separate line?

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
jluhrs: It it is true that sometimes they make the code less clear, but I would argue that the opposite is also true. If I were to remove all type aliases, I would have to replace `HandleType[EventType]` by `StateT[IO, D, (EventResult[ConcreteTypes], Stream[IO, Event[ConcreteTypes]]` all over Engine.scala.
I think that the difference resides in the details that the type alias hides. If they are important to understand the code, then hiding them under a type alias makes the code less clear. If they are  not that relevant, then they become a distraction, and hiding them can make easier to understand the code.
For example, every instance of `StateT` in Engine.scala uses the same state type and the IO monad. The concrete state type is not very relevant, because the access to its attributes is done with Lenses, and IO is only relevant when related to Streams.

cquiroz: i guess it could be done with intermetidate types like
`StateT[IO, D, Results]` where results is a class with a member `(EventResult[ConcreteTypes], Stream[IO, Event[ConcreteTypes]])`

I don't know, perhaps all that should be in a separate PR though

modules/seqexec/model/src/main/scala/seqexec/model/enum/BatchCommandState.scala
cquiroz: this change is breaking the build as boopickle serialization is broken

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -250,9 +254,30 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
+          case BatchExecState.Idle     => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Run(clientId)) >>> {(_, ())}).toHandle *> runQueue(qid, clientId)
+          case BatchExecState.Stopping => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Run(clientId)) >>> {(_, ())}).toHandle
+          case _                       => executeEngine.unit
+        }
+      }.getOrElse(executeEngine.unit)
+    }}.map(_ => StartQueue(qid, clientId)))
   ).map(_.asRight)
 
+  // It assumes only one queue can run at a time
+  private val iterateQueues: PartialFunction[SystemEvent, executeEngine.HandleType[Unit]] = {
```
jluhrs: This is the function that reacts to sequence completion, to start the next sequences in the queue.

PR #658, opened 2018-09-28
Remove legacy code

cquiroz:
There is some legacy code from the time we could send commands in a terminal. This PR removes it and cleans up some code and adds reformatting where appropriate


jluhrs:
Ah, yes, I remember that time, when we were younger and full of optimism, and thought Seqexec would be done in a year :)

cquiroz:
young?

PR #662, opened 2018-10-01
SEQNG-726 Function to add multiples sequences to calibration queue

jluhrs:


cquiroz:
do you also want to add the http endpoint?

jluhrs:
The entry point will be added in SEQNG-728

PR #665, opened 2018-10-01
Refactoring of sequence extracting lenses

cquiroz:
I've been trying to simplify the client side model removing some shorcuts and relying more on lenses as possible

This in one of those refactorings to remove a kludge to avoid `Option` in certain cases, but it was kind of cheating

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/toolbars/SequenceDefaultToolbars.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/toolbars/SequenceDefaultToolbars.scala
@@ -3,7 +3,6 @@
 
 package seqexec.web.client.components.sequence.toolbars
 
-import diode.react.ModelProxy
 import diode.react.ReactConnectProxy
 import cats.implicits._
 import japgolly.scalajs.react.vdom.html_<^._
@@ -247,8 +246,8 @@ object SequenceControl {
       }
       .build
 
-  def apply(p: ModelProxy[SequenceControlFocus]): Unmounted[Props, State, Unit] =
-    component(Props(p()))
+  def apply(p: Props): Unmounted[Props, State, Unit] =
+    component(p)
 }
 
 /**
@@ -256,9 +255,9 @@ object SequenceControl {
   */
 object SequenceDefaultToolbar {
   final case class Props(id: Observation.Id) {
-    val observerReader: ReactConnectProxy[SequenceInfoFocus] =
+    val observerReader: ReactConnectProxy[Option[SequenceInfoFocus]] =
       SeqexecCircuit.connect(SeqexecCircuit.sequenceObserverReader(id))
-    val controlReader: ReactConnectProxy[SequenceControlFocus] =
+    val controlReader: ReactConnectProxy[Option[SequenceControlFocus]] =
       SeqexecCircuit.connect(SeqexecCircuit.sequenceControlReader(id))
   }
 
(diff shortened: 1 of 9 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/toolbars/SequenceDefaultToolbars.scala
```diff
@@ -273,12 +272,19 @@ object SequenceDefaultToolbar {
           SeqexecStyles.shorterRow,
           <.div(
             ^.cls := "ui left floated column eight wide computer eight wide tablet only",
-            p.controlReader(SequenceControl.apply)
+            p.controlReader(_() match {
+                case Some(c) => SequenceControl(SequenceControl.Props(c))
```
cquiroz: This was avoided on the previous model where we would get the value without `Option`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
```diff
@@ -395,17 +395,13 @@ object SequencesOnDisplay {
   def loadingL(id: Observation.Id): Traversal[SequencesOnDisplay, Boolean] =
     SequencesOnDisplay.previewTabById(id) ^|-> PreviewSequenceTab.isLoading
 
-  def tabG(id: Observation.Id): Getter[SequencesOnDisplay, SeqexecTabActive] =
+  def tabG(id: Observation.Id): Getter[SequencesOnDisplay, Option[SeqexecTabActive]] =
```
cquiroz: This is the core change. I cheated a bit by defining an `EmptySeqexecTab` but it's probably better to pass the `Option` up the stack

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/circuit/SeqexecCircuit.scala
```diff
@@ -127,43 +127,31 @@ object SeqexecCircuit extends Circuit[SeqexecAppRootModel] with ReactConnector[S
...
+    this.zoomG(SequenceInfoFocus.sequenceInfoG(id))
 
   def statusAndStepReader(id: Observation.Id): ModelR[SeqexecAppRootModel, Option[StatusAndStepFocus]] =
-    statusReader.zip(sequenceTab(id)).zoom {
-      case (status, SeqexecTabActive(tab, _)) =>
-        tab.sequence.map { t =>
-          StatusAndStepFocus(status.isLogged, t.metadata.instrument, t.id, tab.stepConfigDisplayed, t.steps.length, tab.isPreview)
-        }
-    }
+    this.zoomG(StatusAndStepFocus.statusAndStepG(id))
```
cquiroz: i'm aiming to remove diode's `ModalR` which is kind of a lens with proper Monocle lenses

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/tabs.scala
```diff
@@ -176,11 +191,11 @@ final case class PreviewSequenceTab(currentSequence: SequenceView,
 @SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
 object PreviewSequenceTab {
   implicit val eq: Eq[PreviewSequenceTab] =
-    Eq.by(x => (x.currentSequence, x.stepConfig, x.isLoading, x.tableState, x.tabOperations))
-}
-
-final case object EmptySequenceTab extends SequenceTab {
```
cquiroz: Let's pretend this class didn't happen

PR #667, opened 2018-10-02
SEQNG-729 Implemented stopQueue function.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -299,6 +299,22 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
     }}.map(_ => StartQueue(qid, clientId)))
   ).map(_.asRight)
 
+  private def stopSequencesInQueue(qid: QueueId): executeEngine.HandleType[Unit] =
+    executeEngine.get.map(st => (EngineState.queues ^|-? index(qid)).getOption(st).map(_.queue.filter(sid => (EngineState.executionState ^|-> Engine.State.sequences ^|-? index(sid)).getOption(st).map(_.status.isRunning).getOrElse(false))).getOrElse(List.empty)).flatMap(_.map(executeEngine.pause).fold(executeEngine.unit)(_ *> _))
```
cquiroz: Could you reorg this line a bit, to make it a bit more vertical
You can also replace `map().getOrElse(Nil)` with `foldMap` I think

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -299,6 +299,22 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
+
+
+  def stopQueue(q: EventQueue, qid: QueueId, clientId: ClientID): IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
+    Event.modifyState[executeEngine.ConcreteTypes](executeEngine.get.flatMap{ st =>
+      (EngineState.queues ^|-? index(qid)).getOption(st).map {
+        _.status(st) match {
+          case BatchExecState.Running => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Stop) >>> {(_, ())}).toHandle *> stopSequencesInQueue(qid)
+          case BatchExecState.Waiting => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Stop) >>> {(_, ())}).toHandle
```
cquiroz: What does it mean for a queue to be `Waiting`?

jluhrs: The user commanded it to start, but no sequence is running because the required resources are in use. That means that there is no need to stop sequences, only to mark the queue.

cquiroz: 👍

PR #671, opened 2018-10-03
SEQNG-676: Adding GHOST configuration.

sraaphorst:
Here are some changes I made to try to incorporate the GHOST configuration.

I'm not sure if this only comprises SEQNG-676 or another task as well.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -39,10 +39,12 @@ import edu.gemini.spModel.core.{Peer, SPProgramID}
 import edu.gemini.spModel.obscomp.InstConstants
 import edu.gemini.spModel.seqcomp.SeqConfigNames.OCS_KEY
 import fs2.{Scheduler, Stream}
+import giapi.client.ghost.GHOSTClient
 import org.http4s.client.Client
 import org.http4s.Uri
 import knobs.Config
 import mouse.all._
+import seqexec.server.ghost.GHOSTController
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
@@ -56,6 +58,8 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
 
   val gpiGDS: GDSClient = GDSClient(settings.gpiGdsControl.command.fold(httpClient, GDSClient.alwaysOkClient), settings.gpiGDS)
 
+  val ghostGDS: GDSClient = GDSClient(settings.ghostControl.command.fold(httpClient, GDSClient.alwaysOkClient), settings.ghostGDS)
+
   private val systems = SeqTranslate.Systems(
     odbProxy,
     settings.dhsControl.command.fold(DhsClientHttp(settings.dhsURI), DhsClientSim(settings.date)),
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTHeader.scala
```diff
@@ -0,0 +1,28 @@
...
+import gem.Observation
+import cats.effect.IO
+import seqexec.model.dhs.ImageFileId
+import seqexec.server.{InstrumentSystem, SeqAction}
+import seqexec.server.keywords._
+import seqexec.server.tcs.TcsKeywordsReader
+
+object GHOSTHeader {
+
+  def header(inst: InstrumentSystem[IO],
```
cquiroz: We don't need all the params if the `Header` is empty

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -457,7 +463,9 @@ object SeqexecEngine extends SeqexecConfiguration {
                             failAt: Int,
                             odbQueuePollingInterval: Duration,
                             gpiGiapi: Giapi[IO],
-                            gpiGDS: Uri)
+                            ghostGiapi: Giapi[IO],
+                            gpiGDS: Uri,
+                            ghostGDS: Uri)
   def apply(httpClient: Client[IO], settings: Settings, c: SeqexecMetrics): SeqexecEngine = new SeqexecEngine(httpClient, settings, c)
 
   // Couldn't find this on Scalaz
```
cquiroz: this comment is likely stale

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTHeader.scala
sraaphorst: Can we get rid of all of them?

cquiroz:
i think we are missing the part where `Settings` is populated
see https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala#L693

Also we need sensible defaults set in `app.conf`

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
sraaphorst: I'll delete that :-)

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTHeader.scala
sraaphorst: I'm not sure which ones are safe to get rid of at this point... we have TCS keywords, I think, and I'm not sure about ObsKeywords. Do we need a GDSClient?

jluhrs:
I think the changes ended up covering SEQNG-680

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTHeader.scala
cquiroz: We are not using any of them now. i'm surprsied WartRemover doesn't complaint. i'd remove them all and put them back when needed

sraaphorst: Sounds good to me!

sraaphorst:
@jluhrs That's what I was thinking, too... I can rename this as 680, make another PR, and then do 676.

sraaphorst:
Yes, you're right!

PR #673, opened 2018-10-03
cats effect 1.0

tpolecat:
In-progress upgrades to cats-effect 1.0 which includes updates to fs2, http4s, and doobie. In a few places I'm generalizing things to `F[_]` where we had hardcoded `IO` because it's possible to do so more easily now; and I'm replacing single-element `Stream.bracket` with `Resource`, which is a really nice simplification. Anyway this is going to be a big one, sorry.

No need to review yet unless you're just curious. I just wanted to be able to watch the diff.

Todo:
- [x] track down the problem with the port change in `TelnetdConfiguration.scala`
- [x] figure out empty responses in the gem webserver
- [x] live tests of seqexec
- [x] fill in `null` placeholder in `WebServerLauncher.scala`
- [x] figure out why slf4j-simple was needed


modules/db/src/main/scala/gem/dao/DatasetDao.scala
```diff
@@ -31,7 +31,7 @@ object DatasetDao {
     private final case class StepId(toInt: Int)
     private object StepId {
       implicit val StepIdMeta: Meta[StepId] =
-        Distinct.integer("id_index").xmap(StepId(_), _.toInt)
+        Distinct.integer("id_index").timap(StepId(_))(_.toInt)
```
cquiroz: `timap`?

cquiroz:
Wow, this is huge but great you're working on this.

Where can I learn more about `Resource`?

modules/db/src/main/scala/gem/dao/DatasetDao.scala
tpolecat: It's a doobie thing. Identical to `imap` but asks for a type tag for `B`, which doobie can use in error messages.

tpolecat:
> Where can I learn more about Resource?

https://typelevel.org/cats-effect/datatypes/resource.html

It's really good.

swalker2m:
> https://typelevel.org/cats-effect/datatypes/resource.html

It seems to be written from the perspective of someone who already knows what it is.

tpolecat:
Ok this compiles and the tests pass, but it doesn't run because we're waiting on a knobs release. But I think we should go ahead and start to review these changes because there's a lot of churn right now and the last rebase was a challenge.

cquiroz:
What about the scala update. How is that you got the client to compile?

project/Settings.scala
```diff
@@ -58,7 +58,8 @@ object Settings {
       "-Ywarn-unused:patvars",             // Warn if a variable bound in a pattern is unused.
       "-Ywarn-unused:privates",            // Warn if a private member is unused.
       "-Ywarn-value-discard",              // Warn when non-Unit expression results are unused.
-      "-Ybackend-parallelism", JRuntime.getRuntime.availableProcessors.toString // Run some tasks in parallel
+      "-Ybackend-parallelism", JRuntime.getRuntime.availableProcessors.toString, // Run some tasks in parallel
+      "-Yrangepos"
```
cquiroz: 👍 That gives nicer errors with LSP

project/Settings.scala
```diff
@@ -166,6 +167,7 @@ object Settings {
     val UnboundId              = "com.unboundid"             %   "unboundid-ldapsdk-minimal-edition" % LibraryVersions.unboundId
     val JwtCore                = "com.pauldijou"             %%  "jwt-core"                          % LibraryVersions.jwt
     val Slf4j                  = "org.slf4j"                 %   "slf4j-api"                         % LibraryVersions.slf4j
+    val Slf4jSimple            = "org.slf4j"                 %   "slf4j-simple"                      % LibraryVersions.slf4j
```
cquiroz: Why this?

tpolecat:
> What about the scala update. How is that you got the client to compile?

I changed back to 2.12.6 after you warned me.

project/Settings.scala
tpolecat: Yeah it's a big help.

project/Settings.scala
tpolecat: http4s logging was getting discarded because there was no log handler on the classpath. We can swap in a better one if we want to, I just went for the easy one.

project/Settings.scala
cquiroz: This settings adds like 10% overhad on compilation on 2.12.6, we may need to wait for 2.12.8. WDYT?

tpolecat: Honestly I'm ok with that because it makes error reporting so much better, but we can turn it back off if it's noticeable for you.

cquiroz: Mayb there is a fix by the time we can merge this PR

project/Settings.scala
cquiroz: This is a bit strange, we do have logging from http4s at the moment. I wonder what changed

modules/telnetd/src/main/scala/TelnetdConfiguration.scala
```diff
@@ -8,6 +8,6 @@ final case class TelnetdConfiguration(port: Int)
 object TelnetdConfiguration {
 
   val forTesting: TelnetdConfiguration =
-    apply(9091)
+    apply(9092)
```
cquiroz: Why the port chang?

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -187,33 +229,38 @@ object WebServerLauncher extends StreamApp[IO] with LogInitialization with Seqex
...
-        out    <- Stream.eval(async.topic[IO, SeqexecEvent](NullEvent))
-        pe     <- Stream.eval(PrometheusExportService.build[IO])
-        engine <- Stream.eval(engineIO(cli, pe.collectorRegistry))
-        web    <- Stream.eval(webServerIO(inq, out, engine, pe))
-        exit   <- Stream(
-                    engine.eventStream(inq).to(out.publish),
-                    web
-                  ).join(2).drain ++ Stream.emit(ExitCode.Success)
-      } yield exit
-    }
```
cquiroz: `null`?

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -187,33 +229,38 @@ object WebServerLauncher extends StreamApp[IO] with LogInitialization with Seqex
...
+        cli    <- BlazeClientBuilder[IO](null).resource
```
cquiroz: `null`?

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
```diff
@@ -127,10 +129,10 @@ class SeqexecUIApiRoutes(site: String, devMode: Boolean, auth: AuthenticationSer
         // Create a client specific process
 
         def initialEvent(clientId: ClientId): Stream[IO, WebSocketFrame] =
-          Stream.emit(Binary(trimmedArray(ConnectionOpenEvent(user.toOption, clientId): SeqexecEvent)))
+          Stream.emit(Binary(ByteVector(trimmedArray(ConnectionOpenEvent(user.toOption, clientId): SeqexecEvent))))
```
cquiroz: Why this?

modules/telnetd/src/main/scala/TelnetdConfiguration.scala
tpolecat: For some reason on my machine 9091 didn't work and I have not yet figured out why. Any other port seems ok. I'll add a checklist above.

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
tpolecat: Thanks, I needed to pass an EC there and stubbed it out. I think that explains another issue I'm seeing.

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/SeqexecUIApiRoutes.scala
tpolecat: `Binary` wants a `ByteVector` now.

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
tpolecat: Thanks, I needed to pass an EC there and stubbed it out. I think that explains another issue I'm seeing.

tpolecat:
> I don't see big problems but given our history with fs2 and the seqexec being in production I think we need to put this into a round of testing before merging

Yeah that's my thinking as well. We're blocked on the absence of a knobs release so I may do a local build and check in the jarfile temporarily.

project/Settings.scala
tpolecat: I think it was in the gem launcher, not the seqexec one. I'll take a closer look.

modules/telnetd/src/main/scala/TelnetdConfiguration.scala
tpolecat: So, the problem here and with empty responses on the webserver were due to another version of the app running locally in a docker container. Somehow docker snarfs traffic and routes it to containers that expose the ports, without actually binding the port locally. So you can bind the port and it won't work, but if you kill the container it will start working. Very gross. Anyway, reverting this change.

tpolecat: @cquiroz sorry, I deleted the comment you responded to above. I'm not changing the ports after all. So 9090 and 9091.

project/Settings.scala
tpolecat: I removed this and added the `Logging` settings you're using in seqexec.

tpolecat:
Ok I think this is done, pending live tests with GPI and Epics.

modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
```diff
@@ -134,21 +134,28 @@ object StaticConfigDao {
...
-      implicit val GmosCommonStaticComposite: Composite[GmosCommonStaticConfig] =
-        Composite[(GmosDetector, MosPreImaging)].imap(
-          (t: (GmosDetector, MosPreImaging)) => GmosCommonStaticConfig(t._1, t._2, None, Set.empty))(
+      implicit val GmosCommonStaticRead: Read[GmosCommonStaticConfig] =
+        Read[(GmosDetector, MosPreImaging)].map {
+          (t: (GmosDetector, MosPreImaging)) => GmosCommonStaticConfig(t._1, t._2, None, Set.empty)
+        }
+
+      implicit val GmosCommonStaticWrite: Write[GmosCommonStaticConfig] =
+        Write[(GmosDetector, MosPreImaging)].contramap(
```
swalker2m: So `Composite` becomes `Read` and `Write`?

modules/db/src/test/scala/gem/dao/DaoTest.scala
```diff
@@ -15,6 +15,9 @@ import scala.collection.immutable.TreeMap
 trait DaoTest extends gem.Arbitraries {
   val pid = Program.Id.fromString.unsafeGet("GS-1234A-Q-1")
 
+  private implicit val contextShift: ContextShift[IO] =
+    IO.contextShift(scala.concurrent.ExecutionContext.global)
```
swalker2m: What's happening here?

modules/db/src/main/scala/gem/dao/StaticConfigDao.scala
tpolecat: Yes, `Meta` becomes `Get` + `Put` and `Composite` becomes `Read` + `Write`. There's a description a few paragraphs down [here](https://tpolecat.github.io/doobie/changelog.html).

modules/db/src/test/scala/gem/dao/DaoTest.scala
tpolecat: `ContextShift[F]` is two things:

1. The ability to reschedule a computation via `shift` … this is kind of like `Thread.sleep()` where you want to relax and let something else have time to make progress, only in this case the work may get picked back up by another thread. `Future` does this every time you map or flatMap; `ContextShift[F]` does it only when you ask. In order to do this it needs an execution context, which it closes over. This is important for scala-js where there's a single event loop and you need to insert `shift` sometimes to ensure fairness. Less of an issue on the JVM.

2. The ability to perform a computation on *another* execution context via `evalOn`, and shift back on completion. doobie uses this to shift all the blocking JDBC primitives onto a dedicated thread pool, then shift back onto the CPU pool when the blocking call completes.

`ContextShift` is not a typeclass because a given program might choose a different execution context, but we assume that it's *unique* per program, which makes it ok to pass implicitly. Think of it as part of the runtime system.

`Timer` is a similar kind of thing, by the way. It gives you a way to "sleep" without blocking and is similarly unique for any given program so we can reasonably pass it implicitly.

`Concurrent` adds on the ability to run things in parallel via `.start` which gives you a fiber back; it's like `forkIO` in Haskell. If you have a `ContextShift[IO]` you get `Concurrent[IO]` for free.

tpolecat:
@swalker2m the only doc I know of is the Scaladoc and what's at https://typelevel.org/cats-effect and I admit it's sometimes hard to follow. If you have feedback on the doc you can open an issue, they would appreciate it.

modules/db/src/test/scala/gem/dao/DaoTest.scala
swalker2m: Thanks, that makes sense.  I wish _you_ had time and inclination to write the documentation for cats effect!

jluhrs:
Tested today with GMOS. The EPICS stuff worked OK, but we had some issues. First, the Seqexec server didn't show any of the sequences in the session queue. I had to restart it. Then we were able to run a GMOS sequence. But at some point the sequence was paused, and we tried to resume it from another computer, and the server did not started it. We tried to chose another sequence for GMOS, and again the server did not respond. That was done from two browser in two different machines.
The log file (attached) shows similar errors both times.

[app_seqexec_server_gs-20181026-dcd0cd54.log](https://github.com/gemini-hlsw/ocs3/files/2520761/app_seqexec_server_gs-20181026-dcd0cd54.log)

tpolecat:
Likely related to https://github.com/http4s/blaze/pull/238

tpolecat:
Ok this is updated to http4s 0.20.0-M1 so we need to re-test whatever was surfacing the exceptions above, and we need to re-test the Prometheus stuff, whatever that is, because the http4s API changed a bunch and I may have broken it trying to get it to compile.

If there's a way for me to do this testing from here let me know.

jluhrs:
I just tried to run commit 19c96b8. I run it with a local ODB and the web browser in the same machine. To run it, I used the command `startSeqexecAll` from inside sbt. The result was a failure. The client shows only a blank page. In the server side, I see these error messages:

[error] [HPM] Error occurred while trying to proxy request /api/seqexec/start from localhost:9090 to http://localhost:7070 (ECONNREFUSED) (https://nodejs.org/api/errors.html#errors_common_system_errors)
[error] [HPM] Error occurred while trying to proxy request /api/seqexec/site from localhost:9090 to http://localhost:7070 (ECONNREFUSED) (https://nodejs.org/api/errors.html#errors_common_system_errors)

Those errors repeat every time I reload the page in the web browser.

tpolecat:
merged

modules/seqexec/server/src/main/scala/seqexec/server/keywords/GDSClient.scala
```diff
@@ -21,9 +25,23 @@ import seqexec.server.SeqActionF
...
   * Gemini Data service client
   */
-final case class GDSClient(client: Client[IO], gdsUri: Uri)
+final case class GDSClient(base: Client[IO], gdsUri: Uri)(implicit timer: Timer[IO])
     extends Http4sClientDsl[IO] {
 
+  @SuppressWarnings(Array("org.wartremover.warts.Var"))
+  private val client = {
+    val max = 2
+    var attemptsCounter = 1
```
cquiroz: This likely needs an exception for scalastyle

tpolecat:
Ok this is back up to date with develop. What needs to happen before merging?

cquiroz:
Great, thanks I'll ask for instrument time and test ASAP
In the meanwhile we should try not to push stuff hard to merge, meaning no changes to the seqexec engine

cquiroz:
in the meanwhile I'll deploy a version with this PR for testing at GN along the current seqexec

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -193,18 +193,27 @@ class SeqexecEngine(httpClient: Client[IO], settings: Settings[IO], sm: SeqexecM
...
+  def seqQueueRefreshStream: Stream[IO, executeEngine.EventType] = {
+    val fd = Duration(settings.odbQueuePollingInterval.toSeconds, TimeUnit.SECONDS)
+    Stream.fixedDelay[IO](fd).evalMap(_ => odbProxy.queuedSequences.value).map { x =>
+      Event.getState[executeEngine.ConcreteTypes](st =>
+        x.map(odbLoader.refreshSequenceList(_)(st)).valueOr(r =>
+          List(Event.logWarningMsg(SeqexecFailure.explain(r)))
+        ).some.filter(_.nonEmpty).map(Stream.emits(_).covary[IO])
+      )
+  // def seqQueueRefreshStream: Stream[IO, executeEngine.EventType] =
+  //   Scheduler[IO](corePoolSize = 1).flatMap { scheduler =>
```
cquiroz: Do you know why we don't need the `Scheduler` anymore?

tpolecat: It was folded into cats-effect `Timer`

tpolecat:
Ok, re-merged with `develop`.

project/Settings.scala
```diff
@@ -82,16 +82,16 @@ object Settings {
...
     val mouseVersion            = "0.20"
-    val fs2Version              = "0.10.5"
+    val fs2Version              = "1.0.0"
     val shapelessVersion        = "2.3.3"
     val attoVersion             = "0.6.4"
     val scalaParsersVersion     = "1.1.1"
     val scalaXmlVerson          = "1.1.1"
 
-    val http4sVersion           = "0.18.21"
+    val http4sVersion           = "0.20.0-M3"
```
cquiroz: Why not `0.20.0-M4`

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -728,7 +745,7 @@ object SeqexecEngine extends SeqexecConfiguration {
 
     // TODO: Review initialization of EPICS systems
     @SuppressWarnings(Array("org.wartremover.warts.Throw"))
-    def initEpicsSystem[T](sys: EpicsSystem[T], tops: Map[String, String]): IO[Unit] =
+    def initEpicsSystem(sys: EpicsSystem[_], tops: Map[String, String]): IO[Unit] =
```
cquiroz: Are these equivalent?

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -295,11 +300,19 @@ object GnirsControllerEpics extends GnirsController {
       GnirsEpics.instance.abortCmd.mark *>
       GnirsEpics.instance.abortCmd.post.void
 
-  override def observeProgress(total: Time): Stream[IO, Progress] =
+  private def removePartName(s: String) = {
```
cquiroz: I think this was changed upstream. It is now at
https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala#L271

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -190,33 +229,38 @@ object WebServerLauncher extends StreamApp[IO] with LogInitialization with Seqex
...
+        inq    <- Resource.liftF(Queue.bounded[IO, executeEngine.EventType](10))
+        out    <- Resource.liftF(Topic[IO, SeqexecEvent](NullEvent))
+        cr     <- Resource.liftF(IO(new CollectorRegistry))
+        engine <- Resource.liftF(engineIO(cli, cr))
+        bec    <- blockingExecutionContext
+        _      <- webServerIO(inq, out, engine, cr, bec)
+        _      <- Resource.liftF(engine.eventStream(inq).to(out.publish).compile.drain.start)
+      } yield ExitCode.Success
+
+    r.use(_ => IO.never)
```
cquiroz: What does this do?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
tpolecat: Yes. If the type parameter isn't mentioned again it can be existential. I prefer it this way because it's instantly clear that the type argument is irrelevant.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
tpolecat: Ok thanks I'll have a look.

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
tpolecat: The resource is a server running asynchronously on another fiber (logical thread). So here we use `IO.never` to halt the current fiber forever, or until it's cancelled by the user hitting ^C, at which point the cleanup computation runs and shuts everything down.

tpolecat:
image

PR #680, opened 2018-10-08
Replace diode lenses by Monocle lenses

cquiroz:
I've been replacing diode lenses by monocle ones. This PR changes a few more and splits the large `circuit` package classes into their own files
There isn't any change on functionality

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/circuit/HeaderSideBarFocus.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/circuit/HeaderSideBarFocus.scala
@@ -1,0 +1,50 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.web.client.circuit
+
+import cats.Eq
+import cats.implicits._
+import gem.Observation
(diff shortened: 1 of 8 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/circuit/HeaderSideBarFocus.scala
```diff
@@ -0,0 +1,50 @@
...
+  implicit val eq: Eq[HeaderSideBarFocus] =
+    Eq.by(x => (x.status, x.conditions, x.operator, x.observer))
+
+  val headerSideBarG: Getter[SeqexecAppRootModel, HeaderSideBarFocus] =
+    Getter[SeqexecAppRootModel, HeaderSideBarFocus] { c =>
+      val clientStatus = ClientStatus(c.uiModel.user, c.ws)
+      val obs = c.uiModel.sequencesOnDisplay.selectedOperator match {
+        case Some(x) => x.asRight
+        case _       => c.uiModel.defaultObserver.asLeft
+      }
```
swalker2m: I think this is fine, but I probably would have used `val obs = c.uiModel.sequencesOnDisplay.selectedOperator.toRight(c.uiModel.defaultObserver)`.

cquiroz: good suggesion

PR #683, opened 2018-10-08
SEQNG-749: Show the queue state on the tab

cquiroz:
This PR shows the state of a cal queue on the tab like for any instrument:

seqexec - daycal queue 2018-10-08 16-53-08

The UI part is mostly ready but I'm having troubles getting the queue state from the server. This PR is to request comments about that


modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
```diff
@@ -8,17 +8,19 @@ import cats.implicits._
...
 import seqexec.model.enum.BatchCommandState
+import seqexec.model.enum.BatchExecState
 
 @Lenses
-final case class ExecutionQueue(name:     String,
-                                cmdState: BatchCommandState,
-                                queue:    List[Observation.Id])
+final case class ExecutionQueue(name:      String,
+                                cmdState:  BatchCommandState,
+                                execState: BatchExecState,
```
cquiroz: Do you think it is ok to do this? Passing the state here?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -304,10 +297,10 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
 
   def startQueue(q: EventQueue, qid: QueueId, clientId: ClientID): IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
     Event.modifyState[executeEngine.ConcreteTypes](executeEngine.get.flatMap{ st => {
-      (EngineState.queues ^|-? index(qid)).getOption(st).map {
-        _.status(st) match {
-          case BatchExecState.Idle     => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Run(clientId)) >>> {(_, ())}).toHandle *> runQueue(qid, clientId)
-          case BatchExecState.Stopping => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Run(clientId)) >>> {(_, ())}).toHandle
+      queueO(qid).getOption(st).map { q =>
+        q.status(st) match {
+          case BatchExecState.Idle     => (cmdStateO(qid).set(BatchCommandState.Run(clientId)) >>> execStateO(qid).set(q.status(st)) >>> {(_, ())}).toHandle *> runQueue(qid, clientId)
```
cquiroz: I'm setting the state here in `startQueue` and on `stopQueue` but I'm not getting these updates on the client. I must be missing something

modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
jluhrs: `execState` is a function of `cmdState` and the execution state of all the sequences identified in `queue`. It is recalculated inside `SeqexecEngine` every time it is needed, so it cannot be out of sync from its dependencies. But if it is set inside `ExecutionQueue`, either it is recalculated at the same time any of its dependencies change, or we agree that its value can only be trusted at the output.
I would feel more comfortable with this alternative: put `ExecutionQueue` back into package `seqexec.server`, without this new `execState`, and create an `ExecutionQueueView` in `seqexec.model` with the extra parameters required by the web server. This View class will have a snapshot of the state of the queue at the output of `SeqexecEngine`.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: It is because the sequence states are updated inside `runQueque`. Your call to `q.status` should be after the call to `runQueue` to see the changes.

modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueue.scala
cquiroz: I agree, the way I did it is forcing it. We should do  `ExecutionQueueView` as you suggest

cquiroz:
Rebased onto master and can be reviewed again

PR #684, opened 2018-10-09
SEQNG-745 Select sequences for instruments when running queue.

jluhrs:
Also, set observer for each sequence as they are run.


modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -162,8 +162,8 @@ class SeqexecCommandRoutes(auth:       AuthenticationService,
       se.clearQueue(inputQueue, qid) *>
         Ok(s"All sequences removed from queue $qid")
 
-    case POST -> Root / "queue" / QueueIdVar(qid) / "run" / ClientIDVar(clientId) as _ =>
-      se.startQueue(inputQueue, qid, clientId) *>
+    case POST -> Root / "queue" / QueueIdVar(qid) / "run" / ObserverVar(observer) / ClientIDVar(clientId) as user =>
```
cquiroz: So, this will break the client, but that's ok

PR #690, opened 2018-10-12
SEQNG-751: Send the observer when starting a queue

cquiroz:
The observer should be set when starting a queue. This PR implements this partially as the backend is not complete yet. There are a couple of edge cases that need work but this is enough to make a release and let users test the queue

Note that I'm changin some `Map` to `SortedMap` to avoid using the unsafe Monocle lenses

https://monosnap.com/file/oOUEz2jAfa6AhrcEgSaeGx5kZFXoGI#


modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueueView.scala
```diff
@@ -7,17 +7,24 @@ import cats.Eq
 import cats.implicits._
 import gem.Observation
 import monocle.macros.Lenses
-import seqexec.model.enum.{BatchCommandState, BatchExecState}
+import seqexec.model.enum.BatchCommandState
+import seqexec.model.enum.BatchExecState
```
sraaphorst: Why separate these? Is this a style thing?

cquiroz: yes, otherwise I easily have to split the import into several lines, this is handled automatically

modules/seqexec/model/src/main/scala/seqexec/model/ExecutionQueueView.scala
```diff
@@ -7,17 +7,24 @@ import cats.Eq
...
 
 @Lenses
 final case class ExecutionQueueView(id:        QueueId,
                                     name:      String,
                                     cmdState:  BatchCommandState,
                                     execState: BatchExecState,
-                                    queue:     List[Observation.Id])
+                                    queue:     List[Observation.Id]) {
+  val observer: Option[Observer] = cmdState match {
```
cquiroz: I'm not happy about this but it will do for now. Ideally we'd have the observer in the model for real

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/HeadersSideBar.scala
```diff
@@ -27,26 +44,38 @@ import scala.concurrent.duration._
 object HeadersSideBar {
   final case class Props(model: ModelProxy[HeaderSideBarFocus]) {
     def canOperate: Boolean = model().status.canOperate
-    def sequenceObserver: Either[Observer, SequenceObserverFocus] = model().observer
+    def selectedObserver
+      : Either[Observer, Either[DayCalObserverFocus, SequenceObserverFocus]] =
```
cquiroz: Ugly type to have the options:
* Default observer
* Daycal observer
* instrument observer

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
```diff
@@ -57,6 +64,19 @@ final case class SequencesOnDisplay(tabs: Zipper[SeqexecTab]) {
       case InstrumentSequenceTab(_, Some(curr), _, _, _, _) => curr.id
     }
 
+  /**
+    * Update the observer for the cal tab
+    */
+  def updateCalTabObserver(o: Observer): SequencesOnDisplay = {
```
cquiroz: We update the local Cal queue observer here

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
```diff
@@ -265,14 +300,23 @@ final case class SequencesOnDisplay(tabs: Zipper[SeqexecTab]) {
...
     */
-  def selectedOperator: Option[SequenceObserverFocus] =
-    SequencesOnDisplay.focusSequence.getOption(this).collect {
-      case InstrumentSequenceTab(_, Some(s), _, _, _, _) =>
-        SequenceObserverFocus(s.metadata.instrument,
-                              s.id,
-                              s.allStepsDone,
-                              s.metadata.observer)
-    }
+  def selectedObserver
```
cquiroz: Return the observer for the selected tab

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/HeadersSideBar.scala
```diff
@@ -106,71 +152,108 @@ object HeadersSideBar {
...
-          f.backend.updateStateOb(o.value.some, Callback.empty)
-      })
+    .componentWillMount(f =>
+      f.backend.$.props >>= { p =>
+        p.model()
+          .operator
+          .map(op => f.backend.updateStateOp(op.value.some, Callback.empty))
+          .getOrEmpty *>
+          (p.selectedObserver match {
+            case Right(Right(a)) =>
```
sraaphorst: 😆 
The dreaded nested `Either`.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/tabs.scala
```diff
@@ -81,24 +90,34 @@ object SeqexecTab {
...
       case i: InstrumentSequenceTab => i
     }(identity)
 }
 
-final case class CalibrationQueueTab(tableState: TableState[StepsTable.TableColumn], state: BatchExecState) extends SeqexecTab {
+@Lenses
+final case class CalibrationQueueTab(
+  tableState: TableState[StepsTable.TableColumn],
+  state:      BatchExecState,
+  observer:   Option[Observer])
```
cquiroz: This should be on the core model, but lacking that we need to keep it on the client model, hence this field

PR #694, opened 2018-10-16
SEQNG-762 Removed OIWFS from TCS configuration if instrument does not have it

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/keywords/ObsKeywordReader.scala
```diff
@@ -159,7 +159,10 @@ final case class ObsKeywordReaderImpl(config: Config, site: Site) extends ObsKey
 
   override def getOiwfsGuide: SeqAction[StandardGuideOptions.Value] =
     SeqAction.either(config.extractAs[StandardGuideOptions.Value](new ItemKey(TELESCOPE_KEY, GUIDE_WITH_OIWFS_PROP))
-      .leftMap(explainExtractError))
+      .recoverWith[ConfigUtilOps.ExtractFailure, StandardGuideOptions.Value] {
+        case ConfigUtilOps.KeyNotFound(_)         => StandardGuideOptions.Value.park.asRight
+        case e@ConfigUtilOps.ConversionError(_,_) => e.asLeft
+      }.leftMap(explainExtractError))
```
jluhrs: This gives a default value instead of failing. Keywords OIWFS_ST is mandatory, but the sequence value is not always present.

PR #696, opened 2018-10-17
Fixed some issues with calibration queue status

jluhrs:
* Calibration queue status is Completed as long has it has run at least once, and there are no unrun sequences in the queue.
* It goes back to Idle if it is Completed, but a new sequence is added.
* Now it is possible to add or remove sequences to the Calibration queue while it is running.
* It now attempts to start a new added sequence, if the queue is running.
* Fixed bug where sequences removed from the Session queue were not removed from the calibration queue.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -450,7 +450,8 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
             st
           } else {
             (EngineState.sequences.modify(ss => ss - seqId) >>>
-             EngineState.selected.modify(ss => ss.toList.filter{case (_, x) => x =!= seqId}.toMap))(st)
+             EngineState.selected.modify(ss => ss.toList.filter{case (_, x) => x =!= seqId}.toMap) >>>
+             EngineState.queues.modify(_.mapValues(ExecutionQueue.queue.modify(_.filterNot(_ === seqId)))))(st)
           }
       } withEvent UnloadSequence(seqId)).toHandle
     )
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -249,34 +249,38 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
   def cmdStateO(qid: QueueId): Optional[EngineState, BatchCommandState] =
     queueO(qid) ^|-> ExecutionQueue.cmdState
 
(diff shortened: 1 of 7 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -249,34 +249,38 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
+      for {
+        q <- st.queues.get(qid)
+        seqs <- seqIds.filter(sid => st.executionState.sequences.get(sid)
+          .map(seq => !seq.status.isRunning && !seq.status.isCompleted && !q.queue.contains(sid))
+          .getOrElse(false)).some.filter(_.nonEmpty)
+        if (!seqs.isEmpty)
+      } yield executeEngine.modify(queueO(qid).modify(_.addSeqs(seqs))) *>
+        ((q.cmdState, q.status(st)) match {
+        case (_, BatchExecState.Completed)       => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState)
+          .set(BatchCommandState.Idle) >>> {(_, ())}).toHandle
```
cquiroz: So this changes the cmdState right?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -328,10 +332,12 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
     Event.modifyState[executeEngine.ConcreteTypes](executeEngine.get.flatMap{ st => {
-      queueO(qid).getOption(st).map {
+      queueO(qid).getOption(st).filter(!_.queue.isEmpty).map {
         _.status(st) match {
-          case BatchExecState.Idle     => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Run(observer, user, clientId)) >>> {(_, ())}).toHandle *> runQueue(qid, observer, user, clientId)
-          case BatchExecState.Stopping => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState).set(BatchCommandState.Run(observer, user, clientId)) >>> {(_, ())}).toHandle
+          case BatchExecState.Idle |
+               BatchExecState.Stopping => ((EngineState.queues ^|-? index(qid) ^|-> ExecutionQueue.cmdState)
+            .set(BatchCommandState.Run(observer, user, clientId)) >>> {(_, ())}).toHandle *>
+            runQueue(qid, observer, user, clientId)
```
jluhrs: `runQueue` must be called when state is Stopping, because some sequence could have already been stopped

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: Yes, back to `Idle`, which makes the `execState` change to `Idle`.

PR #698, opened 2018-10-17
SEQNG-766 Allow restarting failed sequence in running queue.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -275,17 +279,25 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
+        if (q.status(st) =!= BatchExecState.Running ||
+          sstOp.map(sst => !sst.isRunning && !sst.isCompleted).getOrElse(true))
+      } yield executeEngine.modify(queueO(qid).modify(_.removeSeq(seqId))) *>
+        ((q.cmdState, q.status(st)) match {
+          case (_, BatchExecState.Completed)         => executeEngine.unit
+          case (BatchCommandState.Run(o, u, c), _)
+            if sstOp.map(_.isError).getOrElse(false) => runQueue(qid, o, u, c)
+          case _                                     => executeEngine.unit
+        })
+    ).getOrElse(executeEngine.unit)}
```
jluhrs: If the removed sequence was in a failed state, it might have been stalling the queue, so try to run it.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -97,9 +97,13 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
     val used = st.sequences.mapValues(_.seq.resources)
       .map(Function.tupled(filterSeqResources(_.status.isRunning))).toList.foldMap(identity)
     // Resources that will be used by sequences in running queues
+    def failedInQueue(q: ExecutionQueue): Set[Resource] = q.queue.map(sid => st.sequences.get(sid).map(x => (sid, Set(x.seq.instrument: Resource))).map(Function.tupled(filterSeqResources(_.status.isError)))
```
cquiroz: could you expand this vertically?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -275,17 +279,25 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
...
+          case (_, BatchExecState.Completed)         => executeEngine.unit
+          case (BatchCommandState.Run(o, u, c), _)
+            if sstOp.map(_.isError).getOrElse(false) => runQueue(qid, o, u, c)
+          case _                                     => executeEngine.unit
+        })
+    ).getOrElse(executeEngine.unit)}
 
   def removeSequenceFromQueue(q: EventQueue, qid: QueueId, seqId: Observation.Id): IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
-    Event.modifyState[executeEngine.ConcreteTypes]((removeSeq(qid, seqId) withEvent UpdateQueueRemove(qid, List(seqId))).toHandle)
+    Event.modifyState[executeEngine.ConcreteTypes](removeSeq(qid, seqId).map[executeEngine.ConcreteTypes#EventData](_ => UpdateQueueRemove(qid, List(seqId))))
```
cquiroz: Why this?

jluhrs: Because I changed `removeSeq` to produce a `Handle[Unit]` instead of a `Endo`, but I still need to output the `SeqEvent`, so I added it at the end with the `map`.

jluhrs: And the reason why I changed `removeSeq`, was to be able to start sequences inside with the call to `runQueue`.

PR #700, opened 2018-10-18
Make all BooPicklers explicit

cquiroz:
I've noted we don't always explicitly define all boo picklers leading to bigger size. From now on this should fail to compile rather than use the auto generated pickler via macros

### DIFF
--- a/modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
+++ b/modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
@@ -4,6 +4,7 @@
 package seqexec.model.boopickle
 
 import boopickle.DefaultBasic._
+import cats.Traverse
 import cats.implicits._
 import gem.Observation
 import java.time.Instant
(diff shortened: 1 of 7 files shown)


cquiroz:
This changes reduces in about 1% the js size just by virtue of removing the boopickle macros

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
```diff
@@ -30,13 +31,28 @@ trait ModelBooPicklers extends GemModelBooPicklers {
   // IMPORTANT The order of the picklers is very relevant to the generated size
   // add them with care
   //**********************
```
tpolecat: Does this comment still apply?

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
```diff
@@ -30,13 +31,28 @@ trait ModelBooPicklers extends GemModelBooPicklers {
...
-  implicit val instrumentPickler = generatePickler[Instrument]
+  implicit val instrumentPickler = transformPickler(
+    (t: Int) =>
+      Instrument.all
+        .find(_.ordinal === t)
+        .getOrElse(throw new RuntimeException("Failed to decode instrument")))(
```
tpolecat: It seems like we could define `enumPickler[A: Enumerated]` and factor out the repetition here and below. Can we add the ordinal to the error message?

tpolecat: It also might be worth making a map or array or something so the lookup isn't linear with `.find` … I don't know how much of a bottleneck this is.

cquiroz: Yeah, though these are not `Enumerated` yet

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
cquiroz: I'll remove it

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
```diff
@@ -26,17 +29,38 @@ import seqexec.model.events._
     "org.wartremover.warts.OptionPartial"
   ))
 trait ModelBooPicklers extends GemModelBooPicklers {
-  //**********************
-  // IMPORTANT The order of the picklers is very relevant to the generated size
-  // add them with care
-  //**********************
-  implicit val instrumentPickler = generatePickler[Instrument]
+  def valuesMap[F[_]: Functor: Foldable, A, B](c: F[A], f: A => B): Map[B, A] =
```
cquiroz: @tpolecat, does this make sense? Do you know why I can’t do just `valueMap(List, _.ordinal)` when calling it?

Instead I have to declare explicitly the arg type

PR #704, opened 2018-10-22
SEQNG-773: Remove button sometimes doesn't work in chrome

cquiroz:
this fixes a bug in chrome that sometimes prevents the remove from queue button of clicking

Tested in OSX, FF and Chrome version 69

/cc @jluhrs Could you test it in Linux?


jluhrs:
Tested in Chrome 67.0.3396.99, SO Linux Centos 3.10.0-693.5.2.el7.x86_64

cquiroz:
Great, thanks for testing

PR #705, opened 2018-10-23
SEQNG-738: Support reordering of the queue

cquiroz:
This is a relatively large change to support reordering of the elements on the cal queue:

https://monosnap.com/file/aVgw2RtxrzbxRnnLJ6dePVpCyx7T61#

The change produced several regressions on animation, and updates, they should be fixed by the most part but some may have escaped

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -321,10 +321,10 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
   def moveSequenceInQueue(q: EventQueue, qid: QueueId, seqId: Observation.Id, delta: Int, cid: ClientId)
   : IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
     Event.modifyState[executeEngine.ConcreteTypes](
-      (moveSeq(qid, seqId, delta) withEvent UpdateQueueMoved(qid, cid)).toHandle)
-  ).map(_.asRight)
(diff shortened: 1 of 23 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -311,20 +311,20 @@ class SeqexecEngine(httpClient: Client[IO], settings: SeqexecEngine.Settings, sm
           .map(_.queue.indexOf(seqId)).toList))))
   ).map(_.asRight)
 
-  private def moveSeq(qid: QueueId, seqId: Observation.Id, d: Int): Endo[EngineState] = st => (
+  private def moveSeq(qid: QueueId, seqId: Observation.Id, delta: Int): Endo[EngineState] = st => (
```
jluhrs: Yes, `d` was for `delta`. I thought it was obvious :smile:

cquiroz: Maybe here but not in other places

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -170,7 +171,7 @@ package object server {
     def addSeq(sid: Observation.Id): ExecutionQueue = q.copy(queue = q.queue :+ sid)
     def addSeqs(sids: List[Observation.Id]): ExecutionQueue = q.copy(queue = q.queue ++ sids)
     def removeSeq(sid: Observation.Id): ExecutionQueue = q.copy(queue = q.queue.filter(_ =!= sid))
-    def moveSeq(sid:Observation.Id, idx: Int): ExecutionQueue = q.copy(queue = moveElement(q.queue, sid, idx))
+    def moveSeq(sid:Observation.Id, delta: Int): ExecutionQueue = q.copy(queue = moveElement(q.queue, sid, delta))
```
cquiroz: This use was not very obvious

PR #707, opened 2018-10-24
Fixes for files placement according to Hydra suggestions

cquiroz:
Testing hydra reveals that some packages could be better placed at different locations. This PR fixes that:

https://docs.triplequote.com/troubleshooting/#misplaced-package-object-warnings


tpolecat:
Directories for empty path segments are pretty common in Scala projects … I emailed Iulian about this yesterday. Seems like they could deal with it. But this change is fine.

PR #708, opened 2018-10-25
Back-end implementation of exposure countdown

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/SequenceGen.scala
```diff
@@ -3,23 +3,54 @@
...
-  val resources: Set[Resource] = steps.foldMap(_.resources)
+  val resources: Set[Resource] = steps.collect{
+    case SequenceGen.PendingStep(_, _, resources, _) => resources
+  }.foldMap(identity(_))
 }
 
 object SequenceGen {
 
-  final case class Step(id: Int, config: StepConfig, resources: Set[Resource], generator: HeaderExtraData => EngineStep)
+  sealed trait Step {
```
cquiroz: We have a bunch of `Step` classes defined, could we have something else `StepGen` or something?

jluhrs:
@cquiroz No, you are free to change it as you see fit.

cquiroz:
It would be much easier to review this if the final  tagless change were in one PR and the countdowns in another

Regardless, where would the webserver listen for updates? Would they come on the some event stream?

jluhrs:
@cquiroz Yes, they will come as `engine.PartialResult(_, _, Progress(total, remaining))` (which for now just produces a `SequenceUpdated` message. The progress value is also present in the `Action` state, in `state.partials`. The head is the most recent one.

PR #709, opened 2018-10-25
SEQNG-778: GPI configuration not being picked up

cquiroz:
There is a serious bug on the current seqexec which doesn't let GPI to run as the configuration of GHOST is passed instead

This is an easy mistake to make, so besides a fix I added tagged types to ensure the correct type is passed

Also we have `SeqexecServer.Settings` and `SeqTranslate.Settings` which is quite confusing. I moved both to their own classes and renamed the latter to `TranslateSettings`


cquiroz:
Tested on GPI real

PR #710, opened 2018-10-25
SEQNG-779: Fixed GHOST to parse angles

sraaphorst:
The support to take a `Config` object and parse angles from it was not implemented; GHOST needs the target information to send to the instrument.

This adds:
1. The logic to GHOST to parse `Angle` and `HourAngle`; and
2. Now also follows the same pattern as GPI, where it can result in errors arising from incorrect `Config` entries.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
@@ -6,10 +6,12 @@ package seqexec.server.ghost
 import cats.data.Reader
 import cats.data.EitherT
 import cats.effect.{IO, Sync}
+import cats.implicits._
 import edu.gemini.spModel.config2.Config
 import edu.gemini.spModel.seqcomp.SeqConfigNames._
 import edu.gemini.spModel.gemini.ghost.Ghost
 import gem.math.{Angle, HourAngle}
+import gem.optics.Format
 
 import scala.concurrent.duration._
 import seqexec.model.dhs.ImageFileId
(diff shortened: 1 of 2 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOST.scala
```diff
@@ -65,51 +67,53 @@ object GHOST {
...
+      // 1. content = None: nothing to process, so Right(None).
+      // 2. process content = Some(a) indicating success, so Right(Some(a))
+      // 3. process content = None indicating failure, so Left(error)
+      val content: Option[String] = extractor[String](propName)
+      val result: Option[Option[A]] = content.map(fmt.getOption)
+      result.map {
+        case None       => Left(ConversionError(INSTRUMENT_KEY / propName,
+                                                s"Could not parse $propName content: ${content.getOrElse("")}"))
+        case other      => Right(other)
+      }.getOrElse(Right(None))
```
sraaphorst: I'm sure there's a much nicer way to do this that I overlooked, so by all means, please teach me what it is :-).

cquiroz: Is there a `toRight(leftValue)` in cats?

tpolecat: I think you could say

```scala
def formatExtractor[A](
  fmt: Format[String, A]
): String => Either[ExtractFailure, Option[A]] = { propName =>
  extractor[String](propName) match {
    case None    => Right(None)
    case Some(c) => fmt.getOption(c).toRight(ConversionError(...))
  }
}
```
The `Angle` upper bound seems unnecessary, and if you return a function type you won't need the η-expansion operator `_` on `raExtractor` and `decExtractor` below. You could also write it as `.fold` instead of `match` but I think `match` is easier to read here.

tpolecat: I have not tested this, it may not even typecheck.

sraaphorst: Thanks, @tpolecat. I'll give that a try after the seqexec meeting is done.

PR #711, opened 2018-10-29
SEQNG-678: GHOST parameter setting, take one

sraaphorst:
This is my first attempt at translating ODB parameter values for GHOST to instrument-specific parameter values. Several notes:

1. I still have some questions as to what needs to be set when, which are detailed in the comments before.

2. This is rather hacky at this point, as it is not the final version and our goal is only to use a small subset of the parameters to implement the most basic GHOST functionality (i.e. passing target information to GHOST). As this advances, we will implement type safety to go along with it.


modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -22,10 +22,85 @@ import squants.time.Time
...
+    def determineType(name: Option[String]): IFUTargetType = name match {
+      case None        => IFUTargetNone
+      case Some("Sky") => IFUTargetSky
+      case _           => IFUTargetObject
+    }
+  }
+  private final case object IFUTargetNone extends IFUTargetType(0)
+  private final case object IFUTargetSky  extends IFUTargetType(1)
+  private final case object IFUTargetObject extends IFUTargetType(2)
+
```
tpolecat: Is there any benefit in these being private? We usually don't bother.

I don't think these need to be nested classes. This creates path-dependent types which aren't what we want. So I would move them to the companion object. 

Also by convention we place the constructors for an ADT inside the companion. 

So:

```scala
object GHOSTController {
  ...
  sealed abstract class IFUTargetType(val targetType: Int)
  object IFUTargetType {
    case object None extends IFUTargetType(0)
    case object Sky  extends IFUTargetType(1)
    ...
   }

}
```

Then your class looks like

```scala
final case class GHOSTController... {
  import GHOSTController._

  // in here you can now say
  IFUTargetType.None

}
```

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -22,10 +22,85 @@ import squants.time.Time
...
+                        bundleConfig: BundleConfig): Configuration =
+      (for {
+        name <- nameOpt
+        ra   <- raOpt
+        dec  <- decOpt
+      } yield {
+        Configuration.single(ifuNum.ifuStr + ".target", IFUTargetType.determineType(nameOpt).targetType) |+|
+          Configuration.single(ifuNum.ifuStr + ".ra", ra.toDoubleDegrees) |+|
+          Configuration.single(ifuNum.ifuStr + ".dec", dec.toDoubleDegrees) |+|
+          Configuration.single(ifuNum.ifuStr + ".bundle", bundleConfig.configName)
```
tpolecat: I would factor out `Configuration.single(ifuNum.ifuStr` into a local method. Stylistically I would assign the `Option[Configuration]` to a variable and then return `config.getOrElse(Configuratin.Zero)` to avoid the awkward parens around the `for`-comp.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -22,10 +22,85 @@ import squants.time.Time
...
+
+  private def hrifuConfig(config: GHOSTConfig): Configuration =
+    ifuConfig(IFU2, config.hrifu1Name, config.hrifu1CoordsRAHMS, config.hrifu1CoordsDecDMS, HighResBundleConfig) |+|
+      ifuConfig(IFU2, config.srifu2Name, config.srifu2CoordsRAHMS, config.srifu2CoordsDecDMS, StandardBundleConfig)
+
+  // If the srifu parameters are defined, use them; otherwise, use the hrifu parameters.
+  // TODO: What do we do with the base position explicit override?
+  // TODO: This was not on the list of provided parameter names.
+  private
+  def ghostConfig(config: GHOSTConfig): SeqActionF[F, CommandResult] = {
```
tpolecat: Non line break after `private` ;-)

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -22,10 +22,85 @@ import squants.time.Time
...
+  }
 
   def applyConfig(config: GHOSTConfig): SeqActionF[F, Unit] =
-    SeqActionF.apply(log.info(s"Configuring $config"))
+    for {
+      _ <- EitherT.liftF(Sync[F].delay(Log.debug("Start GHOST configuration")))
+      _ <- EitherT.liftF(Sync[F].delay(Log.debug(s"GHOST configuration $config")))
+      _ <- ghostConfig(config)
+      _ <- EitherT.liftF(
+        Sync[F].delay(Log.debug("Completed GHOST configuration")))
```
tpolecat: I would factor out `EitherT.liftF(Sync[F].delay(Log.debug(message)))`. Is there not an existing lifted logger already?

sraaphorst:
Thanks, @tpolecat! I appreciate the suggestions. I'll go ahed and make these changes.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
jluhrs: There is `SeqActionF.apply`

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: There's no particular reason to make them private. I guess it's a product of the C++ liquifying my brain. 🤯

Okay, those changes are easy enough to make and make sense; having the case objects in the companion object to the sealed class / trait also allows for much simpler names to be used.

Thanks for the tips... working on old Scala code and Java code in the old OCS has me lagging a bit.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: I have to admit that I just snagged this right off of `GPIController` directly (with the name changes for GHOST, but that's it). I'm not entirely sure how this code works: I'm looking into @jluhrs' suggestion now and reading some Cats docs.

sraaphorst: Excellent. Thanks, @jluhrs. That cleans it up considerably and I understand a lot better how this code is working now. We should make the same change to `GPIController` to clean it up. I'll add a JIRA task and do that quickly.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -22,10 +22,71 @@ import squants.time.Time
...
+  // Question: Do we need to, for (2), touch IFU2 at all (i.e. to standard mode, no target)?
+  // Do we need to always set target type for IFUs? How about bundle configuration?
+  //
+  // Right now, this only generate a config if the name, RA, and dec are defined: otherwise, we get the empty config.
+  private def ifuConfig(ifuNum: IFUNum,
+                        nameOpt: Option[String],
+                        raOpt: Option[HourAngle],
+                        decOpt: Option[Angle],
+                        bundleConfig: BundleConfig): Configuration = {
+    def cfg[P: Show](paramName: String, paramVal: P) =
```
cquiroz: Do we use `Show` in this method?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -22,10 +22,71 @@ import squants.time.Time
...
+      ifuConfig(IFUNum.IFU2, config.srifu2Name, config.srifu2CoordsRAHMS, config.srifu2CoordsDecDMS,
+        BundleConfig.HighRes)
+
+  // If the srifu parameters are defined, use them; otherwise, use the hrifu parameters.
+  // TODO: What do we do with the base position explicit override?
+  // TODO: This was not on the list of provided parameter names.
+  private def ghostConfig(config: GHOSTConfig): SeqActionF[F, CommandResult] = {
+    val giapiApply = srifuConfig(config) |+| hrifuConfig(config)
+
+    EitherT(ghostClient.genericApply(giapiApply).attempt)
```
cquiroz: This feel like common code with `GPI` could we factor it out?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -39,11 +100,40 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
...
+  object BundleConfig {
+    case object Standard extends BundleConfig("lo")
+    case object HighRes extends BundleConfig("hi")
+  }
+
+  sealed abstract class IFUNum(val ifuNum: Int) {
+    val ifuStr: String = s"ghost:cc:cu:ifu$ifuNum"
+  }
+  object IFUNum {
+    case object IFU1 extends IFUNum(1)
```
cquiroz: i guess we should name this param as in the line below

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -39,11 +100,40 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
...
+    case object IFU2 extends IFUNum(ifuNum = 2)
+  }
+
+  sealed abstract class IFUTargetType(val targetType: Int)
+  object IFUTargetType {
+    case object NoTarget extends IFUTargetType(0)
```
cquiroz: probably using named params here too?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
cquiroz: I think this maybe taken from `GPIController`. I used `Show` there but I think it is a bad idea. we should instead create another typeclass decoupling giapi formatting from regular formatting. I'll add a Jira task about it

sraaphorst: Yes, it is taken from `GPIController`, and won't compile unless `P` has `Show`. I wasn't too sure why that was the case, so that's probably a good idea.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: Agreed. Maybe we need a `GIAPIInstrument` or something?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: Sure. I'm not sure why one ended up named and not the others. I'll make them all named.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
cquiroz: Yeah, though we can do that in another PR

PR #712, opened 2018-10-30
SEQNG-782 Avoid processing empty streams.

jluhrs:
Every processed event can produce new events, in the form of an`Option[Stream]`. The code was queuing a `Stream` even if the result was `None`. I removed that.
There is also some code formatting.

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -263,10 +263,11 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
       case Null                    => pure(SystemUpdate(se, EventResult.Ok))
     }
 
-    (ev match {
+    ev match {
       case EventUser(ue)   => handleUserEvent(ue)
-      case EventSystem(se) => handleSystemEvent(se).flatMap(x => userReact.applyOrElse(se, (_:SystemEvent) => unit).map(_ => x))
-    })
+      case EventSystem(se) => handleSystemEvent(se).flatMap(x =>
+        userReact.applyOrElse(se, (_:SystemEvent) =>  unit).map(_ => x))
+    }
   }
 
   /** Traverse a process with a stateful computation. */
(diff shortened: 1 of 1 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -263,38 +263,45 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
...
+  def mapEvalState[A, S, B](input: Stream[IO, A],
+                            initialState: S, f: (A, S) => IO[(S, B, Option[Stream[IO, A]])])
+                           (implicit ec: ExecutionContext): Stream[IO, B] = {
     Stream.eval(fs2.async.unboundedQueue[IO, Stream[IO, A]]).flatMap { q =>
       Stream.eval_(q.enqueue1(input)) ++
         q.dequeue.joinUnbounded.evalMapAccumulate(initialState) { (s, a) =>
           f(a, s).flatMap {
-            case (ns, b, st) =>
-              q.enqueue1(st) >> IO.pure((ns, b))
+            case (ns, b, None)     => IO.pure((ns, b))
```
cquiroz: is this the core change?

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -263,38 +263,45 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
...
+            case (ns, b, Some(st)) => q.enqueue1(st) >> IO.pure((ns, b))
```
cquiroz: It is a bit strange that you could change from `q.enqueue1(Option[A])` to `q.enqueue1(A)`. Am I missing something?

jluhrs: Yes, you are missing something. Previously, `st` was a `Stream` that most of the time was `Stream.empty`. Now it is an `Option[Stream]`, where the case when it was `Stream.empty` was replaced by `None`.

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
jluhrs: This, in conjunction with the change in line 298.

PR #715, opened 2018-11-07
SEQNG-722: Upgrade to react 16.5.1

cquiroz:
This took a while to complete as required some testing effort and there were a few things to fix on the steps table

Otherwise is mostly mechanical changes and I relocated a few files to the correct directory


cquiroz:
9K increase on the output, wasn't that terrible at the end

PR #716, opened 2018-11-07
SEQNG-796 Merged engine and server sequence data in a single Map.

jluhrs:
The sequence data used by the execution engine and the sequence data used by the seqexec server (the upper level) are kept in separate Maps. This PR uses Lenses magic join those Maps in a single one, while keeping the execution engine agnostic of the extra data required by upper level.

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -12,14 +12,13 @@ import seqexec.engine.Result.{PartialVal, PauseContext, RetVal}
 import seqexec.model.{ClientId, SequenceState}
 import fs2.Stream
 import gem.Observation
-import monocle.{Lens, Optional}
-import monocle.function.Index.index
+import monocle.Optional
 import mouse.boolean._
 import org.log4s.getLogger
 
 import scala.concurrent.ExecutionContext
 
-class Engine[D, U](stateL: Lens[D, Engine.State]) {
+class Engine[D, U](stateL: Engine.State[D]) {
 
   class ConcreteTypes extends Engine.Types {
     override type StateType = D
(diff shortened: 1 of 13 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -344,17 +330,8 @@ class Engine[D, U](stateL: Lens[D, Engine.State]) {
...
-    def empty: State = State(Map.empty)
-
-    def sequences: Lens[State, Map[Observation.Id, Sequence.State[IO]]] =
-      Lens[State, Map[Observation.Id, Sequence.State[IO]]](_.sequences)(ss => _ => State.apply(ss))
-    private def atSequence(id: Observation.Id)
-    : Optional[Map[Observation.Id, Sequence.State[IO]], Sequence.State[IO]] = index(id)
-    def sequenceState(id: Observation.Id): Optional[State, Sequence.State[IO]] =
-      sequences ^|-? atSequence(id)
+  trait State[D] {
+    def sequenceStateIndex(sid: Observation.Id): Optional[D, Sequence.State[IO]]
```
jluhrs: This is the core change. Instead of a `case class` containing the engine view of the sequences, `State` is now a trait that must implement a Lens to the sequences. So now the container can keep the engine sequence with its own sequence data, which is invisible to the engine. The only complication I found was with the creation and destruction of the sequences at the engine level.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -145,17 +145,17 @@ class SeqexecEngine(httpClient: Client[IO], settings: Settings[IO], sm: SeqexecM
...
 
   def selectSequenceEvent(i: Instrument, sid: Observation.Id, observer: Observer, user: UserDetails, clientId: ClientId): executeEngine.EventType= {
     val lens =
-      (EngineState.sequences ^|-? index(sid)).modify(ObserverSequence.observer.set(observer.some)) >>>
+      (EngineState.sequences ^|-? index(sid)).modify(SequenceData.observer.set(observer.some)) >>>
        EngineState.instrumentLoadedL(i).set(sid.some) >>>
        refreshSequence(sid)
     def testRunning(st: EngineState):Boolean = (for {
-      sels <- st.selected.get(i)
-      sstate <- st.executionState.sequences.get(sels)
```
jluhrs: Now it is not necessary to go to a second `Map` to find the rest of the sequence info.

PR #718, opened 2018-11-07
SEQNG-794: Fix bad navigation when on config page

cquiroz:
While testing i noted a small bug where if you are on a config page of seq X and the seq is updated, e.g. when running the view will jump back to the main table

This PR fixes this and do some refactoring


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
```diff
@@ -128,10 +128,11 @@ final case class SequencesOnDisplay(tabs: Zipper[SeqexecTab]) {
           .getOrElse(StepsTable.State.InitialTableState)
         val completed = tab
           .flatMap(_.completedSequence)
+        val stepConfig = tab.flatMap(_.stepConfig)
```
cquiroz: This is the actual bug fix, the rest is refactorings

PR #721, opened 2018-11-08
SEQNG-800 Clipped negative countdown values.

jluhrs:


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -241,15 +241,24 @@ class Engine[D, U](stateL: Engine.State[D]) {
...
-      case Failed(id, i, e)        => logError(e) *> fail(id)(i, e) *> pure(SystemUpdate(se, EventResult.Ok))
-      case Busy(id, _)             => Logger.warning(s"Cannot run sequence ${id.format} because required systems are in use.") *> pure(SystemUpdate(se, EventResult.Ok))
-      case BreakpointReached(_)    => Logger.debug("Engine: Breakpoint reached") *> pure(SystemUpdate(se, EventResult.Ok))
-      case Executed(id)            => Logger.debug("Engine: Execution completed") *> next(id) *> pure(SystemUpdate(se, EventResult.Ok))
-      case Executing(id)           => Logger.debug("Engine: Executing") *> execute(id) *> pure(SystemUpdate(se, EventResult.Ok))
-      case Finished(id)            => Logger.debug("Engine: Finished") *> switch(id)(SequenceState.Completed) *> pure(SystemUpdate(se, EventResult.Ok))
+      case Completed(id, i, r)     => Logger.debug(s"Engine: From sequence ${id.format}: Action " +
+        s"completed ($r)") *> complete(id, i, r) *> pure(SystemUpdate(se, EventResult.Ok))
+      case PartialResult(id, i, r) => Logger.debug(s"Engine: From sequence ${id.format}: Partial " +
```
cquiroz: This gets a bit noisy on the logs with updates every second. Unfortunately I cann to pattern match on `r` as `Progress` is on the `seqexec.server`package

modules/seqexec/server/src/main/scala/seqexec/server/ProgressUtil.scala
```diff
@@ -7,38 +7,32 @@ import cats.Applicative
...
       sch.awakeEvery(PollPeriod).evalMapAccumulate(s0){case (st, t) => fs(t).run(st)}.map(_._2))
 
   def countdown[F[_]: Effect: Applicative](total: Time, elapsed: Time): Stream[F, Progress] =
-    ProgressUtil.fromFParam[F] {
+    ProgressUtil.fromF[F] {
       t: FiniteDuration => {
           val progress = Milliseconds(t.toMillis) + elapsed
-          Applicative[F].pure(Progress(total, RemainingTime(total - progress)))
+          val remaining = total - progress
+          val clipped = if(remaining.value >= 0.0) remaining else Seconds(0.0)
```
cquiroz: Will this be applied to simulation only?

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -241,15 +241,24 @@ class Engine[D, U](stateL: Engine.State[D]) {
...
+      case Busy(id, _)             => Logger.warning(s"Cannot run sequence ${id.format} because " +
+        s"required systems are in use.") *> pure(SystemUpdate(se, EventResult.Ok))
+      case BreakpointReached(_)    => Logger.debug("Engine: Breakpoint reached") *>
+        pure(SystemUpdate(se, EventResult.Ok))
+      case Executed(id)            => Logger.debug("Engine: Execution completed") *>
+        next(id) *> pure(SystemUpdate(se, EventResult.Ok))
+      case Executing(id)           => Logger.debug("Engine: Executing") *>
+        execute(id) *> pure(SystemUpdate(se, EventResult.Ok))
+      case Finished(id)            => Logger.debug("Engine: Finished") *>
+        switch(id)(SequenceState.Completed) *> pure(SystemUpdate(se, EventResult.Ok))
```
jluhrs: Changes here are some extra debug info, and formatting.

modules/seqexec/server/src/main/scala/seqexec/server/ProgressUtil.scala
```diff
@@ -7,38 +7,32 @@ import cats.Applicative
...
+          Applicative[F].pure(Progress(total, RemainingTime(clipped)))
         }
-    }
-
+    }.takeThrough(_.remaining.self.value > 0.0)
```
jluhrs: No point on producing more progress events after remaining time reaches 0.

cquiroz: Makes sense

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
jluhrs: But, it is debug level. Do you want it removed?

modules/seqexec/server/src/main/scala/seqexec/server/ProgressUtil.scala
jluhrs: This countdown generator is used for simulated instruments, and for instruments that don't have a countdown.

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
cquiroz: i think it is ok. I just wanted to check what you thought

PR #723, opened 2018-11-08
SEQNG-807 Fixed problem with sync not adding new steps.

jluhrs:


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -60,14 +60,20 @@ class Engine[D, U](stateL: Engine.State[D]) {
     */
   def load(seq: Sequence[IO]): Sequence.State[IO] = Sequence.State.init(seq)
 
+  /**
+    * Redefines an existing sequence. Changes the step actions, removes steps, adds new steps.
+    */
+  def reload(seq: Sequence.State[IO], steps: List[Step[IO]]): Sequence.State[IO] =
```
cquiroz: can you use `F[_]` here?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -427,11 +427,11 @@ class SeqexecEngine(httpClient: Client[IO], settings: Settings[IO], sm: SeqexecM
...
 
     def loadSequenceEvent(seqg: SequenceGen): executeEngine.EventType =
-      Event.modifyState[executeEngine.ConcreteTypes](( {st:EngineState => st.sequences.get(seqId)
-        .map(os => {
-          val newseq = os.copy(seqGen = seqg)
-          EngineState.atSequence(seqId).set(newseq) >>> updateSequenceEndo(seqId, newseq)
-        }).getOrElse(loadSequenceEndo(seqId, seqg))(st)} withEvent LoadSequence(seqId)).toHandle)
+      Event.modifyState[executeEngine.ConcreteTypes](( { st:EngineState =>
+        st.sequences.get(seqId).fold(loadSequenceEndo(seqId, seqg))(
+          _ => reloadSequenceEndo(seqId, seqg)
```
jluhrs: Refresh the sequence with the new `reloadSequenceEndo`, which rebuilds the steps list, instead of using `updateSequenceEndo`, which only updates existing steps.

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
jluhrs: I didn't propagate the `F[_]` to `Engine`

PR #724, opened 2018-11-09
SEQNG-811 Added step id to PartialResult event.

jluhrs:


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -172,10 +175,10 @@ class Engine[D, U](stateL: Engine.State[D]) {
 
   private def actionResume(id: Observation.Id, i: Int, cont: Stream[IO, Result])
   : HandleType[Unit] =
-    getS(id).flatMap(_.map { s =>
-    if (Sequence.State.isRunning(s) && s.current.execution.lift(i).exists(Action.paused))
-      modifyS(id)(_.start(i)) *> Handle.fromStream(act(id, (cont, i)))
-    else unit
+    getS(id).flatMap(_.collect {
+      case s@Sequence.State.Zipper(z, _)
```
jluhrs: The other alternative is `Final`, but `actionResume` should never be called on a sequence in that state.

modules/seqexec/engine/src/main/scala/seqexec/engine/SystemEvent.scala
```diff
@@ -11,9 +11,12 @@ import gem.Observation
   * Events generated internally by the Engine.
   */
 sealed trait SystemEvent
-final case class Completed[R<:RetVal](id: Observation.Id, i: Int, r: OK[R]) extends SystemEvent
-final case class PartialResult[R<:PartialVal](id: Observation.Id, i: Int, r: Partial[R]) extends SystemEvent
-final case class Paused[C <: PauseContext](id: Observation.Id, i: Int, r: Result.Paused[C]) extends SystemEvent
+final case class Completed[R<:RetVal](id: Observation.Id, stepId: Step.Id, i: Int, r: OK[R])
+  extends SystemEvent
+final case class PartialResult[R<:PartialVal](sid: Observation.Id, stepId: Step.Id, i: Int,
+                                             r: Partial[R]) extends SystemEvent
```
jluhrs: Added step id to `PartialResult` and `Completed`, although right now only the former is used.

PR #725, opened 2018-11-09
SEQNG-791: Show obs progress bar total time on the UI

cquiroz:
This PR supports animating the progress bar as the server indicates how much time is left to complete an observation. This ended up being fairly involved, especially to produce somewhat smooth animations.

Here are some examples
GMOS being paused:
https://monosnap.com/file/TlUyuwXzEkJPXLZfBmn8bPtIBWM1is#

F2
https://monosnap.com/file/pUhlG1KKooCMHFCXijb6iBG8jjwCdp#

Along the way I fixed a couple of small bugs SEQNG-767 and SEQNG-807


arturog8m:
Sorry for getting late to this. Two suggestions:
1. Can we use the same units to time remaining and exposure time? The exposure time is in seconds (I guess in all instruments it's the same?), but the remaining time is shown in minutes.  I think consistency would be useful for operations.
2. Can we add text to the remaining time that says: Remaining Time or something like that? or just "Remaining"

arturog8m:
And yes,I really like this! ;)

PR #729, opened 2018-11-12
SEQNG-786: Better change handling on the session queue

cquiroz:
A bug in a previous refactoring produced that not all changes to the global state be reflected on the session queue, e.g. when going from one step to the next

This PR fixes this and adds some improvements to reduce the amount of refreshes

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/circuit/StatusAndLoadedSequencesFocus.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/circuit/StatusAndLoadedSequencesFocus.scala
@@ -4,7 +4,6 @@
 package seqexec.web.client.circuit
 
 import cats.Eq
-import cats.Order
 import cats.implicits._
 import gem.Observation
 import monocle.Getter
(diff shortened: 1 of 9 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/circuit/StatusAndLoadedSequencesFocus.scala
```diff
@@ -22,34 +21,57 @@ final case class SequenceInSessionQueue(id:            Observation.Id,
...
-                                        targetName:    Option[TargetName],
+                                        targetName:    Option[String],
                                         runningStep:   Option[RunningStep],
                                         nextStepToRun: Option[Int])
 
 object SequenceInSessionQueue {
-  implicit val order: Order[SequenceInSessionQueue] = Order.by(_.id)
-  implicit val ordering: scala.math.Ordering[SequenceInSessionQueue] =
-    order.toOrdering
+  implicit val eq: Eq[SequenceInSessionQueue] =
```
cquiroz: This is the core fix. `Order` wasn't consistent with a proper `Eq`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/circuit/StatusAndLoadedSequencesFocus.scala
```diff
@@ -22,34 +21,57 @@ final case class SequenceInSessionQueue(id:            Observation.Id,
                                         active:        Boolean,
                                         loaded:        Boolean,
                                         name:          String,
-                                        targetName:    Option[TargetName],
+                                        targetName:    Option[String],
```
jluhrs: Why the change from `TargetName` to `String`?

cquiroz: it was trying to fix a bug but i'll revert it

PR #733, opened 2018-11-14
SEQNG-812 Refine the estimated observation time

cquiroz:
Estimation of obs time for GPI, good for the progress bar

https://www.dropbox.com/s/bld44uy99hh8pxr/screencast%202018-11-14%2012-08-47.mp4?dl=0

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPI.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPI.scala
@@ -4,28 +4,41 @@
 package seqexec.server.gpi
 
 import cats.data.EitherT
-import cats.effect.{Effect, IO, Sync}
+import cats.effect.Effect
+import cats.effect.IO
+import cats.effect.Sync
 import cats.implicits._
 import cats.data.Reader
 import edu.gemini.spModel.config2.Config
 import edu.gemini.spModel.gemini.gpi.Gpi._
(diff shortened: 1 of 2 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gpi/GPI.scala
```diff
@@ -4,28 +4,41 @@
...
+import squants.time.Time
 
 final case class GPI[F[_]: Effect](controller: GPIController[F])
     extends InstrumentSystem[F]
     with GDSInstrument {
+  // Taken from the gpi isd
+  val readoutOverhead: Time  = Seconds(4)
+  val writeOverhead: Time    = Seconds(2)
+  val perCoaddOverhead: Time = Seconds(2.7)
+  val extraObsTimeout: Time  = Seconds(30)
```
cquiroz: This is arbitrary. We want to have a timeout of the calculated time plus some extra as a timeout. We don't want it to be so short that it timeouts when the obs gets a bit late, nor too long that we have to wait to much for errors to be detected. This value worked fine in my tests

jluhrs: Is this a tolerance for the timeout? If that is the case, I would use a name that reflects it., like 'timeoutTolerance'.

cquiroz: Done

PR #734, opened 2018-11-14
SEQNG-744: Support filtering the session queue

cquiroz:
This Pr adds support for filtering by daycal/night time on the session queue. it will use that filtering when adding sequences to the day cal queue.

As suggested an icon is used also to indicate which seqs belong to which type:

image

Here is the feature in action

https://monosnap.com/file/1YZ06FMw8y2AYfIiYwKSdw2j4PJqLs#


modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
```diff
@@ -82,7 +83,7 @@ object actions {
   final case class RunObsResumeFailed(s:   Observation.Id) extends Action
 
   // Queue actions
-  final case class RequestAllDayCal(qid:     QueueId) extends Action
+  final case class RequestAllSelectedSequences(qid:     QueueId) extends Action
```
jluhrs: Alignment ? :rofl:

cquiroz: Revenge is sweet

PR #736, opened 2018-11-15
SEQNG-689 Don't play sounds for logged off users

cquiroz:
This PR removes sounds while executing for logged off users


cquiroz:
A question came up if this should be actually user selectable. Any opinions?

cquiroz:
after user feedback I changed this to make sound user selectable with a control at the bottom of the page

seqexec - gs-2018b-q-0-74 2018-11-16 10-38-27

PR #737, opened 2018-11-15
SEQNG-815 Display None for unknown targets

cquiroz:
The original default was missleading as there can be flats for day and night. This is how it looks now

image


cquiroz:
I removed the bold font too:
seqexec - gs-2018b-q-0-28 2018-11-15 16-56-55

PR #738, opened 2018-11-16
Seqng 740

cquiroz:
This is another take on how to add sequences to the queue from the session table. We show on the session table what is in the day cal and you can add the directly. You can also add all clicking on the header.

This is better explained on the video below:

https://monosnap.com/file/HZHKBPhhXsmesJgEulXRjjlpgl2ykF#


cquiroz:
Changed icons to:

image

PR #742, opened 2018-11-21
SEQNG-718: Support D&D to the cal queue

cquiroz:
After several failed attempts to integrate libraries to do DnD I resorted to do this with the plain HTML5 DnD API. the API is well know for its quirks so it maybe not that robust but seems to work fine in Chrome/FF

This version is also limited and less polished as I'd like but given we have other ways to add seqs to the queue I'll accept this limited version.

Now it is possible to grab a sequenc from the session table and drop it on teh daytime queue tab. This also lets you do this even if other tab is selected.

Here's a video

https://monosnap.com/file/payj5aY8sm1agn62g9VYu2hJtXDh2H#

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/SessionQueueTable.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/SessionQueueTable.scala
@@ -245,8 +245,9 @@ object SessionQueueTable {
         val optimalSizes = sequences.foldLeft(columnsDefaultWidth) {
           case (currWidths,
                 SequenceInSessionQueue(id, st, i, _, _, n, _, t, r, _, _)) =>
-            val idWidth = max(currWidths.getOrElse(ObsIdColumn, ObsIdMinWidth),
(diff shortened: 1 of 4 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/SessionQueueTable.scala
```diff
@@ -626,17 +628,17 @@ object SessionQueueTable {
         SeqexecStyles.headerRowStyle
       case (_, SessionQueueRow(_, s, _, _, _, _, _, _, _, _, _))
           if s == SequenceState.Completed =>
-        SeqexecStyles.stepRow |+| SeqexecStyles.rowPositive
+        SeqexecStyles.stepRow |+| SeqexecStyles.draggableRow |+| SeqexecStyles.rowPositive
```
tpolecat: You have a CSS semigroup? Cool.

cquiroz: Monoid, you can have empty css 😄

PR #743, opened 2018-11-26
SEQNG-678: GHOST configuration

sraaphorst:
HI all,

Last week when speaking with the instrument team from Australia, I received some answers that necessitated changes to the way GHOSTController worked: hence, this PR.

The idea for each IFU is:
1. If the standard resolution configuration is set, use it.
2. Otherwise if the high resolution configuration is set, use it.
3. Otherwise the IFU must be moved out of the way: request it to be parked.

We are also temporarily turning off observes since they will not be involved in tonight's testing, and all we want to do is test configuration applies.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
@@ -114,7 +114,7 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
 
   // We use a dummy observation for now, since at this point, we cannot actually observe using the instrument.
   def observe(fileId: ImageFileId, expTime: Time): SeqActionF[F, ImageFileId] =
-    SeqActionF.apply(fileId)
+    SeqActionF.apply((fileId, expTime)._1) // suppress unused error
   //    EitherT(ghostClient.observe(fileId, expTime.toMilliseconds.milliseconds).map(_ => fileId).attempt)
   //      .leftMap {
   //        case CommandResultException(_, "Message cannot be null") => Execution("Unhandled observe command")
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
@@ -39,18 +39,25 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
     def cfg[P: Show](paramName: String, paramVal: P) =
       Configuration.single(s"${ifuNum.ifuStr}.$paramName", paramVal)
 
-    val config = for {
-      _   <- nameOpt // We only do this because we want Configuration.Zero if there is no target name.
-      ra  <- raOpt
-      dec <- decOpt
-    } yield {
-      cfg("target", IFUTargetType.determineType(nameOpt).targetType) |+|
-        cfg("ra", ra.toDoubleDegrees) |+|
-        cfg("dec", dec.toDoubleDegrees) |+|
-        cfg("bundle", bundleConfig.configName)
-    }
+    (nameOpt, raOpt, decOpt) match {
+      // Case 1: IFU is in use here for an actual position.
+      case (Some(name), Some(ra), Some(dec)) =>
+        val ifuTargetType = IFUTargetType.determineType(nameOpt)
+        cfg("target", ifuTargetType.targetType) |+|
+          cfg("type", DemandType.DemandRADec.demandType) |+|
+          cfg("ra", ra.toDoubleDegrees) |+|
+          cfg("dec", dec.toDoubleDegrees) |+|
+          cfg("bundle", bundleConfig.determineType(ifuTargetType).configName)
 
-    config.getOrElse(Configuration.Zero)
+      // Case 2: IFU is explicitly excluded from use.
+      case (Some(name), None, None) =>
+        cfg("target", IFUTargetType.NoTarget.targetType) |+|
+          cfg("type", DemandType.DemandPark.demandType)
+
+      // Case 3: IFU is not in this mode. Look to other mode (e.g. high res).
+      case _ =>
+        Configuration.Zero
+    }
   }
 
   private def srifuConfig(config: GHOSTConfig): Configuration =
@@ -66,6 +73,7 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
         BundleConfig.HighRes)
 
   // If the srifu parameters are defined, use them; otherwise, use the hrifu parameters.
+  // Which set of parameters is determined completely by which of srifuName and hrifuName is set.
   // TODO: What do we do with the base position explicit override?
   // TODO: This was not on the list of provided parameter names.
   private def ghostConfig(config: GHOSTConfig): SeqActionF[F, CommandResult] = {
@@ -100,10 +108,16 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
 }
 
 object GHOSTController {
-  sealed abstract class BundleConfig(val configName: String)
+  sealed abstract class BundleConfig(val configName: String) {
+    def determineType(t: IFUTargetType): BundleConfig = t match {
+      case IFUTargetType.SkyPosition => BundleConfig.Sky
+      case _                         => this
+    }
+  }
   object BundleConfig {
-    case object Standard extends BundleConfig(configName = "lo")
-    case object HighRes extends BundleConfig(configName = "hi")
+    case object Standard extends BundleConfig(configName = "IFU_LORES")
+    case object HighRes  extends BundleConfig(configName = "IFU_HRES")
+    case object Sky      extends BundleConfig(configName = "IFU_SKY")
   }
 
   sealed abstract class IFUNum(val ifuNum: Int) {
@@ -125,6 +139,13 @@ object GHOSTController {
       case Some("Sky") => SkyPosition
       case _           => Target
     }
+  }
+
+  sealed abstract class DemandType(val demandType: String)
+  object DemandType {
+    // Future DemandTypes: HALT, HOME, XY
+    case object DemandRADec extends DemandType("IFU_DEMAND_RADEC")
+    case object DemandPark extends DemandType("IFU_DEMAND_PARK")
   }
 
   /**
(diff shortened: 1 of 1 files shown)


sraaphorst:
Just FYI, this is the version of `GHOSTController` I was testing with last week. I didn't want to check it in until now as I wasn't sure it was ready, but it looks to be working correctly with tests now.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -39,37 +39,61 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
...
-      ra  <- raOpt
-      dec <- decOpt
-    } yield {
-      cfg("target", IFUTargetType.determineType(nameOpt).targetType) |+|
-        cfg("ra", ra.toDoubleDegrees) |+|
-        cfg("dec", dec.toDoubleDegrees) |+|
-        cfg("bundle", bundleConfig.configName)
+    (nameOpt, raOpt, decOpt) match {
+      // Case 1: IFU is in use here for an actual position.
+      case (Some(name@_), Some(ra), Some(dec)) =>
```
cquiroz: isn't this the same as `Some(name)`?

sraaphorst: Yes, but I wasn't allowed to do that because I wasn't using `name`, so the compiler insisted I do this instead, IIRC.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -39,37 +39,61 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
     def cfg[P: Show](paramName: String, paramVal: P) =
       Configuration.single(s"${ifuNum.ifuStr}.$paramName", paramVal)
```
cquiroz: this is unrelated to the code but I'd expect the config names to be prefixed like:
`s"ghost:${ifName...`

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: I guess I could instead do:
`case Some(ra), Some(dec) if nameOpt.isDefined() => ...`

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -88,22 +112,30 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
...
 
+  // We use a dummy observation for now, since at this point, we cannot actually observe using the instrument.
   def observe(fileId: ImageFileId, expTime: Time): SeqActionF[F, ImageFileId] =
-    EitherT(ghostClient.observe(fileId, expTime.toMilliseconds.milliseconds).map(_ => fileId).attempt)
-      .leftMap {
-        case CommandResultException(_, "Message cannot be null") => Execution("Unhandled observe command")
-        case CommandResultException(_, m)                        => Execution(m)
-        case f                                                   => SeqexecException(f)
-      }
+    SeqActionF.apply((fileId, expTime)._1) // suppress unused error
```
cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -88,22 +112,30 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
...
+  sealed abstract class BundleConfig(val configName: String) {
+    def determineType(t: IFUTargetType): BundleConfig = t match {
+      case IFUTargetType.SkyPosition => BundleConfig.Sky
+      case _                         => this
+    }
+  }
   object BundleConfig {
-    case object Standard extends BundleConfig(configName = "lo")
-    case object HighRes extends BundleConfig(configName = "hi")
+    case object Standard extends BundleConfig(configName = "IFU_LORES")
```
cquiroz: Where do those strings come from? Could they be read from ocs2 code?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -88,22 +112,30 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
...
+    case object HighRes  extends BundleConfig(configName = "IFU_HRES")
+    case object Sky      extends BundleConfig(configName = "IFU_SKY")
   }
 
   sealed abstract class IFUNum(val ifuNum: Int) {
```
cquiroz: Why `abstract class` rather than `sealed trait`?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -155,20 +194,20 @@ object GHOSTController {
     implicit val eq: Eq[GHOSTConfig] = Eq.by(
```
cquiroz: Is this Eq law-tested?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: It's in the `ifuStr:`
```
 sealed abstract class IFUNum(val ifuNum: Int) {
    val ifuStr: String = s"ghost:cc:cu:ifu$ifuNum"
  }
```

sraaphorst: Everything only has to do with IFUs right now, so I threw it in there. I could pull it out, though, if that seems better.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
cquiroz: Dosen't `case (Some(name), Some(ra), Some(dec))` work?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
cquiroz: ah ok, thanks for the clarification

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: No, none of these are in OCS2 code. I just learned about the discovery of these strings last Wednesday. (The documentation I was reading said that these params were actually `int`s, but it turns out they're defined in the instrument software and convert there to `int`s, but you have to pass in these actual strings.

cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: If I use `sealed trait`, I can't pass in the `val ifuNum: Int` parameter, can I?

The compiler complains in `IFU1` and `IFU2` about reassignment to `val`.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
cquiroz: Ah, you're not using it,then `case (Some(_), Some(ra), Some(dec))`

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: Hmmm... I'd still have to change `IFU1` and `IFU2` to then have these as val params instead of passing them up to the superclass. Is there an advantage to this, out of curiosity?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
swalker2m: Maybe just `case (Some(_), Some(ra), Some(dec)) =>`?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: i.e.:
```
 sealed trait IFUNum {
    val ifuNum: Int
    val ifuStr: String = s"ghost:cc:cu:ifu$ifuNum"
  }
  object IFUNum {
    case object IFU1 extends IFUNum {
        override val ifuNum = 1
    }
    case object IFU2 extends IFUNum {
        override val ifuNum = 2
    }
  }
```

cquiroz: TBH there are only details in one case v/s the other. I often fall back to sealed trait and use `abstract class` only on some cases. perhaps @tpolecat has a better explanation

You can leave it as is, it probably doesn't make a big difference

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: I suppose what would be easier (and works) is to just do `Some(_), Some(ra), Some(dec)`

sraaphorst: This seems like it would make the most sense to me, though:

```
case (Some(ra), Some(dec)) if nameOpt.isDefined =>
```

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: Oh, are you talking about `SeqexecServerArbitraries`? Because if so, there is an arbitrary for `GHOSTConfig`.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
swalker2m: But that doesn't match `(nameOpt, raOpt, decOpt)` ...

sraaphorst: No, I'd have to change the match to `(raOpt, decOpt)`, but the effect would be the same otherwise, no?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: Never mind :-). I do appear to have a `GHOSTSpec` that does that. I must have added it in way back.

PR #744, opened 2018-11-26
SEQNG-806: Add system configuration buttons in steps

cquiroz:
This PR adds buttons to request configuring a subsystem for a given step. This lets you select a step and it will show the buttons for that step. The backend is not yet implemented so it is just simulated but it would let @jluhrs to test

Here it is in action

https://monosnap.com/file/CVwpsSr7tsTPGdLJij3fDEgKLJO4b7#

There are several things pending though
* Get the state to be reflected on all clients
* Forbid multiple clients requesting the same resource
* Display configuration error messages
* Show activity on the rest of the UI


cquiroz:
Forgot to mention the buttons are disabled so they don't show up in production

PR #745, opened 2018-11-27
SEQNG-825: Fix GHOST coordinates

sraaphorst:
During GHOST testing, it was discovered that the seqexec had illegal values for declination. This was due to me using `HourAngle` and `Angle` instead of `RightAscension` and `Declination`, as I should have.

This fixes that.


sraaphorst:
@swalker2m That sounds like a very good idea. I probably won't have time to do it today, but I'll leave this PR open and make that change tomorrow.

PR #746, opened 2018-11-27
SEQNG-804 Engine now is able to run a single Action in a Sequence.

jluhrs:
A not so elegant solution to run a single Action in an inactive Sequence. This feature will be the base to implement the ability for the user to apply a step configuration to any system without running the sequence.


modules/seqexec/engine/src/main/scala/seqexec/engine/ActionCoords.scala
```diff
@@ -0,0 +1,34 @@
...
+package seqexec.engine
+
+import cats.Eq
+import cats.implicits._
+import gem.Observation
+import seqexec.model.StepId
+
+final case class ActionIndex(self: Long) extends AnyVal
+object ActionIndex {
+  implicit val actionIndexEq: Eq[ActionIndex] = Eq.instance(_.self === _.self)
```
cquiroz: `Eq.by(_.self)`

modules/seqexec/engine/src/main/scala/seqexec/engine/Sequence.scala
```diff
@@ -377,6 +418,17 @@ object Sequence {
 
       override val toSequence: Sequence[F] = seq
 
+      override def startSingle(c: ActionCoordsInSeq): State[F] = self
```
cquiroz: is `ActionCoordsInSeq` a new class?

modules/seqexec/engine/src/main/scala/seqexec/engine/ActionCoords.scala
```diff
@@ -0,0 +1,35 @@
...
+object ActionIndex {
+  implicit val actionIndexEq: Eq[ActionIndex] = Eq.by(_.self)
+}
+
+final case class ExecutionIndex(self: Long) extends AnyVal
+object ExecutionIndex {
+  implicit val actionIndexEq: Eq[ExecutionIndex] = Eq.by(_.self)
+}
+
+final case class ActionCoordsInSeq(stepId: StepId, execIdx: ExecutionIndex,
```
cquiroz: Somehow I'd expect the resource to be here. Like we want to execute `TCS` on step `stepId`?

jluhrs: Resource control is in the level above Engine

jluhrs: Resource control is in the level above Engine

jluhrs: Resource control is in the level above Engine

PR #748, opened 2018-11-30
Back end side of system config command

jluhrs:
Implementation of the Single System Configuration command, that allows the user to apply a step configuration to a system.
The command was connected to the UI.
The feedback in the UI is not yet implemented.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -94,27 +94,18 @@ class SeqexecEngine(httpClient: Client[IO], settings: Settings[IO], sm: SeqexecM
     q.enqueue(Stream.emits(loadEvents(seqId))).map(_.asRight).compile.last.attempt.map(_.bimap(SeqexecFailure.SeqexecException.apply, _ => ()))
 
   private def checkResources(seqId: Observation.Id)(st: EngineState): Boolean = {
-    def filterSeqResources(f: Sequence.State[IO] => Boolean)
-                          (sid: Observation.Id, resources: Set[Resource]): Set[Resource] =
-      st.sequences.get(sid).filter(s => f(s.seq)).fold(Set.empty[Resource])(_ => resources)
     // Resources used by running sequences
-    val used = st.sequences.mapValues(_.seqGen.resources)
-      .map(Function.tupled(filterSeqResources(_.status.isRunning))).toList.foldMap(identity)
+    val used = resourcesInUse(st)
+
+    // TODO: this is too much guessing. We should have proper tracking of systems' state.
+    val failedInstruments: Set[Resource] = st.sequences.values.toList.mapFilter(s =>
+      s.seq.status.isError.option(s.seqGen.instrument)).toSet
+
     // Resources that will be used by sequences in running queues
-    def failedInQueue(q: ExecutionQueue): Set[Resource] = q.queue.map(sid =>
-      st.sequences.get(sid).map(x => (sid, Set(x.seqGen.instrument: Resource))).map(
-        Function.tupled(filterSeqResources(_.status.isError))
-      )
-    ).collect{case Some(x) => x}.foldMap(identity)
+    val reservedByQueues = resourcesReserved(failedInstruments, st)
 
-    val usedByQueues = st.queues
-      .filter{ case (_, q) => q.status(st) === BatchExecState.Running || q.status(st) === BatchExecState.Waiting }
-      .map{case (_, q) => q.queue.map(sid => st.sequences.get(sid).map(x => (sid, x.seqGen.resources))
-        .filter{case (_, res) => res.intersect(failedInQueue(q)).isEmpty}
-        .map(Function.tupled(filterSeqResources(_.status.isIdle)))
-      ).collect{case Some(x) => x} }.toList.flatten.foldMap(identity)
+    st.sequences.get(seqId).exists(_.seqGen.resources.intersect(used ++ reservedByQueues).isEmpty)
 
-    st.sequences.get(seqId).map(_.seqGen.resources.intersect(used ++ usedByQueues).isEmpty).getOrElse(false)
   }
 
   def start(q: EventQueue, id: Observation.Id, user: UserDetails, clientId: ClientId): IO[Either[SeqexecFailure, Unit]] =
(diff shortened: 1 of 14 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -408,6 +401,46 @@ class SeqexecEngine(httpClient: Client[IO], settings: Settings[IO], sm: SeqexecM
     }).flatMap(_.map(Function.tupled(runQueue)).getOrElse(executeEngine.unit))
   }
 
+  private def configSystemCheck(sid: Observation.Id, sys: Resource)(st: EngineState): Boolean = {
+    // Resources used by running sequences
+    val used = resourcesInUse(st)
+
+    // TODO: this is too much guessing. We should have proper tracking of systems' state.
+    val failedInstruments: Set[Resource] = st.sequences.values.toList.mapFilter(s =>
```
cquiroz: is this duplicated from `SeqexecEngine`?

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
```diff
@@ -173,8 +171,7 @@ class SeqexecCommandRoutes(auth:       AuthenticationService,
         Ok(s"Stopped from queue $qid")
 
     case POST -> Root / "execute" / ObsIdVar(oid) / PosIntVar(step) / ResourceVar(resource) as _ =>
-      // Just simulate a delay
-      IO.sleep(2.seconds) *>
+      se.configSystem(inputQueue, oid, step, resource) *>
```
cquiroz: So we are not using the response. How is it sent to the client?
How could the client know that a response is correlated with this request?

modules/seqexec/server/src/test/scala/seqexec/server/SeqexecServerLensesSpec.scala
```diff
@@ -13,13 +13,19 @@ import seqexec.server.SeqexecServerArbitraries._
...
+import seqexec.engine.{Action, Actions, Result}
 
 /**
   * Tests SeqexecServer Lenses
   */
 final class SeqexecServerLensesSpec extends CatsSuite with ArbObservation {
 
-  implicit val steppEq: Eq[HeaderExtraData => engine.Step[IO]] = Eq.fromUniversalEquals
+  // I tried to go down the rabbit hole with the Eqs, but it is not worth it for what they are used.
+  implicit val streamEq: Eq[fs2.Stream[IO, Result]] = Eq.fromUniversalEquals
```
cquiroz: This doesn't look right. You shouldn't be able to compare `Stream[IO, N]` I think

modules/seqexec/server/src/test/scala/seqexec/server/SeqexecServerLensesSpec.scala
```diff
@@ -13,13 +13,19 @@ import seqexec.server.SeqexecServerArbitraries._
...
+  implicit val actStateEq: Eq[Action.State] = Eq.fromUniversalEquals
+  implicit val actionEq: Eq[Action[IO]] = Eq.by(x => (x.kind, x.gen, x.state))
```
cquiroz: Same here, i don't think you can do `Eq` with `IO` involved

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -617,39 +667,59 @@ object SeqexecEngine extends SeqexecConfiguration {
...
+   */
+  private def resourcesReserved(failedSystems: Set[Resource], st: EngineState): Set[Resource] = {
+    def reserved(q: ExecutionQueue): Set[Resource] = q.queue.fproduct(st.sequences.get).collect{
+      case (_, Some(s)) if s.seq.status.isIdle &&
+        failedSystems.intersect(s.seqGen.resources).isEmpty => s.seqGen.resources
+    }.foldK
+
+    val runningQs = st.queues.values.filter(_.status(st).running)
+
+    runningQs.map(reserved).toList.foldK
```
cquiroz: `foldK`?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -461,11 +494,22 @@ class SeqexecEngine(httpClient: Client[IO], settings: Settings[IO], sm: SeqexecM
...
     val seq = st.toSequence
+
+    def resources(s: SequenceGen.StepGen): List[Resource] = s match {
+      case SequenceGen.PendingStepGen(_, _, resources, _) => resources.toList
+      case _                                              => List.empty
+    }
     def engineSteps(seq: Sequence[IO]): List[Step] = {
 
       // TODO: Calculate the whole status here and remove `Engine.Step.status`
       // This will be easier once the exact status labels in the UI are fixed.
```
cquiroz: is this comment still relevant?

cquiroz:
Some general questions:

* will the result of the state be stored on the server? e.g. if another client arrives after a resurce run, will it see it?
* Has this been tested with the real hardware? Should it be?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: :sparkles: magic :sparkles:

jluhrs:
Some general answers:

- Yes, the server state tracks when a configuration has started and when it completes or fails. Therefore, this state is visible to all the clients.
- It has not been tested with real hardware. It has being proved that the command triggers the 'configure' method of the simulation instances of TCC, GCAL and GMOS. There is no difference between triggering that an the real systems.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
tpolecat: `List(a, b, c).combineAll` uses `Monoid` and gives you `a |+| b |+| c` whereas `.foldK` uses `MonoidK` and gives you `a <+> b <+> c`.

`MonoidK` is a universally-quantified monoid; i.e., `forAll a. Monoid[F[a]]`.  So a hand-wavy intuition is that the element type is unknown so you can't look at the elements. For example `Monoid[Option[Int]]` adds up any `Int`s it sees but `MonoidK[Option]` can't look at the elements so it uses `orElse`.

```scala
@ Option(1) |+| None |+| Option(2) 
res12: Option[Int] = Some(3)

@ Option(1) <+> None <+> Option(2) 
res13: Option[Int] = Some(1)
```

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/SeqexecCommandRoutes.scala
jluhrs: I'll create a Jira task to add the `clientId`

modules/seqexec/server/src/test/scala/seqexec/server/SeqexecServerLensesSpec.scala
```diff
@@ -13,19 +12,18 @@ import seqexec.server.SeqexecServerArbitraries._
...
   */
 final class SeqexecServerLensesSpec extends CatsSuite with ArbObservation {
 
   // I tried to go down the rabbit hole with the Eqs, but it is not worth it for what they are used.
-  implicit val streamEq: Eq[fs2.Stream[IO, Result]] = Eq.fromUniversalEquals
   implicit val actStateEq: Eq[Action.State] = Eq.fromUniversalEquals
-  implicit val actionEq: Eq[Action[IO]] = Eq.by(x => (x.kind, x.gen, x.state))
-  implicit val steppEq: Eq[HeaderExtraData => List[Actions[IO]]] = Eq.fromUniversalEquals
-  implicit val stepActionsGenEq: Eq[StepActionsGen] = Eq.by(x => (x.pre))
+  implicit def actionEq[F[_]]: Eq[Action[F]] = Eq.by(x => (x.kind, x.state))
```
cquiroz: Maybe these shouldn't be `implicit` and pass them eplicitly to the tests instead

PR #749, opened 2018-12-03
SEQNG-826 Catch errors when binding to stop and abort channels.

jluhrs:
This solves the issue with Seqexec crashing on startup if GNIRS was not available (could have happened with other instruments, but was discovered with GNIRS).

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
@@ -129,27 +129,43 @@ public class CaObserveSenderImpl<C extends Enum<C> & CarStateGeneric> implements
             }
         });
 
+        ReadOnlyClientEpicsChannel<Short> stopMark = null;
         if(stopCmd!=null && stopCmd.length()>0) {
-            stopMark = epicsReader.getShortChannel(stopCmd + CAD_MARK_SUFFIX);
-            stopMark.registerListener(stopMarkListener = (String arg0, List<Short> newVals) -> {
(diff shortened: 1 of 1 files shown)


modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -129,27 +129,43 @@ public CaObserveSenderImpl(String name, String applyRecord, String carRecord, St
...
+            } catch(Throwable e) {
+                LOG.warn(e.getMessage());
+            }
+            if(stopMark!=null) {
+                stopMark.registerListener(stopMarkListener = (String arg0, List<Short> newVals) -> {
+                    if (newVals != null && !newVals.isEmpty()) {
+                        CaObserveSenderImpl.this.onStopMarkChange(newVals.get(0));
+                    }
+                });
+            }
```
sraaphorst: I know we still need it, but it feels so strange to see Java code in the ocs3 code base.
I would suggest making `stopMark` an `Optional<ReadOnlyClientEpicsChannel<Short>>` so you can get rid of the `null` check.

modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -129,27 +129,43 @@ public CaObserveSenderImpl(String name, String applyRecord, String carRecord, St
...
         } else {
             stopMark = null;
         }
+        this.stopMark = stopMark;
```
sraaphorst: Ahhhh... I guess this is why you probably didn't do it. Never mind, then.

PR #750, opened 2018-12-06
Several countdown fixes

jluhrs:
Countdowns are not so pretty when dealing with the real world.

1. Seqexec now uses the FLAMINGOS-2 provided countdown.
2. Guarded against limit cases, like the countdown not having a valid value for several seconds after starting the observation.
3. I had to implement a special case for FLAMINGOS-2, because its countdown does not count down the exposition, but the whole observe command.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -253,7 +253,8 @@ object EpicsUtil {
         c <- rem
         s <- st
         if s.isBusy
-      } yield Progress(total, RemainingTime(c))
-    })
+      } yield Progress(if(total>c) total else c, RemainingTime(c))
+    }).dropWhile(_.remaining.self.value === 0.0) // drop leading zeros
+      .takeThrough(_.remaining.self.value > 0.0) // drop all tailing zeros but the first one
 
 }
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -16,8 +16,9 @@ import org.log4s._
 import squants.Time
 import cats._
 import cats.data.EitherT
-import cats.effect.IO
+import cats.effect.{Effect, IO}
 import cats.implicits._
+import fs2.Stream
 
 trait EpicsCommand {
   import EpicsCommand._
@@ -244,5 +245,15 @@ object EpicsUtil {
 
   def smartSetParam[A: Eq](v: A, get: => Option[A], set: SeqAction[Unit]): SeqAction[Unit] =
     if(get =!= v.some) set else SeqAction.void
+
+  def countdown[F[_]: Effect](total: Time, remT: F[Option[Time]],
+                              obsState: F[Option[CarStateGeneric]]): Stream[F, Progress] =
+    ProgressUtil.fromFOption(_ => (remT, obsState).mapN { case (rem, st) =>
+      for{
+        c <- rem
+        s <- st
+        if s.isBusy
+      } yield Progress(total, RemainingTime(c))
+    })
 
 }
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -252,6 +252,7 @@ object EpicsUtil {
       for{
         c <- rem
         s <- st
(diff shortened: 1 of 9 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -245,4 +246,15 @@ object EpicsUtil {
...
+                              obsState: F[Option[CarStateGeneric]]): Stream[F, Progress] =
+    ProgressUtil.fromFOption(_ => (remT, obsState).mapN { case (rem, st) =>
+      for{
+        c <- rem
+        s <- st
+        if s.isBusy
+      } yield Progress(if(total>c) total else c, RemainingTime(c))
+    }).dropWhile(_.remaining.self.value === 0.0) // drop leading zeros
+      .takeThrough(_.remaining.self.value > 0.0) // drop all tailing zeros but the first one
+
```
jluhrs: It is only active while the exposure is in progress, removes leading and tailing zeros, and never produces a `Progress`  where the remaining time is higher than the total.

modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2ControllerEpics.scala
```diff
@@ -41,77 +39,65 @@ object Flamingos2ControllerEpics extends Flamingos2Controller {
...
+    case LyotWheel.H2         => "Hart2_G5834"
+  }
+
+  implicit val encodeGrismPosition: EncodeEpicsValue[Grism, String] = EncodeEpicsValue {
+    case Grism.Open    => "Open"
+    case Grism.R1200HK => "HK_G5802"
+    case Grism.R1200JH => "JH_G5801"
+    case Grism.R3000   => "R3K_G5803"
+    case Grism.Dark    => "DK_G5804"
+  }
```
jluhrs: Idea gracefully pointed that those were easier to write like this.

modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2ControllerEpics.scala
```diff
@@ -148,8 +134,22 @@ object Flamingos2ControllerEpics extends Flamingos2Controller {
...
+          if obst.isBusy
+          rem <- Flamingos2Epics.instance.countdown.map(_.seconds)
+        } yield Progress(m, RemainingTime(rem))
+        (m, p)
+      }
+    })
+    s(total).dropWhile(_.remaining.self.value === 0.0) // drop leading zeros
+      .takeThrough(_.remaining.self.value > 0.0) // drop all tailing zeros but the first one
+
+  }
```
jluhrs: FLAMINGOS-2 starts with a countdown value higher than the exposure time. I used a `State` to carry that value as the new `total`.

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
cquiroz: Neat

modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2ControllerEpics.scala
cquiroz: Should we consider making these enums from the db like on gem?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosEpics.scala
```diff
@@ -194,13 +192,9 @@ class GmosEpics(epicsService: CaService, tops: Map[String, String]) {
...
 
-  private val observeCAttr: Option[CaAttribute[CarState]] = Option(dcState.addEnum("observeC",
-    GMOS_TOP + "dc:observeC", classOf[CarState]))
-  def observeC: Option[CarState] = observeCAttr.map(_.value)
-
-  def waitObserve(t: Time): SeqAction[CarState] = observeCAttr.map(attr =>
-    waitForValues(attr, List(CarState.IDLE, CarState.ERROR, CarState.PAUSED),t , "GMOS observe state")).getOrElse(
-    SeqAction.fail(SeqexecFailure.Unexpected(s"Cannot access ${GMOS_TOP}dc:observeC")))
+  private val observeCAttr: CaAttribute[CarState] = dcState.addEnum("observeC",
+    GMOS_TOP + "dc:observeC", classOf[CarState])
```
cquiroz: `s"$GMOS_TOP...`?

PR #752, opened 2018-12-12
SEQNG-788: GHOST Configuration

sraaphorst:
Hi all,

Here is my refactoring of the `GHOSTConfig` into more type-safe classes.
NOTE that it currently won't pass test cases and I'm hoping someone can give me some direction.

It's still inevitable that we have to check what fields come from the OT to determine asterism type (unless we change the OT), so I check the IFU coordinates and create the appropriate GHOST mode based on that.

Since I've divided up `GHOSTConfig` into a hierarchy of modes, I had to change the `Gen` / `Arbitrary` for `GHOSTConfig` and its subclasses. Now I'm having trouble with the `GhostSpec` and while I think I have a vague understanding of why this error is happening, I'm not sure how to solve it in the best way.

The error is in the `checkAll` in `GhostSpec`:
https://github.com/gemini-hlsw/ocs3/blob/137b30c29d48587ac9838830494b016329261953/modules/seqexec/server/src/test/scala/seqexec/server/ghost/GhostSpec.scala#L16

The error is:

`GhostSpec.scala:16:52: could not find implicit value for parameter arbF: org.scalacheck.Arbitrary[seqexec.server.ghost.GHOSTController.GHOSTConfig => seqexec.server.ghost.GHOSTController.GHOSTConfig]`

I'm not sure what is being referred to as `arbF` but this clearly seems to be based on the `Eq` type class for `GHOSTConfig`. Should I make a `checkAll` for each `GHOSTConfig` concrete subclass, or is there a better approach?

Any help @jluhrs, @tpolecat, @cquiroz, or @swalker2m could offer would be much appreciated.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
@@ -109,12 +109,12 @@ object GHOSTController {
   object IFUTargetType {
     case object NoTarget extends IFUTargetType(targetType = "IFU_TARGET_NONE")
     case object SkyPosition extends IFUTargetType(targetType = "IFU_TARGET_SKY")
-    case object Target extends IFUTargetType(targetType = "IFU_TARGET_OBJECT")
+    case class Target(name: String) extends IFUTargetType(targetType = "IFU_TARGET_OBJECT")
 
     def determineType(name: Option[String]): IFUTargetType = name match {
       case None        => NoTarget
       case Some("Sky") => SkyPosition
-      case _           => Target
+      case Some(x)     => Target(x)
     }
 
     implicit val IFUTargetTypeShow: Show[IFUTargetType] = Show.show(_.targetType)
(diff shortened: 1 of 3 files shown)


swalker2m:
> I'm not sure what is being referred to as arbF but this clearly seems to be based on the Eq type class for GHOSTConfig.

Suspect maybe it is missing `Cogen[GHOSTConfig]`.  Maybe quickly define that as:

```
implicit val cogGhostConfig: Cogen[GHOSTConfig] = ???
```

just to see if the compile issue goes away?

A couple of unrelated trivial observations:
* I'd like to stick with the convention we use elsewhere and make it `GhostConfig` instead of `GHOSTConfig`.  For example, check all the enums in `gem.enum` like `GmosRoi` which would otherwise have to be shouted `GMOSROI`.
* It seems like a lot of this might belong in the instrument model itself, at least longterm, and not just in `seqexec` but I may not be understanding it.  For now I think it's fine since the seqexec is only technically sharing the same codebase without really being part of the `gem` model.

sraaphorst:
My thought was also `Eq` or `cogen` (I'm not sure how to define a `cogen` in this case).

I'm perfectly fine with changing the case and think your suggestion looks better, but we should be consistent then and do the same for th `GPI____` classes in OCS3. (I followed their example.)

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -152,48 +116,179 @@ object GHOSTController {
...
+        case _ =>
+          None
+      }
+      extracted match {
+        case Some(config) => Right(config)
+        case None         => Left(ContentError("Response does not constitute a valid GHOST configuration"))
+      }
+    }
 
+    implicit val eq: Eq[GHOSTConfig] = Eq.fromUniversalEquals
```
cquiroz: Should this by `Eq.instance` and pattern match on the subclasses?

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
```diff
@@ -121,16 +60,39 @@ final case class GHOSTController[F[_]: Sync](ghostClient: GHOSTClient[F],
...
       case _                         => this
     }
   }
+
   object BundleConfig {
     case object Standard extends BundleConfig(configName = "IFU_LORES")
     case object HighRes  extends BundleConfig(configName = "IFU_HRES")
     case object Sky      extends BundleConfig(configName = "IFU_SKY")
+
+    implicit val showBundleConfig: Show[BundleConfig] = Show.show(_.configName)
```
cquiroz: I'd like use not to use `Show` for this and define a `GiapiParam` typeclass instead. That probably belongs to another PR though

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: I'll give tat a try. I'm not familiar with `Eq.instance` but it'll be a good learning experience.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GHOSTController.scala
sraaphorst: Yes, we have a task specifically for this, I think, to do the same for GPI. I was going to look at that one next.

PR #753, opened 2018-12-12
SEQNG-830 Better handling of obsolete FLAMINGOS2 filters

jluhrs:
New Seqexec was not producing the error message that the old one was causing when trying to use an obsolete filter in FLAMINGOS2. The root of the issue was in the instrument, but I still put some checks in Seqexec to reject the obsolete filters.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
@@ -29,6 +29,8 @@ final class CaApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements Ca
     private final CaApplyRecord apply;
     private final CaCarRecord<C> car;
 
+    private final Boolean trace = false;
+
     private long timeout;
     private TimeUnit timeoutUnit;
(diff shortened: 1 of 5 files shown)


modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
```diff
@@ -359,37 +381,45 @@ public State onTimeout() {
 
     private synchronized void onApplyValChange(Integer val) {
         if (val != null) {
+            State oldState = currentState;
             currentState = currentState.onApplyValChange(val);
             if (currentState.equals(IdleState) && timeoutFuture != null) {
                 timeoutFuture.cancel(true);
                 timeoutFuture = null;
             }
+            if(trace) LOG.debug("onApplyValChange(" + val + "): " + oldState.signature() + " -> " + currentState.signature());
```
cquiroz: Wouldn't be better to disable them at the logging config?

jluhrs: Yes, some day.

PR #755, opened 2018-12-14
SEQNG-839 Added NIRI to web client

jluhrs:
There was almost nothing to do, I just added the list of valid operations for NIRI.


jluhrs:
To test: change site = "GN" in app.conf.

PR #756, opened 2018-12-18
Experimenting with OpenAPI

swalker2m:
This is a PR to share a bit of experimentation I've been doing with specifying APIs in OpenAPI.  This work came out of a meeting with LCO in which we were discussing the possibility (or not) of trying to create a common API for observation triggering, etc.  The idea was to explore what Gemini might do without considering the Grand Unifying API idea and then compare how different or similar the LCO approach is to ours.

Especially since *I don't know what I'm doing*, I only want to raise awareness and generate discussion.  I don't think we need to actually merge.  Also keep in mind, as I state at the top of the document:

>  I don't think it is sustainable in the long run having the documentation be separate from the code.

but it is useful to think about APIs before writing the code and OpenAPI provides a way to describe them in a standard way with some tooling (albeit kind of flakey?).

The idea is to address at least some of the requirements for the [ToO Triggering with TOMs user story](https://kb.cl.gemini.edu/display/OSP/ToO+Triggering+with+TOMs).  I think perhaps the user story requirements don't particularly match the story -- it requires very detailed observation descriptions which I hope we can avoid for the case of TOMs.  

At any rate, this is an initial pass at some of the ideas.  You can see the rendering of the documentation at [Swaggerhub](https://app.swaggerhub.com/apis-docs/Gemini-Observatory/to-o_triggering_with_to_ms_sample_api/1#). In particular, take a look at [the observation `POST`](https://app.swaggerhub.com/apis-docs/Gemini-Observatory/to-o_triggering_with_to_ms_sample_api/1#/Observations/post_observations__programId_). In the "Request body" there is a lame "Example Value" but also a "Model" link in which the model is defined.

### DIFF
--- a/api/ToOTriggeringWithTOMs.yml
+++ b/api/ToOTriggeringWithTOMs.yml
@@ -9,7 +9,7 @@ info:
     
     Notes:
       * The servers don't actually exist.  They are meant to be suggestive.
-      * We're not trying for a complete Gemini API nor a complete specification for this example.  Just an attempt to play with API ideas and get a feel for it.
+      * We're not trying for a complete Gemini API nor a complete specification for this example.  Just an attempt to play with API ideas and get a feel for it.  It's missing position angle for example and AGS results.
       * Default generation of the model from the Scala code probably won't produce a very nice model for the API.  I think we'll want to use custom codecs.
       * We might want to explore generating OpenAPI specifications directly from the code using [Rho](https://github.com/http4s/rho) or something.  I don't think it is sustainable in the long run having the documentation be separate from the code.
       
@@ -60,8 +60,6 @@ paths:
       description: Combines availability and observing conditions for both Gemini North and Gemini South into a single result.
       tags:
         - Status
-      parameters:
-        - $ref: '#/components/parameters/siteParam'
       security: [] # no authentication required
       responses:
         '200':
@@ -807,6 +805,8 @@ components:
       properties:
         asterismType:
           type: string
+        overriddenBase:
+          $ref: '#/components/schemas/CoordinatesObject'
       discriminator:
         propertyName: asterismType
         mapping:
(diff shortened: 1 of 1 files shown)


swalker2m:
Supposedly you can generate a simple sever that implements the API providing canned responses directly from the documentation.  I tried a couple of the language options but then didn't, um, compile since they had glaring syntax issues.  Maybe that's due to errors I made in the YAML (?) but it does validate in the Sagger editor.

tpolecat:
This looks good, I like the Swagger browser thing. It's hard to know whether the model definition here is good because it gives us a layer of indirection for stability, or bad because we have to define the model in another place and keep it in sync. A little of both I guess. Seems like we could probably codegen the enums at least.

While you're doing this you might look at https://github.com/twilio/guardrail which will generate code from an OpenAPI document. I haven't looked at it closely but I know the maintainer and I'm sure he would take contributions if we needed to tweak things.

You might also look at the Rho DSL for http4s, which can in principle generate OpenAPI doc from the application endpoints but I have never seen an example of it actually working. Would be good to chase that down and see if it's realistic or not.

cquiroz:
How easy would it be to create clients for OpenAPI for other languages, specifically Python for our users

swalker2m:
> How easy would it be to create clients for OpenAPI for other languages, specifically Python for our users

That's one of the features of Swaggerhub, but I haven't tried it.  I did try to generate a dummy server though (one in Java and one in Scala) but neither compiled because of syntax errors.  That doesn't inspire confidence in their code generation.  Since it is an open specification, there are probably other code generation tools (like guardrail) as well though.

cquiroz:
I think multi language support the main reason for GraphQL/OpenAPI. The scala.js apps don't really need it

swalker2m:
Just a minor update: I took a look at guardrail.  It appears that it [only supports OpenAPI 2](https://github.com/twilio/guardrail/issues/54) at the moment.

swalker2m:
> It's hard to know whether the model definition here is good because it gives us a layer of indirection for stability, or bad because we have to define the model in another place and keep it in sync.

Right.  Maybe this is what you meant, but my thinking was that we would do our own JSON codecs instead of relying upon automatic derivation.  The JSON produced by our `gem.json.instances` might be a bit awkward to use as is if you're writing, say, a python client.

cquiroz:
I'm closing this as we are not going to use it

PR #757, opened 2018-12-18
SEQNG-836 NIRI Epics code.

jluhrs:
Code to set parameters, trigger commands, and read back statuses from NIRI EPICS channels.

### DIFF
--- a/modules/seqexec/server/src/main/java/edu/gemini/seqexec/server/niri/BuiltInROI.java
+++ b/modules/seqexec/server/src/main/java/edu/gemini/seqexec/server/niri/BuiltInROI.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+ * For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+package edu.gemini.seqexec.server.niri;
+
+public enum BuiltInROI {
+    FullFrame("1024"),
+    Central768("768"),
+    Central512("512"),
+    Centrla256("256"),
+    Central128("128"),
+    Spec1024x512("1024x512");
+
+    final String name;
+
+    BuiltInROI(String name) {
+        this.name = name;
+    }
+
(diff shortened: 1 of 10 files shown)


modules/seqexec/server/src/main/java/edu/gemini/seqexec/server/niri/BuiltInROI.java
```diff
@@ -0,0 +1,26 @@
...
+
+public enum BuiltInROI {
+    FullFrame("1024"),
+    Central768("768"),
+    Central512("512"),
+    Centrla256("256"),
+    Central128("128"),
+    Spec1024x512("1024x512");
+
+    String name;
```
sraaphorst: Shouldn't we make these `final`?

modules/seqexec/server/src/main/resources/Niri.xml
```diff
@@ -0,0 +1,297 @@
...
+            <parameter name="exposureTime">
+                <channel>dcSetup:cad.F</channel>
+                <type>DOUBLE</type>
+                <description>Integration time for IS</description>
+            </parameter>
+            <!--<parameter name="builtinROI">-->
+                <!--<channel>roi:menu</channel>-->
+                <!--<type>STRING</type>-->
+                <!--<description>Builtin ROI</description>-->
+            <!--</parameter>-->
```
sraaphorst: Why are so many of the parameters commented out, out of curiosity?

jluhrs: The file is created by importing the old Seqexec configuration files. I have to take out the Enum channels because the XML format does not have support them (in the old Seqexec, all of them are read as strings). Those channels are created in the code, but I left them in the XML because of their info.

modules/seqexec/server/src/main/java/edu/gemini/seqexec/server/niri/BuiltInROI.java
jluhrs: Yes.

PR #758, opened 2018-12-19
SEQNG-818 Fixed GNIRS bias check

jluhrs:
There was an issue with the check to decide if the GNIRS bias had to be reapplied. The value read from the instrument has the opposite sign.
I got a bit sidetracked. I moved the GNIRS function to check parameters before applying them, to `EpicsUtil`, and changed the GMOS code to use it. That is in the first commit.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosController.scala
@@ -56,7 +56,8 @@ object GmosController {
     @SuppressWarnings(Array("org.wartremover.warts.LeakingSealed"))
     case class BuiltInFPU(fpu: T#FPU) extends GmosFPU
 
-    case class GmosDisperser(disperser: T#Disperser, order: Option[DisperserOrder], lambda: Option[Length])
+    case class GmosDisperser(disperser: T#Disperser, order: Option[DisperserOrder],
+      lambda: Option[Length])
 
     case class CCConfig(
       filter: T#Filter,
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -3,31 +3,23 @@
 
 package seqexec.server.gmos
 
-import seqexec.model.dhs.ImageFileId
-import seqexec.server.EpicsCodex.EncodeEpicsValue
-import seqexec.server.gmos.GmosController.Config.{Beam, InBeam, OutOfBeam, ROI}
-import seqexec.server.EpicsCodex
-import seqexec.server.EpicsUtil
-import seqexec.server.EpicsUtil.smartSetParam
-import seqexec.server.ObserveCommand
-import seqexec.server.Progress
-import seqexec.server.SeqAction
-import seqexec.server.SeqexecFailure
-import edu.gemini.spModel.gemini.gmos.GmosCommonType.AmpReadMode
-import edu.gemini.spModel.gemini.gmos.GmosCommonType.AmpGain
-import edu.gemini.spModel.gemini.gmos.GmosCommonType.AmpCount
-import edu.gemini.spModel.gemini.gmos.GmosCommonType.BuiltinROI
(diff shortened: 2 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosController.scala
```diff
@@ -56,7 +56,8 @@ object GmosController {
     @SuppressWarnings(Array("org.wartremover.warts.LeakingSealed"))
     case class BuiltInFPU(fpu: T#FPU) extends GmosFPU
 
-    case class GmosDisperser(disperser: T#Disperser, order: Option[DisperserOrder], lambda: Option[Length])
+    case class GmosDisperser(disperser: T#Disperser, order: Option[DisperserOrder],
+      lambda: Option[Length])
 
     case class CCConfig(
```
sraaphorst: It doesn't look like we're extending any of these, so why not make them `final case class`?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -69,78 +61,117 @@ class GmosControllerEpics[T<:GmosController.SiteDependentTypes](encoders: GmosCo
...
         roiParameters(binning, i + 1, ROIValues.fromOCS(roi))
-      }.sequence.flatMap(_ => SeqAction.void)
+      }
+    }
+
+  private def roiParameters(binning: CCDBinning, index: Int, roi: Option[ROIValues])
+  : List[SeqAction[Unit]] = (roi, DC.rois.get(index)).mapN { (roi, r) =>
+    r.setCcdXstart1(roi.xStart.value) *>
+    r.setCcdXsize1(roi.xSize.value / binning.x.getValue) *>
+    r.setCcdYstart1(roi.yStart.value) *>
```
sraaphorst: I've never known what `*>` was used for before. This was very helpful.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -234,7 +235,8 @@ object GnirsControllerEpics extends GnirsController {
     val coaddsWriter = smartSetParam(config.coadds, epicsSys.numCoadds,
       dcCmd.setCoadds(config.coadds))
 
-    val biasWriter =smartSetDoubleParam(biasTolerance)(encode(config.wellDepth), epicsSys.detBias,
+    // Value read from the instrument is the negative of what was set
+    val biasWriter = smartSetDoubleParam(biasTolerance)(-encode(config.wellDepth), epicsSys.detBias,
```
sraaphorst: Are we sure this should change from `encode` to `-encode`?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: It's a `flatMap` that ignores the input.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
jluhrs: The arguments are: value too set, function to read back the applied value, function to apply the new value. The issue was that the value read back  from the instrument was the negative of the value set. Alternatively, I could have changed the sign of the second argument.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosController.scala
jluhrs: Making them `final` produces a compile error: "The outer reference in this type test cannot be checked at run time."

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
sraaphorst: Thanks for the explanation. That will be useful in the future for me, I'm sure. Looks good!

PR #759, opened 2018-12-20
SEQNG-841 GNIRS checks before observe.

jluhrs:
Check GNIRS connection to DHS and that its detector array is active before trying to start an exposure.


modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -252,37 +252,46 @@ object GnirsControllerEpics extends GnirsController {
 
   override def applyConfig(config: GnirsConfig): SeqAction[Unit] =
     SeqAction(Log.debug("Starting GNIRS configuration")) *>
+      (if(GnirsEpics.instance.dhsConnected.exists(identity)) SeqAction.void
+       else EitherT.right(IO(Log.warn("GNIRS is not connected to DHS")))
+      ) *>
+      (if(GnirsEpics.instance.arrayActive.exists(identity)) SeqAction.void
+       else EitherT.right(IO(Log.warn("GNIRS detector array is not active")))
+      ) *>
```
jluhrs: Put a warning in the log when configuring the instrument.

PR #760, opened 2018-12-20
SEQNG-784: Extract GiapiInstrumentController

sraaphorst:
A lot of code was reused between the `GhostController` and the `GpiController`, leading to the refactoring of this class to extract the common behaviour into a shared superclass.

@swalker2m has also pointed out that naming conventions were not being maintained (e.g. `GHOSTController`, which should have been `GhostController`), so as a minor addition to this PR, I fixed GHOST, GPI, etc. to use the first-letter-capitalized rule.

If you have any strong feelings about the capitalization, lets' address them in the comment section here.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/GiapiInstrumentController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/GiapiInstrumentController.scala
@@ -11,7 +11,7 @@ import giapi.client.commands.{CommandResult, CommandResultException, Configurati
 import org.log4s.getLogger
 import seqexec.model.dhs.ImageFileId
 import seqexec.server.SeqexecFailure.{Execution, SeqexecException}
-import seqexec.server.keywords.GDSClient
+import seqexec.server.keywords.GdsClient
 import squants.time.Time
 
 import scala.concurrent.duration._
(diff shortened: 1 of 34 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/GiapiInstrumentController.scala
```diff
@@ -0,0 +1,58 @@
...
+import seqexec.server.SeqexecFailure.{Execution, SeqexecException}
+import seqexec.server.keywords.GdsClient
+import squants.time.Time
+
+import scala.concurrent.duration._
+
+/**
+  * Superclass for all GIAPI instrument controllers.
+  */
+abstract class GiapiInstrumentController[F[_]: Sync, CFG, C <: GiapiClient[F]] {
```
cquiroz: Do you really need to restrict to `Sync` here?

sraaphorst: My understanding is yes, although I'm not sure why. If I remove the `Sync`, it curries, adding something like an `F => Sync` at the end. (I'm not 100% sure that that's it. I'll check and reply.)

PR #761, opened 2018-12-20
SEQNG-784: Added GhostConfig Eq operators.

sraaphorst:
As per my discussion with @tpolecat, I added the `Eq` operators for the subclasses of `GhostConfig` and `GhostConfig` at large.

(@tpolecat, is this what you had in mind?)

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GhostController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ghost/GhostController.scala
@@ -151,6 +151,8 @@ object GhostController {
     }
   }
 
+  private implicit val durationEq: Eq[Duration] = Eq.by(_.toMillis)
+
   object StandardResolutionMode {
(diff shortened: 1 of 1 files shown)


sraaphorst:
I think I may be overcomplicating it, but I'm not certain.

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GhostController.scala
```diff
@@ -270,7 +291,19 @@ object GhostController {
       }
     }
 
-    implicit val eq: Eq[GhostConfig] = Eq.fromUniversalEquals
+    import GhostController.{StandardResolutionMode => SRM, HighResolutionMode => HRM}
+    implicit val eq: Eq[GhostConfig] = Eq.instance { (a, b) =>
+      (a: Any, b: Any) match {
```
tpolecat: Just `(a, b) match { ... ` here.

PR #764, opened 2018-12-28
Refactor SeqexecEngine extracting the odb loading functions

cquiroz:
`SeqexecEngine` is a fairly large class at the core of the seqexec. This PR splits a part of it, the part reading the ODB queue to simplify things


modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
```diff
@@ -0,0 +1,121 @@
...
+
+    (loads ++ unloads).toList
+  }
+
+}
+
+object ODBSequencesLoader {
+
+  // TODO Parametrize the IO
+  private def toEngineSequence(
```
cquiroz: I tried to parametrize this into `F[_]` but it ripples across many other parts of the code. I may do it in another PR

PR #766, opened 2019-01-04
Functor.as/void

cquiroz:
Mechanical refactoring of places where `Functor.as` or `Functor.void` can be used instead of `map`

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -82,7 +82,7 @@ class Engine[D, U](stateL: Engine.State[D]) {
           case r                 =>
             singleRunFailed(c, Result.Error(s"Unhandled result for single run action: $r"))
         }
-      ).map[EventResult.Outcome](_ => EventResult.Ok)
+      ).as[EventResult.Outcome](EventResult.Ok)
     ).getOrElse(pure[EventResult.Outcome](EventResult.Failure))
 
   }
@@ -312,7 +312,7 @@ class Engine[D, U](stateL: Engine.State[D]) {
     ev match {
       case EventUser(ue)   => handleUserEvent(ue)
       case EventSystem(se) => handleSystemEvent(se)(ec).flatMap(x =>
-        userReact.applyOrElse(se, (_:SystemEvent) =>  unit).map(_ => x))
+        userReact.applyOrElse(se, (_:SystemEvent) =>  unit).as(x))
     }
   }
 
--- a/modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
+++ b/modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
@@ -192,7 +192,7 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
             q.enqueue1(Event.start[executionEngine.ConcreteTypes](seqId, user, clientId, always)),
             startedFlag.decrement,
             q.enqueue1(Event.nullEvent),
-            q.enqueue1(Event.getState[executionEngine.ConcreteTypes] { _ => Stream.eval(finishFlag.increment).map(_ => Event.nullEvent).some })
+            q.enqueue1(Event.getState[executionEngine.ConcreteTypes] { _ => Stream.eval(finishFlag.increment).as(Event.nullEvent).some })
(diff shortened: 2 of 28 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -82,7 +82,7 @@ class Engine[D, U](stateL: Engine.State[D]) {
           case r                 =>
             singleRunFailed(c, Result.Error(s"Unhandled result for single run action: $r"))
         }
-      ).map[EventResult.Outcome](_ => EventResult.Ok)
+      ).as[EventResult.Outcome](EventResult.Ok)
```
sraaphorst: What, out of curiosity, is the advantage of using `Functor.as` over `map`?

EDIT: Never mind. After scrolling through, I can see the obvious advantages. I didn't know about this before. Good to learn something new!

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
```diff
@@ -406,7 +406,7 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
 
     val c = ActionCoordsInSeq(stepId, ExecutionIndex(0), ActionIndex(0))
     val event = Event.modifyState[executionEngine.ConcreteTypes](
-      executionEngine.startSingle(ActionCoords(seqId, c)).map(_ => ())
+      executionEngine.startSingle(ActionCoords(seqId, c)).void
```
sraaphorst: 👍 Much cleaner.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -277,7 +277,7 @@ class SeqexecEngine(httpClient: Client[IO], settings: Settings[IO], sm: SeqexecM
   def addSequencesToQueue(q: EventQueue, qid: QueueId, seqIds: List[Observation.Id])
   : IO[Either[SeqexecFailure, Unit]] = q.enqueue1(
     Event.modifyState[executeEngine.ConcreteTypes](addSeqs(qid, seqIds)
-      .map[executeEngine.ConcreteTypes#EventData](_ => UpdateQueueAdd(qid, seqIds)))
+      .as[executeEngine.ConcreteTypes#EventData](UpdateQueueAdd(qid, seqIds)))
```
sraaphorst: Why this change? Do we want to always evaluate `UpdateQueueAdd`? Or will that not happen?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/RemoteRequestsHandler.scala
```diff
@@ -97,7 +97,7 @@ class RemoteRequestsHandler[M](modelRW: ModelRW[M, Option[ClientId]])
         Effect(
           SeqexecWebClient
             .resumeObs(id, step)
-            .map(_ => RunObsPause(id))
+            .as(RunObsPause(id))
```
sraaphorst: All these: same comment as above.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/SequenceExecutionHandler.scala
```diff
@@ -25,25 +25,31 @@ class SequenceExecutionHandler[M](
...
-          s.copy(metadata = s.metadata.copy(observer = Some(name)))
-        case s                        => s
-      })
+      val updateObserverE = Effect(
+        SeqexecWebClient.setObserver(sequenceId, name.value).as(NoAction))
+      val updatedSequences =
+        value.copy(sessionQueue = value.sessionQueue.collect {
+          case s if s.id === sequenceId =>
+            s.copy(metadata = s.metadata.copy(observer = Some(name)))
+          case s => s
```
sraaphorst: Not a big deal, but it looks like the spacing alignment was thrown off by your IDE.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: `UpdateQueueAdd` is an output event, that tells the client that a sequence was added to a queue.

PR #768, opened 2019-01-07
SEQNG-795: Split components for steps and config tables

cquiroz:
At the moment there is a single top component for both the steps and step config table and it makes the code more complex than it needs to be and limits our chances for improving performance for large sequences.

This PR removes that and I also did some larger refactoring removing some cruft accumulated along the way. In particular the model contains some optional code that we know is always present thus we can simplify a lot of code downstream

This PR doesn't remove or add functionality


modules/seqexec/web/client/src/main/resources/less/style.less
```diff
@@ -161,16 +161,12 @@ body {
   height: ~"calc(100vh - 44.8em)";
 }
 
-.SeqexecStyles-stepsListPaneWithControls {
+.SeqexecStyles-tableContainer {
   height: ~"calc(100% - 2.8em)";
```
tpolecat: I haven't seen this before. Does CSS do this now or is it a Less thing?

cquiroz: it is css. Basically it calculates the height as 100% minus something. In this case the height of the area where the buttons are.

It will resize the area when the browser is resized

PR #769, opened 2019-01-08
SEQNG-785: Make GIAPI configuration typeclass

sraaphorst:
Previously, when creating GIAPI configuration strings, we would just rely on the `Show` typeclass. Instead, we would like to have it based on a separate typeclass to deliberately force implementations to decide if they want to base it on `Show` or something else.

This PR introduces a new, simple typeclass called `GiapiConfig`, with default instances for frequently used base types (`String`, `Int`, `Double`) and a function to easily create a `GiapiConfig` from an instance of `Show` (although note that this is not `implicit`, so it must be purposefully done).

I kept this typeclass simple instead of going to the level of what is implemented in `Show`; if anyone feels that this should be more extensive, I don't mind trying to imply the contravariant and what not based on `Show`, although I am not exactly sure what the significance is or what benefit it gives us, I have to admit.


modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package giapi.client
+
+import cats.Show
+
+// Produce a configuration string for Giapi.
+trait GiapiConfig[T] {
+  def configuration(t: T): String
```
cquiroz: Maybe `configurationValue` or `configValue` could be more explicit?

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
```diff
@@ -0,0 +1,26 @@
...
+
+object GiapiConfig {
+  def apply[A](implicit instance: GiapiConfig[A]): GiapiConfig[A] = instance
+
+  implicit class GiapiConfigOps[A](value: A) {
+    def configuration(implicit w: GiapiConfig[A]): String =
+      w.configuration(value)
+  }
+
+  def fromShow[A: Show]: GiapiConfig[A] = Show[A].show(_)
```
cquiroz: 👍

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
```diff
@@ -0,0 +1,26 @@
...
+
+// Produce a configuration string for Giapi.
+trait GiapiConfig[T] {
+  def configuration(t: T): String
+}
+
+object GiapiConfig {
+  def apply[A](implicit instance: GiapiConfig[A]): GiapiConfig[A] = instance
+
+  implicit class GiapiConfigOps[A](value: A) {
```
cquiroz: i think this can be a value class with `extends AnyVal`

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
```diff
@@ -0,0 +1,26 @@
...
+    def configuration(implicit w: GiapiConfig[A]): String =
+      w.configuration(value)
+  }
+
+  def fromShow[A: Show]: GiapiConfig[A] = Show[A].show(_)
+
+  implicit val stringConfig: GiapiConfig[String] = t => t
+  implicit val intConfig: GiapiConfig[Int] = _.toString
+  implicit val doubleConfig: GiapiConfig[Double] = _.toString
```
cquiroz: I wonder if this should take precision. It is not needed in GPI and probably not in GHOST but maybe they should take one. I can imagine an instrument choking when sent a string with many many decimals
We can keep it as it is for now

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
cquiroz: Maybe if we move this to `giapi.client` we can simplify the imports. Or perhaps create a `giapi.client.implicits` package containing it

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
tpolecat: More problematic (maybe) is scientific notation.

```scala
@ val a = 0.0001 
a: Double = 1.0E-4

@ a.toString 
res22: String = "1.0E-4"

@ f"$a%6.6f" 
res23: String = "0.000100"
```

cquiroz: Good point

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
tpolecat: I would prefer to add this as a distinct syntax class following the pattern [here](https://github.com/gemini-hlsw/ocs3/tree/develop/modules/core/shared/src/main/scala/gem/syntax). It seems like a lot of boilerplate but I think it pays off to be consistent in how we handle this.

tpolecat:
A few comments. Good idea 👍

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
cquiroz: Do you mean in a `giapi.client.syntax` package? Sounds good

tpolecat: Yes, so `import giapi.client.syntax.giapiconfig._` or `import giapi.client.syntax.all._`.

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
sraaphorst: Sure. I think that sounds better.

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
sraaphorst: Okay. I'll make that change.

modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
sraaphorst: Yes, that had occurred to me but I wasn't sure what to do with it. I'm not sure how `Show` treats `Double` but I'm curious.

tpolecat:
```scala
@ 0.0001 
res1: Double = 1.0E-4

@ 0.0001.show 
res2: String = "1.0E-4"
```

sraaphorst:
@cquiroz I'm fine with going with `f"..."`. Do you have any specific recommendations? @tpolecat's `f"$a%6.6f"`?

tpolecat:
I think `%1.6f` makes more sense. First digit is the minimum width, second is the number of digits after the decimal point. Nothing special about `6` though, could be more.

sraaphorst:
@tpolecat I'll make it so.

PR #770, opened 2019-01-09
SEQNG-835 Implemented EPICS control for NIRI.

jluhrs:
Whit this we should be able to take images with NIRI. I had to redefine some of the Enums, because I used the wrong file to get their EPICS values. There are also some adjustments to the configuration data model.


modules/seqexec/server/src/main/java/edu/gemini/seqexec/server/niri/Disperser.java
```diff
@@ -6,12 +6,17 @@
 package edu.gemini.seqexec.server.niri;
 
 public enum Disperser {
```
cquiroz: Are these LUT to write to the epics channel?

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -264,4 +264,10 @@ object EpicsUtil {
     }).dropWhile(_.remaining.self.value === 0.0) // drop leading zeros
       .takeThrough(_.remaining.self.value > 0.0) // drop all tailing zeros but the first one
 
+  def removePartName(s: String): String = {
```
cquiroz: Could you document this a bit? The code is clear but not why do you need to remove a part

modules/seqexec/server/src/main/scala/seqexec/server/niri/NiriControllerEpics.scala
```diff
@@ -0,0 +1,308 @@
...
+   */
+  private def currentFilter: Option[String] = {
+    val filter1 = epicsSys.filter1
+    val filter2 = epicsSys.filter2
+    val filter3 = epicsSys.filter3
+    val Open = "open"
+
+    (filter1, filter2, filter3).mapN{ (f1, f2, f3) =>
+      val l = List(f1, f2, f3).filterNot(_ === Open)
+      if(l.length === 1) l.headOption
```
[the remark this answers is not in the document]
jluhrs: Only one can be not open. If more than one is not open, the global filter position is undefined.

PR #773, opened 2019-01-09
Fix to keep completed tabs

cquiroz:
We have to keep tabs when they are completed on the client while they are removed on the server. The last PR broke that but this PR restores that functionality

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
@@ -128,13 +128,14 @@ final case class SequencesOnDisplay(tabs: Zipper[SeqexecTab]) {
         val curTableState = tab
           .map(_.tableState)
           .getOrElse(StepsTable.State.InitialTableState)
-        val s = tag[InstrumentSequenceTab.LoadedSV][SequenceView](x).asRight
-        val seq = tab.map{t => if (t.isComplete) tag[InstrumentSequenceTab.CompletedSV][SequenceView](x).asLeft else s}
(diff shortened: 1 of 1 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
```diff
@@ -128,11 +128,13 @@ final case class SequencesOnDisplay(tabs: Zipper[SeqexecTab]) {
...
           .map(_.tableState)
           .getOrElse(StepsTable.State.InitialTableState)
+        val s = tag[InstrumentSequenceTab.LoadedSV][SequenceView](x).asRight
+        val seq = tab.map{t => if (t.isComplete) tag[InstrumentSequenceTab.CompletedSV][SequenceView](x).asLeft else s}
         val stepConfig   = tab.flatMap(_.stepConfig)
         val selectedStep = tab.flatMap(_.selectedStep)
         InstrumentSequenceTab(
           x.metadata.instrument,
-          tag[InstrumentSequenceTab.LoadedSV][SequenceView](x).asRight,
+          seq.getOrElse(s),
```
tpolecat: The logic here with `s` and `seq` is kind of weird … could you simplify? Something like this maybe?

```scala
lazy val left  = tag[InstrumentSequenceTab.CompletedSV][SequenceView](x)
lazy val right = tag[InstrumentSequenceTab.LoadedSV][SequenceView](x)
val seq = tab.filter(_.isComplete).as(left).toLeft(right)
```

cquiroz: Sure, sounds like a good suggestion

PR #774, opened 2019-01-11
SEQNG-837 NIRI FITS keywords.

jluhrs:
All NIRI specific keywords. Values are read from the instrument.
There was a LUT for DHS keywords that did not conform to FITS keywords rules. NIRI added some on its own. I thought it would be better to I moved the LUT to the DB. This way, it can be easily expanded for new instruments.

Do not merge before PR #673


modules/sql/src/main/resources/db/migration/V058__Dhs_KeywordNames.sql
```diff
@@ -0,0 +1,18 @@
+---------------------------
+-- DHS Keyword Names
+-- This table list the DHS exceptions, of keywords names that don't follow
+-- the keyword name restrictions
+---------------------------
+
+CREATE TABLE e_dhs_keyword_names (
+    keyword identifier  PRIMARY KEY NOT NULL REFERENCES e_fits_keyword_names ON DELETE CASCADE,
+    name text             NOT NULL CHECK (name ~ '[a-zA-Z0-9_-]*')
```
cquiroz: So the reason this is `text` is that ait can be longer than 8 chars, but does it need to be `text`. Could it be `VARCHAR(20)` or the max allowed by the DHS?

modules/sql/src/main/scala/gem/sql/EnumDef.scala
```diff
@@ -79,6 +80,8 @@ object EnumDef {
     implicit def caseMagnitudeValue      [S <: Symbol] = at[(S, MagnitudeValue)        ] { case (s, _) => s"  val ${s.name}: MagnitudeValue" }
     implicit def caseOptionMagnitudeValue[S <: Symbol] = at[(S, Option[MagnitudeValue])] { case (s, _) => s"  val ${s.name}: Option[MagnitudeValue]" }
     implicit def caseGnirsPixelScale[S <: Symbol] = at[(S, GnirsPixelScale)] { case (s, _) => s"  val ${s.name}: GnirsPixelScale" }
+    implicit def caseKeywordName[S <: Symbol] = at[(S, KeywordName)] { case (s, _) => s"  val" +
```
cquiroz: It maybe more readable on the same line?

modules/sql/src/main/resources/db/migration/V058__Dhs_KeywordNames.sql
jluhrs: You'll make me look into the DHS code to search for the limit

cquiroz: How disappointing, I thought you'd know that kind of things by heart

modules/sql/src/main/resources/db/migration/V059__Niri_KeywordNames.sql
```diff
@@ -0,0 +1,47 @@
+---------------------------
+-- NIRI FITS Keyword Names
+---------------------------
+
+INSERT INTO e_fits_keyword_names(id, name)
```
swalker2m: Are keywords for all the instruments going to be put into the same table?

modules/sql/src/main/resources/db/migration/V059__Niri_KeywordNames.sql
```diff
@@ -0,0 +1,47 @@
...
+	('A_TMOUNT', 'A_TMOUNT'),
+	('A_VDDCL1', 'A_VDDCL1'),
+	('A_VDDCL2', 'A_VDDCL2'),
+	('A_VDDUC', 'A_VDDUC'),
+	('A_VDET', 'A_VDET'),
+	('A_VGGCL1', 'A_VGGCL1'),
+	('A_VGGCL2', 'A_VGGCL2'),
+	('A_VSET', 'A_VSET');
+
+INSERT INTO e_dhs_keyword_names(keyword, name)
```
swalker2m: What's the difference between DHS keywords and FITS keywords?

modules/sql/src/main/resources/db/migration/V059__Niri_KeywordNames.sql
```diff
@@ -0,0 +1,47 @@
+---------------------------
+-- NIRI FITS Keyword Names
+---------------------------
+
+INSERT INTO e_fits_keyword_names(id, name)
+VALUES
+	('COADDS', 'COADDS'),
```
swalker2m: The all caps is kind of Java-ish and we've avoided it elsewhere but I think it may be expected if they are FITS keywords.

modules/core/shared/src/main/scala/gem/enum/KeywordName.scala
```diff
@@ -668,10 +668,45 @@ object KeywordName {
...
...QTWN77, REQTWP77, REQTWS78, REQTWD78, REQTWN78, REQTWP78, REQTWS79, REQTWD79, REQTWN79, REQTWP79, REQTWS80, REQTWD80, REQTWN80, REQTWP80, REQTWS81, REQTWD81, REQTWN81, REQTWP81, REQTWS82, REQTWD82, REQTWN82, REQTWP82, REQTWS83, REQTWD83, REQTWN83, REQTWP83, REQTWS84, REQTWD84, REQTWN84, REQTWP84, REQTWS85, REQTWD85, REQTWN85, REQTWP85, REQTWS86, REQTWD86, REQTWN86, REQTWP86, REQTWS87, REQTWD87, REQTWN87, REQTWP87, REQTWS88, REQTWD88, REQTWN88, REQTWP88, REQTWS89, REQTWD89, REQTWN89, REQTWP89, REQTWS90, REQTWD90, REQTWN90, REQTWP90, REQTWS91, REQTWD91, REQTWN91, REQTWP91, REQTWS92, REQTWD92, REQTWN92, REQTWP92, REQTWS93, REQTWD93, REQTWN93, REQTWP93, REQTWS94, REQTWD94, REQTWN94, REQTWP94, REQTWS95, REQTWD95, REQTWN95, REQTWP95, REQTWS96, REQTWD96, REQTWN96, REQTWP96, REQTWS97, REQTWD97, REQTWN97, REQTWP97, REQTWS98, REQTWD98, REQTWN98, REQTWP98, REQTWS99, REQTWD99, REQTWN99, REQTWP99, COADDS, EXPTIME, FILTER3, FOCUSNAM, FOCUSPOS, FPMASK, BEAMSPLT, WINDCOVR, FRMSPCYCL, HDRTIMING, LNRS, MODE, NDAVGS, PVIEW, TDETABS, TIME, TMOUNT, UCODENAM, UCODETYP, VDDCL1, VDDCL2, VDDUC, VDET, VGGCL1, VGGCL2, VSET, A_TDETABS, A_TMOUNT, A_VDDCL1, A_VDDCL2, A_VDDUC, A_VDET, A_VGGCL1, A_VGGCL2, A_VSET)
```
swalker2m: There are so many of these.  Does it make sense to mix instruments, if that is what is going on?

modules/sql/src/main/resources/db/migration/V059__Niri_KeywordNames.sql
jluhrs: FITS keywords follow the FITS keywords standard (i.e. up to eight capital letters). In the DHS, the keyword identifier has no such restrictions, although most of them have valid FITS keyword names. There is an additional field in the DHS dictionary for the actual FITS value that DHS uses when writing the keyword to the file. But inside Seqexec, all the keywords are FITS keywords (it had to handle DHS and GDS systems, so we use the more restrictive convention). The DHS table is actually used to translate the FITS keyword names to the ones DHS understand. It contains only the exceptions.

modules/sql/src/main/resources/db/migration/V059__Niri_KeywordNames.sql
jluhrs: Yes, but my plan is to have the definitions in separate files, one per instrument.

modules/sql/src/main/resources/db/migration/V059__Niri_KeywordNames.sql
jluhrs: Yes, that is exactly the case.

modules/core/shared/src/main/scala/gem/enum/KeywordName.scala
jluhrs: I wouldn't worry too much about this list growing large, because it is generated code. I care more about the keeping things readable and tidy in the sql files.

modules/seqexec/server/src/main/scala/seqexec/server/TranslateSettings.scala
```diff
@@ -9,5 +9,6 @@ final case class TranslateSettings(
   gwsKeywords:   Boolean,
   gcalKeywords:  Boolean,
   gmosKeywords:  Boolean,
-  gnirsKeywords: Boolean
+  gnirsKeywords: Boolean,
+  niriKeywords:  Boolean
```
cquiroz: Do we need to update the `app*.conf` files?

jluhrs: No, this value is calculated from another app.conf setting, that was already added for Niri.

PR #776, opened 2019-01-14
SEQNG-849: Seqexec stops working after closing the web socket

cquiroz:
This bug was solved by upgrading fs2. Version 10.0.1 fixes this issue though it is a bit unclear exactly where. I suspect this fix solves it:
https://github.com/functional-streams-for-scala/fs2/pull/1311

I also updated http4s

This has been tested against an instrument


tpolecat:
\o/

arturog8m:
👏 🎆

PR #777, opened 2019-01-15
SEQNG-853 Science fold name depends on instrument configuration.

jluhrs:
The science fold position name is a String composed by a light source prefix (like "gcal" or "ao"), followed by a light sink (usually an instrument name), and a port number. Until now, the code assumed fixed names for each instrument. But Niri uses a different name depending on the internal camera used. This PR makes the instrument name used to compose the SF position name depend on the instrument configuration.
The possible light sink names were compiled into a table used to produce an enum.

As a note, every instrument class now has a bunch of methods that depend on a `Config`. Maybe it should be moved to be a class parameter? Something to consider in the future.


modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2.scala
```diff
@@ -12,13 +12,16 @@ import edu.gemini.spModel.gemini.flamingos2.Flamingos2._
 import edu.gemini.spModel.obscomp.InstConstants.{DARK_OBSERVE_TYPE, OBSERVE_TYPE_PROP}
 import edu.gemini.spModel.seqcomp.SeqConfigNames._
 import java.lang.{Double => JDouble}
+
+import gem.enum.LightSinkName
```
cquiroz: Why the lines before/after?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -322,11 +327,13 @@ object TcsControllerEpics extends TcsController {
...
   def setScienceFoldConfig(sfPos: ScienceFoldPosition): SeqAction[Unit] = sfPos match {
     case ScienceFoldPosition.Parked => TcsEpics.instance.scienceFoldParkCmd.mark
-    case p@ScienceFoldPosition.Position(LightSource.Sky, inst) => getInstPort(inst).flatMap(port =>
-      if (port === 1) TcsEpics.instance.scienceFoldParkCmd.mark.some
-      else encode(p).map(TcsEpics.instance.scienceFoldPosCmd.setScfold)
-    ).getOrElse(SeqAction(()))
-    case p: ScienceFoldPosition.Position => encode(p).map(TcsEpics.instance.scienceFoldPosCmd.setScfold).getOrElse(SeqAction(()))
+    case p@ScienceFoldPosition.Position(LightSource.Sky, sink) =>
+      instrumentFromSinkName(sink).flatMap(getInstPort).flatMap(port =>
+       if (port === 1) TcsEpics.instance.scienceFoldParkCmd.mark.some
```
cquiroz: Can we use a constant rather than `1`?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -347,7 +354,8 @@ object TcsControllerEpics extends TcsController {
       case HrwfsConfig.Manual(h) => setHRPickupConfig(h).some
       case HrwfsConfig.Auto      => {
         c.sfPos.flatMap{
-          case ScienceFoldPosition.Position(_, inst) => getInstPort(inst).map(_ === 1)
+          case ScienceFoldPosition.Position(_, sink) => instrumentFromSinkName(sink)
+            .flatMap(getInstPort).map(_ === 1)
```
cquiroz: Same as above

modules/sql/src/main/resources/db/migration/V060__Light_Sinks.sql
```diff
@@ -0,0 +1,29 @@
+---------------------------
+-- Science Fold Light Sinks
+-- The science fold directs the light to one of this named positions
+---------------------------
+
+CREATE TABLE e_light_sink_names (
+    id identifier  PRIMARY KEY NOT NULL ,
+    name VARCHAR(15)    NOT NULL CHECK (name ~ '[a-z][a-z0-9]*')
```
cquiroz: if name is unique perhaps we don't need the `id`? @tpolecat @swalker2m WDYT?

modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2.scala
jluhrs: Idea messing with my imports.

modules/sql/src/main/resources/db/migration/V060__Light_Sinks.sql
swalker2m: You need the `id` to map to/from the enumerated type.  I suppose you wouldn't have to have a `name` but I assume the name is changeable and/or doesn't conform to the identifier convention.  In fact, the convention for other enums is to have a `short_name` *and* a descriptive `long_name`.

jluhrs: In this case, the name does not conform to identifier conventions.

PR #780, opened 2019-01-15
Shape Geometry

swalker2m:
This PR is a WIP and not intended to be merged yet, if ever.   I wanted to show what I've been playing with, document it before I leave for a few days, and see whether it makes any sense.

The issue is defining shapes for use in AGS, etc.  This would include probe arms, probe ranges, and detector areas at least.  There are two main operations needed for AGS:

1) Determine whether a point is contained in a shape (guide star reachability)
2) Determine the area of a shape relative to another shape (vignetting comparisons)

```scala
trait Shape {

  /**
   * Determines whether the given position is contained in the shape.
   */
  def contains(o: Offset): Boolean

  /**
   * Area of the shape. The value is in microarcseconds angular separation^2
   * but the absolute value is less useful than comparing across multiple
   * shapes (to determine minimum vignetting).
   */
  def area: Long

}
```

Instead of using AWT though, I wanted to find a geometry library either in Scala (fail) or else with Java and JavaScript implementations (see [JTS](https://github.com/locationtech/jts)/[JSTS](https://github.com/bjornharrtell/jsts)) .  So the idea is to describe the Shape in an ADT and then interpret it down to a specific shape implementation in a language/library-specific interpreter.

```scala
sealed trait ShapeExpression {

  def eval(implicit ev: ShapeInterpreter): Shape =
    ev.interpret(this)

}

object ShapeExpression {

  final case class Ellipse(a: Offset, b: Offset)                         extends ShapeExpression
  final case class Polygon(os: List[Offset])                             extends ShapeExpression
  final case class Rectangle(a: Offset, b: Offset)                       extends ShapeExpression

  final case class Difference(a: ShapeExpression, b: ShapeExpression)    extends ShapeExpression
  final case class Intersection(a: ShapeExpression, b: ShapeExpression)  extends ShapeExpression
  final case class Union(a: ShapeExpression, b: ShapeExpression)         extends ShapeExpression

  final case class Rotate(e: ShapeExpression, a: Angle)                  extends ShapeExpression
  final case class Translate(e: ShapeExpression, o: Offset)              extends ShapeExpression

}
```

Note that the shapes are defined with `Offset` positions, which seemed like the natural way to do it.  Usually they are specified in terms of arcsec sizes and at offset position locations.  With some syntax (not shown) here's the JTS/JVM implementation of the interpreter:

```scala
object JtsShapeInterpreter extends ShapeInterpreter {

  private def toGeometry(e: ShapeExpression): Geometry =
    e match {
      // Constructors
      case Ellipse(a, b)      => a.shapeFactory(b).createEllipse
      case Polygon(os)        => Jts.geometryFactory.createPolygon(os.map(_.coordinate).toArray)
      case Rectangle(a, b)    => a.shapeFactory(b).createRectangle

      // Combinations
      case Difference(a, b)   => toGeometry(a).difference(toGeometry(b))
      case Intersection(a, b) => toGeometry(a).intersection(toGeometry(b))
      case Union(a, b)        => toGeometry(a).union(toGeometry(b))

      // Transformations
      case Rotate(e, a)       =>
        AffineTransformation
          .rotationInstance(a.toDoubleDegrees)
          .transform(toGeometry(e))

      case Translate(e, o)    =>
        val c = o.coordinate
        AffineTransformation
          .translationInstance(c.x, c.y)
          .transform(toGeometry(e))
    }

  override def interpret(e: ShapeExpression): Shape =
    JtsShape(toGeometry(e))
}
```
Hiding some syntax support, the GMOS OIWFS probe arm can be defined as a `ShapeExpression`:
```scala
  val PickoffArmLength: Angle      = 358460.mas
  val PickoffMirrorSize: Angle     =  20000.mas
  val ProbeArmLength: Angle        = PickoffArmLength - (PickoffMirrorSize / 2)
  val ProbeArmTaperedWidth: Angle  =  15000.mas
  val ProbeArmTaperedLength: Angle = 180000.mas

  private val arm: ShapeExpression = {
    val hm  = PickoffMirrorSize    / 2
    val htw = ProbeArmTaperedWidth / 2

    val p0 = (-hm,                           htw  )
    val p1 = (p0._1 - ProbeArmTaperedLength, hm   )
    val p2 = (p0._1 - ProbeArmLength,        p1._2)
    val p3 = (p2._1,                        -hm   )
    val p4 = (p1._1,                         p3._2)
    val p5 = (p0._1,                        -htw  )

    ShapeExpression.polygon(p0, p1, p2, p3, p4, p5, p0)
  }

  private val pickoff: ShapeExpression = {
    val s = PickoffMirrorSize / 2
    ShapeExpression.rectangle((s, s), (s - PickoffMirrorSize, s - PickoffMirrorSize))
  }

  /**
   * Description of the GMOS OIWFS probe arm with the pickoff mirror centered
   * at the base position.
   */
  val shape: ShapeExpression =
    arm ∪ pickoff
```

Note `arm ∪ pickoff` to combine the probe arm proper with the pickoff mirror.  Here's an AWT plot obtained from the resulting shape with the pickoff mirror centered at the origin (marked with a +).

screen shot 2019-01-15 at 16 17 02

where the `toAwt` below is only defined in the `JtsShape` implementation and only needed to visualize the shape for the demo:

```scala
    import gem.geom.jts.interpreter._

    override def paint(g: Graphics): Unit = {
      val g2d = g.asInstanceOf[Graphics2D]

      GmosOiwfsProbeArm.shape.eval match {
        case jts: JtsShape => g2d.draw(jts.toAwt)
        case x             => sys.error(s"Whoa unexpected shape type: $x")
      }
  }
```

### DIFF
--- a/modules/core/jvm/src/main/scala/gem/geom/jts/JtsShape.scala
+++ b/modules/core/jvm/src/main/scala/gem/geom/jts/JtsShape.scala
@@ -1,0 +1,50 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.geom.jts
+
+import gem.geom.jts.syntax.all._
+
+import gem.geom.Shape
+import gem.math.Offset
+import org.locationtech.jts.awt.{PointTransformation, ShapeWriter}
+import org.locationtech.jts.geom.{Coordinate, Geometry}
+
+import java.awt.geom.Point2D;
+
+/**
+ * JVM / JTS implementation of Shape.
+ */
+final case class JtsShape(g: Geometry) extends Shape {
+
+  import JtsShape._
+
+  override def contains(o: Offset): Boolean =
+    g.contains(o.point)
+
+  override def area: Long =
+    g.getArea.round
+
+  /**
+   * Converts to the AWT equivalent at 1 arcsec / pixel.  This is specific to
+   * the Java implementation.  (The scale should probably be explicitly
+   * specified.)
+   */
+  def toAwt: java.awt.Shape =
+    awtWriter.toShape(g)
+}
+
+object JtsShape {
+
+  // Converts points to AWT at arcsec / pixel scale.
+  private val toArcsec: PointTransformation =
+    new PointTransformation {
+      def transform(s: Coordinate, d: Point2D): Unit =
+        // With AWT canvas, y increases towards the bottom
+        d.setLocation(s.x / 1000000.0, - s.y / 1000000.0)
+    }
+
+  private val awtWriter: ShapeWriter =
+    new ShapeWriter(toArcsec)
+
+}
--- a/modules/core/shared/src/main/scala/gem/geom/Shape.scala
+++ b/modules/core/shared/src/main/scala/gem/geom/Shape.scala
@@ -1,0 +1,29 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.geom
+
+import gem.math.Offset
+
+/**
+ * Shape description usable for testing point inclusion and calculating the
+ * area.  Point inclusion is used to determine, for instance, guide star
+ * reachability for a probe range.  Comparative area is usable when calculating
+ * which guide star options produce the minimum vignetting (probe arm shadow
+ * intersected with science area).
+ */
+trait Shape {
+
+  /**
+   * Determines whether the given position is contained in the shape.
+   */
+  def contains(o: Offset): Boolean
+
+  /**
+   * Area of the shape. The value is in microarcseconds angular separation^2
+   * but the absolute value is less useful than comparing across multiple
+   * shapes (to determine minimum vignetting).
+   */
+  def area: Long
+
+}
--- a/modules/core/shared/src/main/scala/gem/syntax/ShapeExpression.scala
+++ b/modules/core/shared/src/main/scala/gem/syntax/ShapeExpression.scala
@@ -1,0 +1,73 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem.syntax
+
+import gem.geom.ShapeExpression
+import gem.geom.ShapeExpression._
+
+import gem.math.{Angle, Offset}
+
+final class ShapeExpressionOps(val self: ShapeExpression) extends AnyVal {
+
+  def difference(that: ShapeExpression): ShapeExpression =
+    Difference(self, that)
+
+  def intersection(that: ShapeExpression): ShapeExpression =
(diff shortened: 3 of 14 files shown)


swalker2m:
Hmm, the AWT imports for the demo don't pass build checks but no matter, they aren't critical to the idea.  Why are they "illegal" though in `coreJVM`?

modules/core/shared/src/main/scala/gem/geom/Shape.scala
```diff
@@ -0,0 +1,29 @@
...
+   * Determines whether the given position is contained in the shape.
+   */
+  def contains(o: Offset): Boolean
+
+  /**
+   * Area of the shape. The value is in microarcseconds angular separation^2
+   * but the absolute value is less useful than comparing across multiple
+   * shapes (to determine minimum vignetting).
+   */
+  def area: Long
```
sraaphorst: What will this be used for? Or is it just a requirement of JTS to have this defined?

modules/core/jvm/src/main/scala/gem/geom/jts/JtsShape.scala
```diff
@@ -0,0 +1,50 @@
...
+  // Converts points to AWT at arcsec / pixel scale.
+  private val toArcsec: PointTransformation =
+    new PointTransformation {
+      def transform(s: Coordinate, d: Point2D): Unit =
+        // With AWT canvas, y increases towards the bottom
+        d.setLocation(s.x / 1000000.0, - s.y / 1000000.0)
+    }
+
+  private val awtWriter: ShapeWriter =
+    new ShapeWriter(toArcsec)
```
sraaphorst: I like the simplicity here of converting to a Java `Shape`. Very nice.

swalker2m: I'm not sure we'd ever use that but maybe so.  I suppose we could draw TPE-like images in the server, convert them to PNG or something with `javax.imageio.ImageIO` and show them in the client.

modules/core/shared/src/main/scala/gem/syntax/ShapeExpression.scala
```diff
@@ -0,0 +1,73 @@
...
+    Rotate(self, a)
+
+  def translate(o: Offset): ShapeExpression =
+    Translate(self, o)
+
+  def union(that: ShapeExpression): ShapeExpression =
+    Union(self, that)
+
+  def -(that: ShapeExpression): ShapeExpression =
+    difference(that)
```
sraaphorst: Isn't the set difference operator typically `∖`?

sraaphorst: I guess it doesn't really matter. People will probably better know what `-` means.

sraaphorst:
Overall, I think this looks really good, and much easier and cleaner to work with than what we had before trying to do this through Java code. Definitely a step in the right direction. Some of the mutability (e.g. `GeometricShapeFactory`) is unfortunate, but I'm guessing that that can't be avoided, and it seems buried away so that we don't have to think about it.

tpolecat:
The DSL approach seems right to me; i.e., define an AST for the kinds of shapes we need to deal with, and interpret into something concrete at the end of the world. It insulates us from the details of JTS, which is good. 👍

cquiroz:
Moved to gsp-math

PR #782, opened 2019-01-16
SEQNG-840: Display NIRI step

cquiroz:
This PR adds support to show NIRI filter and camera

seqexec - gs-2018b-q-0-97 2019-01-16 12-20-45


modules/seqexec/model/src/main/scala/seqexec/model/enumerations.scala
```diff
@@ -93,6 +93,57 @@ object enumerations {
       "DARK" -> "Dark"
     )
 
+    val Niri: Map[String, String] = Map(
```
cquiroz: The reason to have these hardcoded is that we can't pull ocs2 files into the client. We may want to get these from the gem db eventually but at the moment Niri isn't supported there

cquiroz:
That is a known issue the end part will be truncated. 

There is a pending task about calculating the widths and making the cols  resizable

Issue #538, opened 2018-06-14
Remove argonaut dependency


tpolecat:
DHS client and seqexec authentication are using argonaut. We need to switch these to use Circe.

cquiroz:
This doesn’t bother me much, we probably have more pressing issues 😄

cquiroz:
But feel free to send a PR if you want. We could consider switching to hhtp4s client for that

tpolecat:
Yes, that too. Minimizing dependencies is a good use of time, especially since this one pulls in scalaz transitively.

PR #783, opened 2019-01-17
SEQNG-856 Rewrote DhsClientHttp to use http4s and circe.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
```diff
@@ -8,113 +8,85 @@ import java.time.format.DateTimeFormatter
...
-      case "success" =>
-        DecodeResult.ok(TrySeq(()))
-      case "error"   =>
-        (r --\ "errors").as[List[Error]].map(
-          l => TrySeq.fail[Unit](SeqexecFailure.Unexpected(l.mkString(", "))))
-      case r         =>
-        DecodeResult.fail(s"Unknown response: $r", s.history.getOrElse(CursorHistory.empty))
+  @SuppressWarnings(Array("org.wartremover.warts.Var"))
+  private val client = {
+    val max = 2
```
cquiroz: Perhaps we should 3 or even more attempts. We really don't want to stop a sequence if there is an intermittent issue on the DHS, OTOH we don't want to wait too long either

modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
```diff
@@ -124,19 +96,72 @@ object DhsClientHttp {
   object DhsError extends ErrorType("DHS_ERROR")
   object InternalServerError extends ErrorType("INTERNAL_SERVER_ERROR")
 
-  implicit def errorTypeDecode: DecodeJson[ErrorType] = DecodeJson[ErrorType]( c =>  c.as[String].map {
+  implicit def errorTypeDecode: Decoder[ErrorType] = Decoder.instance[ErrorType](c =>  c
```
cquiroz: Any reason not to use semi derived codecs?

cquiroz:
Fixes #538

cquiroz:
Perhaps we should test the case of simultaneous observations

modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
jluhrs: Yes. I don't know how.

PR #784, opened 2019-01-17
SEQNG-630A: Don't send apply if the config won't change

cquiroz:
This is the first part of SEQNG-630. In this task we want to create an in memory db that track status item values. In the second part we'll want to use those value to decide if we really need to configure the instrument or not

This task let me learn quite a bit of `cats-effect` but I'm probably making some mistakes. @tpolecat I'd appreciate if we reviewed it

The creation of the db is fairly straightforward but the integration with `GpiClient` and with the rest of the seqexec involved some challenges. Some are commented inline

This PR doesn't change the functionality and it has been tested against GPI

### DIFF
--- a/modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
+++ b/modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
@@ -1,0 +1,59 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package giapi.client
+
+import cats.effect.Sync
+import cats.effect.concurrent.Ref
+import cats.implicits._
+import cats.Applicative
+import cats.implicits._
+import monocle.Lens
+import monocle.function.At.at
+import monocle.function.At.atMap
+
+sealed trait StatusValue extends Product with Serializable
+
+object StatusValue {
+  final case class IntValue(value:    Int) extends StatusValue
+  final case class StringValue(value: String) extends StatusValue
+  final case class FloatValue(value:  Float) extends StatusValue
+  final case class DoubleValue(value: Double) extends StatusValue
+}
+
+/////////////////////////////////////////////////////////////////
+// The GiapiDb stores values streamed from the GMP
+/////////////////////////////////////////////////////////////////
+trait GiapiDb[F[_]] {
+
+  def value(i: String): F[Option[StatusValue]]
+
+  def update[A: ItemGetter](i: String, s: A): F[Unit]
+}
+
+object GiapiDb {
+  def newDb[F[_]: Sync](): F[GiapiDb[F]] =
+    Ref.of[F, Map[String, StatusValue]](Map.empty).map { ref =>
+      def dbat(i: String): Lens[Map[String, StatusValue], Option[StatusValue]] =
+        at(i)
+
+      new GiapiDb[F] {
+        def value(i: String): F[Option[StatusValue]] =
+          ref.get.map(_.get(i))
+
+        def update[A: ItemGetter](i: String, s: A): F[Unit] =
+          ItemGetter[A].value(s) match {
+            case Some(a: Int) =>
+              ref.update(dbat(i).set(StatusValue.IntValue(a).some))
+            case Some(a: String) =>
+              ref.update(dbat(i).set(StatusValue.StringValue(a).some))
+            case Some(a: Float) =>
+              ref.update(dbat(i).set(StatusValue.FloatValue(a).some))
+            case Some(a: Double) =>
+              ref.update(dbat(i).set(StatusValue.DoubleValue(a).some))
+            case _ =>
+              Applicative[F].unit
+          }
+      }
+    }
+}
--- a/modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
+++ b/modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
@@ -7,9 +7,6 @@ import cats.effect.Sync
 import cats.effect.concurrent.Ref
 import cats.implicits._
 import cats.Applicative
-import monocle.Lens
-import monocle.function.At.at
-import monocle.function.At.atMap
 
 sealed trait StatusValue extends Product with Serializable
 
@@ -31,11 +28,8 @@ trait GiapiDb[F[_]] {
 }
 
 object GiapiDb {
-  def newDb[F[_]: Sync](): F[GiapiDb[F]] =
+  def newDb[F[_]: Sync]: F[GiapiDb[F]] =
     Ref.of[F, Map[String, StatusValue]](Map.empty).map { ref =>
-      def dbat(i: String): Lens[Map[String, StatusValue], Option[StatusValue]] =
-        at(i)
-
       new GiapiDb[F] {
         def value(i: String): F[Option[StatusValue]] =
           ref.get.map(_.get(i))
@@ -43,13 +37,13 @@ object GiapiDb {
         def update[A: ItemGetter](i: String, s: A): F[Unit] =
           ItemGetter[A].value(s) match {
             case Some(a: Int) =>
-              ref.update(dbat(i).set(StatusValue.IntValue(a).some))
+              ref.update(_ + (i -> StatusValue.IntValue(a)))
             case Some(a: String) =>
-              ref.update(dbat(i).set(StatusValue.StringValue(a).some))
+              ref.update(_ + (i -> StatusValue.StringValue(a)))
             case Some(a: Float) =>
-              ref.update(dbat(i).set(StatusValue.FloatValue(a).some))
+              ref.update(_ + (i -> StatusValue.FloatValue(a)))
             case Some(a: Double) =>
-              ref.update(dbat(i).set(StatusValue.DoubleValue(a).some))
+              ref.update(_ + (i -> StatusValue.DoubleValue(a)))
             case _ =>
               Applicative[F].unit
           }
--- a/modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
+++ b/modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
@@ -7,7 +7,6 @@ import cats.effect.Sync
 import cats.effect.concurrent.Ref
 import cats.implicits._
 import cats.Applicative
-import cats.implicits._
 import monocle.Lens
 import monocle.function.At.at
 import monocle.function.At.atMap
--- a/modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
+++ b/modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
@@ -4,7 +4,6 @@
 package giapi.client
 
 import cats.Applicative
-import cats.Functor
 import cats.implicits._
 import cats.effect.Sync
 import cats.effect.Resource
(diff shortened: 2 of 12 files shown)


modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
```diff
@@ -0,0 +1,119 @@
...
+    */
+  def newStatusDb[F[_]: ConcurrentEffect](
+    url:   String,
+    items: List[String]
+  ): F[GiapiStatusDb[F]] =
+    for {
+      c  <- Sync[F].delay(new ActiveMQJmsProvider(url)) // Build the connection
+      ss <- Giapi.statusStreamer[F](c) // giapi artifacts
+      db <- GiapiDb.newDb
+      f  <- streamItemsToDb[F](ss.aggregate, db, items).start // run in the background
```
cquiroz: I want this to run in the background. This seems to work but I'm a bit sketchy on the details of `Fiber`

modules/giapi/src/main/scala/giapi/client/gpi/GpiClient.scala
```diff
@@ -98,22 +110,67 @@ final class GpiClient[F[_]](override val giapi: Giapi[F]) extends GiapiClient[F]
       ),
       DefaultCommandTimeout
     )
+
+  override def genericApply(configuration: Configuration): F[CommandResult] = {
+    // TODO Implement a smarter apply
```
cquiroz: Here we'll eventually make the apply smarter by only sending configs when needed

modules/giapi/src/main/scala/giapi/client/package.scala
```diff
@@ -246,23 +250,21 @@ package client {
 
           }
 
-        private def build(ref: Ref[F, ActiveMQJmsProvider]): F[Giapi[F]] =
+        private def build(c: ActiveMQJmsProvider): F[Giapi[F]] =
```
cquiroz: it seems to me we don't need the `Ref` here

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -784,11 +759,28 @@ object SeqexecEngine extends SeqexecConfiguration {
...
+      }
+
+    def ghostClient: cats.effect.Resource[IO, GhostClient[IO]] =
+      if (gpiControl === FullControl) {
+        GhostClient.ghostClient[IO](ghostUrl, ec)
+      } else {
+        GhostClient.simulatedGhostClient(ec)
+      }
+
+    def settings: IO[Settings] = {
```
cquiroz: This is a somewhat large change as instead of getting `IO[GpiClient]` we'll get `Resource[IO, GPIClient]` instead

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -216,16 +214,12 @@ object WebServerLauncher extends IOApp with LogInitialization with SeqexecConfig
       Resource.make(alloc)(free).map(ExecutionContext.fromExecutor)
     }
 
-    def engineIO(httpClient: Client[IO], collector: CollectorRegistry): IO[SeqexecEngine] =
+    def engineIO(httpClient: Client[IO], collector: CollectorRegistry, bec: ExecutionContext): IO[SeqexecEngine] =
       for {
         _          <- configLog // Initialize log before the engine is setup
         c          <- config
         site       <- IO.pure(c.require[Site]("seqexec-engine.site"))
-        giapiGPI   <- SeqexecEngine.giapiConnection[GpiSettings]("seqexec-engine.systemControl.gpi",
```
cquiroz: i think this is a good simplification

modules/giapi/src/main/scala/giapi/client/gpi/GpiClient.scala
```diff
@@ -98,22 +110,67 @@ final class GpiClient[F[_]](override val giapi: Giapi[F]) extends GiapiClient[F]
...
+    url:     String,
+    context: ExecutionContext): Resource[F, GpiClient[F]] = {
+    val giapi: Resource[F, Giapi[F]] =
+      Resource.make(
+        Giapi.giapiConnection[F](url, context).connect
+      )(_.close)
+
+    val db: Resource[F, GiapiStatusDb[F]] =
+      Resource.make(
+        GiapiStatusDb.newStatusDb[F](url, List("gpi:heartbeat"))
```
cquiroz: The list of monitored items will grow quite a bit to support SEQNG-630

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -786,11 +761,28 @@ object SeqexecEngine extends SeqexecConfiguration {
...
+      val giapiClients = for {
+        now   <- cats.effect.Resource.liftF(IO(LocalDate.now))
+        gpi   <- gpiClient
+        ghost <- ghostClient
+      } yield (now, gpi, ghost)
+      giapiClients.use { case (now, gpi, ghost) =>
```
cquiroz: After some testing I think thi is bogus. The resource will be released after settings is created. Any ideas how to properly do this?

modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
```diff
@@ -0,0 +1,58 @@
...
+/////////////////////////////////////////////////////////////////
+trait GiapiDb[F[_]] {
+
+  def value(i: String): F[Option[StatusValue]]
+
+  def update[A: ItemGetter](i: String, s: A): F[Unit]
+}
+
+object GiapiDb {
+  def newDb[F[_]: Sync](): F[GiapiDb[F]] =
```
sraaphorst: I must say that the return type here completely confuses me. 😕

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
```diff
@@ -0,0 +1,119 @@
...
+
+      override def update[B](item: StatusItem[B]): Unit =
+        // Check the item name and enqueue it
+        if (items.contains(item.getName)) {
+          Option(item.getValue).foreach { a =>
+            q.enqueue1((item.getName, a)).toIO.unsafeRunAsync(_ => ())
+          }
+        }
+
+      override def getName: String = "Giapi staus db"
```
sraaphorst: Just a typo: `status` instead of `staus` 😉

modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
tpolecat: You don't need an empty argument list here.

tpolecat: @sraaphorst the return type says, give me a program in `F` that yields a `GiapiDb[F]`; i.e., a `GiapiDb` whose operations are also defined in the effect `F`. You end up with a type like this when you need to do some work in `F` in order to construct the yielded value. This might not make sense, sorry. I did a talk at Scala Days on effects that you might find helpful if the terminology is confusing. Or ask more questions obvs.

modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
```diff
@@ -0,0 +1,58 @@
...
+        at(i)
+
+      new GiapiDb[F] {
+        def value(i: String): F[Option[StatusValue]] =
+          ref.get.map(_.get(i))
+
+        def update[A: ItemGetter](i: String, s: A): F[Unit] =
+          ItemGetter[A].value(s) match {
+            case Some(a: Int) =>
+              ref.update(dbat(i).set(StatusValue.IntValue(a).some))
```
tpolecat: This might be simpler without the lens, since it's always `some`; i.e.,

```scala
ref.update(_ + (i -> a))
```

modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
```diff
@@ -0,0 +1,58 @@
...
+            case Some(a: String) =>
+              ref.update(dbat(i).set(StatusValue.StringValue(a).some))
+            case Some(a: Float) =>
+              ref.update(dbat(i).set(StatusValue.FloatValue(a).some))
+            case Some(a: Double) =>
+              ref.update(dbat(i).set(StatusValue.DoubleValue(a).some))
+            case _ =>
```
tpolecat: If the `ItemGetter` returns `None` we don't want to remove the value from the map?

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
```diff
@@ -0,0 +1,119 @@
...
+  def close: F[Unit]
+}
+
+object GiapiStatusDb {
+  private def dbUpdate[F[_]: Applicative, A](db:   GiapiDb[F],
+                                             name: String,
+                                             a:    A): F[Unit] =
+    a match {
+      case a: Int =>
+        db.update[Int](name, a)
```
tpolecat: You probably don't need explicit type arguments here. Try `db.update(name, a)` … kind of weird because it looks like duplicated code but it's not.

tpolecat: So, because `A` is unconstrained and doesn't appear in the return type and we're typecasing in the `match` I think it would express the meaning a little more clearly if we get rid of `A` and write `a: Any` in the signature.

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
```diff
@@ -0,0 +1,119 @@
...
+        Effect[F].delay {
+          val sh = statusHandler(q)
+          agg.bindStatusHandler(sh)
+          sh
+        }
+      )(sh => Effect[F].delay(agg.unbindStatusHandler(sh)))
+
+    // Create a queue and put updates to forward them to the db
+    val s = for {
+      q <- Stream.eval(Queue.unbounded[F, (String, _)])
```
tpolecat: Same comment here, `(String, Any)` might express the situation better than using an existential, since we're relying on it getting crushed to `Any` anyway.

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
```diff
@@ -0,0 +1,119 @@
...
+      r <- Stream.resource(bind(q))
+      _ <- q.dequeue.map { case (n, v) => dbUpdate(db, n, v) }
```
tpolecat: I think this needs to be `.evalMap` otherwise you're constructing the action but not evaluating it.

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
```diff
@@ -0,0 +1,119 @@
...
+    } yield r
```
tpolecat: Any reason in particular why you're yielding `r`? We're never using it.

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
```diff
@@ -0,0 +1,119 @@
...
+import edu.gemini.jms.activemq.provider.ActiveMQJmsProvider
+import fs2.concurrent.Queue
+import fs2.Stream
+
+/////////////////////////////////////////////////////////////////
+// Links status streaming with the giapi db
+/////////////////////////////////////////////////////////////////
+trait GiapiStatusDb[F[_]] {
+  def value(i: String): F[Option[StatusValue]]
+  def close: F[Unit]
```
tpolecat: Do we want to rely on the user to call `close`? If `newStatusDb` below yielded a `Resource` it could clean itself up.

modules/giapi/src/main/scala/giapi/client/gpi/GpiClient.scala
```diff
@@ -98,22 +110,67 @@ final class GpiClient[F[_]](override val giapi: Giapi[F]) extends GiapiClient[F]
       ),
       DefaultCommandTimeout
     )
+
+  override def genericApply(configuration: Configuration): F[CommandResult] = {
+    // TODO Implement a smarter apply
+    def smartApply(): F[CommandResult] =
```
tpolecat: No `()` here.

modules/giapi/src/main/scala/giapi/client/gpi/GpiClient.scala
```diff
@@ -98,22 +110,67 @@ final class GpiClient[F[_]](override val giapi: Giapi[F]) extends GiapiClient[F]
...
+                      SequenceCommand.APPLY,
+                      Activity.PRESET_START,
+                      configuration
+                    ),
+                    DefaultCommandTimeout)
+
+    for {
+      _ <- statusDb.value("gpi:fpu") // placeholder
+      a <- smartApply()
+    } yield a
```
tpolecat: This can be `statusDb.value("gpi:fpu") *> smartApply`. Not a big deal.

modules/giapi/src/main/scala/giapi/client/gpi/GpiClient.scala
```diff
@@ -98,22 +110,67 @@ final class GpiClient[F[_]](override val giapi: Giapi[F]) extends GiapiClient[F]
...
+    val db: Resource[F, GiapiStatusDb[F]] =
+      Resource.make(
+        GiapiStatusDb.newStatusDb[F](url, List("gpi:heartbeat"))
+      )(_.close)
+
+    for {
+      c <- giapi
+      d <- db
+    } yield new GpiClient[F](c, d)
+  }
```
tpolecat: `(giapi, db).mapN(...)` if you prefer. No big deal.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
tpolecat: `Resource[IO, Settings]`?

modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
cquiroz: Yeah, the lens mas more useful in a previous incarnation that stored the values in a different structure

modules/giapi/src/main/scala/giapi/client/GiapiDb.scala
cquiroz: good question. If it returns None it means we have a type mismatch. do we want to keep the last valid value or remove it. I'll think about it

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
cquiroz: won't WartRemover balk at `Any`?

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
cquiroz: hmm. It is certainly working but I'll try with `evalMap`

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
cquiroz: No reason, I have to yield something. I guess yielding `()` would be clearer

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
cquiroz: hmm, so make it do the `close` call only in the `resource`. Makes sense

modules/giapi/src/main/scala/giapi/client/gpi/GpiClient.scala
cquiroz: Yeah. this is placeholder only, to avoid complains that `statusDb` isn't being called

modules/giapi/src/main/scala/giapi/client/gpi/GpiClient.scala
cquiroz: 👍

modules/giapi/src/main/scala/giapi/client/GiapiStatusDb.scala
tpolecat: Mmm possibly. But we're using `Any` in the `match` so I think we should be open about it.

PR #788, opened 2019-01-22
Timer implicit and GiapiConfig

cquiroz:
This PR remove usages of `ExecutionContext` by `Timer[IO]` simplifying parts of the giapi code.
I also move some methods of `GiapicConfig` to the companion object

### DIFF
--- a/modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
+++ b/modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
@@ -3,7 +3,18 @@
 
 package giapi.client
 
+import cats.Show
+
 // Produce a configuration string for Giapi.
 trait GiapiConfig[T] {
   def configValue(t: T): String
+}
+
+object GiapiConfig {
+  implicit val stringConfig: GiapiConfig[String] = t => t
+  implicit val intConfig: GiapiConfig[Int]       = _.toString
+  implicit val doubleConfig: GiapiConfig[Double] = d => f"$d%1.6f"
+
+  def fromShow[A: Show]: GiapiConfig[A] = Show[A].show(_)
+  def apply[A](implicit instance: GiapiConfig[A]): GiapiConfig[A] = instance
 }
--- a/modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
+++ b/modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
@@ -3,8 +3,6 @@
 
 package giapi.client
 
-import cats.Show
-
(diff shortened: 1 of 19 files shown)


modules/giapi/src/main/scala/giapi/client/GiapiConfig.scala
```diff
@@ -3,7 +3,18 @@
...
-}
\ No newline at end of file
+}
+
+object GiapiConfig {
+  implicit val stringConfig: GiapiConfig[String] = t => t
+  implicit val intConfig: GiapiConfig[Int]       = _.toString
+  implicit val doubleConfig: GiapiConfig[Double] = d => f"$d%1.6f"
+
+  def fromShow[A: Show]: GiapiConfig[A] = Show[A].show(_)
```
tpolecat: Really? This seems extremely sketchy.

cquiroz: Well it is what we used before. the idea is that you can if needed use `Show` It is explicitly not `implicit` as we don't want to automatically let users use `Show` without thinking about it

I guess we could remove it and make them do it explicitly

tpolecat: Since `Show` has no semantics I really don't want to use it for anything that matters … we got into a lot of trouble in OCS using `toString` this way. What would be the impact of removing this?

cquiroz: it is not a big chage, I admit though that we'll essentially reverse to use `toString` but at least it will be explicit

tpolecat: If you have to construct an explicit instance of `GiapiConfig` and it happens to call `toString` that's fine as long as it's a primitive type and you know that the `toString` is sensible and will remain so for eternity. But there shouldn't be a `.fromToString` constructor on the typeclass because it's too easy to end up with a data type whose `toString` is special and can never be changed (and this is not evident when looking at the source code for the data type). That's the situation we have in OCS and I'm very determined to avoid it in OCS3.

It's fine if you want to leave it as-is here and address later. But we need to get the `Show` situation under control eventually.

cquiroz: I tried to change it and it is quite straightforward so I'll update it

tpolecat:
👍

cquiroz:
I removed the constructor from `Show` and also updated cats-effect to 1.2.0
I tested this with GPI and it is working fine

PR #789, opened 2019-01-23
SEQNG-859 Fixes preview removal upon login

cquiroz:
This fixes a small bug where the preview was lost when logging in/out

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
@@ -84,33 +84,38 @@ final case class SequencesOnDisplay(tabs: Zipper[SeqexecTab]) {
     */
   def updateFromQueue(s: SequencesQueue[SequenceView]): SequencesOnDisplay = {
     // Sequences in view stored after completion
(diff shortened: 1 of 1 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/SequencesOnDisplay.scala
```diff
@@ -84,73 +84,76 @@ final case class SequencesOnDisplay(tabs: Zipper[SeqexecTab]) {
...
-      .map(_.some)
-    val updated = updateLoaded(s.loaded.values.toList.map { id =>
+    val completedIds: List[Option[SequenceView]] =
+      SequencesOnDisplay.completedTabs
+        .getAll(this)
+        .filterNot(x => s.loaded.values.toList.contains(x.id))
+        .map(_.some)
+    val allIds = s.sessionQueue.map(_.id)
+    val loadedInSessionQueue = s.loaded.values.toList.map { id =>
       s.sessionQueue.find(_.id === id)
```
sraaphorst: I guess here you could now work with `allIds` instead? Not really an issue but just an observation. (Pun not intended.)

cquiroz: 😄

PR #790, opened 2019-01-24
SEQNG-851 GAOS interface.

jluhrs:
Interface to the Gemini Adaptive Optics System, to be implemented for Altair and GeMS.
It will probably change when implementing the real thing.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
@@ -1,0 +1,45 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server.tcs
+
+import seqexec.server.SeqActionF
+import seqexec.server.tcs.TcsController.FocalPlaneOffset
+import squants.Time
+
+/*
+ * Interface to control AO systems. Implemented by Altair and GeMS
+ */
+trait Gaos {
+  import Gaos._
+
+  /*
+   * Pause GAOS guiding. The GAOS system decides what to pause, according to the reasons given
+   */
+  def pause[F[_]](reasons: Set[Reason]): SeqActionF[F, Unit]
+  /*
+   * Resume GAOS guiding. The GAOS system decides what to resume, according to the reasons given
+   *
+   */
+  def resume[F[_]](reasons: Set[Reason]): SeqActionF[F, Unit]
+
+  /*
+   * Notify GAOS system of the start of the observation
+   */
+  def observe[F[_]](expTime: Time): SeqActionF[F, Unit]
+  /*
+   * Notify GAOS system of the end of the observation
+   */
+  def endObserve[F[_]]: SeqActionF[F, Unit]
+
+}
+
+object Gaos {
+  sealed trait Reason
+  // Telescope offset will be applied/was applied
+  final case class BecauseOfOffset(offset: FocalPlaneOffset) extends Reason
+  // OI will be turn off, was turn on
+  case object BecauseOfOiChange extends Reason
+  // PI will be turn off, was turn on
+  case object BecauseOfP1Change extends Reason
+}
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -0,0 +1,45 @@
...
+/*
+ * Interface to control AO systems. Implemented by Altair and GeMS
+ */
+trait Gaos {
+  import Gaos._
+
+  /*
+   * Pause GAOS guiding. The GAOS system decides what to pause, according to the reasons given
+   */
+  def pause[F[_]](reasons: Set[Reason]): SeqActionF[F, Unit]
```
cquiroz: Should we consider simplifying this. In principle it should be possible to directly return `F[Unit]` Of course the worry is that this may ripple across the codebase making it a very big change but it would be interesting to explore it

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -52,9 +54,41 @@ final case class Tcs(tcsController: TcsController, subsystems: NonEmptyList[Subs
...
       if (s0.tc.offsetA === s1.self.tc.offsetA) computeGuideOff(s0, s1)
-      else GuideConfig(MountGuideOff, M1GuideOff, M2GuideOff)
+                else GuideConfig(MountGuideOff, M1GuideOff, M2GuideOff)
     }
 
+  private def pauseGaos(s0: TcsConfig, s1: Requested[TcsConfig]): SeqAction[Unit] = {
+    val aoEvents: Set[Gaos.Reason] =
+      (if (s0.ge.pwfs1.self === GuiderSensorOn && s1.self.ge.pwfs1.self === GuiderSensorOff)
+        Set(Gaos.BecauseOfP1Change)
+      else Set()) ++
```
cquiroz: `Set.empty`?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -52,9 +54,41 @@ final case class Tcs(tcsController: TcsController, subsystems: NonEmptyList[Subs
...
+      (if(s0.ge.oiwfs.self === GuiderSensorOn && s1.self.ge.oiwfs.self === GuiderSensorOff)
+        Set(Gaos.BecauseOfOiChange)
+      else Set()) ++
+      (if(s0.tc.offsetA =!= s1.self.tc.offsetA)
+        Set(Gaos.BecauseOfOffset(s1.self.tc.offsetA.self))
+      else Set())
+
+    if(aoEvents.isEmpty) SeqAction.void
```
cquiroz: This maybe could be simplified?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
jluhrs: `SeqActionF` is an `EitherT[F, SeqexecFailure, A]`, and we are very interested in the failures.

tpolecat: Hmm I don't think it's possible to implement a method like

```scala
def pause[F[_]](reasons: Set[Reason]): SeqActionF[F, Unit]
```

Without constraints on `F`. I think maybe consider

```scala
trait Gaos[F[_]] {
  def pause(reasons: Set[Reason]): F[Unit]
  def resume(reasons: Set[Reason]): F[Unit]
  ...
}
```

And then you can demand `F[_]: MonadError[?[_], Throwable]: Sync` to construct a concrete `Gaos[F]`. This is called tagless style and it's probably a best practice these days.

cquiroz: Why do you need `MonadError`?

tpolecat: If the implementation wants to `raiseError(blah)` then you need at least `ApplicativeError[F, SeqError]` but maybe you don't need to do this. I was just guessing based on the original signature.

cquiroz: Se me bay just use `F[_]` here an at the implementation level put the minimal constrains to make it work

PR #796, opened 2019-01-30
Convert some of the keyword collection to tagless final encoding

cquiroz:
This PR converts a big chunk of the code related to keywords to tagless final encoding.
It started with one line change that cascaded through a bunch of code

Ideally we'd change `SeqActionF[F, A]` for `F[A]` but that is a much bigger change

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
@@ -3,7 +3,6 @@
 
 package seqexec.server
 
-import cats.effect.IO
 import fs2.Stream
 import seqexec.model.dhs.ImageFileId
 import seqexec.server.keywords.KeywordsClient
@@ -21,7 +20,7 @@ trait InstrumentSystem[F[_]] extends System[F] {
       config: Config): SeqObserveF[F, ImageFileId, ObserveCommand.Result]
   //Expected total observe lapse, used to calculate timeout
   def calcObserveTime(config: Config): Time
-  def keywordsClient: KeywordsClient[IO]
+  def keywordsClient: KeywordsClient[F]
   def observeProgress(total: Time, elapsed: InstrumentSystem.ElapsedTime): Stream[F, Progress]
 }
 
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ghost/Ghost.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ghost/Ghost.scala
@@ -5,7 +5,7 @@ package seqexec.server.ghost
 
 import cats.data.Reader
 import cats.data.EitherT
-import cats.effect.{IO, Sync}
+import cats.effect.Sync
 import cats.implicits._
 import fs2.Stream
 import edu.gemini.spModel.config2.Config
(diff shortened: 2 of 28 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -21,7 +20,7 @@ trait InstrumentSystem[F[_]] extends System[F] {
       config: Config): SeqObserveF[F, ImageFileId, ObserveCommand.Result]
   //Expected total observe lapse, used to calculate timeout
   def calcObserveTime(config: Config): Time
-  def keywordsClient: KeywordsClient[IO]
+  def keywordsClient: KeywordsClient[F]
```
cquiroz: This was the starting change and cascaded to the rest of the codebase

modules/seqexec/server/src/main/scala/seqexec/server/ghost/Ghost.scala
```diff
@@ -42,6 +43,10 @@ final case class Ghost[F[_]: Sync](controller: GhostController[F])
   override val observeControl: InstrumentSystem.ObserveControl =
     InstrumentSystem.Uncontrollable
 
+  override def bundleKeywords(
+    ks: List[KeywordBag => SeqActionF[F, KeywordBag]]): SeqActionF[F, KeywordBag] =
+    GdsInstrument.bundleKeywords(ks)
```
sraaphorst: Is this because right now, GHOST doesn't have any other keywords explicitly defined?

modules/seqexec/server/src/main/scala/seqexec/server/gpi/Gpi.scala
```diff
@@ -53,6 +54,10 @@ final case class Gpi[F[_]: Sync: Timer](controller: GpiController[F])
   override val observeControl: InstrumentSystem.ObserveControl =
     InstrumentSystem.Uncontrollable
 
+  override def bundleKeywords(
+    ks: List[KeywordBag => SeqActionF[F, KeywordBag]]): SeqActionF[F, KeywordBag] =
+    GdsInstrument.bundleKeywords(ks)
```
sraaphorst: Ah... never mind. I see the same thing here.
Will there be no instrument-specific keywords, or maybe I am confused.

modules/seqexec/server/src/main/scala/seqexec/server/gws/GwsHeader.scala
```diff
@@ -3,17 +3,18 @@
 
 package seqexec.server.gws
```
sraaphorst: What is GWS?

jluhrs: Gemini Weather Server

modules/seqexec/server/src/main/scala/seqexec/server/gpi/Gpi.scala
cquiroz: This is happening because I can't put `bundleKeywords` on the `GdsInstrument` trait as it requires as the `F[_]` requires a specifc restriction being `Sync` but can't be done on a trait.

I suppose the best solution would be to eschew inheritance and use a typeclass instead but that's beyond the scope

PR #799, opened 2019-02-01
SEQNG-874: Initial implementation of Nifs Instrument

cquiroz:
Basic classes and logic to read the sequence from the odb

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
@@ -3,20 +3,21 @@
 
 package seqexec.server
 
-import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
+import cats.implicits._
+import edu.gemini.spModel.config2.Config
+import edu.gemini.spModel.config2.ConfigSequence
+import edu.gemini.spModel.config2.ItemKey
+import edu.gemini.spModel.seqcomp.SeqConfigNames.INSTRUMENT_KEY
+import edu.gemini.spModel.seqcomp.SeqConfigNames.OBSERVE_KEY
+import java.beans.PropertyDescriptor
+import scala.reflect.ClassTag
+import scala.collection.breakOut
 import seqexec.model.enum.SystemName
 import seqexec.model.StepConfig
 
-import java.beans.PropertyDescriptor
-
-import scala.reflect.ClassTag
-import scala.collection.breakOut
-
-import cats.implicits._
-
 /**
- * Utility operations to work with Configs from the ODB
- */
+  * Utility operations to work with Configs from the ODB
+  */
 object ConfigUtilOps {
 
   /**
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/nifs/Nifs.scala
```diff
@@ -0,0 +1,214 @@
...
+import seqexec.server.keywords.DhsInstrument
+import seqexec.server.keywords.KeywordsClient
+import seqexec.server.InstrumentSystem.AbortObserveCmd
+import seqexec.server.InstrumentSystem.InfraredControl
+import seqexec.server.InstrumentSystem.StopObserveCmd
+import seqexec.server.nifs.NifsController._
+import squants.Time
+import squants.time.TimeConversions._
+
+final case class Nifs[F[_]: LiftIO: Sync](controller: NifsController[IO], // TODO Convert to NifsController[F]
```
cquiroz: I'd like to convert this to tagless style but this particular change cascaded to a very large change

modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsController.scala
```diff
@@ -0,0 +1,82 @@
...
+import seqexec.server.nifs.NifsController.DCConfig
+import seqexec.server.nifs.NifsController.NifsConfig
+import seqexec.server.ObserveCommand
+import seqexec.server.Progress
+import squants.Time
+import shapeless.tag.@@
+
+trait NifsController[F[_]] {
+
+  def applyConfig(config: NifsConfig): F[Unit]
```
cquiroz: This is different from other controllers. Instead of returning `SeqActionF` we just return `F`

modules/seqexec/server/src/main/scala/seqexec/server/nifs/Nifs.scala
```diff
@@ -0,0 +1,214 @@
...
+    Reader { fileId =>
+      SeqActionF
+        .either(getDCConfig(config).asTrySeq)
+        .flatMap(x => SeqActionF.liftIO(controller.observe(fileId, x)))
+    }
+
+  override def calcObserveTime(config: Config): Time =
+    getDCConfig(config)
+      .map(controller.calcTotalExposureTime)
+      .getOrElse(60.seconds)
```
sraaphorst: Seems like a magic number. Is this not available / defined elsewhere?

cquiroz: This is a good comment. @jluhrs After looking around shouldn't we fail if we cannot calculate the exp time?

modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
```diff
@@ -41,31 +43,41 @@ object ConfigUtilOps {
   def explainExtractError(e: ExtractFailure): SeqexecFailure =
     SeqexecFailure.Unexpected(ConfigUtilOps.explain(e))
 
-  implicit class TrySeqed[A] private [server] (r: Either[ExtractFailure, A]) {
+  implicit class TrySeqed[A] private[server] (r: Either[ExtractFailure, A]) {
     def asTrySeq: TrySeq[A] = r.leftMap(explainExtractError)
   }
 
+  implicit class EitherOptionOps[A] private[server] (r: Either[ExtractFailure, Option[A]]) {
```
cquiroz: This is odd, this showed up in my last PR right?

modules/seqexec/server/src/main/scala/seqexec/server/nifs/Nifs.scala
cquiroz: So this should likely be improved but it affects each instrument. I think we should return `Option[Time]`

PR #800, opened 2019-02-01
SEQNG-873: Nifs Instrument

cquiroz:
This implements a minimal NifsConfigurator and integrates it to the `SeqexecEngine`. At this level we can load and simulate sequences

screencast 2019-02-01 13-29-21

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
@@ -93,6 +93,20 @@ object operations {
     def sequenceOperations: List[SequenceOperations] = Nil
   }
 
+  private val NifsSupportedOperations = new SupportedOperations {
+    def observationOperations(
+      isObservePaused: Boolean
+    ): List[ObservationOperations] =
+      isObservePaused.fold(
+        List(ObservationOperations.StopObservation,
+             ObservationOperations.AbortObservation),
+        List(ObservationOperations.StopObservation,
+             ObservationOperations.AbortObservation)
+      )
+
+    def sequenceOperations: List[SequenceOperations] = Nil
+  }
+
   private val NilSupportedOperations = new SupportedOperations {
     def observationOperations(
       isObservePaused: Boolean): List[ObservationOperations] = Nil
@@ -104,7 +118,8 @@ object operations {
     GmosS -> GmosSupportedOperations,
     GmosN -> GmosSupportedOperations,
     Gnirs -> GnirsSupportedOperations,
-    Niri  -> NiriSupportedOperations
+    Niri  -> NiriSupportedOperations,
+    Nifs  -> NifsSupportedOperations
   )
 
   final implicit class SupportedOperationsOps(val i: Instrument)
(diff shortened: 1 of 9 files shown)


modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
```diff
@@ -93,6 +93,20 @@ object operations {
...
+  private val NifsSupportedOperations = new SupportedOperations {
+    def observationOperations(
+      isObservePaused: Boolean
+    ): List[ObservationOperations] =
+      isObservePaused.fold(
+        List(ObservationOperations.StopObservation,
+             ObservationOperations.AbortObservation),
+        List(ObservationOperations.StopObservation,
+             ObservationOperations.AbortObservation)
+      )
```
tpolecat: I really prefer `if` here … I can never remember the order of `fold.`

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
```diff
@@ -119,4 +121,10 @@ class InstrumentControllerSim(name: String, useTimeout: Boolean) {
 object InstrumentControllerSim {
   def apply(name: String): InstrumentControllerSim = new InstrumentControllerSim(name, false)
   def withTimeout(name: String): InstrumentControllerSim = new InstrumentControllerSim(name, true)
+
+  implicit class SimulationExceptionOps[F[_]: Monad, A](s: EitherT[F, SeqexecFailure, A]) {
+    def orSimulationError(implicit ev: ApplicativeError[F, SeqexecFailure]): F[A] =
+      s.getOrElseF(ev.raiseError(SeqexecFailure.FailedSimulation))
```
tpolecat: This will discard any existing error in the `EitherT` … is this your intent? If you want to absorb the existing error into `F` you can say `s.value.flatMap(_.liftTo[F])` … if not maybe at least log it?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecFailure.scala
```diff
@@ -59,6 +64,20 @@ object SeqexecFailure {
     case GdsException(ex, url)        =>
       s"Failure connecting with GDS at $url: ${ex.getMessage}"
     case GdsXmlError(msg, url)        => s"XML RPC error with GDS at $url: $msg"
+    case FailedSimulation             => s"Failed to simulate"
+  }
+
+  implicit def ev(implicit ae: ApplicativeError[IO, Throwable]): ApplicativeError[IO, SeqexecFailure] = new ApplicativeError[IO, SeqexecFailure] {
```
tpolecat: Can we call this `applicativeErrorIOSeqexecFailure` or something like that? It's usually a good idea for implicit definitions to have long and specific names to lessen the risk of shadowing.

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
cquiroz: Good question. I think I need to revisit this. I'm still finding my way around `AppicativeError`. I was less concerned as the simulation never fails and this is explicitly for simulations

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecFailure.scala
cquiroz: Yes, I'll do that

PR #806, opened 2019-02-05
Make SeqexecFailure a Throwable

cquiroz:
The fact that `SeqexecFailure` is not a Throwable makes it harder to use with plain `IO` (e.g. doing `IO.raiseError(SeqexecFailure...)`

This PR adds that and fixes the conversion of plain `IO[A]` to `SeqActionF[F, SeqexecFailure, A]` to properly catch errors


jluhrs:
But, but, I thought `SeqexecFailure` existed to remove the use of `Throwable`. We even use `attempt` to capture `Throwable`s and then convert them into a `SeqexecFailure`.

cquiroz:
That's correct when we use `EitherT[IO,...]` but making it more abstract trying to use `F[_]` instead, being `SeqexecFailure` not a `Throwable` makes this harder

IMHO some of the perceived complexity of the seqexec engine is the pervasive use of Monad Transformers, in particular `EitherT`

Moving the other way should simplify some things though we have to be careful about where the error boundaries are set

tpolecat:
The issue isn't with `Throwable`, it's with `throw` and `catch`, which we don't do. We use `raiseError` and `attempt` and other combinators to handle exceptions purely.

`IO` has to deal with `Throwable` because it has to deal with arbitrary side-effects, so the easiest way to integrate our error types is to make them exceptions. But we don't throw or catch them. 

An alternative is bifunctor IO like ZIO but it's not clear it's worth the complexity.

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecFailure.scala
```diff
@@ -3,12 +3,10 @@
 
 package seqexec.server
 
-import cats.ApplicativeError
-import cats.effect.IO
 import edu.gemini.seqexec.odb.SeqFailure
 import org.http4s.Uri
 
-sealed trait SeqexecFailure extends Product with Serializable
+sealed trait SeqexecFailure extends Throwable with Product with Serializable
```
tpolecat: Can we extend `Exception` or `IOException` or something more fine-grained than `Throwable`?

cquiroz: Yeah. it could be `Exception` not `IOException` as it is not always IO related

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -203,6 +207,11 @@ package object server {
       ab.fold(a => Validated.Invalid(NonEmptyList.of(a)), b => Validated.Valid(b))
   }
 
+  implicit class EitherTFailureOps[F[_]: FlatMap, A](s: EitherT[F, SeqexecFailure, A]) {
```
cquiroz: This can be simplified using `Monad` as a constraint

jluhrs:
I guess we should be careful to not end with expressions like
`IO.raiseError(UnrecognizedInstrument("Dummy")).attempt.map(_.leftMap(_.getMessage)).unsafeRunSync(SeqexecException)`
because we would lose information.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -178,8 +179,11 @@ package object server {
     def apply[F[_]: Sync, A](a: => A): SeqActionF[F, A] =
       EitherT(Sync[F].delay(TrySeq(a)))
     def liftF[F[_]: Functor, A](a:          => F[A]): SeqActionF[F, A] = EitherT.liftF(a)
-    def liftIO[F[_]: LiftIO: Functor, A](a: => IO[A]): SeqActionF[F, A] =
-      EitherT.liftF(LiftIO[F].liftIO(a))
+    def liftIO[F[_]: LiftIO, A](a: => IO[A]): SeqActionF[F, A] =
+      EitherT(LiftIO[F].liftIO(a.attempt.map(_.leftMap {
+        case e: SeqexecFailure => e
+        case r                 => SeqexecFailure.SeqexecException(r)
+      })))
```
tpolecat: I would call this something else because we already have `LiftIO[EitherT[IO, ...` that has a different behavior. Maybe `embed` or something, to say we're absorbing this `IO` action but we're turning it into a `SeqAction` with different exception-handling behavior. Kind of hope this goes away when we get rid of all the stuff in `EitherT`.

You might want to `flatMap(_.leftMap` and only absorb `NonFatal(r)` and then have a catch-all for fatal exceptions that re-raises them.

PR #808, opened 2019-02-07
Rename toSeqAction to toSeqActionDefault

cquiroz:
The method name `toSeqAction` is a bit misleading as it not only converts a value to `SeqAction` but also defaults if the value is empty

This confused me quite a bit so I'm proposing to rename it

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
@@ -227,7 +227,8 @@ package object keywords {
   }
 
   implicit class A2SeqAction[A: DefaultHeaderValue](val v: Option[A]) {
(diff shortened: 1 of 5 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -227,7 +227,8 @@ package object keywords {
   }
 
   implicit class A2SeqAction[A: DefaultHeaderValue](val v: Option[A]) {
-    def toSeqAction: SeqAction[A] = SeqAction(v.orDefault)
+    // Convert to a SeqAction or use the default
+    def toSeqActionDefault: SeqAction[A] = SeqAction(v.orDefault)
```
cquiroz: This is the core change

PR #809, opened 2019-02-07
Header tagless style

cquiroz:
This is a largish PR that (hopefully) simplifies the Headers using `F[A]` rather than `SeqActionF[F, A]`. There are some tricky business related to error handling but I think this preserves the current semantics

Some notes:
* This has been thoroughly tested with GPI and the behaviour is correct
* I could have changed the existing `KeywordReaders` but i didn't want to make more noise
* NifsHeader will however use the new style

@jluhrs Should we test this with the DHS?

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -6,8 +6,8 @@ package seqexec.server
 import cats._
 import cats.data.{EitherT, NonEmptyList, Reader}
 import cats.effect.{ IO, Concurrent, Timer }
+import cats.effect.LiftIO
 import cats.implicits._
-
 import edu.gemini.seqexec.odb.{ExecutedDataset, SeqexecSequence}
 import edu.gemini.spModel.ao.AOConstants._
 import edu.gemini.spModel.config2.{Config, ItemKey}
@@ -39,6 +39,8 @@ import seqexec.server.tcs.TcsController.ScienceFoldPosition
 import seqexec.server.gnirs._
 import seqexec.server.niri._
 import seqexec.server.nifs._
+import seqexec.server.SeqexecFailure._
+import seqexec.server._
 import squants.Time
 import squants.time.TimeConversions._
 
@@ -50,8 +52,8 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
   import SeqTranslate._
 
   // All instruments ask the DHS for an ImageFileId
-  private def dhsFileId(inst: InstrumentSystem[IO]): SeqAction[ImageFileId] =
-    systems.dhs.createImage(DhsClient.ImageParameters(DhsClient.Permanent, List(inst.contributorName, "dhs-http")))
+  private def dhsFileId[F[_]: LiftIO](inst: InstrumentSystem[F]): SeqActionF[F, ImageFileId] =
+    systems.dhs.createImage(DhsClient.ImageParameters(DhsClient.Permanent, List(inst.contributorName, "dhs-http"))).embed
 
   private def sendDataStart(obsId: Observation.Id, imageFileId: ImageFileId, dataId: String): SeqAction[Unit] =
     systems.odb.datasetStart(obsId, dataId, imageFileId).flatMap{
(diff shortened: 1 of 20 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/keywords/Header.scala
```diff
@@ -5,12 +5,11 @@ package seqexec.server.keywords
...
 import seqexec.model.dhs.ImageFileId
-import seqexec.server.SeqActionF
 
 /**
   * Header implementations know what headers sent before and after an observation
   */
 trait Header[F[_]] {
-  def sendBefore(obsId: Observation.Id, id: ImageFileId): SeqActionF[F, Unit]
-  def sendAfter(id: ImageFileId): SeqActionF[F, Unit]
+  def sendBefore(obsId: Observation.Id, id: ImageFileId): F[Unit]
```
cquiroz: This is the root change of this PR

modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -237,26 +238,36 @@ package object keywords {
...
-  def buildInt16[F[_]: Functor](get: SeqActionF[F, Short], name: KeywordName): KeywordBag => SeqActionF[F, KeywordBag]     = buildKeyword(get, name, Int16Keyword)
-  def buildInt32[F[_]: Functor](get: SeqActionF[F, Int], name: KeywordName): KeywordBag => SeqActionF[F, KeywordBag]       = buildKeyword(get, name, Int32Keyword)
-  def buildFloat[F[_]: Functor](get: SeqActionF[F, Float], name: KeywordName): KeywordBag => SeqActionF[F, KeywordBag]     = buildKeyword(get, name, FloatKeyword)
-  def buildDouble[F[_]: Functor](get: SeqActionF[F, Double], name: KeywordName): KeywordBag => SeqActionF[F, KeywordBag]   = buildKeyword(get, name, DoubleKeyword)
-  def buildBoolean[F[_]: Functor](get: SeqActionF[F, Boolean], name: KeywordName): KeywordBag => SeqActionF[F, KeywordBag] = buildKeyword(get, name, BooleanKeyword)
-  def buildString[F[_]: Functor](get: SeqActionF[F, String], name: KeywordName): KeywordBag => SeqActionF[F, KeywordBag]   = buildKeyword(get, name, StringKeyword)
-
-  def sendKeywords[F[_]: Sync](
+  def buildKeyword[F[_]: MonadError[?[_], Throwable], A](get: SeqActionF[F, A], name: KeywordName, f: (KeywordName, A) => Keyword[A]): KeywordBag => F[KeywordBag] =
```
cquiroz: We keep two set of functions to support the existing headers built around `SeqActionF`.
Should disappear at some moment

modules/seqexec/server/src/main/scala/seqexec/server/keywords/GdsClient.scala
```diff
@@ -21,12 +20,11 @@ import scala.concurrent.duration._
 import scala.xml.Elem
 import seqexec.model.dhs.ImageFileId
 import seqexec.server.SeqexecFailure
-import seqexec.server.SeqActionF
 
 /**
   * Gemini Data service client
   */
-final case class GdsClient[F[_]: Effect](base: Client[F], gdsUri: Uri)(implicit timer: Timer[F])
+final case class GdsClient[F[_]: Effect: cats.Functor](base: Client[F], gdsUri: Uri)(implicit timer: Timer[F])//, ae: ApplicativeError[F, SeqexecFailure])
```
tpolecat: Fix the import and remove commented code?

modules/seqexec/server/src/main/scala/seqexec/server/keywords/Header.scala
tpolecat: :+1:

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -88,15 +90,15 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
+    def closeImage[F[_]: LiftIO](id: ImageFileId): SeqActionF[F, Unit] =
+      SeqActionF.embed(inst.keywordsClient.closeImage(id))
 
     def doObserve(fileId: ImageFileId): SeqAction[Result] =
       for {
         d   <- dataId
         _   <- sendDataStart(obsId, fileId, d)
         _   <- notifyObserveStart
-        _   <- headers(ctx).map(_.sendBefore(obsId, fileId)).sequence
+        _   <- headers(ctx).map(_.sendBefore(obsId, fileId)).sequence.embed
```
jluhrs: What does `embed` do?

cquiroz: embed will take a `F[_]` and convert it to `SeqActionF[_]`. It is like a boundary to avoid propagating the `F[_]` change upstream in the code (yet 😄 )

It also will preserve the error handling semantics

PR #811, opened 2019-02-09
SEQNG-878 NifsKeywords provider

cquiroz:
This is the basis for the NIFS keywords.

I noted that the way we read EPICS channels is not referentially transparent and this PR improves that. I'll be porting the reader for other instrument

Pending are the WCS keywords that need a special treatment and will be done in another PR


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -245,6 +245,9 @@ object EpicsUtil {
     os.map(_.setTimeout(t.toMilliseconds.toLong, MILLISECONDS).asRight).getOrElse(SeqexecFailure.Unexpected("Unable to set timeout for EPICS command.").asLeft)
   }
 
+  def safeAttribute[F[_]: Sync, A](get: => CaAttribute[A]): F[Option[A]] =
+    Sync[F].delay(Option(get).flatMap(x => Option(x.value)))
```
jluhrs: I had a reason to not put the `get` in an `Option`. The functions `getXXX` use a `String` name to find the attribute that was created from the xml files. If it fail, it is because there was an error in the channel definition, most likely a typo in the channel name. If that happens, I wanted to know it immediately.  That is an error with the code, not an execution error, and protecting the `get` with an `Option` would make it harder to find.

cquiroz: I understand. I think this would be better handled at compile time but that is a bigger change. I'll revert it to `.get.value`

modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsEpics.scala
```diff
@@ -3,15 +3,19 @@
...
 import edu.gemini.seqexec.server.nifs.DhsConnected
+import java.lang.{Double => JDouble}
 import org.log4s.{Logger, getLogger}
 import seqexec.server.EpicsCommand.setParameter
 import seqexec.server.{EpicsCommand, EpicsSystem, ObserveCommand, SeqAction}
+import seqexec.server.EpicsUtil.safeAttribute
 
-class NifsEpics(epicsService: CaService, tops: Map[String, String]) {
+class NifsEpics[F[_]: Sync](epicsService: CaService, tops: Map[String, String]) {
   val NifsTop = tops.getOrElse("niri", "niri:")
```
sraaphorst: Is this correct that `NifsTop` should use `niri`?

modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsEpics.scala
```diff
@@ -111,78 +115,103 @@ class NifsEpics(epicsService: CaService, tops: Map[String, String]) {
...
-    .map(_.toDouble)
+  def readTime: F[Option[Double]] =
+    Nested(safeAttribute(dcStatus.getDoubleAttribute("readTime")))
+      .map(_.toDouble).value
 
-  def biasPwr: Option[Double] = Option(dcStatus.getDoubleAttribute("biasPwr").value)
-    .map(_.toDouble)
+  def biasPwr: F[Option[Double]] =
+    Nested(safeAttribute(dcStatus.getDoubleAttribute("biasPwr")))
+      .map(_.toDouble).value
```
sraaphorst: It seems that, for a type `T`,
```
Nested(safeAttribute(dcStatus.getTAttribute("name"))).map(_.toT).value
```
is repeated so many times. Is there a way to reduce all the code copying?

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
sraaphorst: I see you took care of this while I was reviewing 👍

modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsEpics.scala
cquiroz: Doh' good catch

modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsEpics.scala
cquiroz: Right, we could add some utility methods. I expect to use them in other `XXXEpics` objcets

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
jluhrs: I agree that the best scenario would be to have it handled at compile time. We could write a tool that the names used in the code exist in the xml file.

PR #812, opened 2019-02-11
Seqng 581: Use log4s on the client side

cquiroz:
we have java util logging on the client and log4s on the server. I'd rather only have one but log4s didn't support scala.js well. With version 1.7.0 it is now possible


project/plugins.sbt
```diff
@@ -36,7 +36,7 @@ addSbtPlugin("de.heikoseeberger"  % "sbt-header"               % "4.1.0")
 addSbtPlugin("com.typesafe.sbt"   % "sbt-git"                  % "0.9.3")
 addSbtPlugin("com.dwijnand"       % "sbt-dynver"               % "3.0.0")
 
-addSbtPlugin("org.wartremover"    % "sbt-wartremover"          % "2.4.0")
+addSbtPlugin("org.wartremover"    % "sbt-wartremover"          % "2.4.1")
```
sraaphorst: Can't we remove this line? 😁

cquiroz: I wouldn't mind

PR #814, opened 2019-02-11
calcTotalExposureTime should be referentially transparent

cquiroz:
In `InstrumentSystem` the method `calcObserveTime` should IMHO return an `F[Time]`. This because for some instruments, in particular Niri and apprently Nifs we need to read an epics channel to calculate the observe time

Doing it this way respects referential transparency

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
@@ -19,7 +19,7 @@ trait InstrumentSystem[F[_]] extends System[F] {
   def observe(
       config: Config): SeqObserveF[F, ImageFileId, ObserveCommand.Result]
   //Expected total observe lapse, used to calculate timeout
-  def calcObserveTime(config: Config): Time
+  def calcObserveTime(config: Config): F[Time]
   def keywordsClient: KeywordsClient[F]
   def observeProgress(total: Time, elapsed: InstrumentSystem.ElapsedTime): Stream[F, Progress]
 }
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -121,15 +121,25 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
         case ObserveCommand.Success => successTail
         case ObserveCommand.Stopped => stopTail
         case ObserveCommand.Aborted => abortTail
-        case ObserveCommand.Paused  => SeqAction(Result.Paused(ObserveContext(observeTail(id, dataId), inst.calcObserveTime(config))))
+        case ObserveCommand.Paused  =>
+          SeqActionF.liftF(inst.calcObserveTime(config))
+            .map(e => Result.Paused(ObserveContext(observeTail(id, dataId), e)))
       }
     }
 
-    Stream.eval(dhsFileId(inst).value).flatMap{
-      case Right(id) => Stream.emit(Result.Partial(FileIdAllocated(id))).covary[IO] ++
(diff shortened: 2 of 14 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -19,7 +19,7 @@ trait InstrumentSystem[F[_]] extends System[F] {
   def observe(
       config: Config): SeqObserveF[F, ImageFileId, ObserveCommand.Result]
   //Expected total observe lapse, used to calculate timeout
-  def calcObserveTime(config: Config): Time
+  def calcObserveTime(config: Config): F[Time]
```
cquiroz: This is the root change

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -121,15 +121,25 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
     }
 
-    Stream.eval(dhsFileId(inst).value).flatMap{
-      case Right(id) => Stream.emit(Result.Partial(FileIdAllocated(id))).covary[IO] ++
-        inst.observeProgress(inst.calcObserveTime(config), ElapsedTime(0.0.seconds))
-          .map(Result.Partial(_))
-          .mergeHaltR(Stream.eval[IO, Result](doObserve(id).value.map(_.toResult)))
+    Stream.eval(dhsFileId(inst).value).flatMap {
+      case Right(id) =>
+        val observationProgressStream =
```
cquiroz: I've split these for clarity

modules/seqexec/server/src/main/scala/seqexec/server/niri/NiriControllerEpics.scala
```diff
@@ -300,7 +300,8 @@ object NiriControllerEpics extends NiriController {
   override def observeProgress(total: Time): fs2.Stream[IO, Progress] =
     ProgressUtil.countdown[IO](total, 0.seconds)
 
-  override def calcTotalExposureTime(cfg: DCConfig): Time = {
+  override def calcTotalExposureTime(cfg: DCConfig): IO[Time] = IO {
```
cquiroz: Niri reads an epics channel to calculate the exposure time

swalker2m: We come up with some kind of planned observing time estimate in the program model, but I suppose this is more accurate?

modules/seqexec/server/src/main/scala/seqexec/server/niri/NiriControllerSim.scala
```diff
@@ -29,7 +30,7 @@ object NiriControllerSim extends NiriController {
   override def observeProgress(total: Time): fs2.Stream[IO, Progress] =
     sim.observeCountdown(total, ElapsedTime(0.seconds))
 
-  override def calcTotalExposureTime(cfg: DCConfig): Time = {
+  override def calcTotalExposureTime(cfg: DCConfig): IO[Time] = IO.pure {
```
sraaphorst: Why are some `calcTotalExposureTime` methods parameterized with `F[Time]`, and others use `IO[Time]`?

cquiroz: Because I'm still trying to convert parts from `IO` to `F`. A full change is next to impossible so in places I keep `IO`

modules/seqexec/server/src/main/scala/seqexec/server/niri/NiriControllerEpics.scala
cquiroz: I'm not totally sure but I think in the Niri case there are some values calculated in the instrument that need to be taken into account

PR #815, opened 2019-02-12
SEQNG-672 Split ocs2 package to make it available to the scala.js side

cquiroz:
I want to integrate more of gem enums onto the web ui. I'm splitting the ocs2pcakage to make it available on the js side


tpolecat:
I thought the OCS2 module was just to allow us to import old programs into OCS3 and to prevent OCS2 stuff from leaking into the new model … why do you need it on the JS side?

cquiroz:
The seqexec uses ocs2 strings on the front end. I want to map those strings into `gem` enums via the `Parsers` class

PR #817, opened 2019-02-13
SEQNG-879A: Nifs epics DC configuration

cquiroz:
This is the first PR to get NIFS epics working. Up to this point it supports configuring the DC but the CC is pending

This PR is to get some early feedback

Note that I'm trying to simplify the code by using `F[_]` rather than `Either[F, E, _]` but this needs testing

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsControllerEpics.scala
@@ -3,31 +3,39 @@
 
 package seqexec.server.nifs
 
-// import cats.data.EitherT
-import cats.effect.{ IO, Timer }
+import cats.data.OptionT
+import cats.effect.IO
+import cats.effect.Timer
 import cats.implicits._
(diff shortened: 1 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsControllerEpics.scala
```diff
@@ -0,0 +1,174 @@
...
+import squants.Seconds
+import squants.Time
+import squants.time.TimeConversions._
+
+object NifsLookupTables {
+
+  val readModeLUT: Map[LegacyReadMode, Int] = Map(
+    LegacyReadMode.BRIGHT_OBJECT_SPEC -> 1,
+    LegacyReadMode.MEDIUM_OBJECT_SPEC -> 1,
+    LegacyReadMode.FAINT_OBJECT_SPEC -> 1
```
sraaphorst: Why do they all map to 1? Because they are legacy?

cquiroz: Yeah, isn't that silly but that's what the original TCL code does. I prefer to keep it though we could just default to 1

PR #819, opened 2019-02-15
SEQNG-879: Nifs CC setup

cquiroz:
This is the code to configure NIFS CC. It is more or less a direct translation of the tcl code though the two pass setup of the CC is now made explicit

This of course requires intensive testing to be verified

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -19,9 +19,10 @@ import java.util.{Timer, TimerTask}
 import java.util.concurrent.locks.ReentrantLock
 import java.util.concurrent.TimeUnit.MILLISECONDS
 import edu.gemini.epics.acm._
+import mouse.boolean._
+import org.log4s._
 import seqexec.server.EpicsCommand.safe
 import seqexec.server.SeqexecFailure.SeqexecException
-import org.log4s._
 import squants.Time
 import scala.math.abs
 import scala.collection.JavaConverters._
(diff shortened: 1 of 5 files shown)


cquiroz:
I see, that explains some details I didn't understand in the code

modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsControllerEpics.scala
```diff
@@ -96,9 +98,23 @@ trait NifsEncoders {
 object NifsControllerEpics extends NifsController[IO] with NifsEncoders {
   private val Log = getLogger
 
+  private val ConfigTimeout: Time  = Seconds(180)
```
jluhrs: Tcl Seqexec uses 400[s] (NifsInts.itcl:83)

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -339,8 +339,14 @@ object EpicsUtil {
...
-  def smartSetParamF[F[_]: Monad, A: Eq](v: A, get: => F[Option[A]], set: F[Unit]): F[Unit] =
-    get.map(_ =!= v.some).ifM(set, Applicative[F].unit)
+  // The return signature indicates this programs calculates if we maybe need an action
+  // e.g. it checks that a value in epics compares to a reference and if so returns an optional
+  // action
+  def smartSetParamF[F[_]: Monad, A: Eq](v: A, get: => F[Option[A]], set: F[Unit]): F[Option[F[Unit]]] =
+    get.map(_ =!= v.some).map {
+      case true => set.some
+      case false => none
+    }
```
jluhrs: `(_ =!= v.some).option(set)` ?

cquiroz: `get.map(_ =!=.some)` gives `F[Boolean]`

jluhrs: `get.map((_ =!= v.some).option(set))`

cquiroz: I think you need an extra map, but yeah that will work

PR #820, opened 2019-02-18
Added GMOS config lenses

swalker2m:
This PR adds lenses to GMOS config, both static and dynamic.  Ultimately the goal is to use these lenses to simplify sequence generation but I'll need a few more lenses/prisms in `Step` to get there. A few questions that I anticipate:

* Why write them out manually instead of using `@Lenses` or `GenLens`?

I'm not married to this but I thought I'd try it when @tpolecat mentioned he preferred it that way.  Because I wrote them out manually, I felt obligated to write out the tests as well which revealed missing `Eq`, `Order`, `Arbitrary`, and `Cogen` instances and a problem with arbitrary `Duration`.  So that seemed worthwhile in the end.

* Why not use the weathervane operators like `^|->`?

I don't like them and can't remember them.  `composeLens` seems a lot clearer to me.  If there is a consensus about it I'll switch though.

* Why only `Getter` instances for `GmosShuffleOffset.detectorRows`, etc.?

Because `detectorRows` is constrained to be a positive int, you can't make a normal `Lens[GmosShuffleOffset, Int]` because the setter might not work.  I can make a `Prism[Int, GmosShuffleOffset]` but that doesn't compose with anything that contains a `GmosShuffleOffset`.  Maybe there's a correct way to do this, idk.

### DIFF
--- a/modules/core/shared/src/test/scala/gem/arb/ArbTime.scala
+++ b/modules/core/shared/src/test/scala/gem/arb/ArbTime.scala
@@ -13,7 +13,6 @@ import org.scalacheck.Arbitrary._
 import scala.collection.JavaConverters._
 import scala.concurrent.duration.{Duration => SDuration}
 import java.time._
-import java.time.temporal.ChronoUnit
 
 // Arbitrary but resonable dates and times.
 trait ArbTime {
@@ -78,12 +77,7 @@ trait ArbTime {
     }
 
   implicit val arbDuration: Arbitrary[Duration] =
-    Arbitrary {
-      for {
-        a <- Gen.choose(-10000L, 10000L)
-        u <- Gen.oneOf(ChronoUnit.values.filterNot(_.isDurationEstimated))
-      } yield Duration.of(a, u)
-    }
+    Arbitrary(Gen.posNum[Long].map(Duration.ofMillis))
 
   implicit val arbSDuration: Arbitrary[SDuration] = Arbitrary {
     for {
--- a/modules/core/shared/src/test/scala/gem/arb/ArbTime.scala
+++ b/modules/core/shared/src/test/scala/gem/arb/ArbTime.scala
@@ -100,6 +100,9 @@ trait ArbTime {
   implicit val cogLocalDate: Cogen[LocalDate] =
     Cogen[(Int, Int)].contramap(d => (d.getYear, d.getDayOfYear))
 
+  implicit val cogDuration: Cogen[Duration] =
+    Cogen[(Long,Int)].contramap(d => (d.getSeconds, d.getNano))
+
   implicit val cogTimestamp: Cogen[Timestamp] =
     Cogen[Instant].contramap(_.toInstant)
 
--- a/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
@@ -4,11 +4,17 @@
 package gem
 package config
 
-import cats.Eq
 import gem.CoAdds
 import gem.enum._
 import gem.math.Wavelength
+
+import cats.Eq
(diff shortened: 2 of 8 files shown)


modules/core/shared/src/test/scala/gem/arb/ArbTime.scala
```diff
@@ -78,12 +77,7 @@ trait ArbTime {
     }
 
   implicit val arbDuration: Arbitrary[Duration] =
-    Arbitrary {
-      for {
-        a <- Gen.choose(-10000L, 10000L)
-        u <- Gen.oneOf(ChronoUnit.values.filterNot(_.isDurationEstimated))
-      } yield Duration.of(a, u)
-    }
+    Arbitrary(Gen.posNum[Long].map(Duration.ofMillis))
```
cquiroz: Why this change?

swalker2m: We had two conflicting definitions of `Arbitrary[Duration]` before.  The one I changed it to was in `gem.config.Arbitraries` but really belonged in `ArbTime`.  Like `Arbitrary[Timestamp]`, we have to limit the possible values to those that will work when written to the database (say in an `exposure_time` column). In this case, that's duration values that can be mapped to/from milliseconds:

```scala
  implicit val DurationMeta: Meta[Duration] =
    Distinct.long("milliseconds").timap(Duration.ofMillis)(_.toMillis)
```

I actually don't see a constraint on positive exposure times but it seems like there should be one!  We could probably also keep the definition that was there but constrain it so that `ChronoUnit` is at least `MILLIS`.

I suppose another idea would be to use an explicit `Duration` and `Timestamp` generator for database tests.

cquiroz:
I don't have a strong opinion about `@Lenses` though I tend to use the annotation to avoid writing extra code. There maybe a good reason to use `GenLens` instead, perhaps to avoid macros or a slow compilation phase?

I'm ok with the symbols to compose lenses, I just got used to them

swalker2m:
> There maybe a good reason to use `GenLens` instead, perhaps to avoid macros or a slow compilation phase?

Isn't `GenLens` *also* a macro?

swalker2m:
> I'm ok with the symbols to compose lenses, I just got used to them

I was okay with `/:` and `:\` for `foldLeft` and `foldRight` but once we switched away from them it did seem better.  Either way is fine though.

cquiroz:
At any rate I can use either way but we should try to be consistent

The seqexec is full of lenses with the symbolic operators

cquiroz:
You're right `GenLenses` is a macro too

modules/core/shared/src/test/scala/gem/arb/ArbTime.scala
```diff
@@ -100,6 +94,9 @@ trait ArbTime {
   implicit val cogLocalDate: Cogen[LocalDate] =
     Cogen[(Int, Int)].contramap(d => (d.getYear, d.getDayOfYear))
 
+  implicit val cogDuration: Cogen[Duration] =
+    Cogen[(Long,Int)].contramap(d => (d.getSeconds, d.getNano))
+
```
sraaphorst: I'm still not entirely sure what a `Cogen` is. Does anyone have any references that can help me understand better?

tpolecat: TL;DR if you have `Cogen[A]` and `Gen[B]` you get `Gen[A => B]`. The job of `Cogen[A]` is to take an `A` and absorb some information in order to perturb the random seed that's used to generate the return value `B`. This talk explains it very well. You can probably skip ahead to 36:00 if you already understand `Gen`.

    https://www.youtube.com/watch?v=UlZGZh7hfbs

(I'm off today but wanted to share that link.)

PR #822, opened 2019-02-19
SEQNG-888 Approximate calculation of WCS for NIFS

cquiroz:
This PR contains a simplified calculation of the WCS for NIFS

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsHeader.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsHeader.scala
@@ -7,6 +7,9 @@ import cats.MonadError
 import cats.implicits._
 import gem.Observation
 import gem.enum.KeywordName
+import scala.math.cos
+import scala.math.sin
+import scala.math.toRadians
 import seqexec.model.dhs.ImageFileId
 import seqexec.server.InstrumentSystem
 import seqexec.server.keywords._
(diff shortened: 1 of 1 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/nifs/NifsHeader.scala
```diff
@@ -38,9 +43,22 @@ object NifsHeader {
...
+          buildStringS("RA---TAN".pure[F], KeywordName.CTYPE1),
+          buildDoubleS(15.0.pure[F], KeywordName.CRPIX1),
+          buildDoubleS(tcsKeywordsReader.getSourceATarget.getRA, KeywordName.CRVAL1),
+          buildStringS("DEC--TAN".pure[F], KeywordName.CTYPE2),
+          buildDoubleS(34.0.pure[F], KeywordName.CRPIX2),
+          buildDoubleS(tcsKeywordsReader.getSourceATarget.getDec, KeywordName.CRVAL2),
+          buildDoubleS(φ.map(φ => -4.7e-5 * sin(φ)), KeywordName.CD1_1),
+          buildDoubleS(φ.map(φ =>  1.9e-5 * cos(φ)), KeywordName.CD1_2),
+          buildDoubleS(φ.map(φ => -4.7e-5 * cos(φ)), KeywordName.CD2_1),
+          buildDoubleS(φ.map(φ => -1.9e-5 * sin(φ)), KeywordName.CD2_2),
```
jluhrs: Magic numbers ?

cquiroz: I don't have good names for this as they are just calculated factors for NIFS
CD1_1_FACTOR?

sraaphorst: I also wonder about using non-ASCII symbols like `φ`. Unless I'm missing something, it seems like this may complicate things down the road, since it means we have to copy - paste or hunt through unicode to get these symbols.

I think we should be very discerning in our use of them.

PR #824, opened 2019-02-20
SEQNG-845 Show the step subsystem control if we pause the observation

cquiroz:
Small UI fix. It shows the subsytem buttons if the sequence is paused


cquiroz:
seqexec - gs-2018b-q-0-96 2019-02-19 15-31-03

PR #825, opened 2019-02-20
More config lenses

swalker2m:
Adds lenses for `GcalConfig` and `TelescopeConfig` along with a few `Eq` and one `Order` instance. Along the way I'm moving stuff out of `gem.Arbitraries` and `gem.config.Arbitraries` in favor of specific instances in `gem.arb`.

### DIFF
--- a/modules/core/shared/src/test/scala/gem/arb/ArbCoAdds.scala
+++ b/modules/core/shared/src/test/scala/gem/arb/ArbCoAdds.scala
@@ -1,0 +1,21 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+package arb
+
+import gem.syntax.prism._
+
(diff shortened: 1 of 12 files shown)


modules/core/shared/src/test/scala/gem/arb/ArbCoAdds.scala
```diff
@@ -0,0 +1,21 @@
...
+package gem
+package arb
+
+import gem.syntax.prism._
+
+import org.scalacheck.{ Arbitrary, Cogen, Gen }
+
+trait ArbCoAdds {
+
+  implicit val arbCoAdds: Arbitrary[CoAdds] =
```
cquiroz: Should `CoAdds` be restricted to only positive numbers in the type?

swalker2m: Ideally yes.  We have a number of types like this but we haven't settled on a better approach.  At any rate here I'm just moving this code out of `gem.Arbitraries` into an individually importable module like we've done for a number of other types.

PR #828, opened 2019-02-22
Missing NIFS keywords

cquiroz:
We have a few missing keywords for NIFS


modules/sql/src/main/resources/db/migration/V062__More_Nifs_KeywordNames.sql
```diff
@@ -0,0 +1,8 @@
+---------------------------
+-- NIFS FITS Keyword Names
+---------------------------
+
+INSERT INTO e_fits_keyword_names(id, name)
```
cquiroz: I created a new file though I did another one a few days ago. Should I use a new file or patch the old one?
/cc @tpolecat

swalker2m: A new file I believe. Flyway won’t reapply a migration number that it has already applied, even if the content changes if I understand it correctly.

PR #830, opened 2019-02-22
More arbitrary cleanup

swalker2m:
This PR removes `gem.Arbitraries`, moving its content into appropriate `gem.arb.Arb*` traits.  It also removes the duplicate `Observation` arbitrary definitions.  Along the way it:

* Updates enum generation (and the generated enums) to display the enum type when there is a failure to read a tag in `unsafeFromTag`.

* Fixes a bug in `AsterismDao` where a dual target asterism is treated as single target.

* Fixes a bug in `StepDao` for GNIRS.


swalker2m:
In the end I left the `Arbitrary[Observation]` definition to choose between F2, GMOS, and GNIRS.  We are missing schema and [DAO code to write steps](https://github.com/gemini-hlsw/ocs3/blob/develop/modules/db/src/main/scala/gem/dao/StepDao.scala#L156-L174), even if empty, for the other instruments.

swalker2m:
This appears to be the issue with the tests.  I don't _think_ this update is related.

```
[info] GiapiCommandSpec:
[info] - Test sending a command with no handlers !!! IGNORED !!!
[info] - Test sending a command with no answer *** FAILED ***
[info]   giapi.client.commands.CommandResultException was not equal to giapi.client.commands.CommandResultException (GiapiCommandSpec.scala:92)
[info] - Test sending a command with immediate answer
[info] - Test sending a command with accepted but never completed answer
```

tpolecat:
Is that test failure sporadic? It's passing now.

tpolecat:
> In the end I left the `Arbitrary[Observation]` definition to choose between F2, GMOS, and GNIRS

That seems fine, we just need to keep an eye on this as we build out the other instruments.

swalker2m:
> Is that test failure sporadic? It's passing now.

Yes, I ran it again and it passed.

PR #831, opened 2019-02-22
Step Lenses

swalker2m:
This PR adds some Lenses, Prisms, and `Eq` instances for `Step` and its friends.  For context, in order to do sequence generation I’ll need to set up a `Step.GmosN` (for example) and then modify it repeatedly to form a sequence.  Doing it without all of the optics I’ve now added was going to be brutal.

Nevertheless I can’t help but feel I’ve driven into a ditch in this one.  I’ll point out a few specific concerns in comments.

### DIFF
--- a/modules/core/shared/src/main/scala/gem/Step.scala
+++ b/modules/core/shared/src/main/scala/gem/Step.scala
@@ -3,9 +3,12 @@
 
 package gem
 
-import cats.Eq
 import gem.config._
 import gem.enum.SmartGcalType
+
+import cats.Eq
+import cats.implicits._
+import monocle._
 
 sealed trait Step {
   def dynamicConfig: DynamicConfig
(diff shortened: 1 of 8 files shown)


modules/core/shared/src/main/scala/gem/Step.scala
```diff
@@ -16,32 +19,140 @@ object Step {
...
+  object Base extends BaseOptics {
+
     final case object Bias extends Base
     final case object Dark extends Base
     final case class  Gcal(gcal: GcalConfig) extends Base
     final case class  Science(telescope: TelescopeConfig) extends Base
     final case class  SmartGcal(smartGcalType: SmartGcalType) extends Base
+
+    implicit val EqBias: Eq[Bias.type] =
+      Eq.allEqual
```
swalker2m: It seems weird to define `Eq` for `case object`s.

modules/core/shared/src/main/scala/gem/Step.scala
```diff
@@ -16,32 +19,140 @@ object Step {
...
+
+    implicit val EqBase: Eq[Base] =
+      Eq.instance {
+        case (a: Bias.type, b: Bias.type) => a === b
+        case (a: Dark.type, b: Dark.type) => a === b
+        case (a: Gcal,      b: Gcal     ) => a === b
+        case (a: Science,   b: Science  ) => a === b
+        case (a: SmartGcal, b: SmartGcal) => a === b
+        case _                            => false
+      }
```
swalker2m: Is this the right way to do `Eq` for an ADT?  I did the same thing for `DynamicConfig`.

modules/core/shared/src/main/scala/gem/Step.scala
```diff
@@ -16,32 +19,140 @@ object Step {
...
+    /** @group Optics */
+    val base: Lens[GmosS, Base] =
+      Lens[GmosS, Base](_.base)(a => _.copy(base = a))
+
+  }
 
-  implicit val EqStep: Eq[Step] =
-    Eq.fromUniversalEquals
+  implicit def EqStep[S <: Step]: Eq[S] =
+    Eq.by(s => (s.dynamicConfig, s.base))
```
swalker2m: This may be sketchy, idk.

modules/core/shared/src/test/scala/gem/arb/ArbStep.scala
```diff
@@ -5,28 +5,71 @@ package gem
...
 
+  implicit val cogStepBase: Cogen[Step.Base] =
+    Cogen[(Option[Bias.type], Option[Dark.type], Option[Gcal], Option[Science], Option[SmartGcal])]
+      .contramap {
+        case Bias         => (Some(Bias), None,       None,    None,    None   )
+        case Dark         => (None,       Some(Dark), None,    None,    None   )
+        case b: Gcal      => (None,       None,       Some(b), None,    None   )
+        case b: Science   => (None,       None,       None,    Some(b), None   )
+        case b: SmartGcal => (None,       None,       None,    None,    Some(b))
+      }
```
swalker2m: This feels particularly wrong, but I'm not sure how to write the `Cogen` otherwise.

modules/core/shared/src/main/scala/gem/Step.scala
cquiroz: I normally would do `Eq.instance` for `Base` in this case

cquiroz: Oh I see it was `Eq[Base]` base was defined. Why do you need `Eq[Bias.type]`

modules/core/shared/src/main/scala/gem/Step.scala
```diff
@@ -16,32 +19,140 @@ object Step {
...
+
+      /** @group Optics */
+      val smartGcalType: Lens[SmartGcal, SmartGcalType] =
+        Lens[SmartGcal, SmartGcalType](_.smartGcalType)(a => _.copy(smartGcalType = a))
+
+    }
+
+    implicit val EqBase: Eq[Base] =
+      Eq.instance {
+        case (a: Bias.type, b: Bias.type) => a === b
```
cquiroz: I bet you can do `case (_: Bias.type, _: Bias.type) => true`

modules/core/shared/src/main/scala/gem/Step.scala
cquiroz: At least I do the same for Eq in an ADT

modules/core/shared/src/main/scala/gem/Step.scala
cquiroz: It seems fine to me

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
```diff
@@ -123,6 +164,11 @@ object DynamicConfig {
       Step.Niri(this, b)
```
cquiroz: Why `case class Niri()` rather than `case object Niri`?

modules/core/shared/src/test/scala/gem/arb/ArbStep.scala
cquiroz: I think I usually do `Either[Bias, Either[Dark, Either[...`

Not that it feels much better

modules/core/shared/src/main/scala/gem/Step.scala
jluhrs: Or `case (Bias, Bias) => true` ?

modules/core/shared/src/main/scala/gem/Step.scala
jluhrs: Me too, except that I make the `object` cases return always `true`.

modules/core/shared/src/main/scala/gem/Step.scala
swalker2m: I did it so that all the cases of `Eq[Base]` could be defined as `a === b`. I thought about `case (Base, Base) => true` instead but went for symmetry all the way through to the arbitraries and cogens.  idk.

modules/core/shared/src/main/scala/gem/config/DynamicConfig.scala
swalker2m: This was done some time ago.  It does make all the instrument cases the same at least.  Eventually they'll all need to be filled out with configuration details like `GmosN`, etc.  If some of the instrument configs are objects and some are classes then it becomes awkward like `Step.Bias` and `Step.Dark`.  For example, the steps need a `DynamicConfig` instance and so it'd have to be something like `sealed case class Niri(dynamicConfig: DynamicConfig.Niri.type, base: Base) extends Step` until eventually we add the actual config fields and have to switch it over.

There may have been other reasons.

swalker2m: Making it, say,

```
sealed case class Niri(base: Base) extends Step {
  override def dynamicConfig: DynamicConfig =
    Niri
}
```
would be even more fiddling later when `Niri` is ultimately defined.

cquiroz: 👍

modules/core/shared/src/main/scala/gem/Step.scala
jluhrs: In three years, when a new kind of `Step` is needed with a different type of equality, will your future self hate you ?
IMO it is a tiny bit sketchy, but the base trait is sealed, so at least when someone needs to add a new type of Step in the future, he/she will find this quickly.

PR #833, opened 2019-02-27
Seqng 894: Fix ACM to support NIFS

cquiroz:
NIFS doesn't follow the action model as it should be thus we need to change the whole Action Mode code. This PR fixes that and it has been tested on NIFS and also checked for regressions with the other instruments


cquiroz:
We'll check that during testing

PR #835, opened 2019-02-27
SEQNG-907 Use a shorter request timeout for the http client

cquiroz:
The default timeout for the http client is 60s which is too long for example when connecting to the DHS. This reduces the timeout to 5s but the actual amount is subject to testing

### DIFF
--- a/modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
+++ b/modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
@@ -17,7 +17,6 @@ import java.util.concurrent.{ExecutorService, Executors}
 import knobs.{Resource => _, _}
 import mouse.all._
 import org.asynchttpclient.DefaultAsyncHttpClientConfig
-import org.http4s.util.threads.threadFactory
 import org.http4s.client.asynchttpclient.AsyncHttpClient
 import org.http4s.client.Client
 import org.http4s.HttpRoutes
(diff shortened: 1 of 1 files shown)


modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -232,7 +232,17 @@ object WebServerLauncher extends IOApp with LogInitialization with SeqexecConfig
...
-    def engineIO(httpClient: Client[IO], guideConfigDb: GuideConfigDb[IO], collector: CollectorRegistry): Resource[IO, SeqexecEngine] =
+  // Override the default client config
+  val clientConfig = new DefaultAsyncHttpClientConfig.Builder()
+    .setMaxConnectionsPerHost(200)
+    .setMaxConnections(400)
+    .setRequestTimeout(5000) // Change the timeout to 5 seconds
+    .setThreadFactory(threadFactory(name = { i =>
+      s"http4s-async-http-client-worker-${i}"
+    }))
+    .build()
```
swalker2m: Maybe could also be

```scala
val clientConfig = new DefaultAsyncHttpClientConfig.Builder(AsyncHttpClient.defaultConfig)
  .setRequestTimeout(5000) // Change the timeout to 5 seconds
  .build()
```

to avoid resetting the other values?

cquiroz: Oh good idea

PR #839, opened 2019-02-28
SEQNG-914 Play either mp3 or webm depending on browser support

cquiroz:
Audio support is always a bit iffy. MP3 cannot be played on firefox in linux (e.g. on the ops consoles) but we can play webm instead. However webm isn't supported on Safari/iOS

This PR will attempt to play either format as supported by the browser


cquiroz:
Agreed it its a pain

PR #842, opened 2019-03-01
SEQNG-927 Fixed problem with loading Altair sequences.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -238,14 +238,15 @@ object TcsControllerEpics extends TcsController {
 
   def pauseGaos(gaos: Option[Either[Altair[IO], Gems[IO]]], current: EpicsTcsConfig, demand: TcsConfig)
   : SeqAction[(EpicsTcsConfig, TcsController.TcsConfig) => SeqAction[Unit]] = (gaos, demand.gaos).mapN {
-    case (Left(g), c)  => SeqAction.lift(g.pause(c, calcAoPauseConditions(current, demand)).map(resumeGaos))
+    case (Left(g), c)  => SeqActionF.embed[IO, (EpicsTcsConfig, TcsController.TcsConfig) => SeqAction[Unit]](
+      g.pause(c, calcAoPauseConditions(current, demand)).map(resumeGaos))
```
jluhrs: This is unrelated to the bug, but as a rule of thumb, I should use embed to go from `IO` to `SeqAction`, to capture possible exceptions.

PR #844, opened 2019-03-01
Switch Wavelength from Å to pm

swalker2m:
GMOS gratings have dispersion values represented in [fractional nm/pixel](https://www.gemini.edu/sciops/instruments/gmos/spectroscopy-overview/gratings).  These values are required to compute GMOS sequences and so will need to be stored in the `GmosNorthDisperser` and `GmosSouthDisperser` enums and database tables.   `Wavelength` currently stores its value in Å, which is not fine grained enough to represent dispersion.  For example, B1200 has a dispersion of 0.023 nm/pixel (or 23 pm/pixel).

This PR would switch `Wavelength` to pm instead of Å.  It was problematic mainly because I couldn't find valid optics for the conversions and ended up writing a lame `Wavelength.UnitConverter` with a similar API.  Other suggestions are welcome.


modules/core/shared/src/main/scala/gem/enum/F2Disperser.scala
```diff
@@ -22,9 +22,9 @@ sealed abstract class F2Disperser(
 
 object F2Disperser {
 
-  /** @group Constructors */ case object R1200JH extends F2Disperser("R1200JH", "R1200JH", "R=1200 (J + H) grism", Wavelength.fromAngstroms.unsafeGet(13900))
-  /** @group Constructors */ case object R1200HK extends F2Disperser("R1200HK", "R1200HK", "R=1200 (H + K) grism", Wavelength.fromAngstroms.unsafeGet(18710))
-  /** @group Constructors */ case object R3000 extends F2Disperser("R3000", "R3000", "R=3000 (J or H or K) grism", Wavelength.fromAngstroms.unsafeGet(16500))
+  /** @group Constructors */ case object R1200JH extends F2Disperser("R1200JH", "R1200JH", "R=1200 (J + H) grism", Wavelength.fromPicometers.unsafeGet(1390000))
+  /** @group Constructors */ case object R1200HK extends F2Disperser("R1200HK", "R1200HK", "R=1200 (H + K) grism", Wavelength.fromPicometers.unsafeGet(1871000))
+  /** @group Constructors */ case object R3000 extends F2Disperser("R3000", "R3000", "R=3000 (J or H or K) grism", Wavelength.fromPicometers.unsafeGet(1650000))
```
swalker2m: A good portion of this PR is just regenerated enums.

modules/core/shared/src/main/scala/gem/math/Wavelength.scala
```diff
@@ -9,33 +9,97 @@ import gem.syntax.prism._
 import monocle.Prism
 
 /**
- * Exact wavelengths represented as unsigned integral angstroms in the range [0 .. Int.MaxValue]
- * which means the largest representable wavelength is 214.7483647 mm.
- * @param toAngstroms This wavelength in integral angstroms (10^-10 of a meter).
+ * Exact wavelengths represented as unsigned integral picometers in the range [0 .. Int.MaxValue]
+ * which means the largest representable wavelength is 2.147483647 mm.
```
swalker2m: If this is problematic we could switch to `long` which would permit wavelengths of over 9 megameters.

modules/db/src/main/scala/gem/dao/StepDao.scala
```diff
@@ -398,6 +398,8 @@ object StepDao {
 
     object Gmos {
 
+      implicit val WavelengthMeta = WavelengthMetaAsNanometers
```
swalker2m: So now there is no implicit `Meta[Wavelength]`, which may be annoying idk.  GMOS seems to like wavelengths expressed in nm while GNIRS uses μm.

modules/sql/src/main/resources/db/migration/V063__Wavelength.sql
```diff
@@ -0,0 +1,47 @@
+--
+-- Adds comments to various tables, columns, types.
+--
+
+-- Switch GOMS to nm, with precision to distinguish pm 
+-- 10,3 can hold [0, 2147483647] pm as nm
+
+ALTER TABLE smart_gmos_north
+  ALTER COLUMN min_wavelength SET DATA TYPE numeric(10,3),
+  ALTER COLUMN max_wavelength SET DATA TYPE numeric(10,3);
```
swalker2m: It's kind of annoying to see, for example, 300 nm appear as `300.000` in the database but clipping it to anything less means an arbitrary `Wavelength` won't roundtrip.

modules/sql/src/main/scala/gem/sql/Wavelength.scala
```diff
@@ -11,19 +11,25 @@ import java.math.RoundingMode.HALF_UP
...
+      case Wavelength.Nm(_) => 3
+      case Wavelength.Um(_) => 6
     }
     toBigDecimal.underlying.scaleByPowerOfTen(n).setScale(0, HALF_UP).intValue
   }
+
 }
 
 object Wavelength {
+  final case class Pm(toBigDecimal: BigDecimal) extends Wavelength
```
swalker2m: This will be used in the dispersion value for GMOS grating enums.

modules/core/shared/src/main/scala/gem/math/Wavelength.scala
cquiroz: Is that necessary on the optical range?

modules/core/shared/src/main/scala/gem/math/Wavelength.scala
```diff
@@ -9,33 +9,97 @@ import gem.syntax.prism._
...
+      w.toPicometers / p
+
+  }
+
+  /**
+   * Creates a Wavelength from an Int in Å.
+   * @param Å in the range (0 to 21474836)
+   * @return a Wavelength provided Å is in range
+   */
+  def fromAngstroms: UnitConverter =
```
cquiroz: Should these be `unsafeFromAngstroms` given they use `UnitConverter`?

modules/core/shared/src/test/scala/gem/math/WavelengthSpec.scala
```diff
@@ -38,8 +46,62 @@ final class WavelengthSpec extends CatsSuite {
 
   test("Construction from an arbitrary Int must not allow negative values") {
     forAll { (n: Int) =>
-      Wavelength.fromAngstroms.getOption(n).isDefined shouldEqual n >= 0
+      Wavelength.fromPicometers.getOption(n).isDefined shouldEqual n >= 0
+    }
+  }
+
+  private def pow10(exp: Int): Int =
+    BigInt(10).pow(exp).toInt
```
cquiroz: Why `BigInt`?

modules/core/shared/src/main/scala/gem/math/Wavelength.scala
swalker2m: It's always safe to produce the value as `Int` Å and there is a safe option for reading an `Int` value in Å so I think `fromAngstroms` is appropriate.  What is unsafe is assuming that an integer value can be represented in the `Wavelength` type.  If it is less than 0 or more than 21474836 we cannot represent it with an `Int` pm.  For that there is `fromAngstroms.unsafeGet` like for actual `Optics`.

modules/core/shared/src/main/scala/gem/math/Wavelength.scala
swalker2m: IR cuts off at 1 mm so this should be fine.

modules/core/shared/src/test/scala/gem/math/WavelengthSpec.scala
swalker2m: Surprisingly there is no integer version of `Math.pow` and I wanted to avoid using the `Double` version and rounding, etc.  I guess it could do though.

modules/db/src/main/scala/gem/dao/StepDao.scala
tpolecat: I think this is what we want.

modules/sql/src/main/resources/db/migration/V063__Wavelength.sql
tpolecat: This seems correct. Keeping stuff stable as it round-trips is really important.

PR #848, opened 2019-03-05
SEQNG-934 Fixed issue with TCS in position

jluhrs:
The problem was that the in position value was filtered to remove spurious changes. That introduces a delay in the changes (the input has to be stable for XX seconds before the output changes). The problem is that when Seqexec checked the filtered value, it had not yet changed to "false" because of that delay.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
@@ -47,6 +47,7 @@ public class CaWindowStabilizer<T> implements CaAttribute<T> {
     }
 
     private synchronized void onValChange(T val) {
+
         if (val != null && val != lastVal) {
             if (timeoutFuture != null) {
                 timeoutFuture.cancel(true);
@@ -65,7 +66,7 @@ public class CaWindowStabilizer<T> implements CaAttribute<T> {
     private synchronized void onTimeout() {
         timeoutFuture = null;
         filteredVal = lastVal;
-        notifier.notifyValueChange(Arrays.asList(filteredVal));
+        notifier.notifyValueChange(values());
     }
 
     void unbind() {
@@ -97,7 +98,10 @@ public class CaWindowStabilizer<T> implements CaAttribute<T> {
 
     @Override
     public List<T> values() {
-        return Arrays.asList(filteredVal);
+        if(filteredVal != null)
+            return Arrays.asList(filteredVal);
+        else
+            return Arrays.asList();
     }
 
     @Override
@@ -141,6 +145,11 @@ public class CaWindowStabilizer<T> implements CaAttribute<T> {
     @Override
     public void removeListener(CaAttributeListener<T> listener) {
         notifier.removeListener(listener);
+    }
+
+    public CaWindowStabilizer<T> reset() {
+        filteredVal = lastVal = null;
+        return this;
     }
 
 }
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -37,8 +37,8 @@ object TcsControllerEpics extends TcsController {
 
   implicit private val encodeNodChopOption: EncodeEpicsValue[NodChopTrackingOption, String] =
     EncodeEpicsValue {
-      case NodChopTrackingOption.NodChopTrackingOn  => "on"
-      case NodChopTrackingOption.NodChopTrackingOff => "off"
+      case NodChopTrackingOption.NodChopTrackingOn  => "On"
+      case NodChopTrackingOption.NodChopTrackingOff => "Off"
```
jluhrs: These came from testing with the real system.

modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
```diff
@@ -143,4 +147,9 @@ public void removeListener(CaAttributeListener<T> listener) {
         notifier.removeListener(listener);
     }
 
+    public CaWindowStabilizer<T> reset() {
+        filteredVal = lastVal = null;
```
sraaphorst: It seems like this class relies fairly heavily upon `null`. Could we make it use `Optional<T>` instead?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
```diff
@@ -364,17 +364,17 @@ final class TcsEpics[F[_]: Sync](epicsService: CaService, tops: Map[String, Stri
...
   private val tcsStabilizeTime = 1.seconds
 
-  private val filteredInPositionAttr: CaAttribute[String] = new CaWindowStabilizer[String](inPositionAttr, java.time.Duration.ofMillis(tcsStabilizeTime.toMillis))
+  private val filteredInPositionAttr: CaWindowStabilizer[String] = new CaWindowStabilizer[String](inPositionAttr, java.time.Duration.ofMillis(tcsStabilizeTime.toMillis))
   def filteredInPosition:F[Option[String]] = safeAttribute(filteredInPositionAttr)
 
   // This functions returns a Task that, when run, will wait up to `timeout`
   // seconds for the TCS in-position flag to set to TRUE
-  def waitInPosition(timeout: Time): SeqAction[Unit] = waitForValue(filteredInPositionAttr, "TRUE", timeout,
+  def waitInPosition(timeout: Time): SeqAction[Unit] = waitForValue(filteredInPositionAttr.reset, "TRUE", timeout,
```
swalker2m: I'm wondering about the `.reset` which mutates `filteredInPositionAttr` but happens outside of `IO`?  If you call `waitInPosition` while a previously created `SeqAction` instance is running for example what happens?

modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
jluhrs: I agree with both points.

modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
```diff
@@ -97,7 +98,10 @@ public T value() {
 
     @Override
     public List<T> values() {
-        return Arrays.asList(filteredVal);
+        if(filteredVal != null)
+            return Arrays.asList(filteredVal);
+        else
+            return Arrays.asList();
```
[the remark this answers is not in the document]
jluhrs: You are right. The other implementation is `CaAttributeImpl`, which reads an EPICS channel directly. In that case you either have a single value, an array, or a `null`. I should think on how to make both conform to the same specification.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
jluhrs: You are right. It works now because the function happens to be called inside an `IO`, but it should be explicit.

PR #850, opened 2019-03-05
SEQNG-935 Completed guider probe configuration.

jluhrs:
Implemented the follow and park control for probes.
The code became a bit convoluted when handling PWFS2. It is a special case, because PWFS2 cannot be used when using Altair or GeMS. I am not happy with this first implementation, but I am bit pressed of time and need to move to Altair control code, so I'll leave the code improvements for another task (SEQNG-936)

### DIFF
--- a/modules/seqexec/server/src/main/resources/Tcs.xml
+++ b/modules/seqexec/server/src/main/resources/Tcs.xml
@@ -1821,6 +1821,11 @@
             <description>AO fold position</description>
             <top>ag</top>
         </attribute>
+        <!--<attribute name="useAo">-->
+            <!--<channel>im:AOConfigFlag.VAL</channel>-->
+            <!--<type>STRING</type>-->
+            <!--<description>Using AO flag</description>-->
+        <!--</attribute>-->
         <attribute name="airmass">
             <channel>sad:airMass</channel>
             <type>DOUBLE</type>
--- a/modules/seqexec/server/src/main/scala/seqexec/server/altair/Altair.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/altair/Altair.scala
@@ -41,6 +41,8 @@ class Altair[F[_]: Sync] private (controller: AltairController[F],
          Ngs(true) => true
     case _         => false
   }
+
+  def isFollowing: F[Option[Boolean]] = controller.isFollowing
 }
 
 object Altair {
--- a/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
@@ -316,4 +316,6 @@ object AltairControllerEpics extends AltairController[IO] {
    */
   def convertTcsAction[T](a: SeqAction[T]): IO[T] = a.value.flatMap(_.fold(IO.raiseError(_), IO(_)))
 
+  // This is a bit convoluted. AO follow state is read from Altair, but set as part of TCS configuration
+  override def isFollowing: IO[Option[Boolean]] = AltairEpics.instance.aoFollow
 }
--- a/modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
@@ -35,16 +35,19 @@ final case class Tcs private (tcsController: TcsController,
 
   override val resource: Resource = Resource.TCS
 
+  private val usesAltair = gaos.flatMap(_.swap.toOption).isDefined
+
   // Helper function to output the part of the TCS configuration that is actually applied.
   private def subsystemConfig(tcs: TcsConfig, subsystem: Subsystem): List[String] = subsystem match {
     case Subsystem.M1     => List(tcs.gc.m1Guide.show)
     case Subsystem.M2     => List(tcs.gc.m2Guide.show)
     case Subsystem.OIWFS  => List((tcs.gds.oiwfs:GuiderConfig).show)
     case Subsystem.PWFS1  => List((tcs.gds.pwfs1:GuiderConfig).show)
-    case Subsystem.PWFS2  => List((tcs.gds.pwfs2:GuiderConfig).show)
+    case Subsystem.PWFS2  => List(tcs.gds.pwfs2OrAowfs.swap.getOrElse(
+      GuiderConfig(ProbeTrackingConfig.Parked, GuiderSensorOff)).show)
     case Subsystem.Mount  => List(tcs.tc.show)
     case Subsystem.AGUnit => List(tcs.agc.sfPos.show, tcs.agc.hrwfs.show)
-    case Subsystem.Gaos   => List("") //TODO: show Gaos configuration
+    case Subsystem.Gaos   => List(tcs.gds.pwfs2OrAowfs.toOption).collect{ case Some(x) => (x:ProbeTrackingConfig).show }
   }
 
   override def configure(config: Config): SeqAction[ConfigResult[IO]] =
@@ -89,9 +92,9 @@ final case class Tcs private (tcsController: TcsController,
         case Some(Left(_))         => true
         case _                     => false
       }
-      val aoUsesP1 = (gaos.flatMap(_.swap.toOption), c.gaosGuide.flatMap(_.swap.toOption)).mapN(_.usesP1(_))
+      val aoUsesP1 = useAo && (gaos.flatMap(_.swap.toOption), c.gaosGuide.flatMap(_.swap.toOption)).mapN(_.usesP1(_))
         .getOrElse(false)
-      val aoUsesOI = (gaos.flatMap(_.swap.toOption), c.gaosGuide.flatMap(_.swap.toOption)).mapN(_.usesOI(_))
+      val aoUsesOI = useAo && (gaos.flatMap(_.swap.toOption), c.gaosGuide.flatMap(_.swap.toOption)).mapN(_.usesOI(_))
         .getOrElse(false)
 
       TcsConfig(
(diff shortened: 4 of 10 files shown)


modules/seqexec/server/src/main/resources/Tcs.xml
```diff
@@ -1821,6 +1821,11 @@
             <description>AO fold position</description>
             <top>ag</top>
         </attribute>
+        <!--<attribute name="useAo">-->
+            <!--<channel>im:AOConfigFlag.VAL</channel>-->
+            <!--<type>STRING</type>-->
+            <!--<description>Using AO flag</description>-->
+        <!--</attribute>-->
```
jluhrs: Added the attribute only to find that it was an Enum (which are not supported in the xml configuration file). I left it commented to keep the information about the channel.

modules/seqexec/server/src/main/scala/seqexec/server/altair/Altair.scala
```diff
@@ -41,6 +41,8 @@ class Altair[F[_]: Sync] private (controller: AltairController[F],
          Ngs(true) => true
     case _         => false
   }
+
+  def isFollowing: F[Option[Boolean]] = controller.isFollowing
```
jluhrs: I needed this method to access this value from TCS, because even if the following state is in Altair, the command to change the state is in TCS

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsController.scala
```diff
@@ -356,9 +357,8 @@ object TcsController {
   @Lenses
   final case class GuidersConfig(
     pwfs1: GuiderConfig@@P1Config,
-    pwfs2: GuiderConfig@@P2Config,
-    oiwfs: GuiderConfig@@OIConfig,
-    aowfs: Boolean@@AOGuide
+    pwfs2OrAowfs: Either[GuiderConfig@@P2Config, ProbeTrackingConfig@@AoGuide],
```
jluhrs: Not the best option. There is code dealing with this `Either` all over the place.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -50,18 +50,111 @@ object TcsControllerEpics extends TcsController {
...
+  : Option[EpicsTcsConfig => SeqAction[EpicsTcsConfig]] =
+    if(subsystems.contains(Subsystem.PWFS1)) {
+      val actions = List(
+        (c.getNodChop =!= d.getNodChop)
+          .option(setNodChopProbeTrackingConfig(TcsEpics.instance.pwfs1ProbeGuideCmd)(d.getNodChop)),
+        d match {
+          case ProbeTrackingConfig.Parked => (c =!= ProbeTrackingConfig.Parked).option(TcsEpics.instance.pwfs1Park.mark)
+          case _                          => (c.follow =!= d.follow)
+            .option(TcsEpics.instance.pwfs1ProbeFollowCmd.setFollowState(encode(d.follow)))
+        }
```
jluhrs: This is the part that was missing.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsConfigRetriever.scala
```diff
@@ -193,26 +203,26 @@ object TcsConfigRetriever {
...
-      aof  <- getAoFold
-      sf   <- getScienceFoldPosition
-      hr   <- getHrwfsPickupPosition
-      ports <- getInstrumentPorts
+      iaa    <- getIAA
+      offX   <- getOffsetX
+      offY   <- getOffsetY
+      wl     <- getWavelength
+      p1     <- getPwfs1
+      p2OrAo <- getPwfs2OrAowfs(getAoFollow)
```
jluhrs: This structure s used to retrieve the current TCS configuration from the actual TCS EPICS system. It has to deal with mutually exclusive P2 or AO configuration.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -35,16 +35,19 @@ final case class Tcs private (tcsController: TcsController,
 
   override val resource: Resource = Resource.TCS
 
+  private val usesAltair = gaos.flatMap(_.swap.toOption).isDefined
```
sraaphorst: Just out of curiosity, is there a reason you use `swap` here on the `Either` instead of just accessing `left`? This may be nitpicky, but it seems like it would be more clear to say:

```
gaos.exists(_.isLeft)
```

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -102,16 +105,16 @@ final case class Tcs private (tcsController: TcsController,
...
             config.guideWithP1)
           ),
-          tag[P2Config](calcGuiderConfig(
-            calcGuiderInUse(c.tcsGuide, TipTiltSource.PWFS2, M1Source.PWFS2),
-            config.guideWithP2)
+          usesAltair.either(
+            tag[P2Config](calcGuiderConfig(calcGuiderInUse(c.tcsGuide, TipTiltSource.PWFS2, M1Source.PWFS2),
+              config.guideWithP2)),
+            tag[AoGuide](calcGuiderConfig(calcGuiderInUse(c.tcsGuide, TipTiltSource.GAOS, M1Source.GAOS),
+              config.guideWithAO).tracking)
```
sraaphorst: I'm not familiar with shapeless yet, so I'll defer to @swalker2m to provide any feedback here if he has any.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
```diff
@@ -320,4 +320,6 @@ object AltairControllerEpics extends AltairController[IO] {
    */
   def convertTcsAction[T](a: SeqAction[T]): IO[T] = a.value.flatMap(_.fold(IO.raiseError(_), IO(_)))
 
+  // This is a bit convoluted. AO follow state is read from Altair, but set as part of TCS configuration
+  override def isFollowing: IO[Option[Boolean]] = AltairEpics.instance.aoFollow
```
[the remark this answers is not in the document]
jluhrs: Yes. Because we are reading the value from an EPICS system, there is the third option: failed to read the value.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
```diff
@@ -145,6 +145,34 @@ final class TcsEpics[F[_]: Sync](epicsService: CaService, tops: Map[String, Stri
...
+    private val follow = cs.map(_.getString("followState"))
+    def setFollowState(v: String): SeqAction[Unit] = setParameter(follow, v)
+  }
+
+  object aoProbeFollowCmd extends EpicsCommand {
+    override protected val cs: Option[CaCommandSender] = Option(epicsService.getCommandSender("aoFollow"))
+
+    private val follow = cs.map(_.getString("followState"))
+    def setFollowState(v: String): SeqAction[Unit] = setParameter(follow, v)
+  }
```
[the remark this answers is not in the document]
jluhrs: They control different mechanism. The parameter to `FollowCmd` is the name of a `CaCommandSender`, and each one is different.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsController.scala
jluhrs: I am not sure. I still need to remember the base class, because that is what is really used, and I must be forced to explicitly set the tag when passing the value to make sure I am using it in the right place.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
jluhrs: Duh!

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -50,18 +50,111 @@ object TcsControllerEpics extends TcsController {
...
+        d match {
+          case ProbeTrackingConfig.Parked => (c =!= ProbeTrackingConfig.Parked).option(TcsEpics.instance.pwfs2Park.mark)
+          case _                          => (c.follow =!= d.follow)
+            .option(TcsEpics.instance.pwfs2ProbeFollowCmd.setFollowState(encode(d.follow)))
+        }
+      ).collect{ case Some(x) => x }
+
+      actions.nonEmpty.option{ x => actions.sequence *>
+        SeqAction(EpicsTcsConfig.pwfs2OrAowfs.modify(_.leftMap(GuiderConfig.tracking.set(d)))(x))
+      }
```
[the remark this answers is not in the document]
jluhrs: I thought about it. The code looks like a candidate to factor out a common function, but it would require 5 additional parameters.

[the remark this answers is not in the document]
jluhrs: Or I could create a `case class` with all the relevant parameters for each guide probe...

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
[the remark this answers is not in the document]
jluhrs: You are right, and doing it like that helped me generalize a function in `TcsControllerEpics`.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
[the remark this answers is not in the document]
jluhrs: Which is what I did.

PR #851, opened 2019-03-06
SEQNG-913 Implemented support for Altair special control records

jluhrs:
Altair uses a special record to control commands, called taskControl. This PR adds support to handle those records, to be able to change parameters, trigger them , and monitor their execution.
Warning: this is in the EPICS support package ACM, in Java.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandSender.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandSender.java
@@ -13,7 +13,7 @@ import java.util.Set;
  * @author jluhrs
  *
  */
-public interface CaCommandSender {
+public interface CaCommandSender extends CaCommandTrigger {
     /**
      * Retrieves the name of this command sender.
      *
@@ -54,40 +54,6 @@ public interface CaCommandSender {
     void clear() throws TimeoutException;
 
     /**
-     * Trigger the EPICS apply record. The command return immediately.
-     *
-     * @return an object that implements <code>CaCommandMonitor</code>, which
-     *         can be used to monitor the command execution and retrieve its
-     *         result.
-     */
-    CaCommandMonitor post();
-
-    /**
-     * Trigger the EPICS apply record and waits until the command processing
-     * completes.
-     *
-     * @return an object that implements <code>CaCommandMonitor</code>, which
-     *         can be used to monitor the command execution and retrieve its
-     *         result.
-     */
-    CaCommandMonitor postWait() throws InterruptedException;
-
-    /**
-     * Trigger the EPICS apply record. The command return immediately.
-     *
-     * @param callback
-     *            an object that implements <code>CaCommandListener</code>,
-     *            which will be notified when the command execution state
-     *            changes. The object will be used only for this execution of
-     *            the command
-     *
-     * @return an object that implements <code>CaCommandMonitor</code>, which
-     *         can be used to monitor the command execution and retrieve its
-     *         result.
-     */
-    CaCommandMonitor postCallback(CaCommandListener callback);
-
-    /**
      * Adds a parameter of type <code>Integer</code> to this command sender. If
      * the parameter already exist, the existing object is used. CaException is
      * thrown if the existing parameter is of a different type or uses a
@@ -103,16 +69,21 @@ public interface CaCommandSender {
      * @throws CaException
      */
     CaParameter<Integer> addInteger(String name, String channel,
-                                    String description) throws CaException;
-
-    CaParameter<Integer> addInteger(String name, String channel)
-            throws CaException;
-
-    CaParameter<Integer> addInteger(String name, String channel,
                                     String description, boolean isCADParameter) throws CaException;
 
-    CaParameter<Integer> addInteger(String name, String channel, boolean isCADParameter)
-            throws CaException;
+    default CaParameter<Integer> addInteger(String name, String channel)
+            throws CaException {
+        return addInteger(name, channel, null, true);
+    }
+
+    default CaParameter<Integer> addInteger(String name, String channel, boolean isCADParameter) throws CaException {
+        return addInteger(name, channel, null, isCADParameter);
+    }
+
+    default CaParameter<Integer> addInteger(String name, String channel,
+            String description) throws CaException {
+        return addInteger(name, channel, description, true);
+    }
 
     /**
      * Adds a parameter of type <code>Double</code> to this command sender. If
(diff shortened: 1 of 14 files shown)


modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControl.java
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+ * For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+package edu.gemini.epics.acm;
+
+public interface CaTaskControl extends CaApplySender, CaCommandSender {
```
jluhrs: So basically a taskControl record is a CAD, an Apply and a CAR all in one.

modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandSender.java
```diff
@@ -13,7 +13,7 @@
  * @author jluhrs
  *
  */
-public interface CaCommandSender {
+public interface CaCommandSender extends CaCommandTrigger {
```
sraaphorst: What is the difference between an Apply and a Command, if you don't mind me asking?

modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandSender.java
```diff
@@ -103,16 +69,21 @@
...
-            throws CaException;
+    default CaParameter<Integer> addInteger(String name, String channel)
+            throws CaException {
+        return addInteger(name, channel, null, true);
+    }
 
-    CaParameter<Integer> addInteger(String name, String channel,
-                                    String description, boolean isCADParameter) throws CaException;
+    default CaParameter<Integer> addInteger(String name, String channel, boolean isCADParameter) throws CaException {
+        return addInteger(name, channel, null, isCADParameter);
```
sraaphorst: Again, these `null` values make me kind of uncomfortable, especially when there is no documentation in the getter that states that the return value can be `null`.

modules/acm/src/main/java/edu/gemini/epics/acm/CaParameterStorage.java
```diff
@@ -0,0 +1,215 @@
...
+
+import edu.gemini.epics.EpicsWriter;
+import gov.aps.jca.CAException;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+class CaParameterStorage {
```
sraaphorst: What is the advantage of separating out `CaParameterStorage` from `CaCommandSenderImpl`?

It seems now like `CaCommandSenderImpl` is possibly just an unnecessary layer of abstraction as an interface to this class, or am I missing something?

Is this so that you can use it in `CaTaskControlImpl` as well? Something feels a bit repetitive here, but maybe I just haven't seen enough of the PR yet (I'm commenting as I go).

modules/acm/src/main/java/edu/gemini/epics/acm/CaService.java
```diff
@@ -123,6 +125,9 @@ public void unbind() {
         for (CaCommandSenderImpl cs : commandSenders.values()) {
             cs.unbind();
         }
+        for (CaTaskControlImpl tcs : taskControlSenders.values()) {
```
sraaphorst: Just a point... this could be written:
```
taskControlSenders.values().forEach(CaTaskControlImpl::unbind);
```

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlImpl.java
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    @Override
+    public String getApply() {
+        return null;
+    }
+
+    @Override
+    public String getCAR() {
+        return null;
```
sraaphorst: I'm a bit confused here... I thought you said a `taskControl` was a CAD and a CAR. Why does it return `null` for `getCAR()`?

(This could just be me not having done more than a modicum of EPICS programming.)

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlImpl.java
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    @Override
+    public CaParameter<Float> addFloat(String name, String channel, String description, boolean isCADParameter)
+            throws CaException {
+        return storage.addFloat(name, channel, description, isCADParameter);
+    }
+
+    @Override
+    public <T extends Enum<T>> CaParameter<T> addEnum(String name, String channel, Class<T> enumType,
```
sraaphorst: These types of methods are what seem repetitive. Is it not possible to abstract this out to a superclass?

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlImpl.java
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    private synchronized void onTimeout() {
+        timeoutFuture = null;
+        State oldState = currentState;
+        currentState = currentState.onTimeout();
+        if(trace) LOG.debug("onTimeout: " + oldState.signature() + " -> " + currentState.signature());
+    }
+
+    private void succedCommand(final CaCommandMonitorImpl cm) {
```
sraaphorst: Minor typo: `succeedCommand`

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlImpl.java
```diff
@@ -0,0 +1,477 @@
...
+        executor.execute(new Runnable() {
+            @Override
+            public void run() {
+                cm.completeSuccess();
+            }
+        });
+    }
+
+    private void failCommand(final CaCommandMonitorImpl cm, final Exception ex) {
+        executor.execute(new Runnable() {
```
sraaphorst: Implementing `Runnable` like this is also overkill now. Since `Runnable` is functional interface, you can write this as:

```
executor.execute(() -> { cm.completeFailure(ex); });
```

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlRecord.java
```diff
@@ -0,0 +1,190 @@
...
+                }
+            } catch (Throwable e) {
+                LOG.warn(e.getMessage());
+            }
+        }
+
+    }
+
+    synchronized void unbind() {
+        assert(epicsReader!=null);
```
sraaphorst: Do we really want `assert` here? I don't think they're intended for production code.

I think it might be better to use:

```
Objects.requireNonNull(epicsReader);
```

modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandSender.java
jluhrs: That is related to Gemini EPICS records. A command is a record (CAD) that has inputs, and can be triggered. An Apply is a record that has no parameters, but connects to one or more CAD records. It can be triggered, and it will trigger the connected CADs in sequence. Another difference is that the Apply record can produce the CLID, the number that identifies one specific instance of execution. Both CADs and Applies can be associated to a CAR record that publishes the execution state of the CAD or Apply.

modules/acm/src/main/java/edu/gemini/epics/acm/CaParameterStorage.java
jluhrs: I separate it to be able to use it in `CaTaskRecord`, the other type of record that has parameters.

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlImpl.java
jluhrs: The CAR record is used to publish the execution state of a CAD or Apply record. A taskControl record has an output that tell the execution state of the record. In that way, it absorbs the function of a CAR and does not need t o be associated with one.

modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandSender.java
jluhrs: Now the documentation states the description can be `null`.

modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandSender.java
sraaphorst: Thanks very much for that great explanation! I really appreciate it.

PR #852, opened 2019-03-07
SEQNG-931 Removed redundant message.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -305,7 +305,7 @@ object EpicsUtil {
                 locked(lock) {
                   attr.removeListener(statusListener)
                 }
-                f(TrySeq.fail(SeqexecFailure.Timeout(s"waiting for $name.")).asRight)
+                f(TrySeq.fail(SeqexecFailure.Timeout(name)).asRight)
```
jluhrs: The 'explain' method already puts the "waiting for" for `Timeout`.

PR #854, opened 2019-03-08
SEQNG-925 Fixed reading of current SF position.

jluhrs:
Reading of the science fold position would sometimes cause an error. It is OK if the error was that it couldn't read an EPICS channel, but in most cases, the error was caused because it could not transform the read position into a certain type, which is not OK. The type used was designed to represent positions that make sense for a given observation, but missed other posible values of the science fold position.
The solution introduces a new type to represent the real science fold position.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
@@ -4,61 +4,46 @@
 package seqexec.server.tcs
 
 import cats.implicits._
-import mouse.boolean._
+import mouse.string._
 import gem.enum.LightSinkName
 import seqexec.server.EpicsCodex.{DecodeEpicsValue, EncodeEpicsValue}
-import seqexec.server.tcs.TcsController.{LightSource, ScienceFoldPosition}
-import seqexec.server.tcs.TcsControllerEpics.InstrumentPorts
+import seqexec.server.tcs.TcsController.LightSource
+import seqexec.server.tcs.TcsControllerEpics.ScienceFold.{Parked, Position}
+import seqexec.server.tcs.TcsControllerEpics.ScienceFold
 
 // Decoding and encoding the science fold position require some common definitions, therefore I
 // put them inside an object
 private[server] object ScienceFoldPositionCodex {
 
   import LightSource._
-  import ScienceFoldPosition._
 
   private val AO_PREFIX = "ao2"
   private val GCAL_PREFIX = "gcal2"
   private val PARK_POS = "park-pos"
 
-  val BottomPort: Int = 1
-  val InvalidPort: Int = 0
-
-  def portFromSinkName(ports: InstrumentPorts)(n: LightSinkName): Option[Int] = {
-    val port = n match {
-      case LightSinkName.Gmos |
-           LightSinkName.Gmos_Ifu => ports.gmosPort
-      case LightSinkName.Niri_f6 |
-           LightSinkName.Niri_f14 |
-           LightSinkName.Niri_f32 => ports.niriPort
-      case LightSinkName.Nifs     => ports.nifsPort
-      case LightSinkName.Gnirs    => ports.gnirsPort
-      case LightSinkName.F2       => ports.flamingos2Port
-      case LightSinkName.Gpi      => ports.gpiPort
-      case LightSinkName.Ghost    => ports.ghostPort
-      case LightSinkName.Gsaoi    => ports.gsaoiPort
-      case LightSinkName.Ac |
-           LightSinkName.Hr       => BottomPort
-      case LightSinkName.Phoenix |
-           LightSinkName.Visitor  => InvalidPort
-    }
-    (port =!= InvalidPort).option(port)
-  }
-
   private def findSinkInSFName(str: String): Option[LightSinkName] =
     LightSinkName.all.find(i => str.startsWith(i.name))
 
-  implicit val decodeScienceFoldPosition: DecodeEpicsValue[String, Option[ScienceFoldPosition]] = DecodeEpicsValue(
+  private def findPortInSFName(str: String): Option[Int] = str.parseIntOption
+
+  implicit val decodeScienceFold: DecodeEpicsValue[String, Option[ScienceFold]] = DecodeEpicsValue(
     (t: String) => if (t.startsWith(PARK_POS)) Parked.some
-    else if (t.startsWith(AO_PREFIX))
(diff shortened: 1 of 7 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsController.scala
```diff
@@ -254,21 +254,13 @@ object TcsController {
     implicit val eq: Eq[LightSource] =  Eq.fromUniversalEquals
   }
 
-  /** Data type for science fold position. */
-  sealed trait ScienceFoldPosition
-  object ScienceFoldPosition {
-    case object Parked extends ScienceFoldPosition
-    final case class Position(source: LightSource, sink: LightSinkName) extends ScienceFoldPosition
+  /* Data type for science fold position. */
+  final case class LightPath(source: LightSource, sink: LightSinkName)
```
jluhrs: This type replaces `ScienceFoldPosition`. The problem with `ScienceFoldPosition` is that it tried to represent the light path request and the science fold position, and failed at both. It couldn't represent some posible values for the science fold position, and added a `Parked` option that didn't make sense for the requested light path.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -440,7 +485,7 @@ object TcsControllerEpics extends TcsController {
     oiwfs: GuiderConfig,
     telescopeGuideConfig: TelescopeGuideConfig,
     aoFold: AoFold,
-    scienceFoldPosition: ScienceFoldPosition,
+    scienceFoldPosition: Option[ScienceFold],
```
jluhrs: `Option` because the value read for the science fold position could be "invalid". If it is `None`, the science fold will be configured.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
```diff
@@ -4,66 +4,52 @@
...
+  private def findPortInSFName(str: String): Option[Int] = str.parseIntOption
+
+  implicit val decodeScienceFold: DecodeEpicsValue[String, Option[ScienceFold]] = DecodeEpicsValue(
     (t: String) => if (t.startsWith(PARK_POS)) Parked.some
-    else if (t.startsWith(AO_PREFIX))
-      findSinkInSFName(t.substring(AO_PREFIX.length)).map(Position(AO, _))
-    else if (t.startsWith(GCAL_PREFIX))
-      findSinkInSFName(t.substring(GCAL_PREFIX.length)).map(Position(GCAL, _))
-    else findSinkInSFName(t).map(Position(Sky, _))
+    else if (t.startsWith(AO_PREFIX)) for {
```
tpolecat: Just FYI you can encode this kind of thing as a pattern-match (rather than a long `if`/`else if` expression) by using regex extractors. It don't know if it's worth it here, but just something to keep in mind.

```scala
@ val ao = "^AO.*".r 
ao: scala.util.matching.Regex = ^AO.*

@ def blah(s: String) =
    s match {
      case ao() => 'yes
      case _    => 'no
    } 
defined function blah

@ blah("AO123") 
res27: Symbol = 'yes

@ blah("xAO123") 
res28: Symbol = 'no
```

modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
```diff
@@ -4,66 +4,52 @@
...
+      port <- findPortInSFName(t.substring(AO_PREFIX.length + sink.name.length))
+    } yield Position(AO, sink, port)
+    else if (t.startsWith(GCAL_PREFIX)) for {
+      sink <- findSinkInSFName(t.substring(GCAL_PREFIX.length))
+      port <- findPortInSFName(t.substring(GCAL_PREFIX.length + sink.name.length))
+    } yield Position(GCAL, sink, port)
+    else for {
+      sink <- findSinkInSFName(t)
+      port <- findPortInSFName(t.substring(sink.name.length))
+    } yield Position(Sky, sink, port)
```
swalker2m: We have @tpolecat 's `Atto` for this kind of thing so we're not having to work with substrings and indices.  I'm not an expert and I haven't actually tried to compile, but it might go something like this:

```scala
import atto._, Atto._

val lightSink: Parser[LightSinkName] =
  many(letter).map { cs =>
    val sinkName = cs.mkString
    LightSinkName.all.find(_.name == sinkName)
  }.collect { case Some(lsn) => lsn }

def prefixed(p: String, s: LightSource): Parser[Position] =
  (string(p) ~> lightSink ~ int).map { case (ls, port) => Position(s, ls, port) }

val park: Parser[Position] =
  (string(PARK_POS) <~ many(anyChar)).as(Parked)

(park | prefixed(AO_PREFIX, AO) | prefixed(GCAL_PREFIX, GCAL) | prefixed("", SKY)).parseOnly(t).option
```

tpolecat: Oh I didn't even look at what was happening in the `find...` methods. Yes this would be much better than mucking around with strings.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
```diff
@@ -4,66 +4,49 @@
...
-      case LightSinkName.Gsaoi    => ports.gsaoiPort
-      case LightSinkName.Ac |
-           LightSinkName.Hr       => BottomPort
-      case LightSinkName.Phoenix |
-           LightSinkName.Visitor  => InvalidPort
-    }
-    (port =!= InvalidPort).option(port)
-  }
+  def prefixed(p: String, s: LightSource): Parser[ScienceFold] =
+    (string(p) ~> lightSink ~ int).map { case (ls, port) => Position(s, ls, port) }
```
cquiroz: Yeah it is much nicer this way

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
cquiroz: Is "Invalid" the value in the epics channel?

jluhrs: Actually, some positions have the value "undefined". And if the current position does not match any position in the LUT, the value is "INVALID". The important point is that the value is read to decide if it necessary to configure it. If the value is unknown, then it doesn't matter, we just need to configure it to the requested value.

PR #855, opened 2019-03-08
SEQNG-896 Fixed calculation of EXPTIME for NIRI.

jluhrs:
Easy one. The value for the NIRI keyword EXPTIME must be the sum of two values read from the instrument.


jluhrs:
I just follow the scientist instructions.

arturog8m:
Let's put those instructions at least as a comment in the code, please.

PR #856, opened 2019-03-08
SEQNG-905 Check GMOS DHS connection before observe.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -263,24 +263,28 @@ class GmosControllerEpics[T<:GmosController.SiteDependentTypes](encoders: GmosCo
...
+
   override def applyConfig(config: GmosController.GmosConfig[T]): SeqAction[Unit] = {
     val params = setDCConfig(config.dc) ++ setCCConfig(config.cc)
 
     EitherT.right[SeqexecFailure](IO(Log.info("Start Gmos configuration"))) *>
     EitherT.right(IO(Log.debug(s"Gmos configuration: ${config.show}"))) *>
-    (if(params.isEmpty) SeqAction.void
-    else params.sequence *>
+    SeqAction(Log.warn("GMOS is not connected to the DHS"))
+      .unlessA(GmosEpics.instance.dhsConnected.exists(_.trim === dhsConnected)) *>
```
tpolecat: Is `GmosEpics.instance.dhsConnected` a constant? If not this needs to be lifted into a `Sync` effect.

jluhrs: Yes, this and all other attributes in GMOS and other instruments. There is a Jira task for that work (SEQNG-887).

PR #858, opened 2019-03-09
SEQNG-939 First batch of Altair bugfixes

jluhrs:
* Predicting the next AO target position before applying an offset requires to know the current value but retrieving it at each step is unreliable. Instead, now TCC captures the position when the Altair is enabled, and passes it to Seqexec.
* Fixed a bug where Altair was not enabled if was not disabled in the same step (it would fail to restart after an unguided step)
* In Altair, check the current status to decide if it necessary to send a command.
* Fixed some typos in the code the retrieves the current P2 and OI configuration from TCS.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
@@ -31,13 +31,8 @@ object AltairControllerEpics extends AltairController[IO] {
   private def inRangeLinear[T <: Ordered[T]](vMin: T, vMax: T)(v: T): Boolean =
     v > vMin && v < vMax
 
-  private def newPosition(currCfg: EpicsAltairConfig)(prev: FocalPlaneOffset, next: FocalPlaneOffset)
-  : (Length, Length) = {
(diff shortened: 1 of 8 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
```diff
@@ -77,46 +72,70 @@ object AltairControllerEpics extends AltairController[IO] {
...
+  }
+
+  private def pauseNgsMode(starPos: (Length, Length), fieldLens: FieldLens, currCfg: EpicsAltairConfig)(reasons: Set[Gaos.PauseCondition])
+  : IO[Set[ResumeCondition] => IO[Unit]] = {
+    pauseNgsOrLgsMode(starPos, fieldLens, isLgs = false, currCfg)(reasons).map(resumeNgsMode(starPos, _))
   }
 
   private val dumbResume = (_: Set[ResumeCondition]) => IO.unit
+  val aoSettledTimeout: Time = 30.0.seconds
+  val matrixPrepTimeout: Time = 10.seconds
```
sraaphorst: Why these numbers specifically? Is there any reason?

PR #859, opened 2019-03-12
SEQNG-940 For Altair, fixed SF set position, prepare matrix logic.,

jluhrs:
More bug fixes for Altair:
* There was a bug in the logic to determine if the control matrix was OK.
* The channels for the current control matrix star position were swapped.
* The science fold position was using the wrong light source for Altair

### DIFF
--- a/modules/seqexec/server/src/main/resources/Altair.xml
+++ b/modules/seqexec/server/src/main/resources/Altair.xml
@@ -103,12 +103,14 @@
             <description>AO loop settled</description>
         </attribute>
         <attribute name="conmatx">
-            <channel>conMatX</channel>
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/resources/Altair.xml
```diff
@@ -103,12 +103,14 @@
             <description>AO loop settled</description>
         </attribute>
         <attribute name="conmatx">
-            <channel>conMatX</channel>
+            <!-- X/Y coordinates in Altair are swapped with respect to X/Y in the telescope focal plane -->
+            <channel>conMatY</channel>
```
cquiroz: That sounds fun to debug

cquiroz: Should the attribute name be changed to match?

PR #860, opened 2019-03-12
SEQNG-942 Keep frozen probes in following state.

jluhrs:
Probes follow demands depending on two settings. The nod/chop guide configuration tells TCS when to send demands to the probes. The follow state tells the probe to follow such demands. When a big offset is applied, probes are frozen: they stay in their last position, and don't move. In this case, Seqexec was setting all nod/chop values to Off, and follow to Off. The last one was unnecessary, but also should have been inocuos. It ended up that it was causing problems with OIWFS, so I changed the code to not turn follow Off when freezing a probe (which is the behavior of the Tcl Seqexec)

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
@@ -137,7 +137,7 @@ object Tcs {
   implicit class GuideWithOps(guideWith: StandardGuideOptions.Value) {
     val toProbeTracking: ProbeTrackingConfig = guideWith match {
       case StandardGuideOptions.Value.park => ProbeTrackingConfig.Parked
-      case StandardGuideOptions.Value.freeze => ProbeTrackingConfig.Off
+      case StandardGuideOptions.Value.freeze => ProbeTrackingConfig.Frozen
       case StandardGuideOptions.Value.guide => ProbeTrackingConfig.On(NodChopTrackingConfig.Normal)
     }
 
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -137,7 +137,7 @@ object Tcs {
   implicit class GuideWithOps(guideWith: StandardGuideOptions.Value) {
     val toProbeTracking: ProbeTrackingConfig = guideWith match {
       case StandardGuideOptions.Value.park => ProbeTrackingConfig.Parked
-      case StandardGuideOptions.Value.freeze => ProbeTrackingConfig.Off
+      case StandardGuideOptions.Value.freeze => ProbeTrackingConfig.Frozen
```
jluhrs: Frozen is a new option, that has follow on and all chop/nod guiding off.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsController.scala
```diff
@@ -204,12 +204,14 @@ object TcsController {
...
     final case class On(ndconfig: ActiveNodChopTracking) extends ProbeTrackingConfig(FollowOn, ndconfig)
+    case object Frozen extends ProbeTrackingConfig(FollowOn, NodChopTrackingConfig.AllOff)
 
     implicit val onEq: Eq[On] = Eq.by(_.ndconfig)
 
     implicit val eq: Eq[ProbeTrackingConfig] = Eq.instance{
       case (Parked, Parked)   => true
       case (Off, Off)         => true
+      case (Frozen, Frozen)   => true
       case (a@On(_), b@On(_)) => a === b
```
cquiroz: I think this can be `case (a: On, b: On) => a === b`

PR #862, opened 2019-03-12
SEQNG-877 Show the appropriate NIFS fields on the steps table

cquiroz:
This PR adds support to show NIFS fields on the steps table

Seqexec - GS-2018B-Q-0-98 2019-03-12 11-07-48

### DIFF
--- a/modules/seqexec/model/src/test/scala/seqexec/model/ModelLensesSpec.scala
+++ b/modules/seqexec/model/src/test/scala/seqexec/model/ModelLensesSpec.scala
@@ -64,4 +64,12 @@ final class ModelLensesSpec extends CatsSuite with ModelLenses {
            OptionalTests(instrumentDisperserLambdaO))
   checkAll("instrument observing mode Optional",
            OptionalTests(instrumentObservingModeO))
+  checkAll("instrument disperser mode lambda Optional",
+           OptionalTests(instrumentDisperserLambdaO))
+  checkAll("instrument decker Optional",
+           OptionalTests(instrumentDeckerO))
+  checkAll("instrument imaging mirror Optional",
+           OptionalTests(instrumentImagingMirrorO))
+  checkAll("instrument mask Optional",
+           OptionalTests(instrumentMaskO))
 }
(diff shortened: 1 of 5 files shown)


modules/seqexec/model/src/test/scala/seqexec/model/ModelLensesSpec.scala
```diff
@@ -64,4 +64,12 @@ final class ModelLensesSpec extends CatsSuite with ModelLenses {
            OptionalTests(instrumentDisperserLambdaO))
   checkAll("instrument observing mode Optional",
            OptionalTests(instrumentObservingModeO))
+  checkAll("instrument disperser mode lambda Optional",
+           OptionalTests(instrumentDisperserLambdaO))
+  checkAll("instrument decker Optional",
+           OptionalTests(instrumentDeckerO))
```
sraaphorst: I guess decker and disperser mode were omitted previously?

cquiroz: Yes, I forgot to add tests to those lenses

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/ModelOps.scala
```diff
@@ -115,6 +120,14 @@ object ModelOps {
     }
   }
 
+  implicit class CamelCaseOps(val s: String) extends AnyVal {
+    def toCamelCase: String =
+      (s.toList match {
+        case Nil       => Nil
+        case x :: rest => x.toUpper :: rest.map(_.toLower)
+      }).mkString
```
sraaphorst: I'm not 100% clear how this works with camel case. Isn't it just capitalizing the first character and making everything else lowercase?

cquiroz: Right, maybe this shouldn't be called camelCase. What name would it be more appropriate?

jluhrs: I was going to say the same, but Sebastian beat me.

jluhrs: 'PhraseCase' ?

sraaphorst: I think it's generally called TitleCase.

cquiroz: i  looked around and settled on `Sentence case`. There are multiple versions but it seems `Title Case` will imply more than one capitalized letter and I only wanted the first one capitalized

PR #865, opened 2019-03-13
SEQNG-922A: Unit test for the observation state machine

cquiroz:
There is a report of NIRI observations not working after the last updates
This PR adds unit tests to reproduce the issue and tests for NIFS and GMOS. This was a fix can be done without having regressions

These tests require epics but for speed sake the epics layer has been mocked using ScalaMock

### DIFF
--- a/modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
+++ b/modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
@@ -53,48 +53,38 @@ final class ObserveStateSpec extends FunSuite with MockFactory {
     // VAL change
     observe.onApplyValChange(358)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
     // CAR CLID change
     observe.onCarClidChange(358)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
     // CAR VAL change
     observe.onCarValChange(CarState.BUSY)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
 
     // APPLY Goes to IDLE
     // Another VAL change
     observe.onApplyValChange(358)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
     // CAR CLID change
     observe.onCarClidChange(358)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
     // CAR VAL change
     observe.onCarValChange(CarState.IDLE)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
 
     // OBSERVE
     // OBSERVE goes BUSY
     // VAL change
     observe.onApplyValChange(359)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
     // Observe CAR VAL change
     observe.onObserveCarValChange(CarState.BUSY)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
     // CAR CLID change
     observe.onCarClidChange(359)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
     // CAR VAL change
     observe.onCarValChange(CarState.BUSY)
     assert(!observe.applyState().isIdle())
-    assert(!observe.observeState().isDone())
 
     // OBSERVE goes IDLE
     // Observe CAR VAL change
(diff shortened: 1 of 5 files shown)


cquiroz:
Yes, we should capture traces of each instrument in all the possible scenarios to reproduce them as tests

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+    observe.onCarClidChange(359)
+    assert(!observe.applyState().isIdle())
+    // CAR VAL change
+    observe.onCarValChange(CarState.BUSY)
+    assert(!observe.applyState().isIdle())
+
+    // OBSERVE goes IDLE
+    // Observe CAR VAL change
+    observe.onObserveCarValChange(CarState.IDLE)
+    // And we are done and IDLE
```
cquiroz: This is a bit odd, should we wait for apply to go IDLE before going back to idleState

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+    // CAR VAL change
+    observe.onCarValChange(CarState.BUSY)
+    assert(!observe.applyState().isIdle())
+    // Another VAL change
+    observe.onApplyValChange(4171)
+    assert(!observe.applyState().isIdle())
+    // Observe CAR VAL change
+    observe.onObserveCarValChange(CarState.PAUSED)
+    // We are now idle
+    assert(observe.applyState().isIdle())
```
cquiroz: @jluhrs Does this seem right? Should the apply go back to idle when paused?

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+    observe.onCarClidChange(4171)
+    assert(observe.applyState().isIdle())
+    // CAR VAL change
+    observe.onCarValChange(CarState.IDLE)
+    assert(observe.applyState().isIdle())
+
+    // RESUME OBSERVE
+    observe.onApplyValChange(4172)
+    // TODO Fix why this is idle here, it should be busy
+    assert(observe.applyState().isIdle())
```
cquiroz: @jluhrs I think this is wrong, we should go back to not idle when resumed, right?

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
jluhrs: Yes, that is correct.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
jluhrs: Observe and resume work almost the same. Resume will not succeed if the observation was not paused. The same post is used to trigger the resume, and from then it should follow the same sequence of events.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
jluhrs: Yes, you are right

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
cquiroz: Perhaps I need to send another post?

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+      "gm:abort",
+      "GMOS Observe",
+      classOf[CarState],
+      epicsService)
+    // Start idle
+    assert(observe.applyState().isIdle())
+    // Post an observe
+    observe.post()
+
+    // CONFIGURATION
```
jluhrs: Configuration is not triggered nor monitored by the `ObserveSenderImpl`, but by an instance of `ApplySenderImpl`. In the case of GMOS, they use the same apply and CAR records, which means the `ObserveSenderImpl` instance will see all the events, but it must stay idle.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+    observe.onApplyValChange(4166)
+    assert(!observe.applyState().isIdle())
+    // CAR CLID change
+    observe.onCarClidChange(4166)
+    assert(!observe.applyState().isIdle())
+    // CAR VAL change
+    observe.onCarValChange(CarState.IDLE)
+    assert(!observe.applyState().isIdle())
+
+    // OBSERVE
```
jluhrs: The observe must be triggered by a call to `post` in the `ObserveSenderImpl` instance.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+    observe.onCarValChange(CarState.IDLE)
+    assert(!observe.applyState().isIdle())
+
+    // OBSERVE goes IDLE
+    // Observe CAR VAL change
+    observe.onObserveCarValChange(CarState.IDLE)
+    // And we are done and IDLE
+    assert(observe.applyState().isIdle())
+
+    // ENDOBSERVE
```
jluhrs: endObserve is triggered through the `ApplySenderImpl` instance. This behavior is correct.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+      "gm:abort",
+      "GMOS Observe",
+      classOf[CarState],
+      epicsService)
+    // Start idle
+    assert(observe.applyState().isIdle())
+    // Post an observe
+    observe.post()
+
+    // CONFIGURATION
```
jluhrs: Again, configuration is triggered through the `ApplySenderImpl` instance, so the `post` in line 230 is wrong. The `ObserveSenderImpl` must remain idle.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,402 @@
...
+    assert(observe.applyState().isIdle())
+    // CAR CLID change
+    observe.onCarClidChange(4171)
+    assert(observe.applyState().isIdle())
+    // CAR VAL change
+    observe.onCarValChange(CarState.IDLE)
+    assert(observe.applyState().isIdle())
+
+    // RESUME OBSERVE
+    observe.onApplyValChange(4172)
```
jluhrs: There should be a call to `post` before this.

jluhrs:
You should also include checks for the `CaCommandMonitor` instance returned by post. Specially check that the `CaCommandListener` registered is called once and only once at the end.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -0,0 +1,389 @@
...
+    // CAR CLID change
+    observe.onCarClidChange(4172)
+    assert(!observe.applyState().isIdle())
+    // Apply goes IDLE
+    observe.onCarValChange(CarState.IDLE)
+    assert(!observe.applyState().isIdle())
+
+    // Observe CAR VAL change
+    observe.onObserveCarValChange(CarState.IDLE)
+
```
jluhrs: `assert(observe.applyState().isIdle())`

cquiroz: Updated

PR #866, opened 2019-03-14
SEQNG-944 Use guider dependent threshold for guiding while offsetting.

jluhrs:
Offsets can be applied without interrupting guiding if they are small enough. This PR implements that, and makes the decision threshold depend on the guider used.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentGuide.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentGuide.scala
@@ -1,0 +1,12 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server
+
+import seqexec.model.enum.Instrument
+import squants.space.Length
+
+trait InstrumentGuide {
+  def instrument: Instrument
+  def oiOffsetGuideThreshold: Option[Length]
+}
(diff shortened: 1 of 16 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -270,8 +272,36 @@ object TcsControllerEpics extends TcsController {
...
-  private def willMove(current: EpicsTcsConfig, demand: TcsConfig): Boolean =
-    demand.tc.offsetA.exists(_ =!= current.instrumentOffset)
+  private val pwfs1OffsetThreshold = Arcseconds(0.01)/FOCAL_PLANE_SCALE
+  private val pwfs2OffsetThreshold = Arcseconds(0.01)/FOCAL_PLANE_SCALE
+  private def aoOffsetThreshold(instrument: Instrument): Option[Length] = instrument match {
+    case Instrument.Nifs  => (Arcseconds(0.01)/FOCAL_PLANE_SCALE).some
+    case Instrument.Niri  => (Arcseconds(3.0)/FOCAL_PLANE_SCALE).some
+    case Instrument.Gnirs => (Arcseconds(3.0)/FOCAL_PLANE_SCALE).some
+    case _                => none
+  }
```
jluhrs: The old Seqexec used different thresholds for different instruments when guiding with Altair. I followed that, but I am not sure why it is like this.

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentGuide.scala
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server
+
+import seqexec.model.enum.Instrument
+import squants.space.Length
+
+trait InstrumentGuide {
```
jluhrs: This is just the information that TCS needs from the instrument

PR #869, opened 2019-03-18
GuideEnvironment (WIP)

swalker2m:
There is a lot of work left to be done but since I'm struggling with the best way to do this I wanted early feedback.  The challenge is assembling an observation with a consistent asterism and guide environment.  There is a lot here but really I'm focused on how to model this.  From the ground up we have:

* `GuideGroup` - this is basically a typed collection of pairs `(Guider, Target)`.  Usually it is just a single entry, for example GMOS North OI or PWFS2.  It can be more complex though if multiple guiders are involved (as for GSAOI/GeMS).  There are types for each valid guide group in order to prevent creating invalid groups like a Michelle observation with GMOS OI.

* `GuideOptions` - a parameterized type that combines multiple `GuideGroup` of the same broad instrument category into the selections for an observation.  For example, a `GuideOptions[GuideGroup.GmosNorth]` can contain GMOS OI or PWFS options that are valid for GMOS North. Basically this is an auto guide group (for AGS) and zero or more manual guide groups (because astronomers always want an override).

* `GuideEnvironment` - one per instrument where each contains an appropriately typed `GuideOptions`.  This is just `GuideOptions` with the type parameter having been eaten in order to create a `TargetEnvironment`.  For example, `GuideEnvironment.Flamingos2` which contains a `GuideOptions[GuideGroup.Flamingos2]`.

So this is a lot of complexity which makes me feel sheepish.  Do you guys have any suggestions for how to simplify it before I go much further?

### DIFF
--- a/modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
+++ b/modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
@@ -3,11 +3,69 @@
 
 package gem
 
-import gem.util.Zipper
+import cats._
+import cats.implicits._
 
-final case class GuideEnvironment[G](auto: Option[G], manual: Either[List[G], Zipper[G]]) {
+sealed trait GuideEnvironment extends Product with Serializable
 
-  def selected: Option[G] =
-    manual.fold(_ => auto, z => Some(z.focus))
+object GuideEnvironment {
+
+  final case class Flamingos2(options: GuideOptions[GuideGroup.Flamingos2]) extends GuideEnvironment
+  final case class GmosNorth(options: GuideOptions[GuideGroup.GmosNorth ])  extends GuideEnvironment
+  final case class GmosSouth(options: GuideOptions[GuideGroup.GmosSouth ])  extends GuideEnvironment
+
+  // placeholder until the other cases are defined
+  case object NoGuiding extends GuideEnvironment
+
+  object Flamingos2 {
+    val empty: Flamingos2 =
+      Flamingos2(GuideOptions.empty[GuideGroup.Flamingos2])
+
+    def typed(o: Option[GuideEnvironment]): Flamingos2 =
+      o match {
+        case Some(e @ Flamingos2(_)) => e
+        case _                       => empty
+      }
+
+    implicit val EqFlamingos2: Eq[Flamingos2] =
+      Eq.instance { (a, b) => a.options  === b.options }
+  }
+
+  object GmosNorth {
+    val empty: GmosNorth =
+      GmosNorth(GuideOptions.empty[GuideGroup.GmosNorth])
+
+    def typed(o: Option[GuideEnvironment]): GmosNorth =
+      o match {
+        case Some(e @ GmosNorth(_)) => e
+        case _                      => empty
+      }
+
+    implicit val EqGmosNorth: Eq[GmosNorth] =
+      Eq.instance { (a, b) => a.options  === b.options }
+  }
+
+  object GmosSouth {
+    val empty: GmosSouth =
+      GmosSouth(GuideOptions.empty[GuideGroup.GmosSouth])
+
+    def typed(o: Option[GuideEnvironment]): GmosSouth =
+      o match {
+        case Some(e @ GmosSouth(_)) => e
+        case _                      => empty
+      }
+
+    implicit val EqGmosSouth: Eq[GmosSouth] =
+      Eq.instance { (a, b) => a.options  === b.options }
+  }
+
+  implicit val EqGuideEnvironment: Eq[GuideEnvironment] =
+    Eq.instance {
+      case (a: Flamingos2, b: Flamingos2) => a === b
+      case (a: GmosNorth,  b: GmosNorth ) => a === b
+      case (a: GmosSouth,  b: GmosSouth ) => a === b
+      case (NoGuiding,     NoGuiding    ) => true
+      case _                              => false
+    }
 
 }
--- a/modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
+++ b/modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
@@ -1,0 +1,13 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+
+import gem.util.Zipper
+
+final case class GuideEnvironment[G](auto: Option[G], manual: Either[List[G], Zipper[G]]) {
+
+  def selected: Option[G] =
+    manual.fold(_ => auto, z => Some(z.focus))
+
+}
--- a/modules/core/shared/src/main/scala/gem/GuideGroup.scala
+++ b/modules/core/shared/src/main/scala/gem/GuideGroup.scala
@@ -5,15 +5,15 @@ package gem
 
 import cats._
 import cats.implicits._
-
+import cats.data.NonEmptyList
 import gem.enum.Guider
 
 /**
  * Associates one or more guide probes or guide windows with guide stars.
  */
-sealed trait GuideGroup {
+sealed trait GuideGroup extends Product with Serializable {
 
-  def toMap: Map[Guider, Target]
+  def guideStars: NonEmptyList[(Guider, Target)]
 
 }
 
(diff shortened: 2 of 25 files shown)


modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  object Flamingos2 {
+    val empty: Flamingos2 =
+      Flamingos2(GuideOptions.empty[GuideGroup.Flamingos2])
+
+    def typed(o: Option[GuideEnvironment]): Flamingos2 =
+      o match {
+        case Some(e @ Flamingos2(_)) => e
+        case _                       => empty
+      }
```
swalker2m: So this is problematic because it eats an error case where we need a `Flamingos2` environment but are given a `Some(GmosNorth(opts))` for example.  I could make this `unsafe` I suppose or it could return an `Option[Flamingos2]` or whatever but that pushes the problem into `TargetEnvironment`.

modules/core/shared/src/main/scala/gem/TargetEnvironment.scala
```diff
@@ -10,51 +10,60 @@ import scala.collection.immutable.SortedSet
...
+    a: Asterism,
+    g: Option[GuideEnvironment],
+    u: SortedSet[UserTarget]
+  ): TargetEnvironment =
     a match {
-      case a @ Asterism.Phoenix(_)            => Phoenix(Some(a), u)
-      case a @ Asterism.Michelle(_)           => Michelle(Some(a), u)
+      case a @ Asterism.AcqCam(_)             => AcqCam(Some(a), u)
+      case a @ Asterism.Bhros(_)              => Bhros(Some(a), u)
+      case a @ Asterism.Flamingos2(_)         => Flamingos2(Some(a), GuideEnvironment.Flamingos2.typed(g), u)
```
swalker2m: Here the `Option[GuideEnvironment]` we get could be anything but we can only assemble a valid `TargetEnvironment` if it is either `None` or a `Some(GuideEnvironment.Flamingos2(opts))`.

modules/db/src/main/scala/gem/dao/GuideEnvironmentDao.scala
```diff
@@ -0,0 +1,161 @@
...
+    for {
+      gs <- Statements.selectProg(pid).to[List]
+      ss <- GuideStarDao.selectProg(pid)
+    } yield TreeMap.grouping(gs)(_.obsIndex).mergeMatchingKeys(ss) { (obsGroups, obsStars) =>
+      obsStars.flatMap(guideEnvironment(obsGroups, _))
+    }
+
+
+  // Assemble all the groups and guide stars into an actual GuideEnvironment
+  private object guideEnvironment {
```
swalker2m: As you can see, this is essentially a big function but it is kind of hairy.

modules/core/shared/src/main/scala/gem/GuideGroup.scala
```diff
@@ -0,0 +1,123 @@
...
+    /** Ids ordered by wrapped integer value. */
+    implicit val IdOrder: Order[Id] =
+      Order.by(_.toInt)
+  }
+
+  /**
+   * Guide "group" supertype for the majority of cases where there is a single
+   * guide star.
+   */
+  sealed trait SingleGuider extends GuideGroup {
```
tpolecat: I kind of wonder if it would be worth it here to have `SingleGuider[G <: Guider]` and `def guider: G` and `Flamingos2.OI[Guider.F2OI.type]` so at least in the single-target case we would have that association statically. But I don't know if this will end up being useful or not.

modules/core/shared/src/main/scala/gem/GuideGroup.scala
```diff
@@ -0,0 +1,123 @@
...
+
+    def guider: Guider
+    def target: Target
+
+    override val guideStars: NonEmptyList[(Guider, Target)] =
+      NonEmptyList.one(guider -> target)
+
+  }
+
+  sealed trait Flamingos2 extends SingleGuider
```
tpolecat: You might try with `sealed abstract class Flamingos2(val guider: Guider) ...` and `final case class OI(target: Target) extends Flamingos2(Guider.F2OI)` and see if that looks any better.

modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
```diff
@@ -0,0 +1,71 @@
...
+package gem
+
+import cats._
+import cats.implicits._
+
+sealed trait GuideEnvironment extends Product with Serializable
+
+object GuideEnvironment {
+
+  final case class Flamingos2(options: GuideOptions[GuideGroup.Flamingos2]) extends GuideEnvironment
```
cquiroz: I wonder if we should use the same class name as for `Instrument`. Perhaps `FlamingosGE`?

cquiroz: Should `options` be specified on the `trait`?

modules/core/shared/src/main/scala/gem/GuideOptions.scala
```diff
@@ -0,0 +1,57 @@
...
+ * @tparam G guide group type
+ */
+final case class GuideOptions[G](
+  auto:   Option[G],
+  manual: Either[List[G], Zipper[G]]
+) {
+
+  def all: List[G] = {
+    def ms = manual.fold(identity, _.toList)
+    auto.fold(ms)(_ :: ms)
```
tpolecat: FYI you can say `auto.foldRight(ms)(_ :: _)` if you want to feel clever.

modules/core/shared/src/main/scala/gem/GuideGroup.scala
```diff
@@ -0,0 +1,123 @@
...
+import cats.implicits._
+import cats.data.NonEmptyList
+import gem.enum.Guider
+
+/**
+ * Associates one or more guide probes or guide windows with guide stars.
+ */
+sealed trait GuideGroup extends Product with Serializable {
+
+  def guideStars: NonEmptyList[(Guider, Target)]
```
tpolecat: Would it make sense for this to be `NonEmptyMap`? It's implemented with a `TreeMap` so we get a canonical ordering by `Guider`.

modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
tpolecat: Yeah this seems weird. Is it just used for reading from the database? Is it always an error to pass the wrong thing? I think I might prefer `def narrow(g: GuideEnvironment): Option[Flamingos2]` or `def narrow[F[_]: ApplicativeError[_[_], Throwable]: F[Flamingos2]` here, or `unsafe...` over in the db code somewhere.

swalker2m:
> I'm a little curious whether the complexity of multiple guide groups really pays for itself. How often do we have multiples, and what's the benefit?

I think having a manual override is non-negotiable but perhaps we could get away with just one manual option.  In other words, `GuideOptions` becomes essentially either the auto group or a single manual group. That would simplify things a bit.  We might have to throw away some manual options when importing from old programs.  I'll check on Slack.

modules/core/shared/src/main/scala/gem/GuideOptions.scala
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2016-2018 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gem
+
+import gem.util.Zipper
```
cquiroz: oh, how didn't I see this? When did we add a `Zipper`?

modules/core/shared/src/main/scala/gem/util/Zipper.scala
```diff
@@ -0,0 +1,85 @@
...
+
+package gem.util
+
+import cats._
+import cats.implicits._
+import cats.data.NonEmptyList
+
+/**
+ * A placeholder zipper for now. Stolen and modified from
+ * https://github.com/tpolecat/tuco/blob/master/modules/core/src/main/scala/tuco/util/zipper.scala
```
cquiroz: I have this one as well:
https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/web/shared/src/main/scala/seqexec/web/common/Zipper.scala

swalker2m: I mostly just wanted a placeholder to keep going, but I did see that.  It has a lot of good stuff in there. I guess we could move it over and add `previous` and `next` which are missing.  I think the `Applicative` instance I wrote could be added as well.  I was mostly frightened off by the filter thing at the end which I didn't understand. Does it make sense move it out to a `yolo.syntax` package or something?

swalker2m: On the other hand, if we don't need multiple manual options that I don't need a `Zipper` anyway.

cquiroz: I cannot claim the Zipper on the seqexec is totally well done nor performant and it has some functions specific for the purpose I need
We could consolidate into one taking the nastier parts apart
One thing I think is good is to have a `Traverse[Zipper]` instance. that should produce an `Applicative` right?

tpolecat: No, you need to define both `Traverse` and `Applicative`. And we should also define `Comonad`. And should contribute it to cats-collections.

modules/core/shared/src/main/scala/gem/GuideOptions.scala
swalker2m: Nice. 👍  That would enable a one-liner implementation:

`auto.fold(manual.map(_.toList).merge)(_ :: _)`

modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
swalker2m: I'd like to move it out into the db code but I need to do something for the `TargetEnvironment` constructors like `forAsterism`:

```scala
object TargetEnvironment {

  def fromAsterism(
    a: Asterism,
    g: Option[GuideEnvironment],
    u: SortedSet[UserTarget]
  ): TargetEnvironment =
...
```
Here the `Asterism` and `GuideEnvironment` instruments may not match and it isn't clear what to do in that case.  This is where I began to suspect that I was "doing it wrong". (Okay, actually I always start out with that suspicion.)

modules/core/shared/src/main/scala/gem/GuideEnvironment.scala
swalker2m: > Should `options` be specified on the trait?

The point of the `GuideEnvironment` is to eat the type parameter so it isn't clear to me what the type of `options` would be if I include it on the `trait`.  I suppose we could widen the individual `options` to `GuideOptions[GuideGroup]` though if it becomes useful?  Something like:

```scala
sealed trait GuideEnvironment extends Product with Serializable {
  def options: GuideOptions[GuideGroup]
}

object GuideEnvironment {
  final case class Flamingos2(flamingos2Options: GuideOptions[GuideGroup.Flamingos2]) extends GuideEnvironment {
    override def options: GuideOptions[GuideGroup] =
      flamingos2Options.widen[GuideGroup]
  ...
}
```

modules/core/shared/src/main/scala/gem/GuideOptions.scala
swalker2m: Hmm, I ended up needing an explicit conversion to `Iterator` to avoid picking up the cats `Foldable` version with `Eval`.  Short of fine grained import management, is there a better way to do this?

```scala
  def all: List[G] =
    auto.iterator.foldRight(manual.map(_.toList).merge)(_ :: _)
```

tpolecat: Ah, yeah. Stdlib has a conversion to `Iterable` and cats has a conversion to `Foldable.Ops` or something and that's the one you get since it's imported explicitly. At this point what you had first is probably easier to understand. :-\

cquiroz:
This probably be moved to `lucuma-core` if needed.

PR #870, opened 2019-03-18
Restore http Logging

cquiroz:
In one of the many refactorings we lost http logging on the server
This PR restores it and updates the http4s libraries


tpolecat:
How do you look at the Prometheus output? Where does it go?

cquiroz:
It goes to an http route, You can see it with curl
`curl -v http://localhost:7070/metrics`

The prometheus server reads that every 1m to store metrics

PR #874, opened 2019-03-20
SEQNG-928: Protect reading odb observations

cquiroz:
We've seen several times the session queue being empty. Some experiments demonstrate that if there is an exception when reading the odb sequences the stream that polls the odb dies. In that case the seqexec keeps running but there are no sequences

I want that if this happens (some error on the odb reading process) happens we log and kill the seqexec. While it may seem drastic it will ensure we fix any problems and note it right away

I tried to make the code exit gracefully but a `cats-effect` bug https://github.com/typelevel/cats-effect/issues/487 makes it harder, having to resort to `sys.exit`. When the bug is fixed this can be revisited

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
@@ -6,6 +6,7 @@ package seqexec.server
 import cats.Applicative
 import cats.Monad
 import cats.Endo
+import cats.data.Nested
 import cats.implicits._
 import cats.effect.{ Concurrent, IO, Timer }
 import edu.gemini.spModel.obscomp.InstConstants
@@ -58,21 +59,17 @@ final class ODBSequencesLoader[F[_]: Monad](odbProxy: OdbProxy[F], translator: S
           )(st)
       }.withEvent(LoadSequence(seqId)).toHandle)
 
-    t.map {
-      _.map {
-        case (err :: _, None) =>
-          List(Event.logDebugMsg(SeqexecFailure.explain(err)))
-        case (errs, Some(seq)) =>
-          loadSequenceEvent(seq) :: errs.map(e =>
-            Event.logDebugMsg(SeqexecFailure.explain(e)))
-        case _ => Nil
-      }
-      .valueOr(e => List(Event.logDebugMsg(SeqexecFailure.explain(e))))
-    }
+    Nested(t).map {
+      case (err :: _, None) =>
+        List(Event.logDebugMsg(SeqexecFailure.explain(err)))
+      case (errs, Some(seq)) =>
+        loadSequenceEvent(seq) :: errs.map(e =>
+          Event.logDebugMsg(SeqexecFailure.explain(e)))
+      case _ => Nil
+    }.value.map(_.valueOr(e => List(Event.logDebugMsg(SeqexecFailure.explain(e)))))
   }
 
-  def refreshSequenceList(odbList: List[Observation.Id])(
-    st: EngineState)(
+  def refreshSequenceList(odbList: List[Observation.Id], st: EngineState)(
       implicit cio: Concurrent[IO],
                tio: Timer[IO]
     ): F[List[executeEngine.EventType]] = {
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
```diff
@@ -31,20 +33,22 @@ final class ODBSequencesLoader[F[_]: Applicative](odbProxy: OdbProxy[F], transla
...
-                          .config
-                          .extractAs[String](OCS_KEY / InstConstants.PROGRAMID_PROP)
-                          .leftMap(ConfigUtilOps.explainExtractError)
-        _            <- Either.catchNonFatal(
-                          Applicative[F].pure(SPProgramID.toProgramID(progIdString)))
-                        .leftMap(e => SeqexecFailure.SeqexecException(e): SeqexecFailure)
-      } yield translator.sequence(seqId, odbSeq)
+    implicit cio: Concurrent[IO],
+             tio: Timer[IO]
+  ): F[List[executeEngine.EventType]] = {
```
cquiroz: This should be inside `F` as we connect to the odb

modules/seqexec/server/src/main/scala/seqexec/server/OdbProxy.scala
```diff
@@ -13,13 +13,15 @@ import edu.gemini.wdba.session.client.WDBA_XmlRpc_SessionClient
 import gem.Observation
 import seqexec.model.dhs.ImageFileId
 
-class OdbProxy[F[_]](val loc: Peer, cmds: OdbProxy.OdbCommands[F]) {
+class OdbProxy[F[_]: Sync](val loc: Peer, cmds: OdbProxy.OdbCommands[F]) {
 
   def host(): Peer = loc
-  def read(oid: Observation.Id): Either[SeqexecFailure, SeqexecSequence] =
-    SeqExecService.client(loc).sequence(new SPObservationID(oid.format)).leftMap(SeqexecFailure.OdbSeqError)
+  def read(oid: Observation.Id): F[Either[SeqexecFailure, SeqexecSequence]] =
```
cquiroz: Read should return an `F[_]` as it is a side effect

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -203,26 +206,28 @@ class SeqexecEngine(httpClient: Client[IO], gpi: GpiClient[IO], ghost: GhostClie
...
+        Stream.emit(SeqexecFailure.OdbSeqError(e).asLeft)
+      case e: Exception =>
+        Stream.emit(SeqexecFailure.SeqexecException(e).asLeft)
     }
   }
 
-  def eventStream(q: EventQueue): Stream[IO, SeqexecEvent] = {
-    stream(q.dequeue.mergeHaltBoth(seqQueueRefreshStream))(EngineState.default).flatMap(x =>
+  def eventStream(q: EventQueue): Stream[IO, SeqexecEvent] =
+    stream(q.dequeue.mergeHaltBoth(seqQueueRefreshStream.rethrow))(EngineState.default).flatMap(x =>
```
cquiroz: the `rethrow` will wait for a left and exit stopping the queue as well

modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -236,6 +237,14 @@ object WebServerLauncher extends IOApp with LogInitialization with SeqexecConfig
     asyncAppender
   }
 
+  // Logger of error of last resort.
+  // Once https://github.com/typelevel/cats-effect/issues/487 is fixed we
+  // could remove the sys.exit calls
+  def logError: PartialFunction[Throwable, IO[Unit]] = {
+    case e: SeqexecFailure => IO(logger.error(e)(s"Seqexec global error handler ${SeqexecFailure.explain(e)}")) *> IO(sys.exit(-1))
```
cquiroz: Eventually the `sys.exit` will be removed

PR #875, opened 2019-03-22
SEQNG-893: Altair keywords

cquiroz:
More or less a direct translation of the tcl code. Needs testing


build.sbt
```diff
@@ -78,6 +78,7 @@ addCommandAlias("startSeqexecAll", startSeqexecAllCommands.mkString(";", ";", ""
 addCommandAlias("restartSeqexecWDS", restartSeqexecWDSCommands.mkString(";", ";", ""))
 addCommandAlias("stopSeqexecAll", stopSeqexecAllCommands.mkString(";", ";", ""))
 addCommandAlias("genEnums", "; sql/runMain gem.sql.Main modules/core/shared/src/main/scala/gem/enum")
+addCommandAlias("rebuildEnums", "; sql/flywayClean; sql/flywayMigrate; genEnums; coreJVM/compile")
```
sraaphorst: This isn't unique for Altair, is it?

modules/core/shared/src/main/scala/gem/enum/KeywordName.scala
```diff
@@ -727,10 +727,23 @@ object KeywordName {
   /** @group Constructors */ case object MJD_OBS extends KeywordName("MJD_OBS", "MJD_OBS")
   /** @group Constructors */ case object APERTURE extends KeywordName("APERTURE", "APERTURE")
   /** @group Constructors */ case object FILTER extends KeywordName("FILTER", "FILTER")
+  /** @group Constructors */ case object AOFREQ extends KeywordName("AOFREQ", "AOFREQ")
```
sraaphorst: This just seems to me like a bad idea, with all the repeated strings asking for trouble. Couldn't we have an intermediate class like:

```
sealed abstract class KeywordTagIsName(tag: String) extends KeywordName(tag, tag)
```

and use that when the `tag` and the `name` are the same?

build.sbt
cquiroz: It is unrelated, I just do that command very often so I wanted to have it available on sbt

modules/core/shared/src/main/scala/gem/enum/KeywordName.scala
cquiroz: `KeywordName` is generated code. I suppose we could do as you suggest and have the few cases where the `tag` and `name` are different but it would complicate a lot the code generation part

modules/seqexec/server/src/main/resources/Altair.xml
```diff
@@ -92,6 +92,71 @@
             <!--<type>STRING</type>-->
             <!--<description>Sfo loop</description>-->
         <!--</attribute>-->
+        <attribute name="aoexpt">
+            <channel>wfcs:getStatus.VALK</channel>
+            <type>DOUBLE</type>
```
jluhrs: EPICS type of this one is DBR_FLOAT

modules/seqexec/server/src/main/resources/Altair.xml
```diff
@@ -92,6 +92,71 @@
...
+            <description>Altair one over sampling frequency during exposure (Hz)</description>
+        </attribute>
+        <attribute name="aocounts">
+            <channel>wfcs:getStatus4.VALA</channel>
+            <type>DOUBLE</type>
+            <description>Altair GS counts at end of exposure (ADU/subaperture)</description>
+        </attribute>
+        <attribute name="aoseeing">
+            <channel>wfcs:getStatus.VALD</channel>
+            <type>DOUBLE</type>
```
jluhrs: FLOAT

modules/seqexec/server/src/main/resources/Altair.xml
```diff
@@ -92,6 +92,71 @@
...
+            <type>STRING</type>
+            <description>Altair flexure file</description>
+        </attribute>
+        <attribute name="lgustage">
+            <channel>cc:lgsustAssembly.VALA</channel>
+            <type>STRING</type>
+            <description>LGS U stage position</description>
+        </attribute>
+        <attribute name="aobs">
+            <channel>ao:bs.VAL</channel>
```
jluhrs: ~~ao:~~bs.VAL

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairEpics.scala
```diff
@@ -80,6 +80,32 @@ class AltairEpics[F[_]: Async](service: CaService, tops: Map[String, String]) {
     s"${AltairTop}cc:lgszoomSfoLoop.VAL", classOf[LgsSfoControl])
   def sfoLoop: F[Option[LgsSfoControl]] = safeAttribute(sfoLoopAttr)
 
+  def aoexpt: F[Option[Double]] = safeAttributeSDouble(status.getDoubleAttribute("aoexpt"))
```
jluhrs: I think this will fail, because the channel type is Float.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairEpics.scala
```diff
@@ -80,6 +80,32 @@ class AltairEpics[F[_]: Async](service: CaService, tops: Map[String, String]) {
...
+
+  def aocounts: F[Option[Double]] = safeAttributeSDouble(status.getDoubleAttribute("aocounts"))
+
+  def aoseeing: F[Option[Double]] = safeAttributeSDouble(status.getDoubleAttribute("aoseeing"))
```
jluhrs: Same here, `getDoubleAttribute` will return `null` because the channel was created as Float.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairKeywordReader.scala
```diff
@@ -0,0 +1,95 @@
...
+    case AOCRFollow.Fixed     => "no"
+  }
+}
+
+class AltairKeywordReaderImpl[F[_]: Sync: LiftIO](config: Config)
+    extends AltairKeywordReader[F]
+    with AltairKeywordReaderLUT {
+  val sys = AltairEpics.instance
+
+  override def aoexpt: F[Double] =
```
jluhrs: This should be called `aofreq`, as it is calculated as 1/aoexpt (ao exposure time)

PR #880, opened 2019-03-27
SEQNG-883 Fixed issue with resuming queued sequence, added tests.

jluhrs:
The issue is fixed, but I think that the queue logic is getting a bit too complicated, and I can think of some cases where it may not work as expected.


modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -105,22 +107,24 @@ class SeqexecEngine(httpClient: Client[IO], gpi: GpiClient[IO], ghost: GhostClie
...
     // Resources used by running sequences
     val used = resourcesInUse(st)
 
     // Resources that will be used by sequences in running queues
-    val reservedByQueues = resourcesReserved(failedInstruments(st), st)
-
-    st.sequences.get(seqId).exists(_.seqGen.resources.intersect(used ++ reservedByQueues).isEmpty)
+    val reservedByQueues = resourcesReserved(st)
 
+    st.sequences.get(seqId).exists(x =>
```
cquiroz: I guess this is the core change right?. looks good

jluhrs: Yes, I check if the sequence is in the queue, in which case the next one, the check against the resources reserved by the running queue, does not apply.

PR #881, opened 2019-03-28
SEQNG-951 Stop and resume Altair guiding in LGS mode with P1 or OI

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
```diff
@@ -119,15 +119,15 @@ object AltairControllerEpics extends AltairController[IO] {
     reasons: Set[Gaos.ResumeCondition]): Option[IO[Unit]] = {
     val offsets = reasons.collectFirst { case OffsetReached(o) => o }
     val newPosOk = offsets.forall(v => newPosInRange(newPosition(starPos)(v)))
-    val guideOk = reasons.contains(GaosStarOn)
+    val guideOk = reasons.contains(GaosGuideOn)
```
jluhrs: I renamed because in LGS+OI and LGS+P1 there is no guide star, so the old name didn't make sense.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -46,29 +46,27 @@ object Gaos {
   case object OiOn extends ResumeCondition
   // PI will be turn off
   case object P1On extends ResumeCondition
-  // GAOS star guide will be turn off
-  case object GaosStarOn extends ResumeCondition
+  // Guided step
+  case object GaosGuideOn extends ResumeCondition
```
jluhrs: I renamed it because in LGS+OI and LGS+P1 there is no guide star, so the old name didn't make sense.

PR #882, opened 2019-03-28
SEQNG-952 Consistently sort the subsystems

cquiroz:
This is a small bug fix. The subsystem buttons should be sorted consistently

Before:
Seqexec - GS-2018B-Q-0-71 2019-03-28 14-55-24

After:
image


jluhrs:
What is the sort criteria ?

cquiroz:
It uses Resource's index, see
https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/model/src/main/scala/seqexec/model/enum/Resource.scala#L16

Note that this was the criteria used when displaying subsystems being configured but it wasn't used for the subsystems control buttons

PR #887, opened 2019-04-01
SEQNG-964: GSAOI gem enums

cquiroz:
In prepartion to support GSAOI in the seqexec I think we could use adding enums into the model


modules/core/shared/src/main/scala/gem/enum/GsaoiFilter.scala
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Enumerated type for GSAOI Filter.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GsaoiFilter(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val wavelength: Double,
```
swalker2m: I think this should be mapped to a `Wavelength` by the enum definition code in `gem.sql.enum`?

modules/core/shared/src/main/scala/gem/enum/GsaoiFilter.scala
```diff
@@ -0,0 +1,76 @@
...
+  val readMode: GsaoiReadMode,
+  val expsoureTime5050: Double,
```
swalker2m: Same here, but for `Duration`.  At the moment we're using Java `Duration` so I suppose it should be that but we'll probably move to a Scala `FiniteDuration` at some point.

modules/core/shared/src/main/scala/gem/enum/GsaoiReadMode.scala
```diff
@@ -0,0 +1,54 @@
...
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GsaoiReadMode(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val ndr: Int,
+  val readNoise: Int,
+  val minimumExposureTime: Duration,
+  val overhead: Int
```
swalker2m: This is also a `Duration`.

modules/sql/src/main/resources/db/migration/V066__Add_Gsaoi.sql
```diff
@@ -0,0 +1,128 @@
...
+-- Name: e_gsaoi_read_mode; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE e_gsaoi_read_mode (
+    id                        identifier            PRIMARY KEY,
+    short_name                character varying(8)  NOT NULL,
+    long_name                 character varying(40) NOT NULL,
+    ndr                       smallint              NOT NULL CHECK (ndr >= 0),
+    read_noise                smallint              NOT NULL CHECK (read_noise >= 0),
+    minimum_exposure_time     double precision      NOT NULL CHECK (minimum_exposure_time >= 0),
```
swalker2m: This should probably be a `numeric(3,1)`?

modules/sql/src/main/scala/gem/sql/enum/GsaoiEnums.scala
```diff
@@ -0,0 +1,45 @@
...
+        type E = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'ndr -> Int, 'readNoise -> Int, 'minimumExposureTime -> Duration, 'overhead -> Int`.T
+        sql"""SELECT id, id tag, short_name, long_name, ndr, read_noise, minimum_exposure_time, overhead FROM e_gsaoi_read_mode""".query[(String, E)]
+      },
+
+      EnumDef.fromQuery("GsaoiFilter", "GSAOI Filter") {
+        val  m = Witness('MagnitudeBand)
+        val  r = Witness('GsaoiReadMode)
+        type M = m.T
+        type R = r.T
+        type E = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'wavelength -> Double, 'readMode -> EnumRef[R], 'expsoureTime5050 -> Double, 'exposureTimeHalfWell -> Double, 'band -> Option[EnumRef[M]]`.T
```
swalker2m: This is where you'd map the wavelength to a `Wavelength.Pm`, `Wavelength.Nm`, or `Wavelength.Um` depending on the units.

modules/core/shared/src/main/scala/gem/enum/GsaoiFilter.scala
cquiroz: Right, that is what I expected ❓

modules/core/shared/src/main/scala/gem/enum/GsaoiReadMode.scala
cquiroz: Why is this?

cquiroz: I mean why `overhead` is a `Duration`?

swalker2m: I believe it is a time in seconds, so we should try to represent it with a more precise type.

cquiroz:
I've updated the code as suggested and added conversions for `FiniteDuration`. We may want to migrate the java `Duration`s in another PR

modules/core/shared/src/main/scala/gem/enum/GsaoiOdgwSize.scala
```diff
@@ -0,0 +1,51 @@
...
+import cats.syntax.eq._
+import gem.util.Enumerated
+
+/**
+ * Enumerated type for Gsaoi ODGW Size.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GsaoiOdgwSize(
+  val tag: String,
+  val size: Int
```
swalker2m: I suppose that is size as in pixels?

modules/sql/src/main/scala/gem/sql/FiniteDuration.scala
```diff
@@ -0,0 +1,23 @@
...
+  def toMillis: Long
+
+}
+
+object FiniteDuration {
+  final case class Seconds(toMillis: Long) extends FiniteDuration
+  final case class Milliseconds(toMillis: Long) extends FiniteDuration
+
+  def fromSeconds(bd: Long): FiniteDuration.Seconds =
+    Seconds(bd)
```
swalker2m: Here I think we should switch to a `BigDecimal` and do the conversion in the trait as above but as is this would need to multiply `bd` by `1000` to get milliseconds.

```scala
object FiniteDuration {
  final case class Seconds(toBigDecimal: BigDecimal) extends FiniteDuration
  final case class Milliseconds(toBigDecimal: BigDecimal) extends FiniteDuration

  from Seconds(bd: BigDecimal): FiniteDuration.Seconds =
    Seconds(bd)

  from Milliseconds(bd: BigDecimal): FiniteDuration.Milliseconds =
    Microseconds(bd)
}
```

modules/core/shared/src/main/scala/gem/enum/GsaoiFilter.scala
```diff
@@ -0,0 +1,78 @@
...
+  val readMode: GsaoiReadMode,
+  val exposureTime5050: FiniteDuration,
+  val exposureTimeHalfWell: FiniteDuration,
+  val band: Option[MagnitudeBand]
+) extends Product with Serializable
+
+object GsaoiFilter {
+
+  /** @group Constructors */ case object Z extends GsaoiFilter("Z", "Z", "Z (1.015 um)", Wavelength.fromPicometers.unsafeGet(1020000), GsaoiReadMode.Faint, new FiniteDuration(26, java.util.concurrent.TimeUnit.SECONDS), new FiniteDuration(4619, java.util.concurrent.TimeUnit.SECONDS), Some(MagnitudeBand.J))
+  /** @group Constructors */ case object HeI extends GsaoiFilter("HeI", "HeI", "HeI (1.083 um)", Wavelength.fromPicometers.unsafeGet(1080000), GsaoiReadMode.VeryFaint, new FiniteDuration(72, java.util.concurrent.TimeUnit.SECONDS), new FiniteDuration(21792, java.util.concurrent.TimeUnit.SECONDS), Some(MagnitudeBand.J))
```
swalker2m: In the old codebase, the `exposureTime5050` is `72.6` seconds.  Did this change or should it be 72600 milliseconds here?  The same question applies for the other subsecond resolution times.

modules/core/shared/src/main/scala/gem/enum/GsaoiReadMode.scala
```diff
@@ -0,0 +1,54 @@
...
+  val longName: String,
+  val ndr: Int,
+  val readNoise: Int,
+  val minimumExposureTime: FiniteDuration,
+  val overhead: FiniteDuration
+) extends Product with Serializable
+
+object GsaoiReadMode {
+
+  /** @group Constructors */ case object Bright extends GsaoiReadMode("Bright", "Bright", "Bright Objects", 2, 28, new FiniteDuration(5, java.util.concurrent.TimeUnit.SECONDS), new FiniteDuration(10, java.util.concurrent.TimeUnit.SECONDS))
```
swalker2m: Here again the `minExposureTime` before was `5.3` so you'd want `FiniteDuration(5300, MILLISECONDS)`. It's fine that they appear as milliseconds here in the `case object`, we just want the fractional second value in the database so that it makes sense to the user doing a query.

modules/sql/src/main/resources/db/migration/V066__Add_Gsaoi.sql
```diff
@@ -0,0 +1,151 @@
...
+);
+
+ALTER TABLE e_gsaoi_read_mode OWNER TO postgres;
+
+--
+-- Data for Name: e_gsaoi_read_mode; Type: TABLE DATA; Schema: public; Owner: postgres
+--
+
+COPY e_gsaoi_read_mode(id, short_name, long_name, ndr, read_noise, minimum_exposure_time, overhead) FROM stdin;
+Bright	Bright	Bright Objects	2	28	5.3	10
```
swalker2m: Yeah it looks like the 300 ms are being rounded down in the conversion to `FiniteDuration`.

modules/sql/src/main/scala/gem/sql/EnumDef.scala
```diff
@@ -105,6 +110,9 @@ object EnumDef {
     implicit val caseWavelengthUm    = at[Wavelength.Um  ](a => s"""Wavelength.fromPicometers.unsafeGet(${a.toPicometers})""")
     implicit val caseMagnitudeSystem = at[MagnitudeSystem](a => s"MagnitudeSystem.${a.id}")
 
+    implicit val caseFiniteDurationSeconds      = at[FiniteDuration.Seconds     ](a => s"""new FiniteDuration(${a.toMillis}, java.util.concurrent.TimeUnit.SECONDS)""")
```
swalker2m: Here the unit should be `MILLISECONDS` as well to match the value?

modules/sql/src/main/scala/gem/sql/FiniteDuration.scala
```diff
@@ -0,0 +1,23 @@
...
+  * tables.
+  */
+sealed trait FiniteDuration extends Product with Serializable {
+  def toMillis: Long
+
+}
+
+object FiniteDuration {
+  final case class Seconds(toMillis: Long) extends FiniteDuration
+  final case class Milliseconds(toMillis: Long) extends FiniteDuration
```
swalker2m: Following the `Wavelength` example, it seems like these should be `BigDecimal` fields to hold the `numeric` values.  Then you could add a `toMilliseconds` in the `trait` that does the conversion.

```scala
sealed trait FiniteDuration extends Product with Serializable {
  def toBigDecimal: BigDecimal

  def toMilliseconds: Long = {
    val n = this match {
      case FiniteDuration.Milliseconds(_) => 0
      case FiniteDuration.Seconds(_)      => 3
    }
    toBigDecimal.underlying.scaleByPowerOfTen(n).setScale(0, HALF_UP).longValue
  }
}
...
```

modules/core/shared/src/main/scala/gem/enum/GsaoiOdgwSize.scala
cquiroz: The old code base doesn't say anything about it nor the UI as far as I can tell

modules/core/shared/src/main/scala/gem/enum/GsaoiFilter.scala
cquiroz: Good question. I need to revisit why the fraction is being lost. Am I right that on the db we want to store `72.6` rather than `72600`?

swalker2m: Right, I think ideally everything will appear in units that the astronomer is expecting.

modules/core/shared/src/main/scala/gem/enum/GsaoiOdgwSize.scala
sraaphorst: This seems like thee kind of thing we should be documenting now, so that we avoid this issue resurfacing.

cquiroz: This is generated code so we can't document it, but I've renamed it to `pixels` and renamed the table too

Issue #888, opened 2019-04-02
Time value updates


swalker2m:
Many of the instrument parameter enums represent time as a Java `Duration` but we've decided that a Scala `FiniteDuration` is more in line with most APIs so they should be switched.  Also, we represent time values in ms in the database in places where they are naturally fractional seconds so that should be switched.

@cquiroz's #887 adds conversion code for milliseconds so we can use that now.

PR #889, opened 2019-04-02
F2Enums to use FiniteDuration

cquiroz:
F2 ReadMode can use `FiniteDuration` rather than `Duration`. See #888


modules/sql/src/main/scala/gem/sql/enum/F2Enums.scala
```diff
@@ -34,7 +34,7 @@ object F2Enums {
       },
 
       EnumDef.fromQuery("F2ReadMode", "Flamingos2 read modes") {
-        type R = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'description -> String, 'minimumExposureTime -> Duration, 'recommendedExposureTime -> Duration, 'readoutTime -> Duration, 'readCount -> Int, 'readNoise -> Double`.T
+        type R = Record.`'tag -> String, 'shortName -> String, 'longName -> String, 'description -> String, 'minimumExposureTime -> FiniteDuration.Milliseconds, 'recommendedExposureTime -> FiniteDuration.Milliseconds, 'readoutTime -> FiniteDuration.Milliseconds, 'readCount -> Int, 'readNoise -> Double`.T
```
swalker2m: We should add the database migration and store these as seconds in a `numeric` while we're at it.

modules/core/shared/src/main/scala/gem/enum/F2ReadMode.scala
```diff
@@ -18,18 +18,18 @@ sealed abstract class F2ReadMode(
...
   val readCount: Int,
   val readNoise: Double
 ) extends Product with Serializable
 
 object F2ReadMode {
 
-  /** @group Constructors */ case object Bright extends F2ReadMode("Bright", "bright", "Bright Object", "Strong Source", Duration.ofMillis(1500), Duration.ofMillis(5000), Duration.ofMillis(8000), 1, 11.7)
-  /** @group Constructors */ case object Medium extends F2ReadMode("Medium", "medium", "Medium Object", "Medium Source", Duration.ofMillis(6000), Duration.ofMillis(21000), Duration.ofMillis(14000), 4, 6.0)
-  /** @group Constructors */ case object Faint extends F2ReadMode("Faint", "faint", "Faint Object", "Weak Source", Duration.ofMillis(12000), Duration.ofMillis(85000), Duration.ofMillis(20000), 8, 5.0)
+  /** @group Constructors */ case object Bright extends F2ReadMode("Bright", "bright", "Bright Object", "Strong Source", new FiniteDuration(1500, java.util.concurrent.TimeUnit.MILLISECONDS), new FiniteDuration(5000, java.util.concurrent.TimeUnit.MILLISECONDS), new FiniteDuration(8000, java.util.concurrent.TimeUnit.MILLISECONDS), 1, 11.7)
```
swalker2m: In `EnumDef` maybe we can change the import corresponding to `FiniteDuration` into `Some("scala.concurrent.duration._")` and then we can write these as `1500 millis` instead of `new FiniteDuration(1500, java.util.concurrent.TimeUnit.MILLISECONDS)`.  I think it would clean up the enumerated code quite a bit.

modules/core/shared/src/main/scala/gem/enum/F2ReadMode.scala
```diff
@@ -18,18 +18,18 @@ sealed abstract class F2ReadMode(
...
 ) extends Product with Serializable
 
 object F2ReadMode {
 
-  /** @group Constructors */ case object Bright extends F2ReadMode("Bright", "bright", "Bright Object", "Strong Source", Duration.ofMillis(1500), Duration.ofMillis(5000), Duration.ofMillis(8000), 1, 11.7)
-  /** @group Constructors */ case object Medium extends F2ReadMode("Medium", "medium", "Medium Object", "Medium Source", Duration.ofMillis(6000), Duration.ofMillis(21000), Duration.ofMillis(14000), 4, 6.0)
-  /** @group Constructors */ case object Faint extends F2ReadMode("Faint", "faint", "Faint Object", "Weak Source", Duration.ofMillis(12000), Duration.ofMillis(85000), Duration.ofMillis(20000), 8, 5.0)
+  /** @group Constructors */ case object Bright extends F2ReadMode("Bright", "bright", "Bright Object", "Strong Source", 1500.millis, 5000.millis, 8000.millis, 1, 11.7)
+  /** @group Constructors */ case object Medium extends F2ReadMode("Medium", "medium", "Medium Object", "Medium Source", 6000.millis, 21000.millis, 14000.millis, 4, 6.0)
+  /** @group Constructors */ case object Faint extends F2ReadMode("Faint", "faint", "Faint Object", "Weak Source", 12000.millis, 85000.millis, 20000.millis, 8, 5.0)
```
swalker2m: That's easier to read. 👍

PR #893, opened 2019-04-05
SEQNG-124 Sequences can be started from arbitrary step.

jluhrs:
This PR modifies the back end to provide the ability to start a sequence from an arbitrary step. This is a feature already present in the old Tcl Seqexec. The implementation uses existing features. It marks all steps preceding the selected one as to be skipped (excluding already completed steps), and then runs the sequence.

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -49,6 +49,22 @@ class Engine[D, U](stateL: Engine.State[D]) {
       case None      => unit
     }
 
+  /*
+   * startFrom starts a sequence from an arbitrary step. It does it by marking all previous steps to be skipped and then
+   * modifying the state sequence as if it was run.
+   * If the requested step is already run or marked to be skipped, the sequence will start from the next runnable step
+   */
+  def startFrom(id: Observation.Id, step: StepId): HandleType[Unit] =
+    getS(id).flatMap {
+      case Some(seq) if (seq.status.isIdle || seq.status.isError) && seq.toSequence.steps.exists(_.id === step) =>
+        val steps = seq.toSequence.steps.takeWhile(_.id =!= step).mapFilter(p => Step.status(p).canRunFrom.option(p.id))
+        val withSkips = steps.foldLeft[Sequence.State[IO]](seq){ case (s, i) => s.setSkipMark(i, true) }
+        putS(id)(
+          Sequence.State.status.set(SequenceState.Running.init)(withSkips.skips.getOrElse(withSkips).rollback)
+        ) *> send(Event.executing(id))
+      case _ => unit
+    }
+
   def pause(id: Observation.Id): HandleType[Unit] = modifyS(id)(Sequence.State.userStopSet(true))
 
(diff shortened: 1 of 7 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -49,6 +49,22 @@ class Engine[D, U](stateL: Engine.State[D]) {
       case None      => unit
     }
 
+  /*
+   * startFrom starts a sequence from an arbitrary step. It does it by marking all previous steps to be skipped and then
+   * modifying the state sequence as if it was run.
+   * If the requested step is already run or marked to be skipped, the sequence will start from the next runnable step
+   */
+  def startFrom(id: Observation.Id, step: StepId): HandleType[Unit] =
```
jluhrs: Another option would have been to implement it as a sequence of events, like MarkSkip(sid, 1), MarkSkip(sid, 2),..,Start(sid), but there is the risk that another event is processed in between. As it is, everything is resolved in one iteration of the engine.

cquiroz: Good call

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -49,6 +49,22 @@ class Engine[D, U](stateL: Engine.State[D]) {
...
+    getS(id).flatMap {
+      case Some(seq) if (seq.status.isIdle || seq.status.isError) && seq.toSequence.steps.exists(_.id === step) =>
+        val steps = seq.toSequence.steps.takeWhile(_.id =!= step).mapFilter(p => Step.status(p).canRunFrom.option(p.id))
```
cquiroz: Should this be `_.id < step`?

jluhrs: Only if you assume that the ids are ordered. Which they are, but it is not necessary to depend on that fact, because the code use the order in the steps list.

cquiroz: 👍

swalker2m:
Would you write a description of what this PR about?  Is the ODB/WDBA compatible with these changes?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -25,7 +25,7 @@ import seqexec.engine.{Step => _, _}
 import seqexec.engine.Handle
 import seqexec.model._
 import seqexec.model.enum._
-import seqexec.model.events._
+import seqexec.model.events.{SequenceStart => EvSequenceStart, _}
```
cquiroz: Maybe this would be clearer if called `ClientSequentStart`?

jluhrs:
@swalker2m This feature is already present in the current Tcl Seqexec, so I would say that the ODB/WDBS already handles it. What I have seen in the ODB is that all preceding steps are marked as skipped.

swalker2m:
> What I have seen in the ODB is that all preceding steps are marked as skipped.
👍

PR #895, opened 2019-04-08
SEQNG-865 Fixed bug in reading the current science fold position

jluhrs:
Most of the work was already done in SEQNG-892. Before sending an AG configuration, the current science fold position and HR pickup mirror position are compared with the requested ones. If they are the same, no command is sent.
This PR fixes two things. The first one is that there was a bug in the code that decoded the current science fold position (it was failing if the instrument part of the position name had numbers in it)
The second fix is that the HR positions OUT and parked are now considered equivalent. The HR will not be moved if it is in the OUT position and the request is for park (or the other way around)

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
@@ -21,11 +21,7 @@ private[server] object ScienceFoldPositionCodex {
   private val GCAL_PREFIX = "gcal2"
   private val PARK_POS = "park-pos"
 
-  val lightSink: Parser[LightSinkName] =
-    many(letter).map { cs =>
-      val sinkName = cs.mkString
-      LightSinkName.all.find(_.name === sinkName)
-    }.collect { case Some(lsn) => lsn }
+  val lightSink: Parser[LightSinkName] = LightSinkName.all.foldMap(x => string(x.name).as(x))
 
   def prefixed(p: String, s: LightSource): Parser[ScienceFold] =
     (string(p) ~> lightSink ~ int).map { case (ls, port) => Position(s, ls, port) }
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/ScienceFoldPositionCodex.scala
```diff
@@ -21,11 +21,7 @@ private[server] object ScienceFoldPositionCodex {
   private val GCAL_PREFIX = "gcal2"
   private val PARK_POS = "park-pos"
 
-  val lightSink: Parser[LightSinkName] =
-    many(letter).map { cs =>
-      val sinkName = cs.mkString
-      LightSinkName.all.find(_.name === sinkName)
-    }.collect { case Some(lsn) => lsn }
+  val lightSink: Parser[LightSinkName] = LightSinkName.all.foldMap(x => string(x.name).as(x))
```
jluhrs: Thanks @tpolecat !

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsControllerEpics.scala
```diff
@@ -268,6 +268,22 @@ object TcsControllerEpics extends TcsController {
     }
   }
 
+  /*
+   * Positions Parked and OUT are equivalent for practical purposes. Therefore, if the current position is Parked and
+   * requested position is OUT (or the other way around), then is not necessary to move the HR pickup mirror.
+   */
+  private def setHrPickup(subsystems: NonEmptySet[Subsystem], c: AGConfig, current: EpicsTcsConfig)
+  : Option[EpicsTcsConfig => SeqAction[EpicsTcsConfig]] =
+    subsystems.contains(Subsystem.AGUnit).fold(
```
cquiroz: do you need a `fold` I think doing a mouse `option` will do the same here?

PR #896, opened 2019-04-09
SEQNG-970: UI to trigger start from actions

cquiroz:
This PR adds the support to start a sequence from any step

There is some code related to the state of the action which could be improved. I'll do that in another PR removing boilerplate

image


cquiroz:
Yes, that is eaxctly right. we don't care about the response as it normally comes on the websocket channel

PR #898, opened 2019-04-11
SEQNG-973 Enabled debug info for CaObserveSenderImpl

jluhrs:
Extra debug info to investigate SEQNG-891


modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -42,7 +42,7 @@
     private final short MRK_PRESET = 2;
     private final short MRK_IDLE = 0;
 
-    private final Boolean trace = false;
+    private final Boolean trace = true;
```
cquiroz: Maybe we should make this configurable to enable/disable more easily
I understand it is deep in the hirerachy but what about using a sys property
`trace = System.getProperty('epics.observe.trace') != null`?

PR #899, opened 2019-04-11
SEQNG-954: Gsaoi configuration model

cquiroz:
First round of doing GSAOI configuration model
I did a little refactoring of `InstrumentSystem` too. That refactoring should be expanded to all instruments but I'll do that in another PR

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/Gsaoi.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/Gsaoi.scala
@@ -1,0 +1,171 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server.gsaoi
+
+import cats.data.Reader
+import cats.effect.Sync
+import cats.effect.LiftIO
+import cats.implicits._
+import edu.gemini.spModel.config2.Config
+import edu.gemini.spModel.gemini.gsaoi.Gsaoi._
+import edu.gemini.spModel.obscomp.InstConstants.DARK_OBSERVE_TYPE
+import edu.gemini.spModel.obscomp.InstConstants.OBSERVE_TYPE_PROP
+import edu.gemini.spModel.seqcomp.SeqConfigNames.OBSERVE_KEY
+import gem.enum.LightSinkName
+import java.lang.{Double => JDouble}
+import java.lang.{Integer => JInt}
+import shapeless.tag
+import seqexec.server.ConfigUtilOps._
+import seqexec.model.dhs.ImageFileId
+import seqexec.model.enum.Instrument
+import seqexec.server.ConfigUtilOps.ExtractFailure
+import seqexec.server.ConfigResult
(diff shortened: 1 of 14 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/Gsaoi.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  override def sfName(config: Config): LightSinkName = LightSinkName.Gsaoi
+
+  override val contributorName: String = "GSAOI"
+
+  override val observeControl: InstrumentSystem.ObserveControl[F] =
+    OpticControl[F](StopObserveCmd[F](SeqActionF.liftF(controller.stopObserve)),
+                 AbortObserveCmd[F](SeqActionF.liftF(controller.abortObserve)),
+                 PauseObserveCmd[F](SeqActionF.void),
+                 ContinuePausedCmd[F](_ => SeqActionF(ObserveCommand.Unsupported)),
```
cquiroz: GSAOI doesn't support pause even being optical, etc. so I'm returning unsupported though it should never be called. We could instead do another trait like `PartialOpticControl`, I don't know if it would be simpler

modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiController.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def stopObserve: F[Unit]
+
+  def abortObserve: F[Unit]
+
+  def observeProgress(total: Time): fs2.Stream[F, Progress]
+
+  def calcTotalExposureTime(cfg: DCConfig)(
+    implicit ev:                 Applicative[F]): F[Time] = {
+    val readFactor  = 2.2
```
cquiroz: This is what the current seqexec does but it could be improved

modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiController.scala
```diff
@@ -0,0 +1,80 @@
...
+  // DC
+  type ReadMode           = edu.gemini.spModel.gemini.gsaoi.Gsaoi.ReadMode
+  type Roi                = edu.gemini.spModel.gemini.gsaoi.Gsaoi.Roi
+  type Coadds             = Int @@ CoaddsI
+  type ExposureTime       = Time
+  type NumberOfFowSamples = Int @@ NumberOfFowSamplesI
+
+  // CC
+  type Filter       = edu.gemini.spModel.gemini.gsaoi.Gsaoi.Filter
+  type OdgwSize     = edu.gemini.spModel.gemini.gsaoi.Gsaoi.OdgwSize
```
jluhrs: This one will be tricky, because it can be overridden from TCC.

cquiroz: I suppose we still have to read it from the sequence and override it on the controller?

jluhrs: It will probably be included in the GuideConfig stored in GuideConfigDb. Then, at the configuration time, the final value has to be calculated

PR #902, opened 2019-04-14
Refactoring Instrument controller implementations

cquiroz:
This PR refactors all the Instrument controller implementations making them use the same style, basically returning `F[A]` instead of `SeqAction[A]`

We already have some instruments using this style but others use `SeqAction[A]`. This PR unifies them

As all these refactorings I'm trying to limit the effect taking baby steps. Thus now all controllers traits and simulators are parametrized on `F` while the concrete Epics implementations still have `IO`


modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
```diff
@@ -3,27 +3,24 @@
...
-import scala.concurrent.ExecutionContext
+import seqexec.model.dhs.ImageFileId
+import seqexec.server.SeqexecFailure.SeqexecException
 import seqexec.server.InstrumentSystem.ElapsedTime
 import squants.time.{Seconds, Time}
 
 import scala.annotation.tailrec
 
-class InstrumentControllerSim(name: String, useTimeout: Boolean) {
+class InstrumentControllerSim[F[_]: Sync](name: String, useTimeout: Boolean) {
```
tpolecat: In general I prefer a model like this

```scala
// an unconstrained tagless interface
trait InstrumentControllerSim[F[_]] {
  def observe(fileId: ImageFileId, expTime: Time): F[ObserveCommand.Result] 
  ...
}

// and then a factory method that does constrain F
  def apply[F[_]: Sync](name: String, useTimeout: Boolean): InstrumentControllerSim[F] =
    ...
}
```

over a concrete class with constrained `F`, but what you have here is fine for now. I know you're trying to do things systematically.

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
```diff
@@ -37,90 +34,82 @@ class InstrumentControllerSim(name: String, useTimeout: Boolean) {
...
 
   @tailrec
-  private def observeTic(stop: Boolean, abort: Boolean, pause: Boolean, remain: Int, timeout: Option[Int]): TrySeq[ObserveCommand.Result] =
+  private def observeTic(stop: Boolean, abort: Boolean, pause: Boolean, remain: Int, timeout: Option[Int]): F[ObserveCommand.Result] =
     if(remain < tic) {
       Log.info(s"Simulate $name observation completed")
-      TrySeq(ObserveCommand.Success)
-    } else if(stop) TrySeq(ObserveCommand.Stopped)
-      else if(abort) TrySeq(ObserveCommand.Aborted)
+      (ObserveCommand.Success: ObserveCommand.Result).pure[F]
```
tpolecat: You can sometimes get away with `ObserveCommand.Success.pure[F].widen` in cases like this. It kind of depends on what else is in the code path. You can certainly say `.pure[F].widen[ObserveCommand.Result]` but that's not much better. This will be better in Scala 3.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsController.scala
```diff
@@ -3,34 +3,33 @@
...
+  def stopObserve: F[Unit]
 
-  def abortObserve: SeqAction[Unit]
+  def abortObserve: F[Unit]
 
-  def observeProgress(total: Time): fs2.Stream[IO, Progress]
+  def observeProgress(total: Time): fs2.Stream[F, Progress]
 
-  def calcTotalExposureTime[F[_]: Applicative](cfg: GnirsController.DCConfig): F[Time] = {
+  def calcTotalExposureTime(cfg: GnirsController.DCConfig)(implicit ev: Applicative[F]): F[Time] = {
```
tpolecat: When I need a constraint in order to implement a method on a tagless interface I will push it down to the implementations, or maybe introduce an abstract class.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -195,6 +195,17 @@ package object server {
         case r                 => SeqexecFailure.SeqexecException(r)
       })))
 
+    // embedF will take an F attempt it and convert it to SeqActionF
+    // The semantics of error handling are:
+    // * if the F has failed with a known exception type put it on the left of the EitherT
+    // * For non fatal exceptions, they are wrapped in a SeqexecException
+    // * Fatal exceptions are propagated
+    def embedF[F[_]: ApplicativeError[?[_], Throwable], A](a: => F[A]): SeqActionF[F, A] =
```
tpolecat: `a` is strict here so there's need for it to be by-name.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -223,6 +234,11 @@ package object server {
       SeqActionF.embed(ioa)
   }
 
+  implicit class SeqActionOps[F[_]: MonadError[?[_], Throwable], A](a: SeqActionF[F, A]) {
+    def actionF: F[A] =
+      a.leftWiden[Throwable].rethrowT
```
tpolecat: We can define this a little more generically by letting the error type be any subtype of throwable.

```scala
implicit class MoreEitherTOps[F[_]: MonadError[?[_], Throwable], A <: Throwable, B](fa: EitherT[F, A, B]) {
    def widenRethrowT: F[B] = 
      fa.leftWiden[Throwable].rethrowT
  } 
```

And with a bit more work to get it to infer right we can generalize it to any subtype of any error type.

```scala
implicit class MoreEitherTOps[F[_], A, B](fa: EitherT[F, A, B]) {
  def widenRethrowT[T](
    implicit me: MonadError[F, T],
             at: A <:< T
  ): F[B] = 
    fa.leftMap(at).rethrowT
} 
```

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
cquiroz: That seems good but how do you know in the implementation. would you need to put the implementation class inside the `object`?

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
cquiroz: 👍

PR #903, opened 2019-04-15
SEQNG-958 Support loading GSAOI sequences

cquiroz:
Minimal code to get GSAOI sequences to load

image


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -430,6 +433,7 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
     case Instrument.GmosN => true
     case Instrument.Nifs  => true
     case Instrument.Niri  => true
+    case Instrument.Gsaoi => true
```
jluhrs: ODGW are not treated as an OIWFS. This should be false.

PR #904, opened 2019-04-16
Controller refactorings

cquiroz:
Yet another refactor of the instrument controllers. While it looks large it focuses in two issues

* It hides the implementation on the controllers. e.g `NifsControllerSim` is not visible anymore, instead you get a regular `NifsController` using `apply` on the object, thanks for the suggestion @tpolecat 
* While EPICS instrument controllers run on `F[_]`, Giapi controllers run on `SeqActionF[F, _]`. This PR unifies them


modules/seqexec/server/src/main/scala/seqexec/server/ghost/GhostController.scala
```diff
@@ -11,53 +11,12 @@ import giapi.client.commands.Configuration
...
-    def determineType(t: IFUTargetType): BundleConfig = t match {
-      case IFUTargetType.SkyPosition => BundleConfig.Sky
-      case _                         => this
-    }
-  }
+import seqexec.server.ConfigUtilOps.ContentError
+import seqexec.server.ConfigUtilOps.ExtractFailure
+import seqexec.server.keywords.GdsClient
 
+package ghost {
```
tpolecat: It's weird to put this stuff in package blocks and in the package object … can we pull out classes into their own source files and put `Eq[GhostConfig]` into the companion for `GhostConfig`?

cquiroz: Sure, I think package objects are going to be removed in dotty right?

tpolecat: Yeah, but also `package foo { ... }` is virtually unknown and unused even though the language allows it.

cquiroz:
Tested with GPI as the changes pertain the Giapi instruments. All seems fine including error handling

PR #905, opened 2019-04-16
SEQNG-911: Subsystem button colors

sraaphorst:
This PR changes the color of the subsystem buttons in an observation step, using:

- Blue: Idle / not configured.
- Yellow: Configuring.
- Green: Configured.

This is done by introducing a new state for subsystem configuration, namely `ResourceRunComplete`.

One thing I'm not entirely happy about is that when the table is first loaded, the map of resources to steps is an empty map; after that point, it becomes populated, and resources are moved to `ResourceRunIdle`.

One thought, given that I don't think we want to assume what resources are going to be used in advance: we use `Map.empty.withDefaultValue(ResourceRunOperation.ResourceRunIdle)` for the resource state map in `TabOperations.Default`. The issue with this is that the return from `SortedMap.withDefaultValue` is a `Map` instead of a `SortedMap` (although it will still be sorted: the underlying implementation remains a `SortedMap`).

I'm not entirely sure why we are using a `SortedMap` here: does it give us a significant advantage? In any case, the `SortedMap` is spread across a number of files (but as far as I can tell, not required), I'm not going to touch this until it's discussed further.

I'm including a video showing the change to this PR.

seqng911

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/OperationsStateHandler.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/OperationsStateHandler.scala
@@ -75,7 +75,12 @@ class OperationsStateHandler[M](modelRW: ModelRW[M, SequencesOnDisplay])
                                .set(StartFromOperation.StartFromInFlight)))
 
     case RunResourceComplete(id, _, r) =>
-      updated(value.markOperations(id, TabOperations.resourceRun(r).set(none)))
+      updated(
+        value.markOperations(
+          id,
+          TabOperations
+            .resourceRun(r).
+            set(ResourceRunOperation.ResourceRunCompleted.some)))
   }
 
   def handleOperationResult: PartialFunction[Any, ActionResult[M]] = {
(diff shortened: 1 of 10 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/OperationsStateHandler.scala
```diff
@@ -75,7 +75,12 @@ class OperationsStateHandler[M](modelRW: ModelRW[M, SequencesOnDisplay])
                                .set(StartFromOperation.StartFromInFlight)))
 
     case RunResourceComplete(id, _, r) =>
-      updated(value.markOperations(id, TabOperations.resourceRun(r).set(none)))
+      updated(
+        value.markOperations(
+          id,
+          TabOperations
+            .resourceRun(r).
```
cquiroz: looks odd to have a dot `.` at both ends.

cquiroz:
Looks great. Could we add a `check mark` along the green color?

sraaphorst:
@cquiroz Sure. That would probably be clearer. I'll go ahead and give that a try.

sraaphorst:
@cquiroz Here is a video with the checkmarks:
checkmarks

PR #907, opened 2019-04-18
SEQNG-959 Implemented GSAOI Epics code.

jluhrs:
I had to create a specialized version of `CaObserveSenderImpl` for GSAOI, called `CaSimpleObserveSenderImpl`. It is actually simpler, because it does not use two CAR records, but only one.
When writing the tests for `CaSimpleObserveSenderImpl` I found a bug in `CaObserveSenderImpl`, and another in the tests that hid the bug in `CaObserveSenderImpl`. They will be fixed in SEQNG-978

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaSimpleObserveSenderImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaSimpleObserveSenderImpl.java
@@ -1,0 +1,775 @@
+/*
+ * Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+ * For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+package edu.gemini.epics.acm;
+
+import edu.gemini.epics.EpicsReader;
+import edu.gemini.epics.EpicsService;
+import edu.gemini.epics.ReadOnlyClientEpicsChannel;
+import edu.gemini.epics.api.ChannelListener;
+import edu.gemini.epics.impl.EpicsReaderImpl;
+import gov.aps.jca.CAException;
+import gov.aps.jca.TimeoutException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/*
+ * Class to handle observe commands in instruments were there is no CAR for the apply. In those cases, The observe CAR
+ * handles all the command state.
+ */
+public class CaSimpleObserveSenderImpl<C extends Enum<C> & CarStateGeneric> implements ApplySenderWithResource {
+    private static final Logger LOG = LoggerFactory.getLogger(CaSimpleObserveSenderImpl.class.getName());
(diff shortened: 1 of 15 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiEpics.scala
```diff
@@ -0,0 +1,182 @@
...
+  def countdown: F[Option[Double]] =  safeAttributeSDouble(status.getDoubleAttribute("countdown"))
+
+  private val dhsConnectedAttr: CaAttribute[DhsConnected] =
+    status.addEnum[DhsConnected]("dhsConnected", s"${GsaoiTop}sad:dc:dhsConnO", classOf[DhsConnected])
+  def dhsConnected: F[Option[DhsConnected]] = safeAttribute(dhsConnectedAttr)
+
+
+}
+
+object GsaoiEpics extends EpicsSystem[GsaoiEpics[IO]] {
```
cquiroz: I kind of want to migrate away from the `XXXEpics` objects and de them as classes parametrized on `F[_]` Do you think we could do that here?

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlImpl.java
```diff
@@ -21,7 +21,9 @@
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
-public class CaTaskControlImpl implements CaTaskControl {
+interface TaskControlWithResoource extends CaTaskControl, CaResource {}
```
cquiroz: `Resoource`?

modules/acm/src/main/java/edu/gemini/epics/acm/CaSimpleObserveSenderImpl.java
```diff
@@ -0,0 +1,775 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/*
+ * Class to handle observe commands in instruments were there is no CAR for the apply. In those cases, The observe CAR
+ * handles all the command state.
+ */
+public class CaSimpleObserveSenderImpl<C extends Enum<C> & CarStateGeneric> implements ApplySenderWithResource {
```
cquiroz: Will this be used only for GSAOI?

modules/acm/src/main/java/edu/gemini/epics/acm/CaTaskControlImpl.java
jluhrs: `CaResource` is just an interface for objects that need to be "unbinded" at the end of their lifetime.

modules/acm/src/main/java/edu/gemini/epics/acm/CaSimpleObserveSenderImpl.java
jluhrs: Yes. It could be used for other instruments thought (NIFS comes to mind)

modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiEpics.scala
jluhrs: `GsaoiEpics` is already a class paramtrized on `F[_]`. You are looking here at the companion object.

PR #909, opened 2019-04-22
SEQNG-978: ObserveStateSpec rework and fix for ignoring fixes to write DIR

cquiroz:
This PR includes several changes to the epics code to handle observations:

* Epics channels can be mocked using `EpicsReader` and `EpicsWriter` rather than `EpicsService`. This way we can test the listeners and avoid certain failures
* Failures to write to DIR are properly handled
* GSAOI ignored tests are now enabled
* GSAOI abort properly ends on failure
* the executors used on the the classes handling state changes will caught exceptions. This should help to debug exceptional situations not yet considered

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
@@ -13,7 +13,8 @@ import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import edu.gemini.epics.EpicsService;
+import edu.gemini.epics.EpicsReader;
+import edu.gemini.epics.EpicsWriter;
 import edu.gemini.epics.api.ChannelListener;
 import gov.aps.jca.CAException;
 import gov.aps.jca.TimeoutException;
@@ -29,7 +30,7 @@ final class CaApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements Ap
     private final CaApplyRecord apply;
     private final CaCarRecord<C> car;
 
-    private final Boolean trace = false;
+    private final Boolean trace = true;
 
     private long timeout;
     private TimeUnit timeoutUnit;
@@ -70,13 +71,14 @@ final class CaApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements Ap
         final String carRecord,
         final String description,
         final Class<C> carClass,
-        final EpicsService epicsService) throws CAException {
+        final EpicsReader epicsReader,
+        final EpicsWriter epicsWriter) throws CAException {
         super();
         this.name = name;
         this.description = description;
         this.currentState = IdleState;
 
-        apply = new CaApplyRecord(applyRecord, epicsService);
+        apply = new CaApplyRecord(applyRecord, epicsReader, epicsWriter);
         apply.registerValListener(valListener = new ChannelListener<Integer>() {
             @Override
             public void valueChanged(String arg0, List<Integer> newVals) {
@@ -86,7 +88,7 @@ final class CaApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements Ap
             }
         });
 
-        car = new CaCarRecord<C>(carRecord, carClass, epicsService);
+        car = new CaCarRecord<C>(carRecord, carClass, epicsReader);
         car.registerClidListener(carClidListener = new ChannelListener<Integer>() {
             @Override
             public void valueChanged(String arg0, List<Integer> newVals) {
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
@@ -126,7 +126,7 @@ final class CaApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements Ap
             }
         });
 
-        executor = new ScheduledThreadPoolExecutor(2, threadFactory);
+        executor = SafeExecutor.safeExecutor(2, LOG);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
```diff
@@ -64,19 +66,39 @@ public State onTimeout() {
...
+
+        @Override
+        public Thread newThread(Runnable r) {
+            final Thread thread = new Thread(r);
+
+            thread.setUncaughtExceptionHandler( new Thread.UncaughtExceptionHandler() {
+
+                @Override
+                public void uncaughtException(Thread t, Throwable e) {
+                    LOG.error("Uncaught exception on CaSimpleObserverSender", e);
```
tpolecat: Can you also log the thread id?

modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -56,6 +56,25 @@
     private ChannelListener<Short> stopMarkListener;
     private CaObserveSenderImpl.ApplyState currentState;
 
+    private ThreadFactory threadFactory = new ThreadFactory(){
```
tpolecat: This looks familiar, would it be worth factoring out?

modules/acm/src/main/java/edu/gemini/epics/acm/CaSimpleObserveSenderImpl.java
```diff
@@ -192,17 +211,17 @@ public synchronized CaCommandMonitor post() {
...
             failCommand(cm, new CaCommandInProgress());
         } else {
-            currentState = new CaSimpleObserveSenderImpl.WaitApplyPreset(cm);
 
             try {
                 apply.setDir(CadDirective.START);
+                currentState = new CaSimpleObserveSenderImpl.WaitApplyPreset(cm);
+                if (timeout > 0) {
+                    timeoutFuture = executor.schedule(() ->  CaSimpleObserveSenderImpl.this.onTimeout(), timeout, timeoutUnit);
+                }
```
tpolecat: This also looks familiar.

cquiroz: It is the same as another change but I don't think we should refactor this one out, It may be confusing as I want to make both `*observerSender*` classes indepenent

modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
cquiroz: 👍

modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
cquiroz: 👍

modules/acm/src/main/java/edu/gemini/epics/acm/CaApplySenderImpl.java
```diff
@@ -29,7 +31,7 @@
     private final CaApplyRecord apply;
     private final CaCarRecord<C> car;
 
-    private final Boolean trace = false;
+    private final Boolean trace = true;
```
jluhrs: `Boolean.getBoolean("epics.apply.trace")` ?

cquiroz: Yes, you're right

modules/acm/src/main/java/edu/gemini/epics/acm/CaSimpleObserveSenderImpl.java
```diff
@@ -610,7 +629,8 @@ public ApplyState onApplyValChange(Integer val) {
 
         @Override
         public ApplyState onCarValChange(CarStateGeneric val)  {
-            return this;
+            failCommand(cm, new CaObserveAborted());
+            return idleState;
```
jluhrs: Strictly, it should be
```
            if(val.isIdle()) {
                failCommand(cm, new CaObserveAborted());
                return idleState;
            }
            else if(val.isError()) {
                failCommandWithCarError(cm);
                return idleState;
            }
            else if(val.isPaused()) {
                pauseCommand(cm);
                return idleState;
            }
            else {
                return this;
            }
```

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -201,25 +214,28 @@ final class ObserveStateSpec extends FunSuite with MockFactory {
...
+    (epicsReader.getIntegerChannel _).expects("gm:apply.VAL").returns(intChannelS)
+    (epicsReader.getStringChannel _).expects("gm:apply.MESS").returns(strChannel)
+    (epicsReader.getIntegerChannel _).expects("gm:applyC.CLID").returns(intChannelS)
+    (epicsReader.getEnumChannel _).expects("gm:applyC.VAL", *).returns(carChannel)
+    (epicsReader.getStringChannel _).expects("gm:applyC.OMSS").returns(strChannel)
+    (epicsReader.getIntegerChannel _).expects("gm:dc:observeC.CLID").returns(intChannel)
+    (epicsReader.getEnumChannel _).expects("gm:dc:observeC.VAL", *).returns(carChannel)
+    (epicsReader.getStringChannel _).expects("gm:dc:observeC.OMSS").returns(strChannel)
+    (epicsReader.getShortChannel _).expects("gm:stop.MARK").returns(shortChannel)
+    (epicsReader.getShortChannel _).expects("gm:abort.MARK").returns(shortChannel)
```
jluhrs: The same mocks are defined on all the GMOS tests. Can they be factored out ? Same for all other instruments.

modules/acm/src/main/java/edu/gemini/epics/acm/CaSimpleObserveSenderImpl.java
cquiroz: 👍

PR #910, opened 2019-04-24
SEQNG-961 Add GSAOI keywords

cquiroz:
Support of GSAOI keywords


modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiHeader.scala
```diff
@@ -3,18 +3,69 @@
...
+    tcsKeywordsReader: TcsKeywordsReader[F],
+    instReader:        GsaoiKeywordReader[F]
+  ): Header[F] = new Header[F] {
     override def sendBefore(obsId: Observation.Id, id: ImageFileId): F[Unit] =
-      Applicative[F].unit
+      sendKeywords(
+        id,
+        inst,
+        List(
+          buildInt32S(tcsKeywordsReader.getGsaoiInstPort, KeywordName.INPORT),
```
tpolecat: So I guess there are cases where the keyword data type isn't uniquely determined by the Scala data type? Normally I would expect to see this done with a typeclass.

cquiroz: The type passed to `buildXX` should match the type of the first parameter but we may do eventually as you suggest
I'm planning a few refactorings that will get us there. In fact I'd like the keyword name to define a type so e.g. in this case you have to use an `Int` for `INPORT` avoiding a whole class of errors

tpolecat: :+1:

PR #914, opened 2019-04-25
SEQNG-979 Better handling of step selection when a sequence is running

cquiroz:
At the moment you can select step x and if you execute run on a previous step you end up with an odd situation where the run subsystem buttons are available though the sequence is running as:

image

This PR fixes that removing the step selection in this case. Additionally we forbid step selection while a sequence is running

### DIFF
--- a/modules/seqexec/model/src/test/scala/seqexec/model/arb/ArbRunningStep.scala
+++ b/modules/seqexec/model/src/test/scala/seqexec/model/arb/ArbRunningStep.scala
@@ -1,0 +1,27 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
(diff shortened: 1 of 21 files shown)


modules/seqexec/model/src/test/scala/seqexec/model/arb/ArbRunningStep.scala
```diff
@@ -0,0 +1,27 @@
...
+import seqexec.model.StepId
+
+trait ArbRunningStep {
+
+  implicit val arbRunningStep: Arbitrary[RunningStep] =
+    Arbitrary {
+      for {
+        l <- arbitrary[StepId]
+        i <- arbitrary[StepId]
+      } yield RunningStep(l, i)
```
sraaphorst: Shouldn't we force `i ≤ l`?

cquiroz: Good suggestion. I'll harden the model

PR #915, opened 2019-04-25
SEQNG-821 Synchronize resource execution across clients

cquiroz:
There are some discrepancies across multiple clients when a resource is executed. this PR fixes two things:

* If a resource in step x is executed the state of execution is reflected in all clients
* If a second client has selected another step y and step x is executed at the first client, the step being executed is selected in all clients

screencast 2019-04-25 17-36-50


jluhrs:
I am not sure about changing the selected step remotely. That can be annoying to the other user. On the other hand, it is expected that the person running the sequence has control.

cquiroz:
I agree it could be annoying. OTOH hand you want to show what step is being executed. I'm open to any suggestions to improve it

PR #918, opened 2019-04-26
SEQNG-981: Alternative ways to set breakpoints

cquiroz:
It has been commented that the controls to set breakpoints can be hard to use

This PR provides two alternative ways to flip a breakpoint

* Click on any part of the row with ALT pressed
* Click on any part of the row with the middle button

It is difficult to demonstrate but the video below shows the feature 

screencast 2019-04-26 17-16-15


andrewwstephens:
That looks great!  This will resolve one of the most common complaints.

PR #920, opened 2019-04-30
SEQNG-960 Implemented GSAOI Epics configuration.

jluhrs:
Apply the GSAOI configuration to the actual instrument


modules/acm/src/main/java/edu/gemini/epics/acm/CaContinuousApplySenderImpl.java
```diff
@@ -0,0 +1,423 @@
...
+import gov.aps.jca.TimeoutException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+public class CaContinuousApplySenderImpl<C extends Enum<C> & CarStateGeneric> implements ApplySenderWithResource {
```
jluhrs: This special ApplySender handles the case of continuous Epics commands, where the corresponding CAR record indicates the activity of the command, and there is no other indication of just the start. The normal ApplySender would not work in these cases, because it would wait forever for the CAR record to go Idle

modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/Gsaoi.scala
```diff
@@ -50,16 +50,16 @@ final case class Gsaoi[F[_]: LiftIO: Sync](
   override val contributorName: String = "GSAOI"
 
   override val observeControl: InstrumentSystem.ObserveControl[F] =
-    UnpausableControl[F](StopObserveCmd[F](SeqActionF.liftF(controller.stopObserve)),
-                 AbortObserveCmd[F](SeqActionF.liftF(controller.abortObserve)))
+    UnpausableControl[F](StopObserveCmd[F](SeqActionF.embedF(controller.stopObserve)),
+                 AbortObserveCmd[F](SeqActionF.embedF(controller.abortObserve)))
```
jluhrs: Some errors were not captured because of this.

modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiController.scala
```diff
@@ -32,8 +32,8 @@ trait GsaoiController[F[_]] {
 
   def calcTotalExposureTime(cfg: DCConfig)(
     implicit ev:                 Applicative[F]): F[Time] = {
-    val readFactor  = 2.2
-    val readOutTime = 300
+    val readFactor  = 1.2
+    val readOutTime = 15
```
jluhrs: Previous values were taken from the timeout calculation, and were a bit too big for total time estimation.

modules/acm/src/main/java/edu/gemini/epics/acm/CaContinuousApplySenderImpl.java
cquiroz: Did we add tests for this case? With epics mocks?

PR #922, opened 2019-05-07
SEQNG-343 GeMS Epics code.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/gems/package.scala
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server
+
+import cats.Eq
+import cats.implicits._
+import edu.gemini.seqexec.server.gems.{ApdState, BinaryOnOff, ReadyState}
+
+package object gems {
```
cquiroz: This may be better in a trait

PR #923, opened 2019-05-08
SEQNG-877C: Convert GWS to do epics reads inside an effect

cquiroz:
Another PR to make GWS do reads as referential reads inside an `F[_]`

There is a bigger change in this PR, we normally return things like `F[Option[A]]` acknowledging that Epics reads can return null. We could simplify this code a bit doing reads as `F[A]` taking care of checking of nulls and raising an error

If this style seems appropriate we could refactor the other classes as well

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsHealth.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsHealth.scala
@@ -3,16 +3,15 @@
 
 package seqexec.server
 
-import scalaz.syntax.equal._
-import scalaz.std.AllInstances._
+import cats.Eq
+import cats.implicits._
 
-/**
-  * Created by jluhrs on 7/18/17.
-  */
-sealed trait EpicsHealth
+sealed trait EpicsHealth extends Product with Serializable
 
 object EpicsHealth {
-  object Good extends EpicsHealth
-  object Bad extends EpicsHealth
+  case object Good extends EpicsHealth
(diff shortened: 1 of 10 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/EpicsHealth.scala
```diff
@@ -3,16 +3,15 @@
 
 package seqexec.server
 
-import scalaz.syntax.equal._
-import scalaz.std.AllInstances._
```
cquiroz: WTF? Amazing this was still here

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -386,15 +385,32 @@ object EpicsUtil {
   def safeAttribute[F[_]: Sync, A](get: => CaAttribute[A]): F[Option[A]] =
     Sync[F].delay(Option(get.value))
 
+  /** Tries to read a value of type A from a channel
+   *  Null results are raised as error and other errors are captured
+   */
+  def safeAttributeF[F[_]: Sync, A >: Null: Eq](name: String, get: => CaAttribute[A]): F[A] =
+    Sync[F].delay(get.value).ensure(NullEpicsError(name))(_ =!= null)
```
cquiroz: So we check for null and raise an error in that case, and adapt any other error

PR #926, opened 2019-05-09
SEQNG-822 Indicate if a user tries to configure a resource in use

cquiroz:
At the moment the server will safely ignore requests to configure a subsystem e.g. TCS if busy. However the client does not get informed thus it looks like it is still trying to run it

This PR sends a notification in this case

Seqexec - GS-2018B-Q-0-20 2019-05-09 14-18-22


arturog8m:
Add something to tell the user what to do... "Try again later"? "Check TCS"? "Pray"?

PR #928, opened 2019-05-10
GWS Read errors not handled properly

cquiroz:
We have a report of an error reading epics channels. This PR attempts to fix it but I'd like to have some discussion about the details

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
@@ -275,7 +275,7 @@ package object keywords {
     // Check if there is an error reading a value and if there is a failure
     //  use the default
     def safeValOrDefault: F[A] =
-      v.attempt.map(_.getOrElse(DefaultHeaderValue[A].default))
+      v.handleError(_ => DefaultHeaderValue[A].default)
   }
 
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -275,7 +275,7 @@ package object keywords {
     // Check if there is an error reading a value and if there is a failure
     //  use the default
     def safeValOrDefault: F[A] =
-      v.attempt.map(_.getOrElse(DefaultHeaderValue[A].default))
+      v.handleError(_ => DefaultHeaderValue[A].default)
```
cquiroz: @tpolecat `handleError` seems to be correct here but my tests at the console show the line above does the same. Perhaps we could discuss the details

tpolecat: Yeah `.attempt` isn't a reliable way to trap errors because following `map/flatMap` may never happen, depending on `F`. If you need to trap or handle errors use the combinators on `ApplicativeError/MonadError`.

tpolecat: We can talk about it in the morning.

cquiroz:
Did some more testing and there were a few extra surprising details to take care of

PR #930, opened 2019-05-13
SEQNG-991 Make active step display on all changes

cquiroz:
On the big step table refactoring #924 the cache was probably too loose and now we are missing some updates, e.g. when a step goes from configuring to observing

This PR fixes that allowing step changes to be displayed properly at the cost of more computation on each change. Note that I'm not computing the step configuration which is the largest part of the code

Additionally I'm updating some components removing the use of `ReactState` which will no longer be recommended on the next scalajs-react update,


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -457,8 +457,15 @@ object StepsTable extends Columns {
     val InitialState: State = State(InitialTableState, None, None, 0)
   }
 
+  val stdStepReuse: Reusability[StandardStep] =
+    Reusability.caseClassExcept('config)
+  implicit val stepReuse: Reusability[Step] =
```
cquiroz: This is the core change which detects when a steps needs to be re-rendered

PR #932, opened 2019-05-17
Removed stopped state

cquiroz:
This PR is more of a question to discuss. We have the `SequenceState.Stopped` value but we never really set it on the server
@jluhrs Are we really using it?


jluhrs:
At least not in the execution engine, because it does not make any distinction between a sequence that is idle and one that was run and stopped in the middle.

cquiroz:
So do you think I could just remove it from the code?

PR #935, opened 2019-05-24
SEQNG-983 GNIRS Focus and Hartmann masks

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/gnirs/Gnirs.scala
```diff
@@ -89,35 +91,49 @@ object Gnirs {
     case _                 => getCCOtherConfig(config)
   }
 
+  // Used for optional keys. If the key is not present, the result is a None. Other extraction errors are passed on.
+  private implicit class KeyAttemptToOption[T](v: Either[ExtractFailure, T]) {
+    def optionalKey: Either[ExtractFailure, Option[T]] = v match {
```
cquiroz: I think this could be done as well with `recoverWith`, but it is fine as it is as well

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/Gnirs.scala
```diff
@@ -180,4 +210,11 @@ object Gnirs {
     case _                      => None
   }
 
+  private def getFocus(config: Config): Either[ExtractFailure, GnirsController.Focus] =
+    config.extractInstAs[Focus](FOCUS_PROP).flatMap{ v =>
+      if (v.toString === FocusSuggestion.BEST_FOCUS.displayValue) GnirsController.Focus.Best.asRight
```
cquiroz: Doesn't wart remover complain about `toString`?

jluhrs: No. I am not happy with having to analyse the string value to figure out with type of focus setting to use, but I couldn't think of a better way.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/Gnirs.scala
jluhrs: `recoverWith` goes on the bad path and does not change the good values. I could have used `v.map(Some).recoverWith{...}`, but I think this way is fine.

PR #937, opened 2019-05-24
REL-3272 Updated definition for GMOS OVI and OVIC filters.

jluhrs:


cquiroz:
Could you review that definitions for the UI are correct at
https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/model/src/main/scala/seqexec/model/enumerations.scala

jluhrs:
Yes they are correct. They did not change.

PR #938, opened 2019-05-27
SEQNG-887: Refactoring of header calculations

cquiroz:
Convert GnirsKeywordReader to the current model plus some refactorings of SeqTranslate out of SEQNG-993

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -46,7 +46,7 @@ import seqexec.server.nifs._
 import seqexec.server.altair.Altair
 import seqexec.server.altair.AltairHeader
 import seqexec.server.altair.AltairLgsHeader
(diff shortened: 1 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -203,8 +203,8 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
     for {
       stepType  <- calcStepType(config)
       inst      <- toInstrumentSys(stepType.instrument)
-      systems   <- calcSystems(config, stepType)
-      headers   <- calcHeaders(config, stepType)
+      systems   <- calcSystems(config, stepType, inst)
```
cquiroz: I realized I could simplify the code a bit passing `InstrumentSystem[F]` rather than reading it everywhere

jluhrs: :+1:

PR #939, opened 2019-05-28
Upgrade scalajs-react to 1.4.2 and react to 16.7.0

cquiroz:
I have been sitting on this upgrade for a while as dependencies (diode) weren't yet updated

With this PR we can upgrade no a newer react version


sraaphorst:
@tpolecat I think the webm file is for Linux browsers (or FireFox). I remember @cquiroz saying that heh had issues with mp3 on all combinations of platforms.

cquiroz:
doh, we don't need those audio files yet. I'll remove them

cquiroz:
about mp3 and webm, we need both to cover a wide spectrum of browser. FF in linux doesn't play mp3 due to licensing issues

PR #940, opened 2019-05-28
SEQNG-993/994: GPI Align & Calib

cquiroz:
This PR adds support for GPI Align And Calib. 
It supports the current mode of operation where you have a special sequence with a calibration and A&C, but it also supports a mode where all the steps are in one sequence:

Gemini OT -  GS-2018B-Q-0  Science Program 2019-05-27 17-06-43

This mode could eventually simplify night time operations but needs testing

There is still work on UI and actual testing but the core functionality is done in this PR

Seqexec - GS-2018B-Q-0-104 2019-05-28 14-37-18

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
@@ -7,6 +7,7 @@ import seqexec.model.enum.Instrument
 
 sealed trait StepType {
   val instrument: Instrument
+  val includesObserve: Boolean = true
 }
 
 final case class CelestialObject(override val instrument: Instrument) extends StepType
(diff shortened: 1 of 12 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server
+
+import seqexec.model.enum.Instrument
+
+sealed trait StepType {
+  val instrument: Instrument
+  val includesObserve: Boolean = true
```
cquiroz: Now step may skip the observation if e.g. they are align and calib which doesn't require one

PR #942, opened 2019-05-30
SEQNG-887 Refactor GmosKeywordReader and cleanup

cquiroz:
Another PR refactoring the gmos keyword reader.
This refactoring was slightly harder due to the handling of ADC and ROI, thus it was tested against the  gmos epics db

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
@@ -1,0 +1,211 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server.gmos
+
+import cats.Applicative
+import cats.effect.LiftIO
+import cats.effect.Sync
+import cats.implicits._
+import seqexec.server.ConfigUtilOps._
+import seqexec.server.keywords._
+import seqexec.server.gmos.GmosEpics.RoiStatus
+import edu.gemini.spModel.config2.Config
+import edu.gemini.spModel.data.YesNoType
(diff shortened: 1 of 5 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
```diff
@@ -0,0 +1,211 @@
...
+  def adcPrismExtSt:             F[Double]
+  def adcPrismExtEnd:            F[Double]
+  def adcPrismExtMe:             F[Double]
+  def adcWavelength1:            F[Double]
+  def adcWavelength2:            F[Double]
+  def detNRoi:                   F[Int]
+  def roiValues:                 F[List[(Int, RoiValues)]]
+  def aExpCount:                 F[Int]
+  def bExpCount:                 F[Int]
+  def isADCInUse:                F[Boolean]
```
cquiroz: We get `isACInUse` inside `F[_]` now

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
```diff
@@ -0,0 +1,211 @@
...
+  def adcPrismEntSt:             F[Double]
+  def adcPrismEntEnd:            F[Double]
+  def adcPrismEntMe:             F[Double]
+  def adcPrismExtSt:             F[Double]
+  def adcPrismExtEnd:            F[Double]
+  def adcPrismExtMe:             F[Double]
+  def adcWavelength1:            F[Double]
+  def adcWavelength2:            F[Double]
+  def detNRoi:                   F[Int]
+  def roiValues:                 F[List[(Int, RoiValues)]]
```
cquiroz: Instead of a `Map` of actions we'll get a fully resolved action with a map of values. I switched to a `List` of tuples to ensure we preserve the order

PR #944, opened 2019-06-01
SEQNG-824: Display icons on the tab while a subsystem is active

cquiroz:
When a system is active in another tab we want to show it to the other tabs
This PR adds little icons to the tabs as shown below:

screencast 2019-05-31 22-22-16


andrewwstephens:
That looks great!

PR #946, opened 2019-06-03
SEQNG-887: Make Gnirs epics reads referentially transparent

cquiroz:
Another PR making the Epics reads referentially transparent. This time is GNIRS

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
@@ -275,12 +275,12 @@ object GnirsControllerEpics extends GnirsEncoders {
 
   val warnOnDhs =
     epicsSys.dhsConnected.map(_.contains(true)).flatMap(
-       IO(Log.warn("GNIRS is not connected to DHS")).whenA
+       IO(Log.warn("GNIRS is not connected to DHS")).unlessA
     )
 
   val warnOnArray =
     epicsSys.arrayActive.map(_.contains(true)).flatMap(
-       IO(Log.warn("GNIRS detector array is not active")).whenA
+       IO(Log.warn("GNIRS detector array is not active")).unlessA
     )
 
   def apply()(implicit ioTimer: Timer[IO]): GnirsController[IO] =
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
@@ -265,21 +265,21 @@ object GnirsControllerEpics extends GnirsEncoders {
 
   val checkDhs =
     failUnlessM(
-      epicsSys.dhsConnected.map(_.forall(identity)),
+      epicsSys.dhsConnected.map(_.contains(true)),
                 SeqexecFailure.Execution("GNIRS is not connected to DHS"))
 
   val checkArray =
     failUnlessM(
-      epicsSys.arrayActive.map(_.forall(identity)),
+      epicsSys.arrayActive.map(_.contains(true)),
                 SeqexecFailure.Execution("GNIRS detector array is not active"))
 
   val warnOnDhs =
-    epicsSys.dhsConnected.map(_.forall(identity)).flatMap(
+    epicsSys.dhsConnected.map(_.contains(true)).flatMap(
        IO(Log.warn("GNIRS is not connected to DHS")).whenA
     )
 
   val warnOnArray =
-    epicsSys.arrayActive.map(_.forall(identity)).flatMap(
+    epicsSys.arrayActive.map(_.contains(true)).flatMap(
        IO(Log.warn("GNIRS detector array is not active")).whenA
     )
 
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
@@ -5,36 +5,25 @@ package seqexec.server.gnirs
 
 import cats.implicits._
 import cats.effect.{ IO, Timer }
+import cats.data.Nested
 import seqexec.model.dhs.ImageFileId
-import seqexec.server._
+import seqexec.server.{Dark => _, _}
 import edu.gemini.spModel.gemini.gnirs.GNIRSParams
 import edu.gemini.spModel.gemini.gnirs.GNIRSParams.{Camera, Decker, Disperser, ReadMode}
 import fs2.Stream
 import org.log4s.getLogger
-import scala.concurrent.ExecutionContext
 import squants.{Length, Seconds, Time}
 import squants.space.LengthConversions._
 import seqexec.model.dhs.ImageFileId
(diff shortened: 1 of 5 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -238,73 +241,84 @@ object GnirsControllerEpics {
...
 
-  def apply(): GnirsController[IO] =
+  val checkDhs =
+    failUnlessM(
+      epicsSys.dhsConnected.map(_.forall(identity)),
+                SeqexecFailure.Execution("GNIRS is not connected to DHS"))
+
+  val checkArray =
+    failUnlessM(
+      epicsSys.arrayActive.map(_.forall(identity)),
```
cquiroz: @jluhrs Previously `arrayActive` was checked with `exists`. Was there a reason for that?

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsEpics.scala
```diff
@@ -126,100 +133,99 @@ class GnirsEpics(epicsService: CaService, tops: Map[String, String]) {
...
 
-  val state: CaStatusAcceptor = epicsService.getStatusAcceptor("nirs::status")
-  val dcState: CaStatusAcceptor = epicsService.getStatusAcceptor("nirs::dcstatus")
-
-  def arrayId: Option[String] = Option(dcState.getStringAttribute("arrayid").value)
+  private val state: CaStatusAcceptor = epicsService.getStatusAcceptor("nirs::status")
+  private val dcState: CaStatusAcceptor = epicsService.getStatusAcceptor("nirs::dcstatus")
 
-  def arrayType: Option[String] = Option(dcState.getStringAttribute("arraytyp").value)
+  def arrayId: F[Option[String]] = safeAttribute(dcState.getStringAttribute("arrayid"))
```
cquiroz: We are leaning to produce items with the form `F[A]` rather than `F[Option[A]]` but I want to do than in a later PR

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsEpics.scala
```diff
@@ -126,100 +133,99 @@ class GnirsEpics(epicsService: CaService, tops: Map[String, String]) {
...
+    .map(_.toInt =!= 0).value
 
   val arrayActiveAttr: Option[CaAttribute[JDetectorState]] = Option(dcState.addEnum(
     "arrayState", s"${GnirsTop}dc:activate", classOf[JDetectorState]
   ))
-  def arrayActive: Option[Boolean] = arrayActiveAttr.flatMap(at => Option(at.value))
-    .map(_.getActive)
 
-  def minInt: Option[Double] = Option(dcState.getDoubleAttribute("minInt").value).map(_.toDouble)
+  def arrayActive: F[Option[Boolean]] =
```
cquiroz: @jluhrs Could you review this one in particular. I'm not sure if I fully understand the original

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -103,17 +92,25 @@ object GnirsControllerEpics {
...
       case Acquisition => "In"
       case _           => "Out"
     }
 
-    smartSetParam(v, epicsSys.acqMirror.map(removePartName), ccCmd.setAcqMirror(v))
+    smartSetParamF(v, Nested(epicsSys.acqMirror).map(removePartName).value, ccCmd.setAcqMirror(v))
   }
 
-  private def setGrating(s: Spectrography, c: Camera): List[SeqAction[Unit]] = {
+  private def setGrating(s: Spectrography, c: Camera): List[IO[Option[IO[Unit]]]] = {
```
jluhrs: In these case an the ones below, couldn't `IO[List[IO[Unit]]]` be used instead of `List[IO[Option[IO[Unit]]]]`.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsEpics.scala
jluhrs: The original takes the attribute, which reads a binary enum, and converts it to a `Boolean`. Of course, the attribute could not exists, that is why it is treated a an Option. Your implementation follows that, but I don't understand the last `_.getOrElse(false)`. There is no default value in the original. If the value cannot be read, then the whole result is `None`. That us why there is a `flatMap`.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
jluhrs: It was `GnirsEpics.instance.arrayActive.exists(identity)`, meaning that to pass, the value must be read properly and must be true. (i.e. it must be `Some(true)`). Now `arrayActive` gives a `F[Option[Boolean]]`. `forall` is true for `None`, therefore your code will not stop the observation if it cannot verify the state of the array. That is wrong.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
cquiroz: I've tried to implement this on a different way but the issue I found is that I need to know if there are elements on the list. If there are no elements on the list I shouldn't execute the `after` part on https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala#L426
That is needed to avoid calling `post` if there are no changes.
I can still look for a better way to do this. I agree it is hard to follow

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
cquiroz: It seems we should we apply the same logic for the DHS checks, WDYT?

jluhrs: Yes. GNIRS can crash if you try to take an image when it is not connected to the DHS or the detector array is not active.

modules/seqexec/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -238,73 +241,84 @@ object GnirsControllerEpics {
...
+                SeqexecFailure.Execution("GNIRS is not connected to DHS"))
+
+  val checkArray =
+    failUnlessM(
+      epicsSys.arrayActive.map(_.contains(true)),
+                SeqexecFailure.Execution("GNIRS detector array is not active"))
+
+  val warnOnDhs =
+    epicsSys.dhsConnected.map(_.contains(true)).flatMap(
+       IO(Log.warn("GNIRS is not connected to DHS")).whenA
```
jluhrs: `.unlessA` ?

cquiroz: You're right

PR #947, opened 2019-06-03
Dependency updates

cquiroz:
Some library updates. The scala.js update should let us get smaller files


cquiroz:
almost 2% size reduction, not bad

PR #951, opened 2019-06-07
Use gsp-math

swalker2m:
Updates `ocs3` to use `gsp-math`, removing duplicated code.


project/Settings.scala
```diff
@@ -142,6 +142,9 @@ object Settings {
     val guava                   = "25.0-jre"
     val prometheusClient        = "0.6.0"
     val geminiLocales           = "0.1.0-2019a"
+
+    // Gemini Libraries
+    val gspMath                 = "0.1.0+5-7322fff9-SNAPSHOT"
```
cquiroz: I don't think we should depend on a SNAPSHOT

tpolecat: Let's tag a real version here. `0.1.1` maybe.

cquiroz:
This PR is in practice a split of the projects. I've been reluctant to do this mainly on the question on how we ensure apps don't get out of track of the libs as we know happens on the old ocs world.

I don't see a mechanic on how we're going to do it. I guess the builds on the app side will do at least some of it but how do we ensure that libs don't diverge in a way that breaks the apps.

e.g. if gsp-math goes to 2.13 we won't be able to use it on the seqexec as many/most of our dependencies are not yet on 2.13

Would scala-steward help? Can it be made so that libs are "blocked" until apps that depend on them  catch up. I recall the discussion was explicit on us devoting time to do such ports

project/Settings.scala
cquiroz: Is the version with the hash published after you merge a PR right? If so the hash would never change if master is protected

tpolecat:
Yes, I think we should add all our repos to scala-steward, which will open PRs when library updates are available. I think we'll want to cross-publish `gsp-math` and other library code for both 2.12 and 2.13 and then applications can catch up when they're ready.

project/Settings.scala
tpolecat: Yes, that's the merge commit. So it's really not a snapshot since it's stable, but I think we probably don't want every downstream project to get an update PR every time there's a merge into master. So, work for a while, use the snapshot in dev if you need to, then tag a release when it seems time.

cquiroz:
One huge timesink on merging ocs3 and gem was harmonizing the builds including things like scalastyle checks, wartremover, dependency checks, etc
`gsp-math` is much simpler removing most of that as far as I can tell
How could we make that the builds match at least on some basics, like using AutomateHeader and removing WartRemover
Presumably with time plugins will evolve but different projects will get out of sync at that level

tpolecat:
We can make a plugin that defines common build options. Scala Steward will keep plugins up to date too so it's no different from a library dependency.

At this point the scalac warnings are good enough that I think we can just do without WartRemover.

project/Settings.scala
swalker2m: We make updates to libraries like `gsp-math` *because we need them* for one or more of our applications. I don't think we're likely to work for a while on a library independently and then tag it.  Instead it seems more likely that we'll be trying to update an application, find that we need some new syntax or a new method and have to switch to the library to add it.  If we need to tag every update before using it we'll be tagging with every tweak.  I suppose that's fine and I will do it if that's the consensus, but I guess I don't see it as really necessary at least while everything is under active development.

cquiroz: We can probably live with just the hash and only tag every now and then when appropriate.

I'd personally not spawn a library out until it is somewhat stable. It is going to be annoying to be switching projects during active development

swalker2m: I think shared libraries like `gsp-math` and whatever we call the library that contains model code are going to be under constant development.  It will be quite a while before everything stabilizes.

swalker2m:
I've made the requested change but the requestor will be out for a month and I don't want this to languish.  Opening a new PR.

I'm working on establishing a shared build settings SBT plugin and Scala Steward should address the other concerns.

PR #953, opened 2019-06-08
Remove use of ReactS

cquiroz:
The use of the State monad for React is no longer recommended. I've already removed a couple of uses but this PR removes the last bits

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/tabs/CalibrationQueueTab.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/tabs/CalibrationQueueTab.scala
@@ -10,7 +10,7 @@ import japgolly.scalajs.react.component.builder.Lifecycle.RenderScope
 import japgolly.scalajs.react.component.Scala.Unmounted
 import japgolly.scalajs.react.vdom.html_<^._
 import japgolly.scalajs.react.Reusability
-import japgolly.scalajs.react.CatsReact._
+import japgolly.scalajs.react.MonocleReact._
 import japgolly.scalajs.react._
 import monocle.macros.Lenses
 import seqexec.model.enum.BatchExecState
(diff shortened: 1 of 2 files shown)


cquiroz:
This indirectly remove a dependency thus we save 10k on output size

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/tabs/CalibrationQueueTab.scala
```diff
@@ -98,10 +95,10 @@ object CalibrationQueueTab {
...
         e.preventDefaultCB *> Callback { e.dataTransfer.dropEffect = "copy" }
       },
-      ^.onDragEnter ==> b.runStateFn(onDragEnter),
-      ^.onDrop ==> b.runStateFn(onDrop),
-      ^.onDragEnd --> b.runState(onDragEnd),
-      ^.onDragLeave --> b.runState(onDragEnd),
+      ^.onDragEnter ==> onDragEnter(b) _,
+      ^.onDrop ==> onDrop(b) _,
+      ^.onDragEnd --> onDragEnd(b),
+      ^.onDragLeave --> onDragEnd(b),
```
sraaphorst: What is the difference between `==>` and `-->`?

cquiroz: In one case `-->` you only pass a `Callback` which essentially is the effect you want to execute.
`==>` is similar but you also get the `ReactEvent` which maybe useful is some cases
It is documented here
https://github.com/japgolly/scalajs-react/blob/master/doc/VDOM.md#event-handlers

sraaphorst: Thanks!

PR #954, opened 2019-06-10
SEQNG-974 Fix for GNIRS observe.

jluhrs:
Sometimes during a GNIRS observation, the event of the main DC apply record going BUSY is not received (or is not generated by the instrument?), causing Seqexec to wait forever for it (ending in a timeout). This PR fixes it. It also adds tests cases for GNIRS, and factors some of the test code.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
@@ -81,6 +81,43 @@ public final class CaCommandMonitorImpl implements CaCommandMonitor {
         }
     }
 
+    private boolean isActive() {
+        return currentState == State.BUSY;
+    }
+
+    @Override
+    public State waitInactive(long timeout, TimeUnit unit)
+            throws TimeoutException, InterruptedException {
+        lock.lock();
+        State state = currentState;
+        try {
+            while (isActive()) {
+                if (!condition.await(timeout, unit)) {
+                    throw new TimeoutException();
+                }
+            }
+            state = currentState;
+        } finally {
+            lock.unlock();
+        }
+        return state;
+    }
+
+    @Override
+    public State waitInactive() throws InterruptedException {
+        lock.lock();
+        State state = currentState;
+        try {
+            while (!isActive()) {
+                condition.await();
+            }
+            state = currentState;
+        } finally {
+            lock.unlock();
+        }
+        return state;
+    }
+
     @Override
     public State state() {
         return currentState;
@@ -106,9 +143,12 @@ public final class CaCommandMonitorImpl implements CaCommandMonitor {
         if (currentState == State.BUSY) {
             currentState = State.IDLE;
             condition.signalAll();
-            lock.unlock();
-            if (callback != null) {
-                callback.onSuccess();
+            try {
+                if (callback != null) {
+                    callback.onSuccess();
+                }
+            } finally {
+                lock.unlock();
             }
         } else {
             lock.unlock();
@@ -121,9 +161,12 @@ public final class CaCommandMonitorImpl implements CaCommandMonitor {
             this.cause = cause;
             currentState = State.ERROR;
             condition.signalAll();
-            lock.unlock();
-            if (callback != null) {
-                callback.onFailure(cause);
+            try {
+                if (callback != null) {
+                    callback.onFailure(cause);
+                }
+            } finally {
+                lock.unlock();
             }
         } else {
             lock.unlock();
@@ -135,9 +178,12 @@ public final class CaCommandMonitorImpl implements CaCommandMonitor {
         if (currentState == State.BUSY) {
             currentState = State.PAUSE;
             condition.signalAll();
-            lock.unlock();
-            if (callback != null) {
-                callback.onPause();
+            try {
+                if (callback != null) {
+                    callback.onPause();
+                }
+            } finally {
+                lock.unlock();
             }
         } else {
             lock.unlock();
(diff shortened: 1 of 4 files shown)


modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
```diff
@@ -106,9 +143,12 @@ public synchronized void completeSuccess() {
...
             condition.signalAll();
-            lock.unlock();
-            if (callback != null) {
-                callback.onSuccess();
+            try {
+                if (callback != null) {
+                    callback.onSuccess();
+                }
+            } finally {
+                lock.unlock();
```
jluhrs: The `unlock` was moved after the callback call, because otherwise it was possible for code calling `waitDone` to not see the effect of the callback. Same for `competeFailure` and `completePause`.

modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -803,7 +803,12 @@ else if (val.isError()) {
                 return idleState;
             }
             else {
-                return this;
+                if(val.isIdle() && observeState != idleObserveState) {
```
jluhrs: If the main CAR is changing to IDLE here, and the observe CAR has already change, then assume that the main CAR change to BUSY was lost.

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -785,12 +928,12 @@ final class ObserveStateSpec extends FunSuite with GsaoiMocks with NifsMocks wit
 class CadDirectiveChannelMock extends ReadWriteClientEpicsChannel[CadDirective] {
   override def setValue(a: java.util.List[CadDirective]): Unit = {}
   override def setValue(a : CadDirective): Unit = {}
-  override def getAll(): java.util.List[CadDirective] = null
-  override def getDBR(): gov.aps.jca.dbr.DBR = null
-  override def getFirst(): CadDirective = null.asInstanceOf[CadDirective]
-  override def getName(): String = null
-  override def getType(): gov.aps.jca.dbr.DBRType = null
-  override def isValid(): Boolean = true
```
jluhrs: The IDE told me to remove the parentheses

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
```diff
@@ -871,48 +1014,69 @@ class StringChannelMock extends ReadWriteClientEpicsChannel[String] {
...
     val m  = mock[ShortChannelMock]
     (m.registerListener(_: ChannelListener[JShort])).expects(*)
     m
   }
 }
 
+@SuppressWarnings(
+  Array("org.wartremover.warts.NonUnitStatements",
+        "org.wartremover.warts.PublicInference"))
+trait GenericInstMocks extends ChannelsFactory {
```
jluhrs: This trait factor out some common code.

modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -716,7 +716,7 @@ public String signature() {
                         failCommandWithCarError(cm);
                         return idleState;
                     } else if (carState.isBusy()) {
-                        return new CaObserveSenderImpl.WaitApplyBusy(cm, val, carState, observeState);
+                        return new CaObserveSenderImpl.WaitApplyIdle(cm, val, carState, observeState);
```
jluhrs: Why did it go to `WaitApplyBusy` if `carState` was already busy? Looks like a hidden bug.

cquiroz: 👍

modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
```diff
@@ -81,6 +81,43 @@ public void waitDone() throws InterruptedException {
...
+        return currentState == State.BUSY;
+    }
+
+    @Override
+    public State waitInactive(long timeout, TimeUnit unit)
+            throws TimeoutException, InterruptedException {
+        lock.lock();
+        State state = currentState;
+        try {
+            while (isActive()) {
```
cquiroz: This loop is  a bit worrisome. Could we ever be stuck in an infinite loop?

modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
cquiroz: is this the one condition that sometimes happens with GNIRS?

modules/acm/src/test/scala/edu/gemini/epics/acm/ObserveStateSpec.scala
cquiroz: 👍

modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
jluhrs: Yes.

modules/acm/src/main/java/edu/gemini/epics/acm/CaCommandMonitorImpl.java
jluhrs: No, because eventually the timeout will be triggered.

jluhrs: On second thought, there is a small issue here, in that if after x seconds the condition is triggered spuriously (the reason of the loop), the `await` will be  called again with the original timeout value, making effectively equal to timeout + x.
I'll put a Jira ticket to fix it.

PR #957, opened 2019-06-11
Use gsp-math

swalker2m:
Currently the `ocs3` build has a copy of shared "math" classes that have been split into a separate repository, `gsp-math`.  This PR updates `ocs3` to use `gsp-math`.  See #951 for comments on the previous attempt to accomplish this feat.


build.sbt
```diff
@@ -85,9 +85,14 @@ addCommandAlias("rebuildEnums", "; sql/flywayClean; sql/flywayMigrate; genEnums;
 addCommandAlias("schemaSpy", "sql/runMain org.schemaspy.Main -t pgsql -port 5432 -db gem -o modules/sql/target/schemaspy -u postgres -host localhost -s public")
 addCommandAlias("gemctl", "ctl/runMain gem.ctl.main")//
 
+resolvers in ThisBuild +=
+  Resolver.sonatypeRepo("public")
+
 resolvers in ThisBuild +=
   Resolver.sonatypeRepo("snapshots")
 
+updateOptions in ThisBuild := updateOptions.value.withLatestSnapshots(false)
```
swalker2m: If we're going to require tagging every update, these settings may not be necessary.

project/Settings.scala
```diff
@@ -142,6 +142,9 @@ object Settings {
     val guava                   = "25.0-jre"
     val prometheusClient        = "0.6.0"
     val geminiLocales           = "0.1.0-2019a"
+
+    // Gemini Libraries
+    val gspMath                 = "0.1.1"
```
swalker2m: This was the change requested in #951.

swalker2m:
This is compiling again and up-to-date with the `develop` branch as of the time of this writing. They are mostly mechanical but still time consuming changes.  Are there any objections beyond the difference of opinion over monorepos vs. multirepos?

cquiroz:
I think this particular PR is technically fine. The changes are fairly straightforward and the `sbt-gsp` work is very good for any choice of repos.

Additionally, `gsp-math` is a good example of a library that can be spawned off so I have no objections to that in particular.

I'm still not happy about splitting core libraries into a multirepo. While the mitigations (scala-steward, ci-release, sbt-gsp) reduce my worries, I'd still prefer a monorepo at least until libraries are stable, like `gsp-math`. IMHO a multirepo will produce a disassociation of library and application development that can be detrimental to the schedule and work distribution

swalker2m:
>  I guess we could use sbt-gsp on ocs3 as well

Yes, I think that would be good.  I can try to set that up next.

PR #965, opened 2019-06-19
SEQNG-986A: Move TelescopeGuideConfig classes to the model

cquiroz:
This is a fairly straightforward change moving `TelescopeGuideConfig` classes to the seqexec model to be shared with the client. The changes are all mechanical but it is still a fairly sized PR thus I prefer to send it separately


modules/seqexec/model/src/main/scala/seqexec/model/TelescopeGuideConfig.scala
```diff
@@ -0,0 +1,23 @@
...
+import seqexec.model.enum.M1GuideConfig
+import seqexec.model.enum.M2GuideConfig
+
+/** Data type for guide config. */
+@Lenses
+final case class TelescopeGuideConfig(mountGuide: MountGuideOption,
+                                      m1Guide:    M1GuideConfig,
+                                      m2Guide:    M2GuideConfig)
+
+@SuppressWarnings(Array("org.wartremover.warts.PublicInference"))
```
swalker2m: Given #967 I suppose it doesn't matter but out of curiosity where is the public inference that is being suppressed here?

modules/seqexec/model/src/main/scala/seqexec/model/enum/M1GuideConfig.scala
```diff
@@ -0,0 +1,28 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.model.enum
+
+import cats.Eq
+import cats.Show
+import cats.implicits._
+
+/** Data type for M1 guide config. */
+sealed trait M1GuideConfig
```
swalker2m: `extends Product with Serializable`?

modules/seqexec/model/src/test/scala/seqexec/model/arb/ArbCommaOption.scala
```diff
@@ -0,0 +1,23 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.model.arb
+
+import org.scalacheck.Arbitrary
+import org.scalacheck.Cogen
+import org.scalacheck.Gen
+import seqexec.model.enum.ComaOption
+
+trait ArbComaOption {
```
swalker2m: It seems like all of these enums should provide `gem.util.Enumerated` instances.  If they did, you'd get the arbitraries and cogens for free from `gem.arb.ArbEnumerated`.

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
```diff
@@ -236,6 +195,47 @@ trait ModelBooPicklers extends GemModelBooPicklers {
...
 
+  private val comaOptionIdx = Map((0 -> ComaOption.ComaOn),
+                                  (1 -> ComaOption.ComaOff))
+
+  implicit val comaPickler = valuesMapPickler(comaOptionIdx)
+
+  private val tipTiltSourceIdx = Map((0 -> TipTiltSource.PWFS1),
+                                     (1 -> TipTiltSource.PWFS2),
+                                     (2 -> TipTiltSource.OIWFS),
+                                     (3 -> TipTiltSource.GAOS))
```
swalker2m: I wonder if this too could be generalized for any `Enumerated` instance.

modules/seqexec/model/src/main/scala/seqexec/model/TelescopeGuideConfig.scala
cquiroz: `WartRemover` and the `@Lenses` don't like each other because of reasons. So we need to suppress it to let `@Lenses` to work. One of the main reasons I disilke `WartRemover`

modules/seqexec/model/src/main/scala/seqexec/model/enum/M1GuideConfig.scala
cquiroz: Good catch

modules/seqexec/model/src/test/scala/seqexec/model/arb/ArbCommaOption.scala
cquiroz: Hmm. That sounds good, I guess we can use that even for non generated enums?

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
cquiroz: Right, that should be possible too

cquiroz:
I added `Enumerated` instances to the Enums as suggested and extended this to all seqexec model enums doing a general cleanup
It ended up being a larger change. It maybe worth doing another review

modules/seqexec/model/src/main/scala/seqexec/model/enum/Guiding.scala
```diff
@@ -18,10 +18,14 @@ object Guiding {
...
-  implicit val equal: Eq[Guiding] =
-    Eq.fromUniversalEquals
-
   def fromString(s: String): Option[Guiding] =
     all.find(_.configValue === s)
 
+  /** @group Typeclass Instances */
+  implicit val GuidingEnumerated: Enumerated[Guiding] =
+    new Enumerated[Guiding] {
+      def all = Guiding.all
```
jluhrs: This is a minor complain, but I would prefer if the different `Enumerated` followed the same implementation style. Right now there are some that define `all` directly as a `List` of all posible values, and other implementations that define the list in the container object, and have the `Enumerated.all` refer to that one.

cquiroz: Fixed, we now have `val all: List[A]` in all enum types

modules/seqexec/model/src/main/scala/seqexec/model/enum/ComaOption.scala
```diff
@@ -0,0 +1,27 @@
...
+  case object ComaOn  extends ComaOption
+  case object ComaOff extends ComaOption
+
+  val all: List[ComaOption] =
+    List(ComaOn, ComaOff)
+
+  /** @group Typeclass Instances */
+  implicit val CommaOptionEnumerated: Enumerated[ComaOption] =
+    new Enumerated[ComaOption] {
+      def all = List(ComaOn, ComaOff)
```
jluhrs: @carlos You missed this one :)

cquiroz: doh

cquiroz:
Added another update with  the latest suggestions

modules/seqexec/model/src/main/scala/seqexec/model/TelescopeGuideConfig.scala
swalker2m: I guess the code it writes doesn't have an explicit type.  Seems like they should, but it's irrelevant now.

modules/seqexec/model/src/main/scala/seqexec/model/enum/BatchExecState.scala
```diff
@@ -21,14 +21,29 @@ object BatchExecState {
...
     case Completed => "Completed"
+  }
 
+  def tag(s: BatchExecState): String = s match {
+    case Idle      => "Idle"
+    case Running   => "Running"
+    case Waiting   => "Waiting"
+    case Stopping  => "Stopping"
+    case Completed => "Completed"
   }
```
swalker2m: I assumed this would go into the `BatchExecState` `trait` itself with a 

```scala
```

.  At any rate, `Show` should probably call this if you leave it this way so it isn't duplicated?  We should probably have a 1-hour team meeting about whether `Show` should be defined at all and if so whether it should ever be defined to anything but `Show.fromToString`.  :-)

Issue #510, opened 2018-04-25
Automated formatting


cquiroz:
This is a place to discuss formatting. I've been using scalafmt but not very systematically. It is not integrated into the build and I've used it only for new code.
I'm using the format rules selected for `http4s` with the only difference that I'm doing vertical alignment as much as possible.

```
style = default

maxColumn = 100

align = most // In http4s they use align = none

rewrite.rules = [
  AvoidInfix
  RedundantBraces
  RedundantParens
  AsciiSortImports
  PreferCurlyFors
]
```

What do you think?
/cc @tpolecat @swalker2m @sraaphorst @jluhrs @jdnavarro 

Some links:
[scalafmt](http://scalameta.org/scalafmt/#Scalafmt)
[http4s config](https://github.com/http4s/http4s/blob/master/.scalafmt.conf)

tpolecat:
I messed around with it and don't like the results, but I won't put up a fight.

cquiroz:
Are you unhappy with the concept of formatting with scalafmt or with the given formatting? Could it be improved by a better config?

tpolecat:
I'm playing around with the config but the interactions between settings are unpredictable.

swalker2m:
Auto-formatting seems like a great idea, if we can get it to produce reasonable results.  My view is that we stick with the [Scala guide](https://docs.scala-lang.org/style/index.html) as much as possible.  It's arbitrary, but since everyone has their own preferences it gives us a way to resolve disagreements.

Two items that have come up where we might want to differ from the guide are comment style and line length.  I have a slight preference for the Scala doc style (option 2):

> Scaladoc style, with gutter asterisks aligned in column three:

```scala
/** Provides a service as described.
  *
  * This is further documentation of what we're documenting.
  * Here are more details about how it works and what it does.
  */
def member: Unit = ()
```

but the Javadoc style that Rob prefers is also fine.  Does anybody else have a preference?  I'll switch to Javadoc style if not.

For line length, going much over 80 will be a hardship for me since I have to magnify the text and like to see two source files side-by-side.  As long as auto-formatting won't *lengthen* 80 character lines that I write I can just deal with the horizontal scrolling induced by other people using longer lines.

sraaphorst:
I switched to the asterisk commenting style for Scala code after this was brought to my attention (I believe by Shane some time ago) and now only use the Javadoc style when writing Java code.

[...]

PR #969, opened 2019-06-19
Adjusted scalastyle settings

cquiroz:
I propose we get rid of the `scalastyle` check. I don't see the point of doing a style check when there is no commonly agreed style (see #510). the `scalastyle` check is normally just hit when buildkite fails making us waste time

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -18,13 +18,7 @@
   </parameters>
  </check>
  <check level="error" class="org.scalastyle.scalariform.EqualsHashCodeChecker" enabled="true"></check>
- <check level="error" class="org.scalastyle.scalariform.IllegalImportsChecker" enabled="true">
-  <parameters>
-   <parameter name="illegalImports"><![CDATA[sun._,java.awt._]]></parameter>
-  </parameters>
- </check>
  <check level="error" class="org.scalastyle.scalariform.ReturnChecker" enabled="true"></check>
- <!-- <check level="error" class="org.scalastyle.scalariform.NullChecker" enabled="true"></check> -->
  <check level="error" class="org.scalastyle.scalariform.CovariantEqualsChecker" enabled="true"></check>
  <check customId="console.io" level="error" class="org.scalastyle.file.RegexChecker" enabled="true">
   <customMessage>Avoid console i/o disable with // scalastyle:off console.io</customMessage>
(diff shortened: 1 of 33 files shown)


swalker2m:
I'm not sure about this one.  I hardly ever trip upon any of the checks it performs and I think that

>  I don't see the point of doing a style check when there is no commonly agreed style (see #510)

greatly overstates the level of disagreement.  The differences of opinion boil down to just two issues:

* preferred doc style - this one is so trivial that i just decided to go with the flow and do JavaDoc style instead of ScalaDoc style
* line length limit (80 vs. 100) - I'm fine with other people using 100 though it does cause horizontal scrolling as long as my own code can be more or less limited to 80 characters.

otherwise all of our code, and most typelevel Scala code for that matter, are similarly formatted I think.  Automated formatting is difficult to get right though.

So for the most part, in my experience, the scalastyle rules are naturally just not encountered but when they are it is something that should be called out for closer examination so it makes sense to flag them.

What do other people think?  If we keep it then it should be applied to `gsp-math` and any future repos as well.

cquiroz:
The one I hit the most is method line count which is a problem especially with tagless final which requires in many cases to embed classes after `apply`. Also scalastyle doesn't recognize the new style of List with a comma at the end and the error message you get is hardly helpful

My strong preference would be to pick a style and apply it with `scalafmt`. We could use the default style, or perhaps `http4s`', or actually anything. I'd rather spend less time formatting and let a tool do it

I hit `scalastyle` issues with frequency (i'd say at least 2-3 times a week) and normally not until build kite tells me

swalker2m:
If there are specific checks that we have legitimate issues with, like the method line count one, we can configure them or turn them off so that they don't come up so much.  I'm fine with automatic formatting if it can be made to produce code that looks like what we would have done by hand.  That said I'm not sure whether that will cover all the rules that `scalastyle` checks.  It doesn't seem like it's just formatting checks.

Can we automatically run `scalastyle` with each compile so that you know about it sooner? I suppose the downside of that is that you pay for the cost of doing the check all the time instead of just when you're trying to put in the PR.

cquiroz:
I counted last week 5 failed builds out of 60 due to scala style (only my builds). You can imagine each takes 5-10 mins to resolve

Some of the failures I don't even understand and my only workaround is to disable scalastyle completely, so extra work to keep a tool happy

Here's a quick sample of builds failed due to scala style
https://buildkite.com/gemini-hlsw-1/ocs3-cquiroz-pipeline/builds/1086#cc305328-2124-491a-bdd1-48724e6bf6ec
https://buildkite.com/gemini-hlsw-1/ocs3-cquiroz-pipeline/builds/1022#59a94c86-08cf-4505-9722-c63e924cc437

jluhrs:
Scalastyle does not give me thos problems, because I diligently run it to verify my commits before pushing them to Github. Well, at least the times I don't forget to do it.

cquiroz:
I have restored `scalastlye` but removed some rules that in my personal opinion are not needed.

scalastyle-config.xml
```diff
@@ -18,59 +18,29 @@
   </parameters>
  </check>
  <check level="error" class="org.scalastyle.scalariform.EqualsHashCodeChecker" enabled="true"></check>
- <check level="error" class="org.scalastyle.scalariform.IllegalImportsChecker" enabled="true">
-  <parameters>
-   <parameter name="illegalImports"><![CDATA[sun._,java.awt._]]></parameter>
-  </parameters>
- </check>
```
swalker2m: I ran into the `awt` constraint once when trying to visualize an experiment with a different Java2D library.  Do you run into the need to import `sun._` or `java.awt._` often?

scalastyle-config.xml
```diff
@@ -18,59 +18,29 @@
...
  <check level="error" class="org.scalastyle.scalariform.ReturnChecker" enabled="true"></check>
- <!-- <check level="error" class="org.scalastyle.scalariform.NullChecker" enabled="true"></check> -->
  <check level="error" class="org.scalastyle.scalariform.CovariantEqualsChecker" enabled="true"></check>
  <check customId="console.io" level="error" class="org.scalastyle.file.RegexChecker" enabled="true">
   <customMessage>Avoid console i/o disable with // scalastyle:off console.io</customMessage>
   <parameters>
    <parameter name="regex"><![CDATA[println]]></parameter>
   </parameters>
  </check>
- <!-- <check level="error" class="org.scalastyle.scalariform.NumberOfTypesChecker" enabled="true">
```
swalker2m: You're removing stuff that was already commented out, which is fine.  Is it different from setting `enabled="false"`? I guess you have to look at it when you open the file.

scalastyle-config.xml
```diff
@@ -18,59 +18,29 @@
...
   <check level="error" class="org.scalastyle.scalariform.NoFinalizeChecker" enabled="true"/>
   <check level="error" class="org.scalastyle.scalariform.NoWhitespaceAfterLeftBracketChecker" enabled="true"/>
-  <!-- <check level="error" class="org.scalastyle.scalariform.NoWhitespaceBeforeLeftBracketChecker" enabled="true"/> -->
   <check level="error" class="org.scalastyle.scalariform.NotImplementedErrorUsage" enabled="true"/>
   <check level="error" class="org.scalastyle.scalariform.ProcedureDeclarationChecker" enabled="true"/>
   <check level="error" class="org.scalastyle.scalariform.RedundantIfChecker" enabled="true"/>
-  <!-- <check level="error" class="org.scalastyle.scalariform.StructuralTypeChecker" enabled="true"/> -->
-  <!-- <check level="error" class="org.scalastyle.scalariform.TodoCommentChecker" enabled="true"/> -->
   <check level="error" class="org.scalastyle.scalariform.UppercaseLChecker" enabled="true"/>
-  <check level="error" class="org.scalastyle.scalariform.VarFieldChecker" enabled="true"/>
```
swalker2m: I take it this comes up a lot in `scala-js`?

scalastyle-config.xml
cquiroz: Oh, I thought it was never used so what was the point. but we could restore it

scalastyle-config.xml
cquiroz: It comes often on `scala.js` and there are certain legitimate uses of `var`. I guess this would be good if it could flag `public` uses of `var` but local uses are certainly valid

cquiroz:
I suppose we want to enable this on `sbt-gsp`?

swalker2m:
> I suppose we want to enable this on sbt-gsp?

If possible yes I think it makes sense to share all the settings across the projects.

scalastyle-config.xml
swalker2m: I guess it is harmless enough, but I'm fine removing it too.

PR #970, opened 2019-06-19
Remove old ui module

cquiroz:
The ui module was a raw experiment with the gem model
We don't really make any use of it and it just uses build time.
I've simplified the `buildkite` part a bit as well


cquiroz:
That was the essential reason but I doubt we ever checked that. It should be made to do fail automatically but that may be disruptive

PR #999, opened 2019-06-27
SEQNG-1015 Read N&S configuration from the sequence

cquiroz:
This is the first Nod & Shuffle PR. We are now able to read the configuration of N&S sequences

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
@@ -73,8 +73,8 @@ abstract class Gmos[F[_]: Sync, T<:GmosController.SiteDependentTypes](controller
     } yield {
       for {
         s <- BeamLabels.lift(i).toRight(ContentError(s"Unknown label at position $i"))
-        p <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-p").flatMap(configToAngle).flatMap(Offset.P(_).asRight)
-        q <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-q").flatMap(configToAngle).flatMap(Offset.Q(_).asRight)
+        p <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-p").flatMap(configToAngle).map(Offset.P.apply)
+        q <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-q").flatMap(configToAngle).map(Offset.Q.apply)
         k = INSTRUMENT_KEY / s"nsBeam${s.name}-guideWithOIWFS"
         g <- config.extractAs[StandardGuideOptions.Value](k).flatMap(r => Guiding.fromString(r.toString).toRight(KeyNotFound(k)))
       } yield NSPosition(s, Offset(p, q), g)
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -163,7 +191,7 @@ object Gmos {
     val rois = for {
       i <- 1 to 5
     } yield attemptROI(i)
-    rois.toList.collect { case Some(x) => x }
+    rois.toList.mapFilter(identity)
```
swalker2m: `rois.toList.flattenOption`?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -59,6 +60,31 @@ abstract class Gmos[F[_]: Sync, T<:GmosController.SiteDependentTypes](controller
...
+  private val BeamLabels = List('A, 'B, 'C, 'D, 'E, 'F, 'G)
+
+  private def nsPosition(config: Config, sc: Int): Either[ExtractFailure, Vector[NSPosition]] = {
+    (for {
+      i <- 0 to scala.math.min(BeamLabels.length, sc) - 1
+    } yield {
+      for {
+        s <- BeamLabels.lift(i).toRight(ContentError(s"Unknwon label at position $i"))
+        p <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-p").map(_.toDouble)
+        q <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-q").map(_.toDouble)
```
swalker2m: The `seqexec.model.TelescopeOffset` predates this PR, but we've been working hard to keep floating point values out of the model. `gsp-math` has an `Offset` class that stores the offset value unambiguously and perhaps in a future task it could be used instead.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
cquiroz: Good, I didn't know about it

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
cquiroz: I saw that, let me see if we can use it. If I recall when we needed `Offset` the projects were not yet integrated so I made a very similar local class.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -63,17 +64,20 @@ abstract class Gmos[F[_]: Sync, T<:GmosController.SiteDependentTypes](controller
...
     (for {
       i <- 0 to scala.math.min(BeamLabels.length, sc) - 1
     } yield {
       for {
-        s <- BeamLabels.lift(i).toRight(ContentError(s"Unknwon label at position $i"))
-        p <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-p").map(_.toDouble)
-        q <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-q").map(_.toDouble)
+        s <- BeamLabels.lift(i).toRight(ContentError(s"Unknown label at position $i"))
+        p <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-p").flatMap(configToAngle).flatMap(Offset.P(_).asRight)
+        q <- config.extractAs[String](INSTRUMENT_KEY / s"nsBeam${s.name}-q").flatMap(configToAngle).flatMap(Offset.Q(_).asRight)
```
swalker2m: Once you have an `Angle` the `Offset.Q` constructor can't fail so it seems like `.flatMap(configToAngle).map(Offset.Q(_))` would be sufficient?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -191,9 +195,15 @@ object Gmos {
     val rois = for {
       i <- 1 to 5
     } yield attemptROI(i)
-    rois.toList.mapFilter(identity)
+    rois.toList.flattenOption
   }
 
+  private def toGain(s: String): Either[ExtractFailure, Double] =
+    s.parseDoubleOption match {
```
swalker2m: Maybe:

```scala
s.parseDoubleOption.toRight {
  ConversionError(INSTRUMENT_KEY / AMP_GAIN_SETTING_PROP, "Bad Amp gain setting")
}
```

PR #1001, opened 2019-07-05
SEQNG-1018 Assign StepType N&S to the corresponding sequences

cquiroz:
This PR does some refactoring delegating to instrument the decision of step type out of `SeqTranslate`

We use this to tell when a step is Nod and Shuffle and show it on the UI:

image

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/ModelLenses.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/ModelLenses.scala
@@ -3,9 +3,9 @@
 
 package seqexec.model
 
-import seqexec.model.enum._
-import seqexec.model.events._
-
+import cats._
+import cats.implicits._
+import gsp.math.syntax.all._
 import monocle.Lens
 import monocle.Optional
 import monocle.Prism
@@ -17,11 +17,10 @@ import monocle.function.At.atMap
 import monocle.function.FilterIndex.filterIndex
 import monocle.unsafe.MapTraversal._
 import monocle.std.option.some
+import monocle.std.string._
 import monocle.Iso
-
-import cats._
-import cats.implicits._
-import mouse.all._
+import seqexec.model.enum._
+import seqexec.model.events._
 
 trait ModelLenses {
   // Some useful Monocle lenses
@@ -160,9 +159,7 @@ trait ModelLenses {
   private[model] def telescopeOffsetQI: Iso[Double, TelescopeOffset.Q] =
     Iso(TelescopeOffset.Q.apply)(_.value)
   val stringToDoubleP: Prism[String, Double] =
-    Prism((x: String) => x.parseDouble.toOption)(_.show)
-  val stringToIntP: Prism[String, Int] =
-    Prism((x: String) => x.parseInt.toOption)(_.show)
+    Prism((x: String) => x.parseDoubleOption)(_.show)
 
   def stepObserveOptional[A](systemName: SystemName, param: String, prism: Prism[String, A]): Optional[Step, A] =
     standardStepP                            ^|-> // which is a standard step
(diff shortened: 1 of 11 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -127,6 +128,11 @@ abstract class Gmos[F[_]: Sync, T<:GmosController.SiteDependentTypes](controller
     } yield new GmosController.GmosConfig[T](configTypes)(cc, dc)
   )
 
+  override def calcStepType(config: Config): Either[SeqexecFailure, StepType] =
+    Gmos.isNodAndShuffle(config).option(NodAndShuffle(instrument).asRight).getOrElse{
+      SequenceConfiguration.calcStepType(config)
```
jluhrs: `Gmos.isNodAndShuffle(config).fold(NodAndShuffle(instrument).asRight, SequenceConfiguration.calcStepType(config))`

modules/seqexec/server/src/main/scala/seqexec/server/gpi/Gpi.scala
```diff
@@ -52,6 +52,11 @@ final case class Gpi[F[_]: Sync: Timer](controller: GpiController[F])
 
   override val contributorName: String = "gpi"
 
+  override def calcStepType(config: Config): Either[SeqexecFailure, StepType] =
+    Gpi.isAlignAndCalib(config).option(AlignAndCalib.asRight).getOrElse{
+      SequenceConfiguration.calcStepType(config)
```
jluhrs: `Gpi.isAlignAndCalib(config).fold(AlignAndCalib.asRight, SequenceConfiguration.calcStepType(config))`

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
cquiroz: wasn't the prevalent opinion that `fold` on option was bad? @swalker2m

jluhrs: This is `fold` on `boolean`.

swalker2m: I think @tpolecat might have expressed some concern about `.fold` on `Boolean` since it can be difficult to remember which comes first?  I know it is very unhip but `if` expression seems the most readable to me personally

```scala
  if (Gmos.isNodAndShuffle(config)) NodAndShuffle(instrument).asRight
  else SequenceConfiguration.calcStepType(config)
```

That said I endorse any of these options.

modules/seqexec/model/src/main/scala/seqexec/model/ModelLenses.scala
```diff
@@ -173,6 +173,9 @@ trait ModelLenses {
   val stringToStepTypeP: Prism[String, StepType] =
     Prism(StepType.fromString)(_.label)
 
+  val stringToBooleanP: Prism[String, Boolean] =
+    Prism[String, Boolean](_.equalsIgnoreCase("true").some)(_.toString)
```
swalker2m: I think this should be a `Format` because "TRUE" => `true` => "true" => `true` ...  That is distinct inputs yield the same `getOption` result.  Does it pass the `Prism` tests?

swalker2m: Thinking about it a bit more, perhaps it should be a `SplitEpi` because all `String` values map to some `Boolean` value.

```scala
val stringToBoolean: SplitEpi[String, Boolean] =
  SplitEpi(_.equalsIgnoreCase("true"), _.toString)
```

cquiroz: Monocle provides a std `stringToBoolean` prism. I'll use that

PR #1004, opened 2019-07-08
SEQNG-887: Make epics reads for GMOS referentially transparent

cquiroz:
Gmos is the last instrument to get the treatment of SEQNG-887. This is a largish update and it would be good to be reviewed in detail and test it on hardware

Main changes are:
* Put all epics reads inside an `F[_]`
* Reform on the way the decision to apply a param is done. Now we take a full snapshot of the instrument state and use that to decide

There is space for further improvements but I'm trying to keep this constrained on the change size and time invested

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -420,6 +420,18 @@ object EpicsUtil {
   def smartSetParam[A: Eq](v: A, get: => Option[A], set: SeqAction[Unit]): List[SeqAction[Unit]] =
     if(get =!= v.some) List(set) else Nil
 
+  def applyParam[F[_]: Applicative, A: Eq](c: A, d: A, f: A => F[Unit]): Option[F[Unit]] =
+    if (c =!= d) f(d).some else none
+
+  def applyParamT[F[_]: Functor](
+    relTolerance: Double
+  )(c: Double, d: Double, set: Double => F[Unit]): Option[F[Unit]] =
+    if (!(d === 0.0 && c === 0.0) || (d === 0.0 || abs((c - d)/d) > relTolerance)) {
+      set(d).some
+    } else {
+      none
+    }
+
   // This method takes a list of actions returning possible actions
   // If at least one is defined it will execute them and then executes `after`
   def executeIfNeeded[F[_]: Monad, A](i:     List[F[Option[F[Unit]]]],
@@ -444,8 +456,11 @@ object EpicsUtil {
     get.map(_ =!= v.some).map(_.option(set))
 
   def smartSetDoubleParam(relTolerance: Double)(v: Double, get: => Option[Double], set: SeqAction[Unit]): List[SeqAction[Unit]] =
-    if(get.forall(x => (v === 0.0 && x =!= 0.0) || abs((x - v)/v) > relTolerance))
-  List(set) else Nil
+    if (get.forall(x => !(v === 0.0 && x === 0.0) || (v === 0.0 || abs((x - v)/v) > relTolerance))) {
+      List(set)
+    } else {
+      Nil
+    }
 
   def smartSetDoubleParamF[F[_]: Functor](relTolerance: Double)(v: Double, get: F[Option[Double]], set: F[Unit]): F[Option[F[Unit]]] =
     get.map(g => (g.forall(x => (v === 0.0 && x =!= 0.0) || abs((x - v)/v) > relTolerance)).option(set))
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -420,13 +420,37 @@ object EpicsUtil {
   def smartSetParam[A: Eq](v: A, get: => Option[A], set: SeqAction[Unit]): List[SeqAction[Unit]] =
     if(get =!= v.some) List(set) else Nil
 
+  /**
+   * Decides to set a param comparing the current value and the value to be set
+   * @param c Current value on the system
+   * @param d Value to be set
+   * @param f Action to set the parameter
+   */
   def applyParam[F[_]: Applicative, A: Eq](c: A, d: A, f: A => F[Unit]): Option[F[Unit]] =
     if (c =!= d) f(d).some else none
 
+  /**
+   * Test if we should set a value d given that the current value c and
+   * a given tolerance
+   * @param t Max relative tolerance allowed between the current and destination value
+   * @param c Current value on the system
+   * @param d Value to be set
+   */
+  private def toleranceTest(t: Double, c: Double, d: Double): Boolean =
+    (!(d === 0.0 && c === 0.0) && (d === 0.0 || abs((c - d)/d) > t))
+
+  /**
+   * Decides to set a param comparing the current value and the value to be set with
+   * a given tolerance
+   * @param relTolerance Max relative tolerance allowed between the current and destination value
+   * @param c Current value on the system
+   * @param d Value to be set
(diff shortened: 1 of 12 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -420,6 +420,18 @@ object EpicsUtil {
   def smartSetParam[A: Eq](v: A, get: => Option[A], set: SeqAction[Unit]): List[SeqAction[Unit]] =
     if(get =!= v.some) List(set) else Nil
 
+  def applyParam[F[_]: Applicative, A: Eq](c: A, d: A, f: A => F[Unit]): Option[F[Unit]] =
```
jluhrs: Can you add a parameter description?

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -420,6 +420,18 @@ object EpicsUtil {
...
+    if (c =!= d) f(d).some else none
+
+  def applyParamT[F[_]: Functor](
+    relTolerance: Double
+  )(c: Double, d: Double, set: Double => F[Unit]): Option[F[Unit]] =
+    if (!(d === 0.0 && c === 0.0) || (d === 0.0 || abs((c - d)/d) > relTolerance)) {
+      set(d).some
```
jluhrs: !(d === 0.0 && c === 0.0) **&&** (d === 0.0 || abs((c - d)/d) > relTolerance)
If current and demanded value are both 0.0, then we don't need to set it.

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -444,8 +456,11 @@ object EpicsUtil {
     get.map(_ =!= v.some).map(_.option(set))
 
   def smartSetDoubleParam(relTolerance: Double)(v: Double, get: => Option[Double], set: SeqAction[Unit]): List[SeqAction[Unit]] =
-    if(get.forall(x => (v === 0.0 && x =!= 0.0) || abs((x - v)/v) > relTolerance))
-  List(set) else Nil
+    if (get.forall(x => !(v === 0.0 && x === 0.0) || (v === 0.0 || abs((x - v)/v) > relTolerance))) {
```
jluhrs: Same here, that is a `&&`

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -444,8 +456,11 @@ object EpicsUtil {
...
+      List(set)
+    } else {
+      Nil
+    }
 
   def smartSetDoubleParamF[F[_]: Functor](relTolerance: Double)(v: Double, get: F[Option[Double]], set: F[Unit]): F[Option[F[Unit]]] =
     get.map(g => (g.forall(x => (v === 0.0 && x =!= 0.0) || abs((x - v)/v) > relTolerance)).option(set))
```
jluhrs: This needs to be fixed too. What about factoring out the test ?

cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
jluhrs: `toleranceTest` ?

cquiroz: Well it is just a name, some better suggestion?

jluhrs: I was too laconic. I was not complaining about the name, but pointing that it should be used here.

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -420,6 +420,42 @@ object EpicsUtil {
   def smartSetParam[A: Eq](v: A, get: => Option[A], set: SeqAction[Unit]): List[SeqAction[Unit]] =
     if(get =!= v.some) List(set) else Nil
 
+  /**
+   * Decides to set a param comparing the current value and the value to be set
+   * @param c Current value on the system
+   * @param d Value to be set
+   * @param f Action to set the parameter
+   */
+  def applyParam[F[_]: Applicative, A: Eq](c: A, d: A, f: A => F[Unit]): Option[F[Unit]] =
```
tpolecat: I don't think you need `Applicative`.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -224,123 +270,135 @@ object GmosControllerEpics {
...
-        ( (params.sequence *>
-          GmosEpics.instance.configCmd.setTimeout(ConfigTimeout) *>
-          GmosEpics.instance.post).widenRethrowT
-        ).unlessA(params.isEmpty) *>
-        IO(Log.info("Completed Gmos configuration"))
+          IO(Log.debug(s"Gmos configuration: ${config.show}")) *>
+          warnOnDHSNotConected *>
+          ((params.sequence *>
+            sys.configCmd.setTimeout[IO](ConfigTimeout) *>
+            sys.post)
```
tpolecat: You can remove the inner parens here.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosEpics.scala
```diff
@@ -3,310 +3,347 @@
...
+import seqexec.server.SeqexecFailure._
+import seqexec.server.EpicsCommandF
+import seqexec.server.ObserveCommandF
 import scala.collection.breakOut
 import scala.concurrent.duration._
-import cats.implicits._
-import mouse.all._
 
-class GmosEpics(epicsService: CaService, tops: Map[String, String]) {
+class GmosEpics[F[_]: Async](epicsService: CaService, tops: Map[String, String]) {
```
tpolecat: Stylistically this would be better as a trait with unconstrained `F` and then a factory method that demands `Async`, but I realize you're trying to minimize the diff so this is fine for now.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
```diff
@@ -131,60 +130,56 @@ object GmosKeywordReaderDummy {
 }
 
 object GmosKeywordReaderEpics {
-  def apply[F[_]: Sync: LiftIO]: GmosKeywordReader[F] = new GmosKeywordReader[F] {
-    private val F = implicitly[Sync[F]]
-    private val sys = GmosEpics.instance
+  def apply[F[_]: Sync](sys: GmosEpics[F]): GmosKeywordReader[F] = new GmosKeywordReader[F] {
```
tpolecat: I tend to give these things names like `fromGmosEpics` rather than using `apply`. Not a big deal.

cquiroz:
@tpolecat I merged this PR but I can follow up your comments on #1005 which closes the loop on this set of changes

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosEpics.scala
cquiroz: yeah. I may do this when we refactor `EpicsSystem` which is in a future update

PR #1005, opened 2019-07-09
SEQNG-887: Final touches on SEQNG-887

cquiroz:
I've been doing this large refactoring to make all epics reads referentially transparent. While some improvements can still be done we should close this

This PR removes some duplicated code that was able to work on the old and new world and rename `EpicsCommandF` to `EpicsCommand` and `ObsecveCommandF` to `ObserveCommand`

As changes go this PR is fairly mechanical

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -10,6 +10,7 @@ import cats.effect.Sync
 import cats.effect.Async
 import cats.implicits._
 import fs2.Stream
+import gem.util.Enumerated
 import java.lang.{ Double => JDouble }
 import java.lang.{ Integer => JInt }
 import java.lang.{ Float => JFloat }
@@ -37,8 +38,8 @@ trait EpicsCommand {
       cs.map { ccs =>
         ccs.postCallback {
           new CaCommandListener {
-            override def onSuccess(): Unit = f(Completed.asRight)
-            override def onPause(): Unit = f(Paused.asRight)
+            override def onSuccess(): Unit = f(Result.Completed.asRight)
+            override def onPause(): Unit = f(Result.Paused.asRight)
             override def onFailure(cause: Exception): Unit = f(cause.asLeft)
           }
         }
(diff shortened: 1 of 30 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -124,33 +90,18 @@ trait EpicsSystem[T] {
...
-      p.map(_.set(f(v)).asRight).getOrElse(SeqexecFailure.Unexpected("Unable to set parameter.").asLeft)
-    } )
-
-  def setParameter[T](p: Option[CaParameter[T]], v: T): SeqAction[Unit] =
-    safe(SeqAction.either {
-      p.map(_.set(v).asRight).getOrElse(SeqexecFailure.Unexpected("Unable to set parameter.").asLeft)
-    } )
+  sealed trait Result extends Product with Serializable
+  case object Paused extends Result
+  case object Completed extends Result
```
tpolecat: I think elsewhere we're using a style where instances are defined inside the companion, so `Result.Paused`, etc.

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -262,6 +179,10 @@ object EpicsCodex {
 }
 
 object EpicsUtil {
+  def safe[A](a: SeqAction[A]): SeqAction[A] = EitherT(a.value.attempt.map {
```
tpolecat: Can you add a comment explaining what this does?

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
cquiroz: yes, I have been changing some but I missed this one

modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -159,41 +110,7 @@ object EpicsCommand {
...
-
-  def mark: SeqAction[Unit] = safe(EitherT(IO.apply {
-    cs.map(_.mark().asRight).getOrElse(SeqexecFailure.Unexpected("Unable to mark command.").asLeft)
-  }))
-
-  def setTimeout(t: Time): SeqAction[Unit] = EpicsUtil.setTimeout(cs.map(_.getApplySender), t)
-}
-
-trait ObserveCommandF {
+trait ObserveCommand     {
```
jluhrs: Why that much space before the `{` ?

cquiroz: just a mistake

PR #1006, opened 2019-07-09
SEQNG-1019: Set nsPairs for all GMOS sequences

cquiroz:
We realized that `nsPairs` should be set always for N&S sequences and for non N&S sequences. Thus this is a bug on the current seqexec

This PR fixes it and also adds a special `NSConfig` case class. This is useful as N&S configuration is neither completely CC nor DC

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -88,6 +88,10 @@ private[gmos] final case class GmosDCEpicsState(
   ccdYBinning: Int
 )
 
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -316,6 +327,11 @@ object GmosControllerEpics extends GmosEncoders {
           // TODO these are not smart about setting them only if needed
           setROI(config.bi, config.roi)
 
+      private def nsParams(state: GmosNSEpicsState, config: NSConfig): List[IO[Unit]] =
+        List(
```
cquiroz: This `List` is unnecessary now but we'll add more elements to the list when configuring for N&S

PR #1013, opened 2019-07-18
SEQNG-331A: Initial refactoring

cquiroz:
As part of N&S support we discussed moving the code that creates the observation actions to a separate trait to reduce the complexity of `SeqTranslate`. That change snowballed to a larger refactoring which is in this PR. This PR includes the following changes:

* Move the code that creates observation actions from `SeqTranslate` to `ObserveActions`
* Include an `F[_]` param on `PauseContext`. This change cascades up the stack as we need to add an effect param at several levels
* Move and rename `EpicsCommand.Result` and `ObserveCommand.Result`. There are at least classes called `Result` and that is quite confusing at times
* Enforce the rule of putting sealed trait implementations into their companion object

In all this change is fairly big but mechanical and will make it simpler to add a special `observe` function for N&S. No functionality has changed besides a bit of code to properly process `NodAndShuffle` step types


modules/seqexec/engine/src/main/scala/seqexec/engine/Event.scala
```diff
@@ -10,51 +10,53 @@ import gem.Observation
...
 /**
   * Anything that can go through the Event Queue.
   */
-sealed abstract class Event[+D<:Engine.Types]
-final case class EventUser[D<:Engine.Types](ue: UserEvent[D]) extends Event[D]
-final case class EventSystem(se: SystemEvent) extends Event[Nothing]
+sealed trait Event[+F[_], +D <: Engine.Types] extends Product with Serializable
 
 object Event {
+  final case class EventUser[D <: Engine.Types](ue: UserEvent[D]) extends Event[Nothing, D]
```
cquiroz: Likely half of the changes are due to putting ADT instances in the companion objects

modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
```diff
@@ -98,15 +101,15 @@ object Result {
   // Base traits for results. They make harder to pass the wrong value.
   trait RetVal
   trait PartialVal
-  trait PauseContext
+  trait PauseContext[F[_]]
```
cquiroz: This change makes it necessary to propagate `F[_]` up to several classes up the code tree

modules/seqexec/engine/src/main/scala/seqexec/engine/package.scala
```diff
@@ -6,62 +6,78 @@ package seqexec
...
+
+      case object Idle extends ActionState[Nothing] {
+        override val isIdle: Boolean = true
+      }
+      case object Started extends ActionState[Nothing]
+      final case class Paused[F[_]](ctx: PauseContext[F]) extends ActionState[F]
+      final case class Completed[V <: RetVal](r: V) extends ActionState[Nothing]
+      final case class Failed(e: Error) extends ActionState[Nothing]
+
+      private def actionStateToStatus[F[_]](s: ActionState[F]): ActionStatus =
```
cquiroz: Extracted from `SeqTranslate`

modules/seqexec/model/src/main/scala/seqexec/model/enum/ApplyCommandResult.scala
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.model.enum
+
+import gem.util.Enumerated
+
+sealed trait ApplyCommandResult extends Product with Serializable
```
cquiroz: This was called `EpicsCommand.Result`. It is confusing that we have at least 3 classes called `Result` and in fact these are not used only in Epics code. Hence I thought about moving them to the model

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -0,0 +1,129 @@
...
+import seqexec.server.ConfigUtilOps._
+import seqexec.server.keywords._
+import seqexec.server.InstrumentSystem.ElapsedTime
+import seqexec.server.SeqTranslate._
+import squants.time.TimeConversions._
+
+/**
+ * Methods usedd to generate observation related actions
+ */
+trait ObserveActions {
```
cquiroz: This trait contains methods previously on `SeqTranslate`. They will be used for a N&S aware `observe` function. Mostly they are copied verbatim from `SeqTranslate` with some improvements with respect to replace `IO` usages with `F[_]`

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -423,26 +326,32 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
-          Tcs.fromConfig(systems.tcs, inst.hasOI.fold(allButGaos, allButGaosNorOi), None, sys, systems.guideDb)(
-            config,
-            LightPath(TcsController.LightSource.Sky, sys.sfName(config)),
-            extractWavelength(config)),
-          Gcal(systems.gcal, site == Site.GS)
-      )).asRight
+      case StepType.CelestialObject(inst) =>
+        (sys :: celestialObjectSys(inst)).asRight
+
+      case StepType.NodAndShuffle(inst) =>
```
cquiroz: This is the only semantically relevant change. It allows us to execute N&S steps though they will behave as regular steps. Previously an attempt to run would cause an error

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -542,32 +451,34 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
-        calcInstHeader(config, sys).map(h => Reader(ctx =>
-          List(commonHeaders(TcsEpics.instance, config, flatOrArcTcsSubsystems(inst).toList, sys)(ctx), gcalHeader(GcalEpics.instance, sys), gwsHeaders(GwsEpics.instance, sys), h)))
-
-    case DarkOrBias(_)      =>
-        calcInstHeader(config, sys).map(h => Reader(ctx => List(commonHeaders(TcsEpics.instance, config, Nil, sys)(ctx), gwsHeaders(GwsEpics.instance, sys), h)))
-    case AlignAndCalib         => TrySeq(Reader(_ => Nil)) // No headers for A&C
-    case st                    => TrySeq.fail(Unexpected(s"Unsupported step type $st"))
+  ): TrySeq[Reader[HeaderExtraData, List[Header[IO]]]] = {
+    stepType match {
+      case StepType.CelestialObject(_) | StepType.NodAndShuffle(_) =>
```
cquiroz: same as before. This lets us execute `N&S` steps

PR #1015, opened 2019-07-24
SEQNG-1025: Gpi Align And Calib

cquiroz:
This PR contains code with some adjustments to allow running align and calib with Gpi.

The main changes are
* Set a longer timeout for the specific command
Seqexec - GS-2018B-Q-0-115 2019-07-24 16-19-35

* Verify that we are not guiding before sending align and calib
Seqexec - GS-2018B-Q-0-116 2019-07-24 16-29-36

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GpiController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GpiController.scala
@@ -17,13 +17,16 @@ import edu.gemini.spModel.gemini.gpi.Gpi.{ Lyot => LegacyLyot }
 import edu.gemini.spModel.gemini.gpi.Gpi.{ ObservingMode => LegacyObservingMode }
 import edu.gemini.spModel.gemini.gpi.Gpi.{ PupilCamera => LegacyPupilCamera }
 import edu.gemini.spModel.gemini.gpi.Gpi.{ Shutter => LegacyShutter }
+import edu.gemini.aspen.giapi.commands.HandlerResponse.Response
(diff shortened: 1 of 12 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gpi/GpiController.scala
```diff
@@ -246,10 +255,41 @@ object GpiController extends GpiLookupTables with GpiConfigEq {
...
+          .map(StatusValue.intValue(_).map(x => GpiGuiding.fromInt(x.toInt)))
+          .ensure(new SeqexecFailure.Execution(s"Channel $GuidingChannel should be an int"))(
+            _.isDefined)
+          .map { _.orNull } // orNull lets us typecheck but it will never be used due to the `ensure` call above
+          .adaptError {
+            case GiapiException(message) => SeqexecFailure.Execution(message) }
+
+      override def alignAndCalib: F[Unit] =
+        guiding
+          .map(_ === GpiGuiding.NotGuiding)
```
cquiroz: I'm having second thoughts on this check. We have had the philosophy of requiring instruments to ensure we do operations safely rather than on the seqexec. I verified the instrument code and indeed this is checked at the lower level
@jluhrs thoughts?

jluhrs:
I don't think this change is necessary, because the instrument itself should complain and cause an execution error. But there are several small changes in this PR that I think should be applied anyways.

cquiroz:
Yes, I'd just remove the check but preserve the other changes. Note that the timeout change is a must

PR #1022, opened 2019-07-29
SEQNG-331 Further refactorings

cquiroz:
This is a last refactoring to remove the usage of monad transformers preferring `F[A]` over `EitherT[F, Throwable, A]`

This seems to simplify the code quite a bit at the expense of requiring care to handle all errors at the appropriate places

This was tested on instrument (GPI) and in simulation in several scenarios. The change is very transparent for regular operations but it is bigger to handle exceptional cases

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -95,13 +95,14 @@ class Engine[D, U](stateL: Engine.State[D]) {
     x.map(p =>
       modifyS(c.sid)(_.startSingle(c.actCoords)) *>
       Handle.fromStream[D, EventType](
-        p.attempt.collect {
-          case Right(r@Result.OK(_))    => singleRunCompleted(c, r)
-          case Right(e@Result.Error(_)) => singleRunFailed(c, e)
-          case Right(r)                 =>
-            singleRunFailed(c, Result.Error(s"Unhandled result for single run action: $r"))
-          case Left(r: Exception)       =>
-            singleRunFailed(c, Result.Error(s"Unhandled error for single run action: ${r.getMessage}"))
+        p.attempt.flatMap {
+          case Right(r @ Result.OK(_))    =>
+            Stream.eval(IO(singleRunCompleted(c, r)))
+          case Right(e @ Result.Error(_)) =>
+            Stream.eval(IO(singleRunFailed(c, e)))
+          case Right(r)                   =>
+            Stream.eval(IO(singleRunFailed(c, Result.Error(s"Unhandled result for single run action: $r"))))
+          case Left(t: Throwable)         => Stream.raiseError[IO](t)
         }
       ).as[Outcome](Outcome.Ok)
     ).getOrElse(pure[Outcome](Outcome.Failure))
@@ -177,13 +178,13 @@ class Engine[D, U](stateL: Engine.State[D]) {
   private def act(id: Observation.Id, stepId: StepId, t: (Stream[IO, Result[IO]], Int))
   : Stream[IO, EventType] = t match {
     case (gen, i) =>
-      gen.attempt.collect {
-        case Right(r@Result.OK(_))        => completed(id, stepId, i, r)
-        case Right(r@Result.OKStopped(_)) => stopCompleted(id, stepId, i, r)
-        case Right(r@Result.Partial(_))   => partial(id, stepId, i, r)
-        case Right(e@Result.Error(_))     => failed(id, i, e)
-        case Right(r@Result.Paused(_))    => paused[IO](id, i, r)
-        case Left(t: Exception)           => failed(id, i, Result.Error(t.getMessage))
+      gen.attempt.flatMap {
+        case Right(r@Result.OK(_))        => Stream.eval(IO(completed(id, stepId, i, r)))
+        case Right(r@Result.OKStopped(_)) => Stream.eval(IO(stopCompleted(id, stepId, i, r)))
+        case Right(r@Result.Partial(_))   => Stream.eval(IO(partial(id, stepId, i, r)))
+        case Right(e@Result.Error(_))     => Stream.eval(IO(failed(id, i, e)))
+        case Right(r@Result.Paused(_))    => Stream.eval(IO(paused[IO](id, i, r)))
+        case Left(t: Throwable)           => Stream.raiseError[IO](t)
       }
   }
 
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -7,6 +7,11 @@ import cats._
 import cats.data.StateT
 import cats.effect.{Concurrent, IO}
 import cats.implicits._
+import fs2.Stream
+import gem.Observation
+import monocle.Optional
+import mouse.boolean._
+import org.log4s.getLogger
 import seqexec.engine.Event._
 import seqexec.engine.EventResult.UserCommandResponse
 import seqexec.engine.EventResult.SystemUpdate
@@ -15,11 +20,6 @@ import seqexec.engine.Result.{PartialVal, RetVal}
 import seqexec.engine.SystemEvent._
 import seqexec.engine.UserEvent._
 import seqexec.model.{ClientId, SequenceState, StepId}
-import fs2.Stream
-import gem.Observation
-import monocle.Optional
-import mouse.boolean._
-import org.log4s.getLogger
 
 class Engine[D, U](stateL: Engine.State[D]) {
 
@@ -54,7 +54,7 @@ class Engine[D, U](stateL: Engine.State[D]) {
       case None      => unit
     }
 
-  /*
+  /**
    * startFrom starts a sequence from an arbitrary step. It does it by marking all previous steps to be skipped and then
    * modifying the state sequence as if it was run.
    * If the requested step is already run or marked to be skipped, the sequence will start from the next runnable step
@@ -95,11 +95,13 @@ class Engine[D, U](stateL: Engine.State[D]) {
     x.map(p =>
       modifyS(c.sid)(_.startSingle(c.actCoords)) *>
       Handle.fromStream[D, EventType](
-        p.map{
-          case r@Result.OK(_)    => singleRunCompleted(c, r)
-          case e@Result.Error(_) => singleRunFailed(c, e)
-          case r                 =>
+        p.attempt.collect {
+          case Right(r@Result.OK(_))    => singleRunCompleted(c, r)
+          case Right(e@Result.Error(_)) => singleRunFailed(c, e)
+          case Right(r)                 =>
             singleRunFailed(c, Result.Error(s"Unhandled result for single run action: $r"))
+          case Left(r: Exception)       =>
+            singleRunFailed(c, Result.Error(s"Unhandled error for single run action: ${r.getMessage}"))
         }
       ).as[Outcome](Outcome.Ok)
     ).getOrElse(pure[Outcome](Outcome.Failure))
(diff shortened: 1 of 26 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -95,11 +95,13 @@ class Engine[D, U](stateL: Engine.State[D]) {
     x.map(p =>
       modifyS(c.sid)(_.startSingle(c.actCoords)) *>
       Handle.fromStream[D, EventType](
-        p.map{
-          case r@Result.OK(_)    => singleRunCompleted(c, r)
-          case e@Result.Error(_) => singleRunFailed(c, e)
-          case r                 =>
+        p.attempt.map{
```
cquiroz: The `attempt` call is key to handle effect errors

modules/seqexec/server/src/main/scala/seqexec/server/System.scala
```diff
@@ -12,11 +12,11 @@ trait System[F[_]] {
   /**
     * Called to configure a system
     */
-  def configure(config: Config): SeqActionF[F, ConfigResult[F]]
+  def configure(config: Config): F[ConfigResult[F]]
```
cquiroz: This is a key change. It also means the caller is responsible for error handling

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -169,66 +165,13 @@ package object server {
     def fail[A](p:  SeqexecFailure): TrySeq[A] = Either.left(p)
   }
 
-  type SeqAction[A]        = EitherT[IO, SeqexecFailure, A]
```
cquiroz: Now we can remove all non local uses of Monad transformers

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -20,7 +21,7 @@ trait InstrumentSystem[F[_]] extends System[F] with InstrumentGuide {
   val observeControl: InstrumentSystem.ObserveControl[F]
 
   def observe(
-      config: Config): SeqObserveF[F, ImageFileId, ObserveCommandResult]
+      config: Config): Kleisli[F, ImageFileId, ObserveCommandResult]
```
cquiroz: This was suggested by @tpolecat as a good replacement for the previous construct

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -95,11 +95,13 @@ class Engine[D, U](stateL: Engine.State[D]) {
     x.map(p =>
       modifyS(c.sid)(_.startSingle(c.actCoords)) *>
       Handle.fromStream[D, EventType](
-        p.map{
-          case r@Result.OK(_)    => singleRunCompleted(c, r)
-          case e@Result.Error(_) => singleRunFailed(c, e)
-          case r                 =>
+        p.attempt.collect {
```
tpolecat: You're using `collect` here to discard `Left(_: Throwable)`?

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
```diff
@@ -87,27 +88,27 @@ object InstrumentControllerSim {
...
+    def applyConfig[C: Show](config: C): F[Unit] = F.delay {
       Log.info(s"Simulate applying $name configuration ${config.show}")
       Thread.sleep(ConfigurationDelay.toMilliseconds.toLong)
     }
 
-    def stopObserve: F[Unit] = Sync[F].delay {
+    def stopObserve: F[Unit] = F.delay {
       Log.info(s"Simulate stopping $name exposure")
-      Thread.sleep(1500)
+      Thread.sleep(StopObserveDelay.toMilliseconds.toLong)
```
tpolecat: Why not use your `Timer` instance here and elsewhere?

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -30,84 +30,86 @@ trait ObserveActions {
...
-      SeqexecFailure.Unexpected(ConfigUtilOps.explain(e))))
+    def dataId: F[String] =
+      EitherT.fromEither[F](
+        config.extractAs[String](OBSERVE_KEY / DATA_LABEL_PROP)
+          .leftMap(e => SeqexecFailure.Unexpected(ConfigUtilOps.explain(e)))
+      ).widenRethrowT
 
-    def notifyObserveStart: SeqActionF[F, Unit] =
+    def notifyObserveStart: F[Unit] =
       otherSys.map(_.notifyObserveStart).sequence.void
```
tpolecat: `otherSys.traverse_(_.notifyObserveStart)` here and elsewhere. I know it's not what you're working on but can we simplify it anyway?

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -30,84 +30,86 @@ trait ObserveActions {
...
+      inst.keywordsClient.closeImage(id)
 
-    def doObserve(fileId: ImageFileId): SeqActionF[F, Result[F]] =
+    def doObserve(fileId: ImageFileId): F[Result[F]] =
       for {
         d   <- dataId
         _   <- sendDataStart(systems, obsId, fileId, d)
         _   <- notifyObserveStart
-        _   <- SeqActionF.embedF(headers(ctx).map(_.sendBefore(obsId, fileId)).sequence)
+        _   <- headers(ctx).map(_.sendBefore(obsId, fileId)).sequence
```
tpolecat: traverse

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -30,84 +30,86 @@ trait ObserveActions {
...
         ret <- observeTail(fileId, d)(r)
       } yield ret
 
-    def observeTail(id: ImageFileId, dataId: String)(r: ObserveCommandResult): SeqActionF[F, Result[F]] = {
-      def okTail(stopped: Boolean): SeqActionF[F, Result[F]] = for {
+    def observeTail(id: ImageFileId, dataId: String)(r: ObserveCommandResult): F[Result[F]] = {
+      def okTail(stopped: Boolean): F[Result[F]] = for {
         _ <- notifyObserveEnd
-        _ <- SeqActionF.embedF(headers(ctx).reverseMap(_.sendAfter(id)).sequence.void)
+        _ <- headers(ctx).reverseMap(_.sendAfter(id)).sequence.void
```
tpolecat: What is `reverseMap`?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -167,16 +167,17 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
+    case Left(e: Throwable) => Event.logErrorMsg(SeqexecFailure.explain(SeqexecFailure.SeqexecException(e)))
     case _       => Event.nullEvent
   }
 
   def stopObserve(seqId: Observation.Id)(
     implicit cio: Concurrent[IO],
              tio: Timer[IO]
   ): EngineState => Option[Stream[IO, executeEngine.EventType]] = st =>{
-    def f(oc: ObserveControl[IO]): Option[SeqAction[Unit]] = oc match {
+    def f(oc: ObserveControl[IO]): Option[IO[Unit]] = oc match {
```
tpolecat: This is a weird type. Seems like you want `IO[Unit]` here where the empty case is `IO()`.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -150,19 +151,23 @@ abstract class Gmos[F[_]: Sync, T<:GmosController.SiteDependentTypes](controller
       SequenceConfiguration.calcStepType(config)
     }
 
-  override def observe(config: Config): SeqObserveF[F, ImageFileId, ObserveCommandResult] =
-    Reader { fileId =>
-      SeqActionF.liftF(calcObserveTime(config)).flatMap { x =>
-        SeqActionF.embedF(controller.observe(fileId, x))
+  override def observe(config: Config): Kleisli[F, ImageFileId, ObserveCommandResult] =
+    Kleisli { fileId =>
```
tpolecat: In the future we might want to absorb the `ImageFileId` into the effect and demand `G[_]: ApplicativeAsk[?[_], ImageFileId]`, and then our return type here could be `G[ObserveCommandResult]`. At the boundary with this code you would instantiate with `Kleisli[F` and run it with the image id, yielding a plain `F` again. Just a thought for the future, this is fine for now.

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
cquiroz: Yes, do you think we should catch `Throwable` too?

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
cquiroz: Yeah I need to refactor this. I'll do it in another PR, this is a known issue but I want to tackl this as part of the N&S simulation support part

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
cquiroz: 👍 A question, why `traverse_` rather than `traverse`?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
cquiroz: I'll review this. In cases we need to have `Option[IO[Unit]]` as we need to know if the `IO` does nothing. OTOH I suppose there is no harm in jut running `IO.unit`

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
cquiroz: It is part of the std library. Like `map` but iterates in reverse

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
jluhrs: Discarding `Throwable`s makes me uneasy.

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
```diff
@@ -204,30 +204,6 @@ class packageSpec extends FlatSpec with NonImplicitAssertions {
     assert(result.isDefined)
   }
 
-  "engine" should "not capture runtime exceptions." in {
```
jluhrs: This test is important, to ensure we are not discarding relevant errors.

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentControllerSim.scala
```diff
@@ -45,7 +45,7 @@ sealed trait InstrumentControllerSim[F[_]] {
 }
 
 object InstrumentControllerSim {
-  private final class InstrumentControllerSimImpl[F[_]: Sync: Timer](name: String, useTimeout: Boolean) extends InstrumentControllerSim[F] {
+  private final class InstrumentControllerSimImpl[F[_]](name: String, useTimeout: Boolean)(implicit val F: Sync[F], T: Timer[F]) extends InstrumentControllerSim[F] {
```
jluhrs: What is the advantage of this style over the previous version ?

cquiroz: It is a small thing and I'm not consistent but rather than doing `Sync[F].delay` I can just do `F.delay`

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
cquiroz: There is a reason for this. The engine needs to do `F[A].attempt` to capture regular errors. Almost by definition that will capture runtime exceptions. If I recall this test was about not capturing things like OOM, perhaps we should change it that way?

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
tpolecat: I think we might want `case Left(NonFatal(e)) => ...` which will catch Throwables other than ones that mean you really do want the application to crash.

modules/seqexec/engine/src/test/scala/seqexec/engine/packageSpec.scala
cquiroz: I mean change the test to not capture e.g OOM

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
cquiroz: Yeah that sounds reasonable. Using `collect` makes me a bit uneasy but `map` would need to do something with e.g. `OOM`

PR #1023, opened 2019-07-30
SEQNG-1032 Added Gems control channels in TCS Epics code.

jluhrs:
Added several EPICS commands and statuses in TCS to control GeMS.


cquiroz:
Is `BinaryEnabledDisabled` a new class? I don't see it

PR #1028, opened 2019-08-05
SEQNG-1028 Apply GeMS guiders tracking configuration.

jluhrs:
This PR is limited to applying the tracking configuration of GeMS guiders (nod/chop and follow state). There is a lot left to do, like the notifications to GeMS of guide changes, and starting/stopping the guiders.
One small complication is that all the settings go through TCS. But in most cases it does not have a dedicated CAR for every possible GeMS guider. Instead, because only 4 can be used at a time, TCS defines 4 virtual telescopes for GeMS (G1 to G4), and a mapping to the actual guiders.

### DIFF
--- a/modules/seqexec/server/src/main/resources/Tcs.xml
+++ b/modules/seqexec/server/src/main/resources/Tcs.xml
@@ -209,11 +209,6 @@
         </command>
         <command name="odgw3Park">
             <description>ODGW 3 park</description>
-            <parameter name="command">
-                <channel>odgw3Park.A</channel>
-                <type>STRING</type>
-                <description>Park command</description>
-            </parameter>
         </command>
         <command name="ngsPr1Park">
             <record>ngsPr1Park</record>
@@ -328,11 +323,6 @@
         </command>
         <command name="odgw1Park">
             <description>ODGW 1 park</description>
-            <parameter name="command">
-                <channel>odgw1Park.A</channel>
-                <type>STRING</type>
-                <description>Park command</description>
-            </parameter>
         </command>
         <command name="pwfs2Park">
             <record>pwfs2Park</record>
@@ -465,11 +455,6 @@
         </command>
         <command name="odgw4Park">
             <description>ODGW 4 park</description>
-            <parameter name="command">
-                <channel>odgw4Park.A</channel>
-                <type>STRING</type>
-                <description>Park command</description>
-            </parameter>
         </command>
         <command name="pwfs1StopObserve">
             <record>pwfs1StopObserve</record>
@@ -947,11 +932,6 @@
         </command>
         <command name="odgw2Park">
             <description>ODGW 2 park</description>
-            <parameter name="command">
-                <channel>odgw2Park.A</channel>
-                <type>STRING</type>
-                <description>Park command</description>
-            </parameter>
         </command>
         <command name="hrwfsPark">
             <record>hrwfsPark</record>
--- a/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
@@ -17,7 +17,7 @@ import edu.gemini.spModel.gemini.altair.AltairParams.FieldLens
 import monocle.macros.Lenses
 import seqexec.server.{SeqexecFailure, TrySeq}
 import seqexec.server.altair.AltairController._
-import seqexec.server.tcs.{FOCAL_PLANE_SCALE, Gaos, TcsEpics}
+import seqexec.server.tcs.{FOCAL_PLANE_SCALE, TcsEpics}
 import seqexec.server.tcs.Gaos._
 import seqexec.server.tcs.TcsController.FocalPlaneOffset
 import squants.{Length, Time}
@@ -73,14 +73,14 @@ object AltairControllerEpics extends AltairController[IO] {
   implicit val fieldLensEq: Eq[FieldLens] = Eq.by(_.ordinal)
 
   private def pauseNgsOrLgsMode(starPos: (Length, Length), fieldLens: FieldLens, currCfg: EpicsAltairConfig)(
-    reasons: Set[Gaos.PauseCondition])
+    reasons: Set[PauseCondition])
   : Option[(EpicsAltairConfig, IO[Unit])] = {
-    val offsets = reasons.collectFirst { case OffsetMove(p, n) => (p, n) }
+    val offsets = reasons.collectFirst { case PauseCondition.OffsetMove(p, n) => (p, n) }
     val newPos = offsets.map(u => newPosition(starPos)(u._2))
     val newPosOk = newPos.forall(newPosInRange)
     val matrixOk = newPos.forall(validateCurrentControlMatrix(currCfg, _)) || fieldLens === FieldLens.IN
     val prepMatrixOk = newPos.forall(validatePreparedControlMatrix(currCfg, _)) || fieldLens === FieldLens.IN
-    val guideOk = !reasons.contains(GaosGuideOff) //It can follow the guide star on this step
+    val guideOk = !reasons.contains(PauseCondition.GaosGuideOff) //It can follow the guide star on this step
 
     val needsToStop = !(newPosOk && matrixOk && guideOk)
 
@@ -103,7 +103,7 @@ object AltairControllerEpics extends AltairController[IO] {
   }
 
   private def pauseResumeNgsMode(starPos: (Length, Length), fieldLens: FieldLens, currCfg: EpicsAltairConfig)(
-    pauseReasons: Set[Gaos.PauseCondition], resumeReasons: Set[Gaos.ResumeCondition])
+    pauseReasons: Set[PauseCondition], resumeReasons: Set[ResumeCondition])
   : PauseResume[IO] = {
     val pause = pauseNgsOrLgsMode(starPos, fieldLens, currCfg)(pauseReasons)
     val resume = resumeNgsOrLgsMode(starPos, pause.map(_._1).getOrElse(currCfg))(resumeReasons)
@@ -115,10 +115,10 @@ object AltairControllerEpics extends AltairController[IO] {
   val matrixPrepTimeout: Time = 10.seconds
 
   private def resumeNgsOrLgsMode(starPos: (Length, Length), currCfg: EpicsAltairConfig)(
-    reasons: Set[Gaos.ResumeCondition]): Option[IO[Unit]] = {
-    val offsets = reasons.collectFirst { case OffsetReached(o) => o }
+    reasons: Set[ResumeCondition]): Option[IO[Unit]] = {
+    val offsets = reasons.collectFirst { case ResumeCondition.OffsetReached(o) => o }
     val newPosOk = offsets.forall(v => newPosInRange(newPosition(starPos)(v)))
-    val guideOk = reasons.contains(GaosGuideOn)
+    val guideOk = reasons.contains(ResumeCondition.GaosGuideOn)
 
     (newPosOk && guideOk).option(
       (epicsAltair.waitMatrixCalc(CarStateGEM5.IDLE, matrixPrepTimeout) *>
@@ -196,7 +196,7 @@ object AltairControllerEpics extends AltairController[IO] {
 
   private def pauseResumeLgsMode(strap: Boolean, sfo: Boolean, starPos: (Length, Length), fieldLens: FieldLens,
                            currCfg: EpicsAltairConfig)(
-    pauseReasons: Set[Gaos.PauseCondition], resumeReasons: Set[Gaos.ResumeCondition]): PauseResume[IO] = {
+    pauseReasons: Set[PauseCondition], resumeReasons: Set[ResumeCondition]): PauseResume[IO] = {
     val pause = pauseLgsMode(strap, sfo, starPos, fieldLens, currCfg)(pauseReasons)
     val resume = resumeLgsMode(strap, sfo, starPos)(pause.map(_._1).getOrElse(currCfg))(resumeReasons)
 
@@ -204,28 +204,28 @@ object AltairControllerEpics extends AltairController[IO] {
   }
 
   private def pauseLgsMode(strap: Boolean, sfo: Boolean, starPos: (Length, Length), fieldLens: FieldLens,
-                           currCfg: EpicsAltairConfig)(reasons: Set[Gaos.PauseCondition])
+                           currCfg: EpicsAltairConfig)(reasons: Set[PauseCondition])
   : Option[(EpicsAltairConfig, IO[Unit])] =
     pauseNgsOrLgsMode(starPos, fieldLens, currCfg)(reasons).filter(_ => strap || sfo)
       .map(_.bimap(ttgsOffEndo, ttgsOff(currCfg) *> _))
 
   private def resumeLgsMode(strap: Boolean, sfo: Boolean, starPos: (Length, Length))(currCfg: EpicsAltairConfig)(
-    reasons: Set[Gaos.ResumeCondition]): Option[IO[Unit]] =
+    reasons: Set[ResumeCondition]): Option[IO[Unit]] =
     resumeNgsOrLgsMode(starPos, currCfg)(reasons).filter(_ => strap || sfo).map(_ *> ttgsOn(strap, sfo, currCfg))
 
   /**
    * Modes LgsWithP1 and LgsWithOi don't use an Altair target. The only action required is to start or stop corrections
    */
   private def pauseResumeLgsWithXX(currCfg: EpicsAltairConfig)(
-      pauseReasons: Set[Gaos.PauseCondition], resumeReasons: Set[Gaos.ResumeCondition])
+      pauseReasons: Set[PauseCondition], resumeReasons: Set[ResumeCondition])
   : PauseResume[IO] = {
-    val pause: Option[IO[Unit]] = (pauseReasons.contains(Gaos.GaosGuideOff) && currCfg.aoLoop).option{
+    val pause: Option[IO[Unit]] = (pauseReasons.contains(PauseCondition.GaosGuideOff) && currCfg.aoLoop).option{
       epicsTcs.aoCorrect.setCorrections(CorrectionsOff) *>
         epicsTcs.targetFilter.setShortCircuit(TargetFilterClosed) *>
         epicsTcs.targetFilter.post[IO].void
     }
-    val resume: Option[IO[Unit]] = (resumeReasons.contains(Gaos.GaosGuideOn) &&
-      (pauseReasons.contains(Gaos.GaosGuideOff) || !currCfg.aoLoop)
+    val resume: Option[IO[Unit]] = (resumeReasons.contains(ResumeCondition.GaosGuideOn) &&
+      (pauseReasons.contains(PauseCondition.GaosGuideOff) || !currCfg.aoLoop)
     ).option{
       epicsTcs.aoCorrect.setCorrections(CorrectionsOn) *>
         epicsTcs.aoFlatten.mark[IO] *>
--- a/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
@@ -73,10 +73,10 @@ object AltairControllerEpics extends AltairController[IO] {
   implicit val fieldLensEq: Eq[FieldLens] = Eq.by(_.ordinal)
 
   private def pauseNgsOrLgsMode(starPos: (Length, Length), fieldLens: FieldLens, currCfg: EpicsAltairConfig)(
-    reasons: Set[PauseCondition])
+    reasons: PauseConditionSet)
   : Option[(EpicsAltairConfig, IO[Unit])] = {
-    val offsets = reasons.collectFirst { case PauseCondition.OffsetMove(p, n) => (p, n) }
-    val newPos = offsets.map(u => newPosition(starPos)(u._2))
+    val newOffset = reasons.offsetO.map(_.newOffset)
+    val newPos = newOffset.map(newPosition(starPos))
     val newPosOk = newPos.forall(newPosInRange)
     val matrixOk = newPos.forall(validateCurrentControlMatrix(currCfg, _)) || fieldLens === FieldLens.IN
     val prepMatrixOk = newPos.forall(validatePreparedControlMatrix(currCfg, _)) || fieldLens === FieldLens.IN
@@ -103,7 +103,7 @@ object AltairControllerEpics extends AltairController[IO] {
   }
 
   private def pauseResumeNgsMode(starPos: (Length, Length), fieldLens: FieldLens, currCfg: EpicsAltairConfig)(
-    pauseReasons: Set[PauseCondition], resumeReasons: Set[ResumeCondition])
+    pauseReasons: PauseConditionSet, resumeReasons: ResumeConditionSet)
   : PauseResume[IO] = {
     val pause = pauseNgsOrLgsMode(starPos, fieldLens, currCfg)(pauseReasons)
     val resume = resumeNgsOrLgsMode(starPos, pause.map(_._1).getOrElse(currCfg))(resumeReasons)
@@ -115,8 +115,8 @@ object AltairControllerEpics extends AltairController[IO] {
   val matrixPrepTimeout: Time = 10.seconds
 
   private def resumeNgsOrLgsMode(starPos: (Length, Length), currCfg: EpicsAltairConfig)(
-    reasons: Set[ResumeCondition]): Option[IO[Unit]] = {
-    val offsets = reasons.collectFirst { case ResumeCondition.OffsetReached(o) => o }
+    reasons: ResumeConditionSet): Option[IO[Unit]] = {
+    val offsets = reasons.offsetO.map(_.newOffset)
     val newPosOk = offsets.forall(v => newPosInRange(newPosition(starPos)(v)))
     val guideOk = reasons.contains(ResumeCondition.GaosGuideOn)
 
@@ -196,7 +196,7 @@ object AltairControllerEpics extends AltairController[IO] {
 
   private def pauseResumeLgsMode(strap: Boolean, sfo: Boolean, starPos: (Length, Length), fieldLens: FieldLens,
                            currCfg: EpicsAltairConfig)(
-    pauseReasons: Set[PauseCondition], resumeReasons: Set[ResumeCondition]): PauseResume[IO] = {
+    pauseReasons: PauseConditionSet, resumeReasons: ResumeConditionSet): PauseResume[IO] = {
     val pause = pauseLgsMode(strap, sfo, starPos, fieldLens, currCfg)(pauseReasons)
     val resume = resumeLgsMode(strap, sfo, starPos)(pause.map(_._1).getOrElse(currCfg))(resumeReasons)
 
@@ -204,20 +204,20 @@ object AltairControllerEpics extends AltairController[IO] {
   }
 
   private def pauseLgsMode(strap: Boolean, sfo: Boolean, starPos: (Length, Length), fieldLens: FieldLens,
-                           currCfg: EpicsAltairConfig)(reasons: Set[PauseCondition])
+                           currCfg: EpicsAltairConfig)(reasons: PauseConditionSet)
   : Option[(EpicsAltairConfig, IO[Unit])] =
     pauseNgsOrLgsMode(starPos, fieldLens, currCfg)(reasons).filter(_ => strap || sfo)
       .map(_.bimap(ttgsOffEndo, ttgsOff(currCfg) *> _))
 
   private def resumeLgsMode(strap: Boolean, sfo: Boolean, starPos: (Length, Length))(currCfg: EpicsAltairConfig)(
-    reasons: Set[ResumeCondition]): Option[IO[Unit]] =
+    reasons: ResumeConditionSet): Option[IO[Unit]] =
     resumeNgsOrLgsMode(starPos, currCfg)(reasons).filter(_ => strap || sfo).map(_ *> ttgsOn(strap, sfo, currCfg))
 
   /**
    * Modes LgsWithP1 and LgsWithOi don't use an Altair target. The only action required is to start or stop corrections
    */
   private def pauseResumeLgsWithXX(currCfg: EpicsAltairConfig)(
-      pauseReasons: Set[PauseCondition], resumeReasons: Set[ResumeCondition])
+      pauseReasons: PauseConditionSet, resumeReasons: ResumeConditionSet)
   : PauseResume[IO] = {
     val pause: Option[IO[Unit]] = (pauseReasons.contains(PauseCondition.GaosGuideOff) && currCfg.aoLoop).option{
       epicsTcs.aoCorrect.setCorrections(CorrectionsOff) *>
@@ -247,7 +247,7 @@ object AltairControllerEpics extends AltairController[IO] {
     None
   )
 
-  override def pauseResume(pauseReasons: Set[PauseCondition], resumeReasons: Set[ResumeCondition],
+  override def pauseResume(pauseReasons: PauseConditionSet, resumeReasons: ResumeConditionSet,
                            fieldLens: FieldLens)(cfg: AltairConfig): IO[PauseResume[IO]] =
     retrieveConfig.map{ currCfg =>
       cfg match {
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
@@ -8,14 +8,14 @@ import seqexec.server.altair.AltairController.AltairConfig
 import seqexec.server.gems.Gems.GemsGuiderStatus
 import seqexec.server.gems.GemsController.{GemsConfig, GemsOff}
 import seqexec.server.tcs.Gaos
-import seqexec.server.tcs.Gaos.{PauseResume, ResumeCondition}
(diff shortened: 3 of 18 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/GuideConfigDb.scala
```diff
@@ -94,7 +94,9 @@ object GuideConfigDb {
             odgw2 <- c.downField("odgw2On").as[Boolean]
             odgw3 <- c.downField("odgw3On").as[Boolean]
             odgw4 <- c.downField("odgw4On").as[Boolean]
-          } yield GemsOn(ttgs1, ttgs2, ttgs3, odgw1, odgw2, odgw3, odgw4)
+            useOI <- c.downField("useOI").as[Boolean].recover{case _ => false}
+            useP1 <- c.downField("useP1").as[Boolean].recover{case _ => false}
```
jluhrs: Used `recover` because TCC is not sending these parameters yet.

modules/seqexec/server/src/main/resources/Tcs.xml
```diff
@@ -209,11 +209,6 @@
         </command>
         <command name="odgw3Park">
             <description>ODGW 3 park</description>
-            <parameter name="command">
-                <channel>odgw3Park.A</channel>
-                <type>STRING</type>
-                <description>Park command</description>
-            </parameter>
```
jluhrs: This parameter comes from the old Seqexec, where they were used to mark the CAD record. The new Seqexec can mark the record using the DIR input, so this parameter is no longer necessary.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
```diff
@@ -73,14 +73,14 @@ object AltairControllerEpics extends AltairController[IO] {
   implicit val fieldLensEq: Eq[FieldLens] = Eq.by(_.ordinal)
 
   private def pauseNgsOrLgsMode(starPos: (Length, Length), fieldLens: FieldLens, currCfg: EpicsAltairConfig)(
-    reasons: Set[Gaos.PauseCondition])
+    reasons: Set[PauseCondition])
   : Option[(EpicsAltairConfig, IO[Unit])] = {
-    val offsets = reasons.collectFirst { case OffsetMove(p, n) => (p, n) }
+    val offsets = reasons.collectFirst { case PauseCondition.OffsetMove(p, n) => (p, n) }
```
jluhrs: PauseConditionn cases were moved to the companion object.

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
```diff
@@ -31,12 +38,35 @@ object Gems {
...
+
+  final case class GemsGuiderStatus[F[_]](
+    ngs1: F[Option[Boolean]],
+    ngs2: F[Option[Boolean]],
+    ngs3: F[Option[Boolean]],
+    odgw1: F[Option[Boolean]],
+    odgw2: F[Option[Boolean]],
+    odgw3: F[Option[Boolean]],
+    odgw4: F[Option[Boolean]]
+  )
```
jluhrs: This class is used to retrieve the GeMS guiders active state (i.e. if they are producing error measurements or not). They were put in a case class so the implementation can be switched between the simulated and real version.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -3,17 +3,20 @@
...
+import seqexec.server.tcs.Gaos.{PauseCondition, PauseResume, ResumeCondition}
 import squants.Time
 
 trait GemsController[F[_]] {
   import GemsController._
 
-  def pause(reasons: Set[PauseCondition]): F[Unit]
-  def resume(reasons: Set[ResumeCondition], config: GemsConfig): F[Unit]
+  def pauseResume(config: GemsConfig, pauseReasons: Set[PauseCondition],
+                  resumeReasons: Set[ResumeCondition]): F[PauseResume[F]]
```
jluhrs: This follows the implementation of Altair. The pause and resume procedures are calculated in the same function. In this way, the resume can depend on what is done in the pause.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -29,8 +32,9 @@ object GemsController {
     odgw1: Boolean,
     odgw2: Boolean,
     odgw3: Boolean,
-    odgw4: Boolean
+    odgw4: Boolean,
+    useOI: Boolean,
+    useP1: Boolean
```
jluhrs: GeMS was designed to use F2 OI, but never commissioned. Maybe it should be removed in the future. GeMS with the new NGS2 will use P1 for focus correction.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -30,43 +30,53 @@ trait Gaos[F[_]] {
...
+    // Telescope offset will be changed
+    final case class OffsetMove(previousOffset: FocalPlaneOffset, newOffset: FocalPlaneOffset) extends PauseCondition
+    // OI will be turn off
+    case object OiOff extends PauseCondition
+    // PI will be turn off
+    case object P1Off extends PauseCondition
+    // Unguided step
+    case object GaosGuideOff extends PauseCondition
+    // Instrument config (affects ODGW)
+    case object InstConfigMove extends PauseCondition
```
jluhrs: Moved to companion object.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -30,43 +30,53 @@ trait Gaos[F[_]] {
...
+    // Telescope offset will be changed
+    final case class OffsetReached(newOffset: FocalPlaneOffset) extends ResumeCondition
+    // OI will be turn off
+    case object OiOn extends ResumeCondition
+    // PI will be turn off
+    case object P1On extends ResumeCondition
+    // Guided step
+    case object GaosGuideOn extends ResumeCondition
+    // Instrument config (affects ODGW)
+    case object InstConfigCompleted extends ResumeCondition
```
jluhrs: Moved to companion object.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsConfigRetriever.scala
```diff
@@ -192,6 +195,75 @@ object TcsConfigRetriever {
     getStatusVal(Nested(TcsEpics.instance.sourceAWavelength).map(v => Wavelength(Angstroms(v))).value,
       "central wavelength")
 
+  private def getGemsMap: IO[Map[GemsSource, VirtualGemsTelescope]] = (for {
```
jluhrs: This function retrieves the GeMS vitual telescope to Gems guider map from TCS.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsConfigRetriever.scala
```diff
@@ -192,6 +195,75 @@ object TcsConfigRetriever {
...
+
+  private def getOdgw4(g: VirtualGemsTelescope, active: IO[Option[Boolean]])
+  : IO[GuiderConfig] = for {
+    prk <- getStatusVal(TcsEpics.instance.odgw4Parked, "ODGW4 parked state")
+    trk <- getStatusVal(getNodChopTrackingConfig(TcsEpics.instance.gemsGuideConfig(g)), "ODGW4 tracking configuration")
+    fol <- getStatusVal(Nested(TcsEpics.instance.odgw4Follow)
+      .map{ if(_) FollowOption.FollowOn else FollowOption.FollowOff}.value, "ODGW4 follow state")
+    wfs <- getStatusVal(Nested(active).map{if(_) GuiderSensorOn else GuiderSensorOff}.value, "ODGW4 Active status")
+  } yield GuiderConfig(prk.fold(ProbeTrackingConfig.Parked, calcProbeTrackingConfig(fol, trk)), wfs)
+
```
jluhrs: I could definitely factor out some code in the getNgsX and getOdgwX functions

swalker2m:
To be clear, by "GeMS guiders" you mean the GSAOI ODGWs and the Canopus WFSs?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsConfigRetriever.scala
swalker2m: Agree. Is there any reason not to do that now?  It looks like you just need 3 parameters:
* parked
* follow
* guider name

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsConfigRetriever.scala
```diff
@@ -218,6 +290,36 @@ object TcsConfigRetriever {
...
+      ngs3    <- mapping.get(Ttgs3).map(getNgs3(_, gemsSt.ngs3))
+        .getOrElse(IO(GuiderConfig(ProbeTrackingConfig.Off, GuiderSensorOff)))
+      odgw1   <- mapping.get(Odgw1).map(getOdgw1(_, gemsSt.odgw1))
+        .getOrElse(IO(GuiderConfig(ProbeTrackingConfig.Off, GuiderSensorOff)))
+      odgw2   <- mapping.get(Odgw2).map(getOdgw2(_, gemsSt.odgw2))
+        .getOrElse(IO(GuiderConfig(ProbeTrackingConfig.Off, GuiderSensorOff)))
+      odgw3   <- mapping.get(Odgw3).map(getOdgw3(_, gemsSt.odgw3))
+        .getOrElse(IO(GuiderConfig(ProbeTrackingConfig.Off, GuiderSensorOff)))
+      odgw4   <- mapping.get(Odgw4).map(getOdgw4(_, gemsSt.odgw4))
+        .getOrElse(IO(GuiderConfig(ProbeTrackingConfig.Off, GuiderSensorOff)))
```
swalker2m: Maybe all of this should be factored out as well.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsConfigRetriever.scala
```diff
@@ -192,6 +195,75 @@ object TcsConfigRetriever {
...
+      .map{ if(_) FollowOption.FollowOn else FollowOption.FollowOff}.value, "NGS1 follow state")
+    wfs <- getStatusVal(Nested(active).map{if(_) GuiderSensorOn else GuiderSensorOff}.value, "NGS1 Active status")
+  } yield GuiderConfig(calcProbeTrackingConfig(fol, trk), wfs)
+
+  private def getNgs2(g: VirtualGemsTelescope, active: IO[Option[Boolean]]): IO[GuiderConfig] = for {
+    trk <- getStatusVal(getNodChopTrackingConfig(TcsEpics.instance.gemsGuideConfig(g)), "NGS2 tracking configuration")
+    fol <- getStatusVal(Nested(TcsEpics.instance.ngs2Follow)
+      .map{ if(_) FollowOption.FollowOn else FollowOption.FollowOff}.value, "NGS2 follow state")
+    wfs <- getStatusVal(Nested(active).map{if(_) GuiderSensorOn else GuiderSensorOff}.value, "NGS2 Active status")
+  } yield GuiderConfig(calcProbeTrackingConfig(fol, trk), wfs)
```
tpolecat: Looks like a lot of copy/paste in this file. Is there anything you can abstract out?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsNorthControllerEpicsAo.scala
```diff
@@ -92,7 +92,7 @@ object TcsNorthControllerEpicsAo {
       )),
       setScienceFold(EpicsTcsAoConfig.base)(subsystems, current, tcs.agc.sfPos),
       setHrPickup(EpicsTcsAoConfig.base)(subsystems, current, tcs.agc)
-    ).collect{ case Some(x) => x }
+    ).mapFilter(identity)
```
swalker2m: `.flattenOption`

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSouthControllerEpicsAo.scala
```diff
@@ -0,0 +1,182 @@
...
+      ).mapFilter(identity)
+
+      actions.nonEmpty.option { x =>
+        actions.sequence *>
+          IO((EpicsTcsAoConfig.cwfs1 ^|-> GuiderConfig.tracking).set(demand)(x))
+      }
+    }
+    else none
+
+  def setNgs2Guide(g: VirtualGemsTelescope,
```
tpolecat: Same here, can we factor out any common logic in this file?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSouthControllerEpicsAo.scala
```diff
@@ -0,0 +1,182 @@
...
+import seqexec.server.tcs.TcsController.{GuiderConfig, ProbeTrackingConfig, Subsystem, wavelengthEq}
+import seqexec.server.tcs.TcsControllerEpicsCommon.{BaseEpicsTcsConfig, GuideControl, agTimeout, applyParam, encodeFollowOption, setGuideProbe, setHrPickup, setNodChopProbeTrackingConfig, setOiwfsProbe, setPwfs1Probe, setScienceFold, setTelescopeOffset, setWavelength, tcsTimeout}
+import seqexec.server.tcs.TcsEpics.VirtualGemsTelescope
+import seqexec.server.tcs.TcsSouthController.{GemsGuiders, TcsSouthAoConfig}
+
+object TcsSouthControllerEpicsAo {
+
+  val Log: Logger = getLogger
+
+  def setNgs1Guide(g: VirtualGemsTelescope,
```
swalker2m: There is quite a bit of repeated code here. Would it be difficult to factor this out?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSouthControllerEpicsAo.scala
```diff
@@ -0,0 +1,182 @@
...
+  def setGemsProbes(subsystems: NonEmptySet[Subsystem], current: EpicsTcsAoConfig, demand: GemsGuiders)
+  : List[EpicsTcsAoConfig => IO[EpicsTcsAoConfig]] = List(
+      current.mapping.get(Ttgs1).flatMap(setNgs1Guide(_, subsystems, current.cwfs1.tracking, demand.ngs1.tracking)),
+      current.mapping.get(Ttgs2).flatMap(setNgs1Guide(_, subsystems, current.cwfs2.tracking, demand.ngs2.tracking)),
+      current.mapping.get(Ttgs3).flatMap(setNgs1Guide(_, subsystems, current.cwfs3.tracking, demand.ngs3.tracking)),
+      current.mapping.get(Odgw1).flatMap(setOdgw1Probe(_)(subsystems, current.odgw1.tracking, demand.odgw1.tracking)),
+      current.mapping.get(Odgw2).flatMap(setOdgw1Probe(_)(subsystems, current.odgw2.tracking, demand.odgw2.tracking)),
+      current.mapping.get(Odgw3).flatMap(setOdgw1Probe(_)(subsystems, current.odgw3.tracking, demand.odgw3.tracking)),
+      current.mapping.get(Odgw4).flatMap(setOdgw1Probe(_)(subsystems, current.odgw4.tracking, demand.odgw4.tracking))
+    ).mapFilter(identity)
```
swalker2m: `.flattenOption`

modules/seqexec/server/src/test/scala/seqexec/server/tcs/GuideConfigDbSpec.scala
```diff
@@ -117,7 +117,7 @@ final class GuideConfigDbSpec extends FlatSpec {
       M1GuideConfig.M1GuideOn(M1Source.GAOS),
       M2GuideConfig.M2GuideOn(ComaOption.ComaOn, Set(TipTiltSource.GAOS))
     ),
-    Some(Right(GemsOn(true, false, false, true, false, true, true)))
+    Some(Right(GemsOn(true, false, false, true, false, true, true, false, false)))
```
swalker2m: I hate to say it, but I think it would better to have explicit types for all these `Boolean`s.  This class takes 9 `Boolean` parameters so it would be easy to get them in the wrong order and not realize it.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
jluhrs: "That won't happen" :)

jluhrs: Now seriously, it won't happen because there is one offset per step. I see your point, though I am not sure how to tighten the data type.

modules/seqexec/server/src/test/scala/seqexec/server/tcs/GuideConfigDbSpec.scala
jluhrs: I. Hate. You.

jluhrs:
@swalker2m Yes, by "GeMS guiders" I mean the Canopus (soon to be NGS2) WFS and ODGWs.

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
```diff
@@ -15,8 +16,14 @@ trait Gems[F[_]] extends Gaos[F] {
 
   def pauseResume(config: GemsConfig, pauseReasons: Set[Gaos.PauseCondition],
                   resumeReasons: Set[ResumeCondition]): F[PauseResume[F]]
-}
 
+  def usesP1(guide: GemsConfig): Boolean
+
+  def usesOI(guide: GemsConfig): Boolean
```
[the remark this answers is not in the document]
jluhrs: I think not. The values read through `GemsConfig` are the current guiders state as read from the external systems. The functions `usesP1` and `usesOI` give information about the GeMS configuration we are trying to apply. I'll rename `GemsConfig` to make it more clear.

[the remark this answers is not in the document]
jluhrs: You are right, I'll do it.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
swalker2m: I'm not sure of the context, but could it be that you need an `Option[OffsetMove]` and a `Set[OtherPauseConditions]`.  By `OtherPauseConditions` I mean something that includes all the `PauseCondition`s that you currently have but not `OffsetMove`.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -30,43 +30,53 @@ trait Gaos[F[_]] {
...
-  final case class OffsetReached(newOffset: FocalPlaneOffset) extends ResumeCondition
-  // OI will be turn off
-  case object OiOn extends ResumeCondition
-  // PI will be turn off
-  case object P1On extends ResumeCondition
-  // Guided step
-  case object GaosGuideOn extends ResumeCondition
-
-  implicit val becauseOffsetMoveEq: Eq[OffsetMove] = Eq.by(x => (x.previousOffset, x.newOffset))
+    implicit val becauseOffsetMoveEq: Eq[OffsetMove] = Eq.by(x => (x.previousOffset, x.newOffset))
```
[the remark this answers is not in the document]
jluhrs: The conditions had longer names before. I shortened them, but forgot to change the `Eq` instance name.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
jluhrs: I could create a specialized `PauseConditions` to use instead of `Set[PauseConditions]`. Something like `case class PauseConditions(offsetO: Option[OffsetMove], other: Set[OtherPauseCondition])`, and with some convenience functions to put and retrieve values.

swalker2m: Yes, I think that makes sense with an appropriate name for `OtherPauseCondition` presumably.

jluhrs:
I think I have properly addressed all the feedbacks.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -3,34 +3,118 @@
...
 }
 
 object GemsController {
 
-  sealed trait GemsConfig
+  sealed trait P1Usage
+  object P1Usage {
+    case object UsesP1 extends P1Usage
+    case object DoesNotUseP1 extends P1Usage
+
```
swalker2m: Since you need it elsewhere when parsing `Boolean`s from the sequence(?), it's probably best to define `fromBoolean` here:

```scala
def fromBoolean(b: Boolean): P1Usage =
  if (b) UsesP1 else DoesNotUseP1
```

modules/seqexec/server/src/main/scala/seqexec/server/tcs/GuideConfigDb.scala
```diff
@@ -94,7 +94,19 @@ object GuideConfigDb {
...
+          } yield GemsOn(
+            if(ttgs1) Ttgs1Active.Ttgs1On else Ttgs1Active.Ttgs1Off,
+            if(ttgs2) Ttgs2Active.Ttgs2On else Ttgs2Active.Ttgs2Off,
+            if(ttgs3) Ttgs3Active.Ttgs3On else Ttgs3Active.Ttgs3Off,
+            if(odgw1) Odgw1Active.Odgw1On else Odgw1Active.Odgw1Off,
+            if(odgw2) Odgw2Active.Odgw2On else Odgw2Active.Odgw2Off,
+            if(odgw3) Odgw3Active.Odgw3On else Odgw3Active.Odgw3Off,
+            if(odgw4) Odgw4Active.Odgw4On else Odgw4Active.Odgw4Off,
+            if(useP1) P1Usage.UsesP1 else P1Usage.DoesNotUseP1,
+            if(useOI) OIUsage.UsesOI else OIUsage.DoesNotUseOI
```
swalker2m: `OIUsage.fromBoolean(useOI)`, etc.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSouth.scala
```diff
@@ -101,14 +140,22 @@ object TcsSouth {
...
+    guideWithP1: Option[StandardGuideOptions.Value],
+    guideWithP2: Option[StandardGuideOptions.Value],
+    guideWithOI: Option[StandardGuideOptions.Value],
+    guideWithCWFS1: Option[StandardGuideOptions.Value],
+    guideWithCWFS2: Option[StandardGuideOptions.Value],
+    guideWithCWFS3: Option[StandardGuideOptions.Value],
+    guideWithODGW1: Option[StandardGuideOptions.Value],
+    guideWithODGW2: Option[StandardGuideOptions.Value],
+    guideWithODGW3: Option[StandardGuideOptions.Value],
+    guideWithODGW4: Option[StandardGuideOptions.Value],
```
swalker2m: I'm not really suggesting anything here for this PR I guess, but this one leaves me unquiet. Long lists of parameters where everything is the same type seem problematic because it is easy to mess up the order. Here there are 10 parameters with the same type. Also, it seems odd that all the different OIs combine down to one thing, `guideWithOI`, but other individual guider types are listed.  If we add a new guider I guess this class grows another `Option[StandardGuideOptions.Value]` parameter.  We could at least wrap them at some point to give them a distinct type.

modules/seqexec/server/src/test/scala/seqexec/server/tcs/GuideConfigDbSpec.scala
```diff
@@ -117,7 +126,17 @@ final class GuideConfigDbSpec extends FlatSpec {
...
+    Some(Right(GemsOn(
+      Ttgs1On,
+      Ttgs2Off,
+      Ttgs3Off,
+      Odgw1On,
+      Odgw2Off,
+      Odgw3On,
+      Odgw4On,
+      DoesNotUseP1,
+      DoesNotUseOI
```
swalker2m: :+1:

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -28,45 +30,125 @@ trait Gaos[F[_]] {
...
-
-  implicit val becauseOffsetMoveEq: Eq[OffsetMove] = Eq.by(x => (x.previousOffset, x.newOffset))
-
-  implicit val pauseReasonEq: Eq[PauseCondition] = Eq.instance{
-    case (a: OffsetMove, b: OffsetMove) => a === b
-    case (OiOff, OiOff)                 => true
-    case (P1Off, P1Off)                 => true
-    case (GaosGuideOff, GaosGuideOff)   => true
-    case _                              => false
+  sealed trait PauseCondition extends Product with Serializable
```
swalker2m: I think the `PauseConditionSet` is a great idea, but it sort of bugs me that `PauseCondition` and `ResumeCondition`, along with `PauseConditionSet` and `ResumeConditionSet` are completely isomorphic and yet require all that duplicated code.  At the same time, having a distinct type for pause set vs. resume set is valuable. Maybe combine them in one `Condition` and `ConditionSet` and then wrap the set to give them distinct types?

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -3,34 +3,118 @@
...
+    implicit val p1UsageEq: Eq[P1Usage] = Eq.instance{
+      case (UsesP1, UsesP1)             => true
+      case (DoesNotUseP1, DoesNotUseP1) => true
+      case _                            => false
```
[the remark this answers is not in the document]
tpolecat: Agree. Scala 3 will have `deriving` like Haskell so this will get easier.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -3,34 +3,118 @@
...
+    implicit val oiUsageEq: Eq[OIUsage] = Eq.instance{
+      case (UsesOI, UsesOI)             => true
+      case (DoesNotUseOI, DoesNotUseOI) => true
+      case _                            => false
+    }
+  }
+
+  sealed trait GemsConfig {
+    val usesP1: Boolean
+    val usesOI: Boolean
```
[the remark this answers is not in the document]
jluhrs: Because I don't want to make the comparison every time I need to make a decision. Those values are always used in if/else clauses.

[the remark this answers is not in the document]
jluhrs: I know, I will rename them to `isP1Used` and `isOIUsed` to make clear they are a true/false query.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -3,34 +3,118 @@
...
   def endObserve: F[Unit]
 
+  val stateGetter: GemsGuiderStatus[F]
+
 }
 
 object GemsController {
 
-  sealed trait GemsConfig
+  sealed trait P1Usage
```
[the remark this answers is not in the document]
jluhrs: I don't understand why and when I should make traits `extend Product with Serializable`.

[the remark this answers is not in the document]
swalker2m: I think always for any sum type. https://typelevel.org/blog/2018/05/09/product-with-serializable.html

jluhrs:
@swalker2m For some reason I don't have the option to reply to your comment on Gems.scala. Anyways, those values have a different meaning. The types I created before are used to indicate a request to use a specific guider for GeMS. The values in `GemsGuiderStatus` are the procedures used to know if a guider is currently on. Actually, that value does not reflect the whole guider state, only if it is producing measurements (the detector could be reading out, but the probe may not be following the target, making the measurements useless). I will rename it.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -3,34 +3,118 @@
...
+
+  case object GemsOff extends GemsConfig {
+    override val usesP1: Boolean = false
+    override val usesOI: Boolean = false
+  }
+
+  sealed trait Ttgs1Active
+  object Ttgs1Active {
+    case object Ttgs1On extends Ttgs1Active
+    case object Ttgs1Off extends Ttgs1Active
```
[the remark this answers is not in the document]
jluhrs: Ok, I'll add them

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -28,45 +30,125 @@ trait Gaos[F[_]] {
...
+    implicit val pauseReasonEq: Eq[PauseCondition] = Eq.instance{
+      case (a: OffsetMove, b: OffsetMove)                   => a === b
+      case (a: FixedPauseCondition, b: FixedPauseCondition) => a === b
+      case _                                                => false
+    }
+
+  }
+
+  // Non-repeatable collection of PauseCondition that admits only one OffsetMove
+  final case class PauseConditionSet private (offsetO: Option[OffsetMove], fixed: Set[FixedPauseCondition]){
```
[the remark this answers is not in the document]
jluhrs: Or by starting from empty and adding values. Basically I wanted to hide the special behavior for `OffsetMove`.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -28,45 +30,125 @@ trait Gaos[F[_]] {
...
+
+    def +(v: PauseCondition): PauseConditionSet = v match {
+      case a:OffsetMove           => PauseConditionSet(a.some, fixed)
+      case b: FixedPauseCondition => PauseConditionSet(offsetO, fixed + b)
+    }
+
+    def contains(v: FixedPauseCondition): Boolean = fixed.contains(v)
```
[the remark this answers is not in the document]
jluhrs: I thought about that, but then I cannot create an `OffsetMove` without an offset. I would need to call it like `x.contains(OffsetMove(dummyValue)`.
The semantics of `contains` is never used for `OffsetMove` anyways. That's it, I never just want to now if exist in the set.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSouth.scala
jluhrs: Ok, I'll put a new task for this.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Gaos.scala
jluhrs: I see your point, but honestly, I don't thing the gain is worth my time.

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
jluhrs: Those values have a different meaning. The types I created before are used to indicate a request to use a specific guider for GeMS. The values in `GemsGuiderStatus` are the procedures used to know if a guider is currently on. Actually, that value does not reflect the whole guider state, only if it is producing measurements (the detector could be reading out, but the probe may not be following the target, making the measurements useless). I see that the names don't match, I'll fix them.

jluhrs:
Here comes another commit, this one removes the use of `Boolean` for GeMS detectors activity.
I will not change the apparent code duplication between `PauseCondition` and `ResumeCondition`, because I cannot guarantee the current symmetry between those types will persist.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsConfigRetriever.scala
```diff
@@ -20,7 +20,14 @@ import seqexec.model.TelescopeGuideConfig
...
 import seqexec.server.SeqexecFailure
-import seqexec.server.gems.Gems.GemsWfsStatus
+import seqexec.server.gems.Gems.{DetectorStateOps, GemsWfsState, Ngs1DetectorState, Ngs2DetectorState, Ngs3DetectorState, Odgw1DetectorState, Odgw2DetectorState, Odgw3DetectorState, Odgw4DetectorState}
+import seqexec.server.gems.Gems.Ngs1DetectorState.{Off => _, On => _, _}
+import seqexec.server.gems.Gems.Ngs2DetectorState.{Off => _, On => _, _}
+import seqexec.server.gems.Gems.Ngs3DetectorState.{Off => _, On => _, _}
+import seqexec.server.gems.Gems.Odgw1DetectorState.{Off => _, On => _, _}
+import seqexec.server.gems.Gems.Odgw2DetectorState.{Off => _, On => _, _}
+import seqexec.server.gems.Gems.Odgw3DetectorState.{Off => _, On => _, _}
+import seqexec.server.gems.Gems.Odgw4DetectorState.{Off => _, On => _, _}
```
swalker2m: Are these imports to pick up the implicits? I thought that would happen by default.

PR #1031, opened 2019-08-08
SEQNG-1033 Set GMOS disperser wavelenght even if disperser does not change

jluhrs:
Fixed bug where GMOS disperser's wavelength was not set between steps with the same disperser.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -220,6 +220,9 @@ object GmosControllerEpics extends GmosEncoders {
   def setOrder(state: GmosCCEpicsState, o: DisperserOrder): Option[IO[Unit]] =
     applyParam(state.disperserOrder, disperserOrderEncoderInt.encode(o), (_: Int) => CC.setDisperserOrder(disperserOrderEncoder.encode(o)))
 
+  def setWavelength(state: GmosCCEpicsState, w: Length): Option[IO[Unit]] =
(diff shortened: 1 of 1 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -234,15 +237,20 @@ object GmosControllerEpics extends GmosEncoders {
         //If disperser or order are set, force mode configuration
         s.foldMap(_ => s ++ List(CC.setDisperserMode(disperserMode0).some))
       case cfg.GmosDisperser.OrderN(d, o, w) =>
-        val s0: Option[IO[Unit]] = setDisperser(state, d)
+        val s0: Option[IO[Unit]] = setDisperser(state, d) // None means disperser is already in the right position
```
swalker2m: I don't know how this works so bear with me here if this is a dumb question.  The `IO` for applying the configuration change makes sense to me but I'm wondering how you know it is already in the right position without checking, and if you check, would that not require being in `IO`?

jluhrs: Just before applying the configuration the current configuration is read from the instrument and passed around. That is the first argument, `state`. `setDisperser` will produce a `Some[IO[_]]` only if the current disperser is different from the demanded one, effectively removing unnecessary interactions with the instrument.

tpolecat: Why not return `IO.unit` in this case?

PR #1033, opened 2019-08-12
SEQNG-1034: GPI Wrong Sample mode

cquiroz:
The sample mode for GPI is a parameter on the odb that cannot be edited and or all intents it is always UTR. However, it is still possible using the engineering to set it to some of the other values. Normally the value would be set automatically when doing an observation but in a rare case, doing a calibration may not set it and if combined with manual changes could produce FR-40693 where the readout mode was set to Single.

This PR will read the value from the ODB, compare it to the actual value and set it on the instrument if needed

Another set of parameters in the same category are the readout area. While there hasn't been a case of using a subarray ever, it is still possible to be set using the engineering level tools. Thus the readout area params receive the same treatment

This PR still needs some real field testing with the instrument


cquiroz:
Testing has been delayed due to instrument schedule. I'll merge anyway and test later on

PR #1036, opened 2019-08-14
SEQNG-331 Refactor of Actions list

cquiroz:
I continuously get confused with the definition of actions as `List[List[Action[F]]`. I also realized that the inner list shouldn't be empty thus I've changed it to `List[NonEmptyList[Action[F]]]` and renamed the type alias to `ParallelActions`

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
@@ -27,11 +27,7 @@ final case class Execution[F[_]](execution: List[Action[F]]) {
   val results: List[Action[F]] = execution.filter(Action.finished)
 
   def toParallelActionsList: List[ParallelActions[F]] =
-    if (execution.isEmpty) {
-      Nil
-    } else {
-      List(NonEmptyList.fromListUnsafe(execution))
-    }
+    NonEmptyList.fromList(execution).toList
 
   /**
     * Calculate `Execution` `Status` based on the underlying `Action`s.
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
@@ -5,6 +5,7 @@ package seqexec.engine
 
 import cats.Eq
 import cats.implicits._
+import cats.data.NonEmptyList
 import monocle.function.Index.{index, listIndex}
 import monocle.syntax.apply._
 import mouse.boolean._
@@ -25,6 +26,13 @@ final case class Execution[F[_]](execution: List[Action[F]]) {
 
   val results: List[Action[F]] = execution.filter(Action.finished)
 
+  def toParallelActionsList: List[ParallelActions[F]] =
+    if (execution.isEmpty) {
+      Nil
+    } else {
+      List(NonEmptyList.fromListUnsafe(execution))
+    }
+
   /**
     * Calculate `Execution` `Status` based on the underlying `Action`s.
     *
@@ -40,7 +48,7 @@ final case class Execution[F[_]](execution: List[Action[F]]) {
     * Obtain the resulting `Execution` only if all actions have been completed.
     *
     */
-  val uncurrentify: Option[Actions[F]] =
+  val uncurrentify: Option[List[Action[F]]] =
     (execution.nonEmpty && finished(this)).option(results)
 
   /**
@@ -63,8 +71,8 @@ object Execution {
     * Make an `Execution` `Current` only if all the `Action`s in the execution
     * are pending.
     */
-  def currentify[F[_]](as: Actions[F]): Option[Execution[F]] =
-    (as.nonEmpty && as.forall(_.state.runState.isIdle)).option(Execution(as))
+  def currentify[F[_]](as: ParallelActions[F]): Option[Execution[F]] =
+    as.forall(_.state.runState.isIdle).option(Execution(as.toList))
 
   def errored[F[_]](ex: Execution[F]): Boolean = ex.execution.exists(_.state.runState match {
     case ActionState.Failed(_) => true
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
@@ -14,12 +14,11 @@ import seqexec.engine.Action.ActionState
   * A list of `Executions` grouped by observation.
   */
 @Lenses
-final case class Step[F[_]](
-                             id: StepId,
-                             breakpoint: Step.BreakpointMark,
-                             skipped: Step.Skipped,
-                             skipMark: Step.SkipMark,
-                             executions: List[List[Action[F]]]
+final case class Step[F[_]](id: StepId,
+                            breakpoint: Step.BreakpointMark,
+                            skipped: Step.Skipped,
+                            skipMark: Step.SkipMark,
+                            executions: List[ParallelActions[F]]
 )
 
 object Step {
@@ -40,8 +39,12 @@ object Step {
   def skippedL[F[_]]: Lens[Step[F], Boolean] = Step.skipped ^|-> Skipped.self
 
   def init[F[_]](id: StepId,
-                 executions: List[List[Action[F]]]): Step[F] = Step(id, BreakpointMark(false),
-    Skipped(false), SkipMark(false), executions)
+                 executions: List[ParallelActions[F]]): Step[F] =
+      Step(id = id,
+           breakpoint = BreakpointMark(false),
+           skipped = Skipped(false),
+           skipMark = SkipMark(false),
+           executions = executions)
 
   /**
     * Calculate the `Step` `Status` based on the underlying `Action`s.
@@ -51,14 +54,14 @@ object Step {
     if(step.skipped.self) StepState.Skipped
     else
       // Find an error in the Step
-      step.executions.flatten.find(Action.errored).flatMap { x => x.state.runState match {
+      step.executions.flatMap(_.toList).find(Action.errored).flatMap { x => x.state.runState match {
         case ActionState.Failed(Result.Error(msg)) => msg.some
         case _                                     => None
         // Return error or continue with the rest of the checks
       }}.map[StepState](StepState.Failed).getOrElse(
         // All actions in this Step were completed successfully, or the Step is empty.
-        if (step.executions.flatten.forall(Action.completed)) StepState.Completed
-        else if (step.executions.flatten.forall(_.state.runState.isIdle)) StepState.Pending
+        if (step.executions.flatMap(_.toList).forall(Action.completed)) StepState.Completed
+        else if (step.executions.flatMap(_.toList).forall(_.state.runState.isIdle)) StepState.Pending
         // Not all actions are completed or pending.
         else StepState.Running
       )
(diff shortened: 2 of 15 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
```diff
@@ -25,6 +26,13 @@ final case class Execution[F[_]](execution: List[Action[F]]) {
 
   val results: List[Action[F]] = execution.filter(Action.finished)
 
+  def toParallelActionsList: List[ParallelActions[F]] =
+    if (execution.isEmpty) {
+      Nil
+    } else {
+      List(NonEmptyList.fromListUnsafe(execution))
+    }
```
tpolecat:
```scala
def toParallelActionsList: List[ParallelActions[F]] = 
  NonEmptyList.fromList(execution).toList
```

modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
```diff
@@ -101,11 +104,10 @@ object Step {
     /**
       * Obtain the resulting `Step` only if all `Execution`s have been completed.
       * This is a special way of *unzipping* a `Zipper`.
-      *
       */
     val uncurrentify: Option[Step[F]] =
       if (pending.isEmpty) focus.uncurrentify.map(
-        x => Step(id, breakpoint, Skipped(false), skipMark, x :: done)
+        x => Step(id, breakpoint, Skipped(false), skipMark, x.prepend(done))
```
tpolecat: `::` works with `NonEmptyList`

modules/seqexec/engine/src/main/scala/seqexec/engine/package.scala
```diff
@@ -104,12 +105,19 @@ package object engine {
...
-  type Results[F[_]] = List[Result[F]]
+  type ParallelActions[F[_]] = NonEmptyList[Action[F]]
+
+  implicit class ListParallelActionsOps[F[_]](val v: List[Action[F]]) extends AnyVal {
+    def prepend(ac: List[ParallelActions[F]]): List[ParallelActions[F]] =
+      if (v.isEmpty) {
+        ac
+      } else {
+        NonEmptyList.fromListUnsafe(v) :: ac
+      }
```
tpolecat: I think you can say `NonEmptyList.fromList(v).foldRight(ac)(_ :: _)` for the body here. If not you can at least do `NonEmptyList.fromList(v) match ...`

modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
```diff
@@ -115,16 +117,16 @@ object Step {
...
-        id,
-        breakpoint,
-        Skipped(false),
-        skipMark,
-        done ++ List(focus.execution) ++ pending
+        id = id,
+        breakpoint = breakpoint,
+        skipped = Skipped(false),
+        skipMark = skipMark,
+        executions = done ++ focus.toParallelActionsList ++ pending
```
sraaphorst: Much easier to read now. 👍

modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
sraaphorst: Isn't this switched around now? Is there any advantage to it?

sraaphorst: I guess it doesn't matter because you want all the `Execution`s to be done, but I'm just curious.

modules/seqexec/engine/src/main/scala/seqexec/engine/package.scala
```diff
@@ -104,12 +105,19 @@ package object engine {
...
     * without interruption. A *sequential* `Execution` can be represented with
     * an `Execution` with a single `Action`.
     */
-  type Actions[F[_]] = List[Action[F]]
-
-  type Results[F[_]] = List[Result[F]]
+  type ParallelActions[F[_]] = NonEmptyList[Action[F]]
+
+  implicit class ListParallelActionsOps[F[_]](val v: List[Action[F]]) extends AnyVal {
+    def prepend(ac: List[ParallelActions[F]]): List[ParallelActions[F]] =
```
sraaphorst: Ahhh... never mind. I think I understand now, seeing this.

modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
cquiroz: x isn't `NonEmptyList` but rather `List[Action[F]]` and `done` is `List[NonEmptyList[F]]` hence the special method

modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
cquiroz: Good, I guess we should refrain to use `fromListUnsafe` even though we know the list is empty

modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
cquiroz: Yeah, I spent some time adding the field names or I'd get quite lost

modules/seqexec/engine/src/test/scala/seqexec/engine/SequenceSpec.scala
```diff
@@ -146,25 +146,36 @@ class SequenceSpec extends FlatSpec {
...
+      id = 1,
+      breakpoint = Step.BreakpointMark(false),
+      Step.SkipMark(false),
+      pending = pending,
+      focus = focus,
+      done = done.map(_.map{r =>
+        val x = fromF[IO](ActionType.Observe, IO(r))
+        x.copy(state = Execution.actionStateFromResult(r)(x.state))
+      }),
+      rollback)
```
sraaphorst: Just being pedantic and stating that you missed `rolledback = ` here. 😄

modules/seqexec/engine/src/test/scala/seqexec/engine/StepSpec.scala
```diff
@@ -382,8 +394,8 @@ class StepSpec extends FlatSpec {
       case Some(Sequence.State.Zipper(zipper, status, _)) =>
         inside (zipper.focus.toStep) {
           // Check that the sequence stopped midway
-          case Step(_, _, _, _, ex1::ex2::ex3::Nil) =>
-            assert( Execution(ex1).results.length == 2 && Execution(ex2).results.length == 1 && Execution(ex3).actions.length == 1)
+          case Step(_, _, _, _, ex1 :: ex2 :: ex3 :: Nil) =>
```
sraaphorst: You changed the rest of them from `ex1 :: ex2 :: ... :: Nil` to `List(ex1, ex2, ...)`. Why not this one?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -487,7 +488,7 @@ object SeqTranslate {
     def toAction(kind: ActionType): Action[F] = fromF[F](kind, x.attempt.map(_.toResult))
   }
 
-  implicit class ConfigResultToAction[F[_]: Functor: ApplicativeError[?[_], Throwable]](val x: F[ConfigResult[F]]) {
```
sraaphorst: This is something I don't recall having seen before. What is a `?[_]`?

modules/seqexec/server/src/main/scala/seqexec/server/SequenceGen.scala
```diff
@@ -52,12 +53,12 @@ object SequenceGen {
...
-  final case class StepActionsGen[F[_]](pre: List[Actions[F]],
+  final case class StepActionsGen[F[_]](pre: List[ParallelActions[F]],
                                   configs: Map[Resource, Action[F]],
-                                  post: HeaderExtraData => List[Actions[F]]) {
-    def generate(ctx: HeaderExtraData): List[Actions[F]] =
+                                  post: HeaderExtraData => List[ParallelActions[F]]) {
+    def generate(ctx: HeaderExtraData): List[ParallelActions[F]] =
       pre ++
-        (if(configs.isEmpty) List() else List(configs.values.toList)) ++
+        NonEmptyList.fromList(configs.values.toList).foldMap(List(_)) ++
```
sraaphorst: Previously, you got a NEL from a `List` unsafely... how can you do it here without using the unsafely and make an assumption that `configs` is not an empty `Map`?

modules/seqexec/engine/src/test/scala/seqexec/engine/SequenceSpec.scala
cquiroz: 👍

modules/seqexec/engine/src/test/scala/seqexec/engine/StepSpec.scala
cquiroz: No reason really, I should use the same style

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
cquiroz: It means (i think) you want an `ApplicativeError` for some `F[_]`. Maybe @tpolecat can explain it ibetter

modules/seqexec/server/src/main/scala/seqexec/server/SequenceGen.scala
cquiroz: I think we should avoid using the unsafe version and use something like this code instead. `foldMap` has a nice property that it will give you the `Monid` empty if needed. In this case `Nil`

cquiroz: As suggested before this can be also `NonEmptyList.fromList(...).toList`

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
tpolecat: Yeah the syntax is a little wacky.

A context bound `[X : Y]` desugars to `(implicit ev: Y[X])` where `ev` is a unique inaccessible name the compiler generates. So in order for this to be well-kinded `Y` must be a type constructor (i.e., a type-level function) with a single parameter of the same kind ("shape") as `X`.

A first-order example would be

- `[A : Monoid]` which is how you normally write it, or
- `[A : Monoid[?]]` which shows the hole explicitly.

These are equivalent. A higher-order example would be

- `[F[_] : Functor]` which is how you normally write it, or
- `[F[_] : Functor[?[_]]]` which shows the hole explicitly.

Note that the hole `?[_]` has the same shape as `F[_]`. They have to match, otherwise it's ill-kinded and you'll get a compile error.

Ok, so what we're dealing with here is `ApplicativeError[F[_], E]` which has *two* parameters, but a context bound requires a type constructor with exactly one. So you have to partially apply it, and to do this you need the second form, where we show the hole explicitly for the type variable we leaving free, and specify `Throwable` for the other one.

- `[F[_]: ApplicativeError[?[_], Throwable]]`

Which in the end desugars to `(implicit ev: ApplicativeError[F, Throwable])`

Make sense? When in doubt you can always write out the implicit parameter list. This is just shorthand.

cquiroz: @tpolecat Thanks for the explanation. This works only because we have the `kind-projector` plugin right? What would the story be in dotty?

tpolecat: Something very close to kind-projector is built into dotty but I don't know exactly how you would write this.

PR #1041, opened 2019-08-21
upgrade sbt-gsp, gsp-math, cats, scalatest

tpolecat:
This adds the `sbt-gsp` plugin and updates `gsp-math` and upstream dependencies including ScalaTest, which is the source of most of the changes here.


project/plugins.sbt
```diff
@@ -9,9 +9,12 @@ val scalaJSVersion =
 libraryDependencies ++= Seq(
   "org.postgresql" % "postgresql"  % "42.2.6", // needed by flyway
   "org.slf4j"      % "slf4j-nop"   % "1.7.28", // to silence some log messages
-  "org.typelevel" %% "cats-core"   % "1.6.1",
-  "org.typelevel" %% "cats-effect" % "0.8"
+  "org.typelevel" %% "cats-core"   % "2.0.0-RC1",
```
geminihlswadmin: Could this be the problem? Doesn't this add cats to the sbt classpath?

tpolecat: No, that's not a big deal. It was happening before I did that. Also you're logged in as @geminihlswadmin for some reason.

tpolecat:
Ok the tests run now and everything passes. Thanks @cquiroz for spotting the problem in gsp-math. I will resolve conflicts and with luck this will be good to go.

tpolecat:
Ok assuming this passes can you have a look @cquiroz since you seem to be around? Good time to merge this since nobody is around.

cquiroz:
I looks fine in general. I'd like to give it a spin starting up the seqexec and do a few simulated runs
I'll try to do it today

tpolecat:
:+1: sounds good, thanks

modules/ephemeris/src/test/scala/gem/horizons/HorizonsEphemerisQueryPagingPropSpec.scala
```diff
@@ -83,7 +80,7 @@ class HorizonsEphemerisQueryPagingPropSpec extends PropSpec with PropertyChecks
     }
   }
 
-  property("each element is spaced exactly by duration") {
+  test("each element is spaced exactly by duration") {
```
cquiroz: I don't recall `property` was it related to scalacheck?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/SeqexecApp.scala
```diff
@@ -31,6 +29,9 @@ final class SeqexecLauncher[F[_]](implicit val F: Sync[F], L: LiftIO[F]) {
     setLoggerThreshold("", AllThreshold)
   }
 
+  private implicit val ioContextShift: ContextShift[IO] =
```
cquiroz: Why didn't we need this before?

modules/ephemeris/src/test/scala/gem/horizons/HorizonsEphemerisQueryPagingPropSpec.scala
tpolecat: Yeah you can write tests with straight scalacheck and no scalatest.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/SeqexecApp.scala
tpolecat: Good question, I will look more closely.

tpolecat: `IO.fromFuture` now takes `ContextShift[IO]`. Daniel added late in the release cycle https://github.com/typelevel/cats-effect/commit/ebab1db7b9a116c939c60213339743d5b9907ac9

PR #1045, opened 2019-08-22
SEQNG-1029 Implemented control of GeMS.

jluhrs:
This PR implements the control of GeMS. GeMS loop control is notified when starting an unguided step,  and when starting/completing a telescope offset. ODGW control is incomplete, but with this we can start testing with the real systems.
The code referred to the Canopus WFS in three different ways. I normalized all references to use just CWFS.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -46,6 +46,7 @@ import seqexec.server.altair.AltairEpics
 import seqexec.server.altair.AltairLgsHeader
 import seqexec.server.altair.AltairKeywordReaderEpics
 import seqexec.server.altair.AltairKeywordReaderDummy
+import seqexec.server.gems.Gems
 import squants.Time
 import squants.time.TimeConversions._
 
@@ -324,9 +325,16 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
 
   private def getTcs(subs: NonEmptySet[TcsController.Subsystem], useGaos: Boolean, inst: InstrumentSystem[IO],
                      lsource: LightSource, config: Config): TrySeq[System[IO]] = site match {
-    case Site.GS => TcsSouth.fromConfig[IO](systems.tcsSouth, subs, None, inst, systems.guideDb)(
-      config, LightPath(lsource, inst.sfName(config)), extractWavelength(config)
-    ).asRight
+    case Site.GS => if(useGaos)
+      Gems.fromConfig[IO](systems.gems, systems.guideDb)(config).map(a =>
+        TcsSouth.fromConfig[IO](systems.tcsSouth, subs, a.some, inst, systems.guideDb)(
+          config, LightPath(lsource, inst.sfName(config)), extractWavelength(config)
+        )
+      )
+      else
+        TcsSouth.fromConfig[IO](systems.tcsSouth, subs, None, inst, systems.guideDb)(
+          config, LightPath(lsource, inst.sfName(config)), extractWavelength(config)
+        ).asRight
     case Site.GN => if(useGaos)
         Altair.fromConfig(config, systems.altair).map(a =>
           TcsNorth.fromConfig[IO](systems.tcsNorth, subs, a.some, inst, systems.guideDb)(
@@ -364,6 +372,11 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
           .map(tcs => sys :: List(tcs, Gcal(systems.gcal, site == Site.GS)))
 
       case StepType.AlignAndCalib         => List(sys).asRight
+
+      case StepType.Gems(inst) =>
+        getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi).add(Gaos), true, sys, TcsController.LightSource.AO, config)
+          .map(tcs => sys :: List(tcs, Gcal(systems.gcal, site == Site.GS)))
+
 
       case _                     => TrySeq.fail(Unexpected(s"Unsupported step type $stepType"))
     }
@@ -443,7 +456,7 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
           calcInstHeader(config, sys).map(h => Reader(ctx =>
             List(commonHeaders(TcsEpics.instance, config, allButGaos.toList, sys)(ctx), gwsHeaders(GwsEpics.instance, sys), h)))
 
-      case StepType.AltairObs(_) =>
+      case StepType.AltairObs(_)    =>
         val tcsKReader = if (settings.tcsKeywords) TcsKeywordsReaderEpics[IO](TcsEpics.instance) else DummyTcsKeywordsReader[IO]
         for {
           gst  <- Altair.guideStarType(config)
@@ -456,14 +469,20 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
                       gwsHeaders(GwsEpics.instance, sys), h)))
         } yield read
 
-      case StepType.FlatOrArc(inst)       =>
+      case StepType.FlatOrArc(inst) =>
           calcInstHeader(config, sys).map(h => Reader(ctx =>
             List(commonHeaders(TcsEpics.instance, config, flatOrArcTcsSubsystems(inst).toList, sys)(ctx), gcalHeader(GcalEpics.instance, sys), gwsHeaders(GwsEpics.instance, sys), h)))
 
-      case StepType.DarkOrBias(_)      =>
+      case StepType.DarkOrBias(_)   =>
           calcInstHeader(config, sys).map(h => Reader(ctx => List(commonHeaders(TcsEpics.instance, config, Nil, sys)(ctx), gwsHeaders(GwsEpics.instance, sys), h)))
-      case StepType.AlignAndCalib         => TrySeq(Reader(_ => Nil)) // No headers for A&C
-      case st                    => TrySeq.fail(Unexpected(s"Unsupported step type $st"))
+
+      case StepType.AlignAndCalib   => TrySeq(Reader(_ => Nil)) // No headers for A&C
+
+      case StepType.Gems(_)         =>
+        calcInstHeader(config, sys).map(h => Reader(ctx =>
+          List(commonHeaders(TcsEpics.instance, config, allButGaos.toList, sys)(ctx), gwsHeaders(GwsEpics.instance, sys), h)))
+
+      case st                       => TrySeq.fail(Unexpected(s"Unsupported step type $st"))
     }
   }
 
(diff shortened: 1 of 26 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -456,14 +469,20 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
 
-      case StepType.DarkOrBias(_)      =>
+      case StepType.DarkOrBias(_)   =>
           calcInstHeader(config, sys).map(h => Reader(ctx => List(commonHeaders(TcsEpics.instance, config, Nil, sys)(ctx), gwsHeaders(GwsEpics.instance, sys), h)))
-      case StepType.AlignAndCalib         => TrySeq(Reader(_ => Nil)) // No headers for A&C
-      case st                    => TrySeq.fail(Unexpected(s"Unsupported step type $st"))
+
+      case StepType.AlignAndCalib   => TrySeq(Reader(_ => Nil)) // No headers for A&C
+
+      case StepType.Gems(_)         =>
```
cquiroz: Why do we need a new `StepType`? Is it equivalent to the `Altair` step type?

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
```diff
@@ -24,25 +33,71 @@ trait Gems[F[_]] extends Gaos[F] {
...
+      }
 
     override val stateGetter: GemsWfsState[F] = controller.stateGetter
   }
 
-  def fromConfig[F[_]: Applicative](controller: GemsController[F])/*(config: Config)*/: F[Gems[F]] =
-    Applicative[F].pure(new GemsImpl[F](controller, GemsOff))
+  // `combine` calculates the final configuration between the configuration coming from the step and the configuration
+  // set by the operator.
+  private def combine(opConfig: GemsConfig, stepConfig: GemsConfig): GemsConfig = GemsController.GemsOn(
```
cquiroz: Could this be a `Monoid`? then you could do `config1 |+| config2`

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
```diff
@@ -24,25 +33,71 @@ trait Gems[F[_]] extends Gaos[F] {
...
+    Odgw4Usage.fromBoolean(opConfig.isOdgw4Used && stepConfig.isOdgw4Used),
+    P1Usage.fromBoolean(opConfig.isP1Used && stepConfig.isP1Used),
+    OIUsage.fromBoolean(opConfig.isOIUsed && stepConfig.isOIUsed)
+  )
+
+  def fromConfig[F[_]: MonadError[?[_], Throwable]](controller: GemsController[F], guideConfigDb: GuideConfigDb[F])
+                                                         (config: Config)
+  : TrySeq[Gems[F]] = {
+    for {
+      p1    <- config.extractAs[StandardGuideOptions.Value](TELESCOPE_KEY / Tcs.GUIDE_WITH_PWFS1_PROP)
```
cquiroz: There are utility methoods `extractObsAs` and `extractInstAs`. Maybe you can add a `extractTelescopeAs`

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
```diff
@@ -63,33 +118,33 @@ object Gems {
...
-    case object On extends Ngs1DetectorState
-    case object Off extends Ngs1DetectorState
+  sealed trait Cwfs1DetectorState extends Product with Serializable
+  object Cwfs1DetectorState {
+    case object On extends Cwfs1DetectorState
+    case object Off extends Cwfs1DetectorState
 
-    implicit val ngs1DetectorStateEq: Eq[Ngs1DetectorState] = Eq.fromUniversalEquals
-    implicit val ngs1DetectorStateDetectorStateOps: DetectorStateOps[Ngs1DetectorState] =
+    implicit val cwfs1DetectorStateEq: Eq[Cwfs1DetectorState] = Eq.fromUniversalEquals
```
cquiroz: We have been converting some of these to `Enumerated` like https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/model/src/main/scala/seqexec/model/enum/FPUMode.scala

It would give you an `Eq` and `Arb` automatically

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -124,9 +135,9 @@ object GemsController {
   }
 
   final case class GemsOn(
-                           ttgs1: Ttgs1Usage,
-                           ttgs2: Ttgs2Usage,
-                           ttgs3: Ttgs3Usage,
+                           cwfs1: Cwfs1Usage,
```
cquiroz: Why did you rename these?

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
```diff
@@ -0,0 +1,197 @@
...
+  : Option[F[Unit]] = {
+    val unguided = pauseConditions.contains(PauseCondition.GaosGuideOff).option(UnguidedCondition)
+    val offset = pauseConditions.offsetO.as(OffsetCondition)
+    val instMove = pauseConditions.contains(PauseCondition.InstConfigMove).option(InstrumentCondition)
+
+    val reasons = List(unguided, offset, instMove).flattenOption
+
+    reasons.nonEmpty.option {
+      epicsSys.LoopControl.setCommand(PauseCmd) *>
+        epicsSys.LoopControl.setReasons(reasons.mkString("|")) *>
```
cquiroz: Is this format required on the Epics side?

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
```diff
@@ -135,6 +146,27 @@ object GemsController {
...
+    override val isCwfs1Used: Boolean = cwfs1 === UseCwfs1
+
+    import Cwfs2Usage._
+    override val isCwfs2Used: Boolean = cwfs2 === UseCwfs2
+
+    import Cwfs3Usage._
+    override val isCwfs3Used: Boolean = cwfs3 === UseCwfs3
+
+    import Odgw1Usage._
+    override val isOdgw1Used: Boolean = odgw1 === UseOdgw1
```
tpolecat: Why not just `Odgw1Usage.UseOdgw1` here and elsewhere? Are you importing any other symbol from `Odgw1Usage`?

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
```diff
@@ -0,0 +1,197 @@
...
+    Nested(
+      (gsaoiSys.guiding, gsaoiSys.odgw2Multiplier).mapN{case (x, y) => (x, y).mapN{case (g, m) => g && m > 0}}
+    ).map(DetectorStateOps.fromBoolean[Odgw2DetectorState]).value,
+    Nested(
+      (gsaoiSys.guiding, gsaoiSys.odgw3Multiplier).mapN{case (x, y) => (x, y).mapN{case (g, m) => g && m > 0}}
+    ).map(DetectorStateOps.fromBoolean[Odgw3DetectorState]).value,
+    Nested(
+      (gsaoiSys.guiding, gsaoiSys.odgw4Multiplier).mapN{case (x, y) => (x, y).mapN{case (g, m) => g && m > 0}}
+    ).map(DetectorStateOps.fromBoolean[Odgw4DetectorState]).value
+  )
```
tpolecat: Can we factor out the copy/paste here and elsewhere in this file?

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
tpolecat: ... in which case why not a shorter name like `Odgw1Usage.Use`?

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
jluhrs: No, because it is not commutative. The configuration set by the operator in TCC takes precedence.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: It is not a new type, but it was ignored until now. I agree it could be merged with `StepType.Altair`, if the code that handles both types ends up being equivalent. But let's wait until this is running before making that decision.

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
tpolecat: Monoids only have to be associative, not commutative. `this orElse that` is a common pattern.

modules/seqexec/server/src/main/scala/seqexec/server/gems/Gems.scala
jluhrs: That makes a lot of sense.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
jluhrs: It is explained in the description of the PR. The code referred to the Canopus detector as Ttgs, Cwfs and Ngs. I changed all of them to Cwfs.

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
jluhrs: Yes

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
jluhrs: About the first question, it is convenient that way because it is always used in `if-else` statements. The alternative would be to do write the comparison inside every `if`.
About the second, you are right. This is the code that motivated to ask in Slack about style, but I was to lazy to go back and fix that code. I'll change it.

tpolecat: Sorry, what I meant was, why do `import FooUsage` and use `UseFoo` rather than skipping the import and saying `FooUsage.UseFoo`.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsController.scala
jluhrs: Aaaahhh. Actually, I only care about the implicit `Eq`.

PR #1051, opened 2019-08-29
GeMS FITS keywords

jluhrs:
The GeMS FITS keywords pull information from several sources: the sequence, TCS, the GeMS subsystems, and GSAOI, so it is a bit more complicated that an instrument keywords code. The EPICS part is not yet tested, so I will hold merging the code until then


jluhrs:
On second thought, I will merge the PR and have another task to test and fix the access to EPICS channels.

PR #1053, opened 2019-08-30
SEQNG-1041 Changed style of GeMS and GSAOI Epics statuses.

jluhrs:
Status read method used the old style of producing a `F[Option[A]]`, where a `None` signaled a null value. They were changed to `F[A]`with a null value signaled by an exception that can be managed up the stack.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -232,26 +232,41 @@ object EpicsUtil {
   /** Tries to read a value of type A from a channel
    *  Null results are raised as error and other errors are captured
    */
-  def safeAttributeF[F[_]: Sync, A >: Null](name: String, get: => CaAttribute[A]): F[A] =
-    Sync[F].delay(Option(get.value)) // Wrap the read on Option to do null check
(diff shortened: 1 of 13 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -232,26 +232,41 @@ object EpicsUtil {
...
-  def safeAttributeF[F[_]: Sync, A >: Null](name: String, get: => CaAttribute[A]): F[A] =
-    Sync[F].delay(Option(get.value)) // Wrap the read on Option to do null check
+  def safeAttributeWrapF[F[_]: Sync, A >: Null](name: String, get: => A): F[A] =
+    Sync[F].delay(Option(get)) // Wrap the read on Option to do null check
       .adaptError{ case e => SeqexecException(e)} // if we have e.g CAException wrap it
       .ensure(NullEpicsError(name))(_.isDefined) // equivalent to a null check
       .map{_.orNull} // orNull lets us typecheck but it will never be used due to the `ensure` call above
 
+  def safeAttributeF[F[_]: Sync, A >: Null](get: => CaAttribute[A]): F[A] =
+    safeAttributeWrapF(get.name, get.value)
```
cquiroz: Nice

PR #1054, opened 2019-09-02
Feature/seqng 1043

jluhrs:
All systems' EPICS statuses are now read as `F[A]` instead of `F[Option[A]]`, leaving to `F` the task of dealing with null values.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -232,17 +232,17 @@ object EpicsUtil {
   /** Tries to read a value of type A from a channel
    *  Null results are raised as error and other errors are captured
    */
-  def safeAttributeWrapF[F[_]: Sync, A >: Null](name: String, get: => A): F[A] =
+  def safeAttributeWrapF[F[_]: Sync, A >: Null](channel: String, get: => A): F[A] =
     Sync[F].delay(Option(get)) // Wrap the read on Option to do null check
       .adaptError{ case e => SeqexecException(e)} // if we have e.g CAException wrap it
-      .ensure(NullEpicsError(name))(_.isDefined) // equivalent to a null check
+      .ensure(NullEpicsError(channel))(_.isDefined) // equivalent to a null check
       .map{_.orNull} // orNull lets us typecheck but it will never be used due to the `ensure` call above
 
   def safeAttributeF[F[_]: Sync, A >: Null](get: => CaAttribute[A]): F[A] =
-    safeAttributeWrapF(get.name, get.value)
+    safeAttributeWrapF(get.channel, get.value)
 
   def safeAttributeSListF[F[_]: Sync, A >: Null](get: => CaAttribute[A]): F[List[A]] =
-    safeAttributeWrapF(get.name, get.values.asScala.toList)
+    safeAttributeWrapF(get.channel, get.values.asScala.toList)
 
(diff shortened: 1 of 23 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -232,17 +232,17 @@ object EpicsUtil {
...
+      .ensure(NullEpicsError(channel))(_.isDefined) // equivalent to a null check
       .map{_.orNull} // orNull lets us typecheck but it will never be used due to the `ensure` call above
 
   def safeAttributeF[F[_]: Sync, A >: Null](get: => CaAttribute[A]): F[A] =
-    safeAttributeWrapF(get.name, get.value)
+    safeAttributeWrapF(get.channel, get.value)
 
   def safeAttributeSListF[F[_]: Sync, A >: Null](get: => CaAttribute[A]): F[List[A]] =
-    safeAttributeWrapF(get.name, get.values.asScala.toList)
+    safeAttributeWrapF(get.channel, get.values.asScala.toList)
```
jluhrs: Having the channel address is more useful for debugging.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairKeywordReader.scala
```diff
@@ -82,14 +81,12 @@ object AltairKeywordReaderEpics extends AltairKeywordReaderLUT {
   def apply[F[_]: Sync](sys: AltairEpics[F]): AltairKeywordReader[F] = new AltairKeywordReader[F] {
 
     override def aofreq: F[Double] =
-      Nested(sys.aoexpt)
-        .filter(_ =!= 0.0f)
-        .map(1 / _.toDouble)
-        .value
+      sys.aoexpt
+        .map(1.0 / _.toDouble)
```
cquiroz: Why is the filter off? Are you assuming that `safeValOrDefalt` would catch a divide by zero?

modules/seqexec/server/src/main/scala/seqexec/server/gcal/GcalKeywordReader.scala
```diff
@@ -41,25 +41,25 @@ object GcalKeywordsReaderEpics {
...
-      val ir   = Nested(sys.lampIr.delay).map{
-        case BinaryOnOff.ON => "IRhigh"
-        case _              => "IRlow"
-      }.value
-      val qh   = sys.lampQH.delay.map(_.exists(onCheck)).ifM("QH".some.pure[F], noValue)
-      val thAr = sys.lampThAr.delay.map(_.exists(onCheck)).ifM("ThAr".some.pure[F], noValue)
-      val xe   = sys.lampXe.delay.map(_.exists(onCheck)).ifM("Xe".some.pure[F], noValue)
+      def onCheck(v: BinaryOnOff): Boolean = v === BinaryOnOff.ON
 
-      ar.orElse(xe).orElse(cuAr).orElse(thAr).orElse(qh).orElse(ir).safeValOrDefault
```
jluhrs: This shouldn't work. The intention is that the result must be the first `Some(_)`, or a default value if none. But the `orElse` here is run over `F`, not teh contained `Option`!

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairKeywordReader.scala
jluhrs: Yes

PR #1055, opened 2019-09-02
SEQNG-1044 Improved the implementation of GCAL.

jluhrs:
The implementation of GCAL was simplified and changed to follow the other systems.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -353,32 +353,36 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
     sys: InstrumentSystem[IO]
   ): TrySeq[List[System[IO]]] = {
     stepType match {
-      case StepType.CelestialObject(inst) =>
-        getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), false, sys, TcsController.LightSource.Sky, config)
-          .map(tcs => sys :: List(tcs, Gcal(systems.gcal, site == Site.GS)))
+      case StepType.CelestialObject(inst) => for {
+        tcs  <- getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), false, sys, TcsController.LightSource.Sky, config)
+        gcal <- Gcal.fromConfig(systems.gcal, site == Site.GS)(config)
+      } yield List(sys, tcs, gcal)
 
-      case StepType.NodAndShuffle(inst) =>
-        getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), false, sys, TcsController.LightSource.Sky, config)
-          .map(tcs => sys :: List(tcs, Gcal(systems.gcal, site == Site.GS)))
+      case StepType.NodAndShuffle(inst)   => for {
+        tcs  <- getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), false, sys, TcsController.LightSource.Sky, config)
+        gcal <- Gcal.fromConfig(systems.gcal, site == Site.GS)(config)
+      } yield List(sys, tcs, gcal)
 
-      case StepType.FlatOrArc(inst)       =>
-        getTcs(flatOrArcTcsSubsystems(inst), false, sys, TcsController.LightSource.GCAL, config)
-          .map(tcs => sys :: List(tcs, Gcal(systems.gcal, site == Site.GS)))
+      case StepType.FlatOrArc(inst)       => for {
+        tcs  <- getTcs(flatOrArcTcsSubsystems(inst), false, sys, TcsController.LightSource.GCAL, config)
+        gcal <- Gcal.fromConfig(systems.gcal, site == Site.GS)(config)
+      } yield List(sys, tcs, gcal)
 
-      case StepType.DarkOrBias(_)      =>  List(sys).asRight
+      case StepType.DarkOrBias(_)         =>  List(sys).asRight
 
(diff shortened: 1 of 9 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -353,32 +353,36 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
     sys: InstrumentSystem[IO]
   ): TrySeq[List[System[IO]]] = {
     stepType match {
-      case StepType.CelestialObject(inst) =>
-        getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), false, sys, TcsController.LightSource.Sky, config)
-          .map(tcs => sys :: List(tcs, Gcal(systems.gcal, site == Site.GS)))
+      case StepType.CelestialObject(inst) => for {
+        tcs  <- getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), false, sys, TcsController.LightSource.Sky, config)
```
cquiroz: Could you name the boolean parameter? Or change if for an ADT

modules/seqexec/server/src/main/scala/seqexec/server/gcal/Gcal.scala
```diff
@@ -37,82 +36,54 @@ final case class Gcal[F[_]](controller: GcalController[F], isCP: Boolean)(implic
...
+      val tharLamp = lamps.map(v => if (v.contains(Lamp.THAR_ARC)) LampState.On else LampState.Off)
+      val qhLamp = lamps.map(v => if (v.contains(Lamp.QUARTZ)) LampState.On else LampState.Off)
+      val xeLamp = lamps.map(v => if (v.contains(Lamp.XE_ARC)) LampState.On else LampState.Off)
       val irLampCP = lamps.map(v => if (v.contains(Lamp.IR_GREY_BODY_HIGH) || v.contains(Lamp.IR_GREY_BODY_LOW)) Some(LampState.On) else None)
       val irLampMK = lamps.map(v => if (v.contains(Lamp.IR_GREY_BODY_HIGH)) Some(LampState.On)
                                     else if (v.contains(Lamp.IR_GREY_BODY_LOW)) Some(LampState.Off) else None)
-      val shutter = config.extractAs[Shutter](CALIBRATION_KEY / SHUTTER_PROP).map(Some(_)).recoverWithDefault(None)
-      val filter = config.extractAs[Filter](CALIBRATION_KEY / FILTER_PROP).map(Some(_)).recoverWithDefault(None)
-      val diffuser = config.extractAs[Diffuser](CALIBRATION_KEY / DIFFUSER_PROP).map(Some(_)).recoverWithDefault(None)
+      val shutter = config.extractAs[Shutter](CALIBRATION_KEY / SHUTTER_PROP)
```
cquiroz: Maybe we could add a `extractCalAs`?

modules/seqexec/server/src/main/scala/seqexec/server/gcal/GcalControllerEpics.scala
```diff
@@ -67,104 +41,93 @@ object GcalControllerEpics {
...
-    GcalEpics.instance.lampsCmd.setCuArLampOn(v)
-  )
-
-  private def setThArLampParams(v: BinaryOnOff): List[IO[Unit]] = List(
-    GcalEpics.instance.lampsCmd.setThArLampName("ThAr"),
-    GcalEpics.instance.lampsCmd.setThArLampOn(v)
-  )
+  def apply[F[_]: Async](epics: => GcalEpics[F]): GcalController[F] = new GcalController[F] {
+    override def applyConfig(config: GcalConfig): F[Unit] =
+      retrieveConfig(epics).flatMap(configure(epics, _, config))
```
cquiroz: So we read the state and then do a smart configure?

modules/seqexec/server/src/main/scala/seqexec/server/gcal/GcalControllerEpics.scala
```diff
@@ -67,104 +41,93 @@ object GcalControllerEpics {
...
+      applyParam(current.shutter, encode(demand.shutter), epics.shutterCmd.setPosition),
+      demand.filterO.flatMap(d => applyParam(current.filter, encode(d), epics.filterCmd.setName)),
+      demand.diffuserO.flatMap(d => applyParam(current.diffuser, encode(d), epics.diffuserCmd.setName))
+    ).flattenOption
+
+    (
+      params.sequence *>
+        epics.lampsCmd.setTimeout[F](SetupTimeout) *>
+        epics.post
+    ).whenA(params.nonEmpty)
```
cquiroz: I always think we should go to make `applyParam` to return `F[Unit]` rather than `F[Option[Unit]]` but we need to skip the `setTimout/post` steps if no param is set. Is that correct?

modules/seqexec/server/src/main/scala/seqexec/server/gcal/GcalControllerEpics.scala
jluhrs: Yes, exactly.

modules/seqexec/server/src/main/scala/seqexec/server/gcal/GcalControllerEpics.scala
jluhrs: Yes, that is correct. Triggering a configuration command when no parameter has changed is a waste of time.

modules/seqexec/server/src/main/scala/seqexec/server/gcal/Gcal.scala
jluhrs: Done.

PR #1056, opened 2019-09-03
Engine support for Nod and shuffle

cquiroz:
This PR is a fairly large change to support running Nod And Shuffle steps. At the moment the engine has the notion of actions and usually an observe is one action. Moreover, now all observe actions are the same for each instrument making it easier to centralize. However Nod and Shuffle requires several actions inside a single observe and they are different depending on the instrument and step type.

N&S Support required an important change on the design where instead of a single place where observe actions are created we delegate to the instrument how to create the actions asking them to return an `InstrumentActions` instance. For most instruments the code won't change but for GPI and GMOS we create special `InstrumentAction` instances.

The methods in `ObserveActions` have been reorganized to let the different pieces be assembled differently between instruments.

This PR supports pushing updates to the client and a new Step type was added (Contributing greatly to the PR size as changes were required on the UI). Extra events are sent to indicate when NS begins and at each substage. This can be somewhat observed on the image below though it is far from being the final UI

ezgif com-video-to-gif(2)

You may notice that we do config/observe/config/observe in the same step

Finally in this PR I'm adding log4cats to reduce the pervasive use of `Sync` and `better-monadic-for`

While this is the core N&S PR we are far from over. The following needs to still be done:

* Add TCS offsets
* Support N&S Darks
* Fix some issues on the client state across observes
* Custom UI for N&S
* Support for the special pause/abort for N&S
* N&S keywords
* Testing

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -515,6 +515,7 @@ lazy val seqexec_server = project
   .settings(
     addCompilerPlugin(Plugins.paradisePlugin),
     addCompilerPlugin(Plugins.kindProjectorPlugin),
+    addCompilerPlugin(Plugins.betterMonadicForPlugin),
     libraryDependencies ++=
       Seq(Http4sCirce,
           Squants.value,
@@ -526,7 +527,8 @@ lazy val seqexec_server = project
           Log4s.value,
           Http4sXml,
           Http4sBoopickle,
-          PrometheusClient
+          PrometheusClient,
+          Log4Cats.value
       ) ++ Http4s ++ Http4sClient ++ SeqexecOdb ++ Monocle.value ++ WDBAClient ++ TestLibs.value ++
         Circe.value
   )
--- a/modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
@@ -8,6 +8,7 @@ import cats.implicits._
 import seqexec.model.enum._
 import monocle.Prism
 import monocle.Lens
+import monocle.Optional
 import monocle.macros.Lenses
 import monocle.macros.GenPrism
 
@@ -21,15 +22,9 @@ sealed trait Step extends Product with Serializable {
 }
 
 object Step {
-  val Zero: Step = StandardStep(id = -1,
-                                config        = Map.empty,
-                                status        = StepState.Pending,
-                                breakpoint    = false,
-                                skip          = false,
-                                fileId        = None,
-                                configStatus  = Nil,
-                                observeStatus = ActionStatus.Pending)
-
+  implicit class LensBooleanOps[A](l: Lens[A, Boolean]) {
+    def negate: A => A = l.modify(!_)
+  }
   val standardStepP: Prism[Step, StandardStep] =
     GenPrism[Step, StandardStep]
 
@@ -86,10 +81,10 @@ object Step {
       }
     }
 
-  val observeStatus: Lens[Step, ActionStatus] =
-    Lens[Step, ActionStatus] {
-      case s: StandardStep      => s.observeStatus
-      case s: NodAndShuffleStep => s.nsStatus.observing
+  val observeStatus: Optional[Step, ActionStatus] =
+    Optional[Step, ActionStatus] {
+      case s: StandardStep      => s.observeStatus.some
+      case s: NodAndShuffleStep => s.nsStatus.observing.some
     } { n => a =>
       a match {
         case s: StandardStep => StandardStep.observeStatus.set(n)(s)
@@ -98,10 +93,10 @@ object Step {
       }
     }
 
-  val configStatus: Lens[Step, List[(Resource, ActionStatus)]] =
-    Lens[Step, List[(Resource, ActionStatus)]] {
-      case s: StandardStep      => s.configStatus
-      case s: NodAndShuffleStep => s.configStatus
+  val configStatus: Optional[Step, List[(Resource, ActionStatus)]] =
+    Optional[Step, List[(Resource, ActionStatus)]] {
+      case s: StandardStep      => s.configStatus.some
+      case s: NodAndShuffleStep => s.configStatus.some
     } { n => a =>
       a match {
         case s: StandardStep      => StandardStep.configStatus.set(n)(s)
@@ -121,14 +116,14 @@ object Step {
 
   implicit class StepOps(val s: Step) extends AnyVal {
     def flipBreakpoint: Step = s match {
-      case st: StandardStep      => StandardStep.breakpoint.modify(!_)(st)
-      case st: NodAndShuffleStep => NodAndShuffleStep.breakpoint.modify(!_)(st)
+      case st: StandardStep      => StandardStep.breakpoint.negate(st)
+      case st: NodAndShuffleStep => NodAndShuffleStep.breakpoint.negate(st)
       case st                    => st
     }
 
     def flipSkip: Step = s match {
-      case st: StandardStep      => StandardStep.skip.modify(!_)(st)
-      case st: NodAndShuffleStep => NodAndShuffleStep.skip.modify(!_)(st)
+      case st: StandardStep      => StandardStep.skip.negate(st)
+      case st: NodAndShuffleStep => NodAndShuffleStep.skip.negate(st)
       case st                    => st
     }
 
@@ -154,6 +149,11 @@ object Step {
     def isRunning: Boolean = s.status match {
       case StepState.Running => true
       case _                 => false
+    }
+
+    def runningOrComplete: Boolean = s.status match {
+      case StepState.Running | StepState.Completed => true
+      case _                                       => false
     }
 
     def isObserving: Boolean = s match {
--- a/modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
@@ -6,8 +6,12 @@ package seqexec.model
 import cats._
 import cats.implicits._
 import seqexec.model.enum._
+import monocle.Prism
+import monocle.Lens
+import monocle.macros.Lenses
+import monocle.macros.GenPrism
 
-sealed trait Step {
+sealed trait Step extends Product with Serializable {
   def id: StepId
   def config: StepConfig
   def status: StepState
@@ -17,40 +21,123 @@ sealed trait Step {
 }
 
 object Step {
-  val Zero: Step = StandardStep(id = -1, config = Map.empty, status = StepState.Pending, breakpoint = false, skip = false, fileId = None, configStatus = Nil, observeStatus = ActionStatus.Pending)
+  val Zero: Step = StandardStep(id = -1,
+                                config        = Map.empty,
+                                status        = StepState.Pending,
+                                breakpoint    = false,
+                                skip          = false,
+                                fileId        = None,
+                                configStatus  = Nil,
+                                observeStatus = ActionStatus.Pending)
+
+  val standardStepP: Prism[Step, StandardStep] =
+    GenPrism[Step, StandardStep]
+
+  val nsStepP: Prism[Step, NodAndShuffleStep] =
+    GenPrism[Step, NodAndShuffleStep]
+
+  val status: Lens[Step, StepState] =
+    Lens[Step, StepState] {
+      _.status
+    } { n => a =>
+      a match {
+        case s: StandardStep      => StandardStep.status.set(n)(s)
+        case s: NodAndShuffleStep => NodAndShuffleStep.status.set(n)(s)
+      }
+    }
+
+  val config: Lens[Step, StepConfig] =
+    Lens[Step, StepConfig] {
+      _.config
+    } { n => a =>
+      a match {
+        case s: StandardStep      => StandardStep.config.set(n)(s)
+        case s: NodAndShuffleStep => NodAndShuffleStep.config.set(n)(s)
+      }
+    }
+
+  val id: Lens[Step, StepId] =
+    Lens[Step, StepId] {
+      _.id
+    } { n => a =>
+      a match {
+        case s: StandardStep      => StandardStep.id.set(n)(s)
+        case s: NodAndShuffleStep => NodAndShuffleStep.id.set(n)(s)
+      }
+    }
+
+  val skip: Lens[Step, Boolean] =
+    Lens[Step, Boolean] {
+      _.skip
+    } { n => a =>
+      a match {
+        case s: StandardStep      => StandardStep.skip.set(n)(s)
+        case s: NodAndShuffleStep => NodAndShuffleStep.skip.set(n)(s)
+      }
+    }
+
+  val breakpoint: Lens[Step, Boolean] =
+    Lens[Step, Boolean] {
+      _.breakpoint
+    } { n => a =>
+      a match {
+        case s: StandardStep      => StandardStep.breakpoint.set(n)(s)
+        case s: NodAndShuffleStep => NodAndShuffleStep.breakpoint.set(n)(s)
+      }
+    }
+
+  val observeStatus: Lens[Step, ActionStatus] =
+    Lens[Step, ActionStatus] {
+      case s: StandardStep      => s.observeStatus
+      case s: NodAndShuffleStep => s.nsStatus.observing
+    } { n => a =>
+      a match {
+        case s: StandardStep => StandardStep.observeStatus.set(n)(s)
+        case s: NodAndShuffleStep =>
+          (NodAndShuffleStep.nsStatus ^|-> NodAndShuffleStatus.observing).set(n)(s)
+      }
+    }
+
+  val configStatus: Lens[Step, List[(Resource, ActionStatus)]] =
+    Lens[Step, List[(Resource, ActionStatus)]] {
+      case s: StandardStep      => s.configStatus
+      case s: NodAndShuffleStep => s.configStatus
+    } { n => a =>
+      a match {
+        case s: StandardStep      => StandardStep.configStatus.set(n)(s)
+        case s: NodAndShuffleStep => NodAndShuffleStep.configStatus.set(n)(s)
+      }
+    }
 
   implicit val equal: Eq[Step] =
     Eq.instance {
       case (x: StandardStep, y: StandardStep) =>
-        (x.id === y.id) &&
-        (x.config === y.config) &&
-        (x.status === y.status) &&
-        (x.breakpoint === y.breakpoint) &&
-        (x.skip === y.skip) &&
-        (x.fileId === y.fileId) &&
-        (x.configStatus === y.configStatus) &&
-        (x.observeStatus === y.observeStatus)
+        x === y
+      case (x: NodAndShuffleStep, y: NodAndShuffleStep) =>
+        x === y
       case _ =>
         false
     }
 
   implicit class StepOps(val s: Step) extends AnyVal {
     def flipBreakpoint: Step = s match {
-      case st: StandardStep => st.copy(breakpoint = !st.breakpoint)
-      case st               => st
+      case st: StandardStep      => StandardStep.breakpoint.modify(!_)(st)
+      case st: NodAndShuffleStep => NodAndShuffleStep.breakpoint.modify(!_)(st)
+      case st                    => st
     }
 
     def flipSkip: Step = s match {
-      case st: StandardStep => st.copy(skip = !st.skip)
-      case st               => st
+      case st: StandardStep      => StandardStep.skip.modify(!_)(st)
+      case st: NodAndShuffleStep => NodAndShuffleStep.skip.modify(!_)(st)
+      case st                    => st
     }
 
     def file: Option[String] = None
 
     def canSetBreakpoint(i: Int, firstRunnable: Int): Boolean = s.status match {
-      case StepState.Pending | StepState.Skipped
-        | StepState.Paused | StepState.Running => i > firstRunnable
-      case _                                   => false
+      case StepState.Pending | StepState.Skipped | StepState.Paused |
+          StepState.Running => i > firstRunnable
+      case _ => false
     }
 
     def canSetSkipmark: Boolean = s.status match {
@@ -70,8 +157,9 @@ object Step {
     }
 
     def isObserving: Boolean = s match {
-      case StandardStep(_, _, _, _, _, _, _, o) => o === ActionStatus.Running
-      case _                                    => false
+      case StandardStep(_, _, _, _, _, _, _, o)      => o === ActionStatus.Running
+      case NodAndShuffleStep(_, _, _, _, _, _, _, o) => o.observing === ActionStatus.Running
+      case _                                         => false
     }
 
     def isObservePaused: Boolean = s match {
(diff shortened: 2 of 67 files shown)


modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
```diff
@@ -80,29 +168,89 @@ object Step {
...
+
+object NodAndShuffleStatus {
+
+  implicit val equalNodAndShuffleStatus: Eq[NodAndShuffleStatus] =
+    Eq.by(x => (x.observing))
+}
+
 // Other kinds of Steps to be defined.
+@Lenses
+final case class NodAndShuffleStep(
```
cquiroz: Client side representation of a Nod and shuffle step, it needs some more data than present now

modules/seqexec/model/src/main/scala/seqexec/model/enum/ObserveCommandResult.scala
```diff
@@ -9,11 +9,12 @@ sealed trait ObserveCommandResult extends Product with Serializable
 
 object ObserveCommandResult {
   case object Success extends ObserveCommandResult
-  case object Paused extends ObserveCommandResult
+  case object Paused  extends ObserveCommandResult
   case object Stopped extends ObserveCommandResult
   case object Aborted extends ObserveCommandResult
+  case object Partial extends ObserveCommandResult // Used for N&S
```
cquiroz: This lets us differentiate between a human induced pause and one from N&S

modules/seqexec/model/src/test/scala/seqexec/model/SeqexecModelArbitraries.scala
```diff
@@ -105,67 +105,6 @@ trait SeqexecModelArbitraries extends ArbObservation with ArbStepState {
     } yield s
   }
 
-  def asciiStr: Gen[String] =
```
cquiroz: Moved to another file

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentActions.scala
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+  * Algebra to generate actions for an observation.
+  * Most instruments behave the same but in some cases we need to customize
+  * behavior. The two prime examples are:
+  * GPI A&C
+  * GMOS N&S
+  * In both cases the InstrumentActions for the instrument can provide the correct behavior
+  */
+trait InstrumentActions[F[_]] {
```
cquiroz: This is a key change. We are now delegating the construction of actions to each instrument

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentActions.scala
```diff
@@ -0,0 +1,108 @@
...
+        Stream.emit(Result.Partial(FileIdAllocated(fileId))) ++ observationCommand
+      case Left(e: SeqexecFailure) => Stream.emit(Result.Error(SeqexecFailure.explain(e)))
+      case Left(e: Throwable) => Stream.emit(Result.Error(SeqexecFailure.explain(SeqexecFailure.SeqexecException(e))))
+    }
+  }
+
+  /**
+    * Default Actions for most instruments it basically delegates to ObserveActions
+    */
+  def defaultInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger]
```
cquiroz: Most instruments just need to use the default `InstrumentAction`

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -20,15 +20,22 @@ trait InstrumentSystem[F[_]] extends System[F] with InstrumentGuide {
...
+  def observe(config: Config): Kleisli[F, ImageFileId, ObserveCommandResult]
+
   //Expected total observe lapse, used to calculate timeout
   def calcObserveTime(config: Config): F[Time]
+
   def keywordsClient: KeywordsClient[F]
+
   def observeProgress(total: Time, elapsed: InstrumentSystem.ElapsedTime): Stream[F, Progress]
+
+  def instrumentActions(config: Config): InstrumentActions[F]
```
cquiroz: Instrument can provide a `InstrumentActions`

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -4,129 +4,201 @@
...
+  ): Stream[F, Result[F]] =
+    for {
+      ot <- Stream.eval(env.inst.calcObserveTime(env.config))
+      pr <- env.inst.observeProgress(ot, ElapsedTime(0.0.seconds))
+    } yield Result.Partial(pr)
+
+  /**
+    * Tell each subsystem that an observe will start
+    */
+  def notifyObserveStart[F[_]: Applicative](
```
cquiroz: Many methods used internally in `observed` are now publicly exposed to let `InstrumentAction` instance use them

modules/seqexec/server/src/main/scala/seqexec/server/ObserveEnvironment.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import cats.data.Reader
+import gem.Observation
+import edu.gemini.spModel.config2.Config
+import seqexec.server.keywords._
+
+/**
+  * Describes the parameters for an observation
+  */
+final case class ObserveEnvironment[F[_]](
```
cquiroz: We need a bunch of parameters to setup an obs. Passing them in a data class makes the signatures simpler

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -77,13 +84,11 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
 
-      def rest(ctx: HeaderExtraData): List[ParallelActions[IO]] =
-        (stepType.includesObserve).option {
-          List(
-            NonEmptyList.one(Action(ActionType.Observe, observe(systems, config, obsId, inst, sys.filterNot(inst.equals),
-              headers)(ctx), Action.State(ActionState.Idle, Nil)))
-          )
-        }.orEmpty
+      def rest(ctx: HeaderExtraData): List[ParallelActions[IO]] = {
+        val env = ObserveEnvironment(systems, config, stepType, obsId, inst, sys.filterNot(inst.equals), headers, ctx)
```
cquiroz: Key change. We now delegate action creation to the instruments

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -544,121 +545,21 @@ object SeqexecEngine extends SeqexecConfiguration {
   def splitWhere[A](l: List[A])(p: A => Boolean): (List[A], List[A]) =
     l.splitAt(l.indexWhere(p))
 
-  def splitAfter[A](l: List[A])(p: A => Boolean): (List[A], List[A]) =
```
cquiroz: These methods were spun off to another class to make this file shorter

modules/seqexec/server/src/main/scala/seqexec/server/StepsView.scala
```diff
@@ -0,0 +1,182 @@
...
+import seqexec.model.enum.Instrument._
+import seqexec.model.enum.SystemName
+import seqexec.model.StepState
+import seqexec.engine
+import seqexec.engine.Action.ActionState
+import seqexec.engine.Action
+import seqexec.engine.ParallelActions
+import seqexec.server.gmos.Gmos
+
+sealed trait StepsView[F[_]] {
```
cquiroz: It can define a view for different types of steps like N&S

modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2.scala
```diff
@@ -30,7 +31,7 @@ import squants.time.{Seconds, Time}
 
 import scala.reflect.ClassTag
 
-final case class Flamingos2[F[_]: Sync: Timer](f2Controller: Flamingos2Controller[F], dhsClient: DhsClient[F]) extends DhsInstrument[F] with InstrumentSystem[F] {
+final case class Flamingos2[F[_]: Sync: Timer: Logger](f2Controller: Flamingos2Controller[F], dhsClient: DhsClient[F]) extends DhsInstrument[F] with InstrumentSystem[F] {
```
cquiroz: We should remove `Sync` in future refactorings if possible and like with a more restricted constraint and `Logger`

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerSim.scala
```diff
@@ -3,24 +3,89 @@
...
+    NSObsState.current ^<-? some ^|-> NSCurrent.fileId
+
+  val exposureCount: Optional[NSObsState, Int] =
+    NSObsState.current ^<-? some ^|-> NSCurrent.exposureCount
+
+  val expTime: Optional[NSObsState, Time] =
+    NSObsState.current ^<-? some ^|-> NSCurrent.expTime
+}
+
+object GmosControllerSim {
```
cquiroz: The simulator got a lot more complex to support N&S. It needs to keep an internal state

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -0,0 +1,206 @@
...
+import seqexec.server.ObserveActions
+import seqexec.server.ObserveContext
+import seqexec.server.SeqexecFailure
+import seqexec.server.ObserveActions._
+import seqexec.server.gmos.GmosController.Config._
+
+/**
+  * Gmos needs different actions for N&S
+  */
+class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger, A <: GmosController.SiteDependentTypes](
```
cquiroz: In this class we can create a list of actions in case we are doing N&S

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -0,0 +1,206 @@
...
+    rows:      Int,
+    post:      (Stream[F, Result[F]], ObserveEnvironment[F]) => Stream[F, Result[F]]
+  ): List[ParallelActions[F]] =
+    Gmos.NsSequence.zipWithIndex
+      .map {
+        case (stage, i) =>
+          val rowsToShuffle = if (stage === StageA) 0 else rows
+          List(
+            // Configure rows to shuffle
+            NonEmptyList.one(
```
cquiroz: Configure of rows to shuffle, we should do the TCS in parallel

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -0,0 +1,206 @@
...
+              inst
+                .configureShuffle(rowsToShuffle)
+                .as(Response.Configured(inst.resource))
+                .toAction(ActionType.Configure(inst.resource))
+              // TODO confgure the TCS
+            ),
+            // Do an obs observe/continue
+            NonEmptyList.one(
```
cquiroz: Send observe or continue as needed

modules/seqexec/server/src/main/scala/seqexec/server/gpi/GpiInstrumentActions.scala
```diff
@@ -0,0 +1,41 @@
...
+import seqexec.engine.Result
+import seqexec.server.StepType
+import seqexec.server.InstrumentActions
+import seqexec.server.ObserveActions
+import seqexec.server.ObserveEnvironment
+
+/**
+  * Gpi needs different actions for A&C
+  */
+class GpiInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger]
```
cquiroz: Gpi also produces custom Actions for A&C but it is much simpler, we just don't do anything

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -217,7 +224,7 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
              tio: Timer[IO]
   ): EngineState => Option[Stream[IO, executeEngine.EventType]] = st => {
 
-    def resumeIO(c: ObserveContext[IO], resumeCmd: IO[ObserveCommandResult]): IO[Result[IO]] =
+    def resumeIO(c: ObserveContext[IO], resumeCmd: IO[ObserveCommandResult]): IO[Stream[IO, Result[IO]]] =
```
tpolecat: This is a slightly weird type. You can wrap it with `Stream.eval(...).flatten` and get `Stream[IO, Result[IO]]` instead.

modules/seqexec/server/src/main/scala/seqexec/server/StepsView.scala
```diff
@@ -0,0 +1,182 @@
...
+        s ++ a ++ r
+    }
+
+    // Find out systems in the future
+    val presentSystems = configStatus.keys.toList
+    // Calculate status of pending items
+    val systemsPending = pending.map {
+      s => separateActions(s).bimap(_.map(_.kind).flatMap(kindToResource), _.map(_.kind).flatMap(kindToResource))
+    }.flatMap {
+      x => x._1.tupleRight(ActionStatus.Pending) ::: x._2.tupleRight(ActionStatus.Completed)
```
tpolecat: In cases like this I would do `flatMap { case (a, b) => ...` to avoid the `._1`, `._2` projections.

modules/seqexec/server/src/main/scala/seqexec/server/StepsView.scala
```diff
@@ -0,0 +1,182 @@
...
+            _.flatMap(a => kindToResource(a.kind).tupleRight(ActionStatus.Running)).toMap,
+            _.flatMap(r => kindToResource(r.kind).tupleRight(ActionStatus.Completed)).toMap)
+        s ++ a ++ r
+    }
+
+    // Find out systems in the future
+    val presentSystems = configStatus.keys.toList
+    // Calculate status of pending items
+    val systemsPending = pending.map {
+      s => separateActions(s).bimap(_.map(_.kind).flatMap(kindToResource), _.map(_.kind).flatMap(kindToResource))
```
tpolecat: Worthwhile abstracting this operation out? It occurs two more times in `pendingConfigStatus`.

modules/seqexec/server/src/main/scala/seqexec/server/StepsView.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  private def defaultStepsView[F[_]]: StepsView[F] = new StepsView[F] {
+    def stepView(
+      stepg: SequenceGen.StepGen[F],
+      step: engine.Step[F],
+      altCfgStatus: List[(Resource, ActionStatus)]
+    ): Step = {
+      val status = engine.Step.status(step)
+      val configStatus =
+        if (status === StepState.Completed || status === StepState.Running) {
```
tpolecat: Kinda prefer pattern-matches in cases like this.

tpolecat: Also it's copy-pasted in `gmosStepsView` below.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -200,6 +215,19 @@ object Gmos {
       ns               <- (if (useNS) nodAndShuffle(config) else NSConfig.NoNodAndShuffle.asRight)
     } yield ns).leftMap(e => SeqexecFailure.Unexpected(ConfigUtilOps.explain(e)))
 
+  def nsConfigF[F[_]: ApplicativeError[?[_], Throwable]](config: Config): F[NSConfig] =
+    ApplicativeError[F, Throwable]
+      .catchNonFatal(
+        nsConfig(config).getOrElse(NSConfig.NoNodAndShuffle)
```
tpolecat: I think we chatted about this, but can you confirm that this and `expTime` are deterministic (i.e., we could memoize the results of these method calls)? If not we need to use `Sync` here.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -0,0 +1,206 @@
...
+    ObserveActions.observationProgressStream(env)
+
+  // This tail is based on ObserveActions.observeTail
+  // But it can understand how to process Partial observations
+  // And can eventually return more than one result
+  private def observeTail(
+    fileId: ImageFileId,
+    dataId: String,
+    env:    ObserveEnvironment[F]
+  )(r:      ObserveCommandResult): F[Stream[F, Result[F]]] =
```
tpolecat: Similar to the comment about `IO[Stream[IO` above. Are you sure you don't want to flatten to `Stream[F`?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -0,0 +1,206 @@
...
+              NonEmptyList.one(
+                Action(ActionType.Undefined,
+                       Stream.emits[F, Result[F]](
+                         List(Result.Partial(NSStart),
+                              Result.OK(Response.Ignored))
+                       ),
+                       Action.State(Action.ActionState.Idle, Nil))
+              ) ::
+                actionPositions(env, rows, post) // Add steps for each cycle
+                  .replicateA(cycles)
```
tpolecat: :fire:

modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
```diff
@@ -17,40 +21,123 @@ sealed trait Step {
 }
 
 object Step {
-  val Zero: Step = StandardStep(id = -1, config = Map.empty, status = StepState.Pending, breakpoint = false, skip = false, fileId = None, configStatus = Nil, observeStatus = ActionStatus.Pending)
+  val Zero: Step = StandardStep(id = -1,
```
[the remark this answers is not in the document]
cquiroz: Now that you mention it I'm not even sure we use `Zero`. perhaps it can be removed, I assume we used it as a default

modules/seqexec/server/src/main/scala/seqexec/server/ObserveEnvironment.scala
```diff
@@ -0,0 +1,23 @@
...
+  systems:  Systems[F],
+  config:   Config,
+  stepType: StepType,
+  obsId:    Observation.Id,
+  inst:     InstrumentSystem[F],
+  otherSys: List[System[F]],
+  headers:  Reader[HeaderExtraData, List[Header[F]]],
```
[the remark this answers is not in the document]
cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/StepsView.scala
cquiroz: Good idea

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
cquiroz: I'll revisit this. I was also a bit worried about this type, but it got drowned on the large set of issues being handled

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
cquiroz: So the same `config` param will always return the same result of `nsConfig(config)`

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
cquiroz: What do you mean?

tpolecat: Sorry, I just really like `replicateA`.

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -4,129 +4,201 @@
...
-        Stream.emit(Result.Partial(FileIdAllocated(id))) ++
-          observationProgressStream.mergeHaltR(observationCommand)
-      case Left(e: SeqexecFailure)   => Stream.emit(Result.Error(SeqexecFailure.explain(e)))
-      case Left(e: Throwable)   => Stream.emit(Result.Error(SeqexecFailure.explain(SeqexecFailure.SeqexecException(e))))
+      )
+      .widenRethrowT
+
+  /**
+    * Preamble for observations. It tells the odb, the subsystems
+    * send the start headers and finall ysends an observe
```
jluhrs: "finally sends"

build.sbt
```diff
@@ -515,6 +515,7 @@ lazy val seqexec_server = project
   .settings(
     addCompilerPlugin(Plugins.paradisePlugin),
     addCompilerPlugin(Plugins.kindProjectorPlugin),
+    addCompilerPlugin(Plugins.betterMonadicForPlugin),
```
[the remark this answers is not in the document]
cquiroz: It lets you write for loops a bit better, for example you can split `IO[(A, B)]`, see https://github.com/oleg-py/better-monadic-for

modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
```diff
@@ -6,8 +6,12 @@ package seqexec.model
...
+import monocle.Prism
+import monocle.Lens
+import monocle.macros.Lenses
+import monocle.macros.GenPrism
 
-sealed trait Step {
+sealed trait Step extends Product with Serializable {
   def id: StepId
   def config: StepConfig
   def status: StepState
```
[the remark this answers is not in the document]
cquiroz: I think the old convention is that at the trait level they are `def` and implementations can decide to override with `val`

modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
```diff
@@ -17,40 +21,123 @@ sealed trait Step {
...
+    Lens[Step, StepId] {
+      _.id
+    } { n => a =>
+      a match {
+        case s: StandardStep      => StandardStep.id.set(n)(s)
+        case s: NodAndShuffleStep => NodAndShuffleStep.id.set(n)(s)
+      }
+    }
+
+  val skip: Lens[Step, Boolean] =
```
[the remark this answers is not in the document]
cquiroz: I understand that, perhaps they should be an `Optional`?

modules/seqexec/server/src/main/scala/seqexec/server/ObserveEnvironment.scala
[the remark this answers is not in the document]
cquiroz: I removed all uses of `Reader` in this class, as you mentioned it is not needed

modules/seqexec/model/src/main/scala/seqexec/model/Step.scala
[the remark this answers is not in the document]
cquiroz: breakpoint and skip are present in the `trait` I think a lens is fine. `observeStatus/configStatus` could be an optional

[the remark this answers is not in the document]
jluhrs: Ups, my bad. I didn't see them.

cquiroz:
I resolved most of the comments on the last commit

PR #1057, opened 2019-09-03
SEQNG-1045 Fixed bug that prevented to load non calibration sequences.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -353,17 +353,11 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
     sys: InstrumentSystem[IO]
   ): TrySeq[List[System[IO]]] = {
     stepType match {
-      case StepType.CelestialObject(inst) => for {
-        tcs  <- getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), useGaos = false, sys,
-          TcsController.LightSource.Sky, config)
-        gcal <- Gcal.fromConfig(systems.gcal, site == Site.GS)(config)
-      } yield List(sys, tcs, gcal)
+      case StepType.CelestialObject(inst) => getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), useGaos = false, sys,
```
cquiroz: There are a bit hard to read, could we perhaps write the `.map` in a different line and the parametrs for `getTcs` each on its own line?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -353,17 +353,11 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
+          TcsController.LightSource.Sky, config).map{ List(sys, _, Gcal.defaultGcal(systems.gcal)) }
 
-      case StepType.NodAndShuffle(inst)   => for {
-        tcs  <- getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), useGaos = false, sys,
-          TcsController.LightSource.Sky, config)
-        gcal <- Gcal.fromConfig(systems.gcal, site == Site.GS)(config)
-      } yield List(sys, tcs, gcal)
+      case StepType.NodAndShuffle(inst)   => getTcs(inst.hasOI.fold(allButGaos, allButGaosNorOi), useGaos = false, sys,
+          TcsController.LightSource.Sky, config).map{ List(sys, _, Gcal.defaultGcal(systems.gcal)) }
```
cquiroz: Why don't you need `site` anymore?

jluhrs: There is a difference in when the IR lamp is turned on at both sites, but it does not matter when GCAL is not used.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: Done

PR #1061, opened 2019-09-04
SEQNG-332 Implemented telescope nod for Gmos N&S

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -69,7 +69,7 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
       headers: HeaderExtraData => List[Header[IO]],
       stepType: StepType
     ): SequenceGen.StepGen[IO] = {
-      val ia = inst.instrumentActions(config)
+      val ia = inst.instrumentActions(config, sys.collectFirst{case x: Tcs[IO] => x})
```
cquiroz: Doesn't this give you an erasure error?

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -29,7 +30,7 @@ trait InstrumentSystem[F[_]] extends System[F] with InstrumentGuide {
 
   def observeProgress(total: Time, elapsed: InstrumentSystem.ElapsedTime): Stream[F, Progress]
 
-  def instrumentActions(config: Config): InstrumentActions[F]
+  def instrumentActions(config: Config, tcsO: Option[Tcs[F]]): InstrumentActions[F]
```
cquiroz: Could we instead pass `ObserveEnvironment[F]`. it contains both `config` and `systems: List[System[F]]` from where you can find `Tcs[F]`

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -135,22 +140,31 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger, A <: Gmos
...
             // Configure rows to shuffle
-            NonEmptyList.one(
+            NonEmptyList(
               inst
                 .configureShuffle(rowsToShuffle)
                 .as(Response.Configured(inst.resource))
-                .toAction(ActionType.Configure(inst.resource))
-              // TODO confgure the TCS
+                .toAction(ActionType.Configure(inst.resource)),
+              (tcsO, nsPositionO).mapN{ case (tcs, nsPos) =>
```
cquiroz: How is it in case of a dark? Would `nsPosition0` be `None`?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/Tcs.scala
```diff
@@ -6,12 +6,14 @@ package seqexec.server.tcs
...
-import seqexec.model.enum.M1Source
-import seqexec.model.M1GuideConfig
-import seqexec.model.M2GuideConfig
+import seqexec.model.{M1GuideConfig, M2GuideConfig, TelescopeGuideConfig}
+import seqexec.model.enum.{M1Source, NodAndShuffleStage, TipTiltSource}
 import seqexec.server.tcs.TcsController._
+import seqexec.server.{ConfigResult, System}
+
+trait Tcs[F[_]] extends System[F] {
+  def nod(stage: NodAndShuffleStage, offset: InstrumentOffset, guided: Boolean): F[ConfigResult[F]]
```
cquiroz: nice

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSouthControllerEpics.scala
```diff
@@ -32,6 +33,13 @@ class TcsSouthControllerEpics private (guideConfigDb: GuideConfigDb[IO]) extends
   override def notifyObserveStart: IO[Unit] = TcsControllerEpicsCommon.notifyObserveStart
 
   override def notifyObserveEnd: IO[Unit] = TcsControllerEpicsCommon.notifyObserveEnd
+
+  override def nod(subsystems: NonEmptySet[Subsystem], tcsConfig: TcsSouthConfig)
+                  (stage: NodAndShuffleStage, offset: InstrumentOffset, guided: Boolean)
+  : IO[Unit] = tcsConfig match {
+    case c: BasicTcsConfig => TcsControllerEpicsCommon.nod(subsystems, offset, guided, c)
+    case _: TcsSouthAoConfig => SeqexecFailure.Execution("N&S not supported when using Altair").raiseError[IO, Unit]
```
cquiroz: Altair at GS?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: I was afraid it would, but it compiles.

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
jluhrs: That is not a bad idea, and having all the systems involved in the step may be useful for future features.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
jluhrs: I think we need a new type of step, but the implementation will be easy.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSouthControllerEpics.scala
jluhrs: 😱

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
tpolecat: That's unsound, you need a better solution for that.

tpolecat: Sorry, I misunderstood that code. It's actually ok, it's just weird. But Scala handles that case ok. You would get a warning for anything other than `IO` in that position.

cquiroz: I recommended to parametrize the method anyway and move it to `ObserveEnvironment[F]`

PR #1063, opened 2019-09-05
SEQNG-1046 Implemented GMOS N&S FITS keywords.

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
```diff
@@ -26,18 +32,66 @@ final case class GmosObsKeywordsReader[F[_]: Sync](config: Config) {
...
       .map(_.toBoolean)
       .safeValOrDefault
+
+  def nodMode: F[String] = "STANDARD".pure[F]
+
+  def nodPix: F[Int] = Gmos.nodAndShuffle(config).map(_.rows).explainExtractError
+
+  def nodCount: F[Int] = Gmos.nodAndShuffle(config).map(_.cycles).explainExtractError
+
+  def nodAxOff: F[Double] = Gmos.nodAndShuffle(config).explainExtractError
```
tpolecat: These four methods are identical other than the stage and angle selector. I think you forgot to change the error message in the last two when you copy/pasted. Can we factor out one method and call it with different arguments?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosHeader.scala
```diff
@@ -96,21 +122,23 @@ object GmosHeader {
             buildString(gmosReader.dcName, KeywordName.GMOSDC),
             buildString(gmosReader.detectorType, KeywordName.DETTYPE),
             buildString(gmosReader.detectorId, KeywordName.DETID),
-            buildDouble(gmosReader.exposureTime, KeywordName.EXPOSURE),
+            buildDouble(
+              gmosObsReader.isNS.ifM(
+                (gmosObsReader.nodCount, gmosReader.exposureTime).mapN{ case (c, t) => 2 * c * t },
```
cquiroz: I tend to put these calculations on the keyword reader and leave the `Header` class just as set of values and keyword names. WDYT?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
```diff
@@ -26,18 +32,66 @@ final case class GmosObsKeywordsReader[F[_]: Sync](config: Config) {
...
+        .getOrElse(SeqexecFailure.Unexpected("Cannot find stage A parameters in step configuration.")
+          .raiseError[F, Double]
+        )
+    )
+
+  def nodBxOff: F[Double] = Gmos.nodAndShuffle(config).explainExtractError
+    .flatMap(
+      _.positions.find(_.stage === StageB)
+        .map(x => (x.offset.p.toAngle.toMicroarcseconds.toDouble/1e6).pure[F])
+        .getOrElse(SeqexecFailure.Unexpected("Cannot find stage A parameters in step configuration.")
```
cquiroz: stage B?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
```diff
@@ -26,18 +32,66 @@ final case class GmosObsKeywordsReader[F[_]: Sync](config: Config) {
...
+        )
+    )
+
+  def isNS: F[Boolean] = config.extractInstAs[java.lang.Boolean](USE_NS_PROP).map(_.booleanValue).explainExtractError
+
+}
+
+object GmosObsKeywordsReader {
+
+  private implicit class ExplainExtractError[F[_]: Sync, A](v: Either[ExtractFailure, A]) {
```
cquiroz: Do you need `Sync` here? I think `MonadError` is enough

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosHeader.scala
jluhrs: I agree in general, but the problem here is that the calculation combines values from two readers.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
jluhrs: Y tried with `MonadError` first, but then I got compilation errors about ambiguous implicits.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosHeader.scala
cquiroz: Good point

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosKeywordReader.scala
cquiroz: That may mean you are missing to explicity set the `F`. e.g. `widenRethrowT[F]`?

jluhrs: I rearranged the type parameters and put some explicitly, and now it compiles with `MonadError`

PR #1074, opened 2019-09-10
SEQNG-1052 Transform N&S biases, flats and arcs into normal ones.

jluhrs:
Flats, arcs and biases steps inserted in a N&S sequence have N&S parameters that the Seqexec might try to apply. This is not supported, and the old Seqexec deals with them by treating them as no N&S steps.
I tried different approaches, like trying to cancel the effects of the N&S parameters everywhere they are used, but at the end I went for the simpler solution of fixing the step configuration by overriding the `useNS` parameter. The class used to read the step configuration is a mutable Java class (`Config`). To avoid mutation I built a wrapper, and passed it instead of the original `Config`.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/CleanConfig.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/CleanConfig.scala
@@ -1,0 +1,71 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.server
+
+import cats.implicits._
+import edu.gemini.spModel.config2.{Config, ItemEntry, ItemKey}
+import seqexec.server.ConfigUtilOps._
+import edu.gemini.spModel.gemini.gmos.InstGmosCommon.USE_NS_PROP
+import edu.gemini.spModel.obscomp.InstConstants.{ARC_OBSERVE_TYPE, FLAT_OBSERVE_TYPE, OBSERVE_TYPE_PROP, BIAS_OBSERVE_TYPE}
+import edu.gemini.spModel.seqcomp.SeqConfigNames.INSTRUMENT_KEY
+import seqexec.model.StepConfig
+import seqexec.model.enum.SystemName
+
+import scala.collection.breakOut
+
+/*
+ * CleanConfig is a wrapper over Config that allows to override some of the Config parameters. It allows to change some
+ * of the Config parameters without mutation.
+ * It is used to correct some inconsistencies in the sequences coming from the ODB. It has only one of those corrections
+ * now, but it allows for more to be added in the future.
+ */
+class CleanConfig(config: Config, overrides: Map[ItemKey, AnyRef]) {
+  def itemValue(k: ItemKey): Option[AnyRef] = overrides.get(k).orElse(Option(config.getItemValue(k)))
+
+  private def sanitizeValue(s: Any): String = s match {
+    case x: edu.gemini.shared.util.immutable.Some[_] => s"${x.getValue}"
+    case _: edu.gemini.shared.util.immutable.None[_] => "None"
+    case _ => s"$s"
+  }
+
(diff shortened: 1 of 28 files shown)


tpolecat:
> Flats, arcs and biases steps inserted in a N&S sequence have N&S parameters that the Seqexec might try to apply.

Why? Can we fix it at the source and avoid hacking around it?

jluhrs:
@tpolecat Yes, I think this would be better fixed in the ODB. I don't like the Seqexec having to second guess the sequence definition. But until then, I think this is a acceptable solution. It is bounded and easy to modify or take out if/when the ODB is changed.

modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
```diff
@@ -85,60 +85,77 @@ object ConfigUtilOps {
...
+      c.extractInstAs[JInt](property)
+        .map(_.toInt.some)
+        .map(_.map(tag[A][Int]))
+        .recoverOption
+
+  }
+
+  implicit class ConfigOps(val c: Config) extends AnyVal {
+
     private def sanitizeValue(s: Any): String = s match {
```
cquiroz: This seems duplicated above

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -164,23 +166,23 @@ object Gmos {
   // In principle we'd expect the OT to send the sequence but instead the
   // sequence is hardcoded in the seqexec and we only read the positions from
   // the OT
-  val NsSequence = NonEmptyList.of(StageB, StageA, StageA, StageB)
+  val NsSequence: NonEmptyList[NodAndShuffleStage] = NonEmptyList.of(StageB, StageA, StageA, StageB)
```
cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/CleanConfig.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import scala.collection.breakOut
+
+/*
+ * CleanConfig is a wrapper over Config that allows to override some of the Config parameters. It allows to change some
+ * of the Config parameters without mutation.
+ * It is used to correct some inconsistencies in the sequences coming from the ODB. It has only one of those corrections
+ * now, but it allows for more to be added in the future.
+ */
+class CleanConfig(config: Config, overrides: Map[ItemKey, AnyRef]) {
```
cquiroz: I wonder if this could be somehow `Map[ItemKey, _]`. I understand why it is `AnyRef` but now that we are going to the trouble of having `CleanConfig` we may try to tighten the screws

Perhaps it could be via a
`trait ItemValue[T]` and the map is `Map[ItemKey, ItemValue[_]]`?

I don't know. existentials give me the creeps. @tpolecat what do you think?

cquiroz: I think we could `CleanConfig` client side but it means getting rid of all old ocs2 references. Would that be hard?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
jluhrs: I cannot take the credit, Idea told me to do it :)

modules/seqexec/server/src/main/scala/seqexec/server/CleanConfig.scala
tpolecat: My suggestion here is to add a check (it could just be a runtime assertion, we need to add a test to be sure it works) to ensure that the runtime type of overridden values are assignable to the same runtime types of the Config's values. So at least we can be confident that we're not turning a filter into an angle or something. I don't think types can really help us here because the underlying config is untyped.

jluhrs: But all this will be solved when we move the Seqexec to the new OCS, right?

Right?

jluhrs:
About making this a full replacement of the java Config, with proper types. I think that is desirable, but out of the scope of this PR. And the problem will resolve itself when we move to the new OCS.

PR #1075, opened 2019-09-10
SEQNG-1055 Use dataId in call to sequenceStart.

jluhrs:
Fixed bug where the notification send to the ODB to signal the start of the sequence execution was not using the dataId.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -4,14 +4,14 @@
 package seqexec.server
 
 import cats._
-import cats.data.NonEmptySet
-import cats.data.NonEmptyList
+import cats.data.{EitherT, NonEmptyList, NonEmptySet}
 import cats.effect.{Concurrent, IO, Timer}
 import cats.effect.Sync
 import cats.effect.LiftIO
 import cats.implicits._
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -568,4 +568,14 @@ object SeqTranslate {
     def toAction(kind: ActionType): Action[F] = fromF[F](kind, x.map(r => Result.OK(Response.Configured(r.sys.resource))))
   }
 
+  def dataIdFromConfig[F[_]: MonadError[?[_], Throwable]](config: Config): F[DataId] =
```
cquiroz: This is already defined at `ObserveActions`. see https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala#L110

jluhrs: Not exactly. The one in `ObserveActions` takes a `ObserveEnvironment[F]`

cquiroz: Could we merge them?

jluhrs: Done

PR #1084, opened 2019-09-13
SEQNG-1048 Fixes actions for N&S on multiple cycles

cquiroz:
N&S with multiple cycles was not properly done, in part because of a misunderstanding  of the  `replicateA`, method. However the simple fix also showed some other issues which led to a larger rewrite.
N&S was originally done with multiple actions for each subexposure but instead it needs to be done with a single large action

More was discussed about this here:
https://gist.github.com/cquiroz/50ae3fe89711243a9f06a0515c257bad

The simulator also needed updates to support multiple cycles

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
@@ -149,7 +149,7 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Concurrent: Logge
     // Configure GMOS rows
     Stream.eval(
       inst.configureShuffle(rowsToShuffle).as(Result.Partial(NSRowsConfigure))
-    ).mergeHaltBoth(
+    ).merge(
       // TCS Nod
       (env.getTcs, nsPositionO).mapN {
         case (tcs, nsPos) =>
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
@@ -5,23 +5,41 @@ package seqexec.server.gmos
 
 import cats._
 import cats.implicits._
-import cats.data.NonEmptyList
 import fs2.Stream
 import io.chrisdavenport.log4cats.Logger
-import seqexec.model.ActionType
 import seqexec.model.dhs._
+import seqexec.model.enum.NodAndShuffleStage
 import seqexec.model.enum.NodAndShuffleStage._
-import seqexec.model.enum.{Guiding, ObserveCommandResult}
-import seqexec.engine.Action
+import seqexec.model.enum.Guiding
+import seqexec.model.enum.ObserveCommandResult
 import seqexec.engine.ParallelActions
 import seqexec.engine.Result
-import seqexec.server.{CleanConfig, FileIdAllocated, FileIdProvider, InstrumentActions, ObserveActions, ObserveContext, ObserveEnvironment, Response, StepType}
-import seqexec.server.SeqTranslate._
+import seqexec.server.CleanConfig
+import seqexec.server.FileIdAllocated
+import seqexec.server.FileIdProvider
+import seqexec.server.InstrumentActions
+import seqexec.server.ObserveActions
+import seqexec.server.ObserveContext
+import seqexec.server.ObserveEnvironment
+import seqexec.server.StepType
+import seqexec.server.InstrumentActions._
 import seqexec.server.ObserveActions._
-import seqexec.server.gmos.GmosController.Config.{NSConfig, NSPosition}
-import seqexec.server.tcs.TcsController.{InstrumentOffset, OffsetP, OffsetQ}
+import seqexec.server.gmos.GmosController.Config._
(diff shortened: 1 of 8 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -78,130 +96,128 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger, A <: Gmos
...
-                  env
+    env:       ObserveEnvironment[F],
+    post:      (Stream[F, Result[F]], ObserveEnvironment[F]) => Stream[F, Result[F]]
+  ): Stream[F, Result[F]] = {
+    val rowsToShuffle = if (sub.stage === StageA) 0 else rows
+    val nsPositionO   = positions.find(_.stage === sub.stage)
+    // Configure GMOS rows
+    Stream.eval(
+      inst.configureShuffle(rowsToShuffle).as(Result.Partial(NSRowsConfigure))
+    ) ++
```
jluhrs: `merge` instead of `++` allows to run the nod and shuffle in parallel.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -78,130 +96,128 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger, A <: Gmos
...
+                sub.stage,
+                InstrumentOffset(
+                  tag[OffsetP](nsPos.offset.p.toRadians.radians),
+                  tag[OffsetQ](nsPos.offset.q.toRadians.radians)
                 ),
-                Action.State(Action.ActionState.Idle, Nil)
+                nsPos.guide === Guiding.Guide
               )
-            )
+              .as(Result.Partial(NSTCSNod))
```
jluhrs: Shouldn't the result have the stage?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -78,130 +96,128 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger, A <: Gmos
...
+    fileId: ImageFileId,
+    env:    ObserveEnvironment[F],
+    post:   (Stream[F, Result[F]], ObserveEnvironment[F]) => Stream[F, Result[F]]
+  ): Stream[F, Result[F]] =
     Gmos
       .nsConfig(config)
-      .map {
+      .foldMap {
         case NSConfig.NoNodAndShuffle =>
-          // This shouldn't happen but we need to code it anyway
```
jluhrs: The data types used for GMOS configuration could be changed to avoid this.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/package.scala
```diff
@@ -8,6 +8,8 @@ import seqexec.engine.Result.PartialVal
 package gmos {
   // N&S Partials. TODO add more details about the current step
   case object NSStart extends PartialVal
+  case object NSRowsConfigure extends PartialVal
+  case object NSTCSNod extends PartialVal
```
jluhrs: `case class NSTCSNod(stage: NodAndShuffleStage)` ? I would expect you will need that for the UI.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
cquiroz: I want to add more data once the UI is being developed. but yes we should add more info about where we are

modules/seqexec/server/src/main/scala/seqexec/server/gmos/package.scala
cquiroz: See above

cquiroz:
Yes it will, the plan was not to care too much about UI at this level

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -78,130 +97,128 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Logger, A <: Gmos
...
-                  env
+    env:       ObserveEnvironment[F],
+    post:      (Stream[F, Result[F]], ObserveEnvironment[F]) => Stream[F, Result[F]]
+  ): Stream[F, Result[F]] = {
+    val rowsToShuffle = if (sub.stage === StageA) 0 else rows
+    val nsPositionO   = positions.find(_.stage === sub.stage)
+    // Configure GMOS rows
+    Stream.eval(
+      inst.configureShuffle(rowsToShuffle).as(Result.Partial(NSRowsConfigure))
+    ).mergeHaltBoth(
```
jluhrs: Should be `merge`, or you will get only one result. And what is worst, the subexposure will start after the first one finishes.

cquiroz: I don't think that will happen, but it could start the next subexposure before the nod completes as `inst.configureShuffle` maybe faster that the nod
I changed it to `merge`

PR #1085, opened 2019-09-13
SEQNG-1040 Fixed issue with mixing real GeMS with simulated GSAOI.

jluhrs:
The real GeMS code needs access to GSAOI statuses and commands. The issue was that it was using `GsaoiEpics.instance` even if Gsaoi was internally simulated (and therefore, `GsaoiEpics.instance` was not initialized). I opted for creating an interface for GeMS to access GSAOI functionality, one that is instantiated following the Gsaoi simulation parameter. Besides fixing the bug, this implementation has the advantage of allowing mixing real GeMS control with simulated GSAOI.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -81,31 +81,35 @@ class SeqexecEngine(
   val ghostGDS: GdsClient[IO] = GdsClient(settings.ghostControl.command.fold(httpClient, GdsClient.alwaysOkClient),
     settings.ghostGDS)
 
-  private val systems = Systems[IO](
-    odbProxy,
-    settings.dhsControl.command.fold(DhsClientHttp(httpClient, settings.dhsURI),
-      DhsClientSim.unsafeApply(settings.date)),
-    (settings.tcsControl.command && settings.site === Site.GS).fold(TcsSouthControllerEpics(guideConfigDb), TcsSouthControllerSim[IO]),
-    (settings.tcsControl.command && settings.site === Site.GN).fold(TcsNorthControllerEpics(), TcsNorthControllerSim[IO]),
-    settings.gcalControl.command.fold(GcalControllerEpics(GcalEpics.instance), GcalControllerSim[IO]),
-    settings.f2Control.command.fold(Flamingos2ControllerEpics[IO](Flamingos2Epics.instance),
-      settings.instForceError.fold(Flamingos2ControllerSimBad.unsafeApply[IO](settings.failAt),
-        Flamingos2ControllerSim.unsafeApply[IO])),
-    settings.gmosControl.command.fold(GmosSouthControllerEpics(), GmosControllerSim.unsafeSouth[IO]),
-    settings.gmosControl.command.fold(GmosNorthControllerEpics(), GmosControllerSim.unsafeNorth[IO]),
-    settings.gnirsControl.command.fold(GnirsControllerEpics(), GnirsControllerSim.unsafeApply[IO]),
-    settings.gsaoiControl.command.fold(GsaoiControllerEpics(), GsaoiControllerSim.unsafeApply[IO]),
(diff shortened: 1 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerSim.scala
```diff
@@ -40,6 +40,15 @@ object GsaoiControllerSim {
       override def observeProgress(total: Time): fs2.Stream[F, Progress] =
         sim.observeCountdown(total, ElapsedTime(0.seconds))
 
+      override def currentState: F[GsaoiGuider.GuideState] = (new GsaoiGuider.GuideState {
+        override def isGuideActive: Boolean = false
+
+        override def isOdgwGuiding(odgwId: GsaoiGuider.OdgwId): Boolean = false
+      }).pure[F]
+
+      override def guide: F[Unit] = ().pure[F]
```
cquiroz: `Applicative[F].unit`?

modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerSim.scala
```diff
@@ -18,8 +18,8 @@ import squants.Time
 import squants.time.TimeConversions._
 
 object GsaoiControllerSim {
-  def unsafeApply[F[_]: Sync: Logger: Timer]: GsaoiController[F] =
-    new GsaoiController[F] {
+  def unsafeApply[F[_]: Sync: Logger: Timer]: GsaoiController[F] with GsaoiGuider[F] =
+    new GsaoiController[F] with GsaoiGuider[F] {
```
cquiroz: Should we have a new name for this? rather than `A with B`?

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
```diff
@@ -156,9 +153,9 @@ object GemsControllerEpics {
   val Log = getLogger
 
   def apply[F[_]: Async](epicsSys: => GemsEpics[F],
-                         gsaoiSys: => GsaoiEpics[F]
+                         gsaoiGuider: => GsaoiGuider[F]
```
cquiroz: I think this may be pass by value. `GmosEpics` needs to be passed by ref due to the initalization process

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -81,31 +81,35 @@ class SeqexecEngine(
...
-    settings.niriControl.command.fold(NiriControllerEpics(), NiriControllerSim.unsafeApply[IO]),
-    settings.nifsControl.command.fold(NifsControllerEpics(), NifsControllerSim.unsafeApply[IO]),
-    (settings.altairControl.command && settings.tcsControl.command).fold(AltairControllerEpics, AltairControllerSim),
-    (settings.gemsControl.command && settings.gemsControl.command).fold(
-      GemsControllerEpics(GemsEpics.instance, GsaoiEpics.instance),
-      GemsControllerSim[IO]
-    ),
-    guideConfigDb
-  )
+  private val systems = {
```
cquiroz: Maybe we should move to `Systems[F]`. it is getting cumbersome here in `SeqexecEngine`

modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerSim.scala
jluhrs: That was the one!

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
jluhrs: But Systems defines the generic type, while SeqexecEngine instances the concrete object to be used in runtime.

PR #1087, opened 2019-09-13
Seqexec small UI fixes

rpiaggio:
- Reinstate (un)setting breakpoint by pressing middle button in Seqexec UI
- Don't disable auto scroll in the step table if the user scrolls manually

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
@@ -896,15 +896,16 @@ object StepsTable extends Columns {
   )(e:          ReactMouseEvent): Callback =
     // If alt is pressed or middle button flip the breakpoint
     if (e.altKey || e.button === MIDDLE_BUTTON) {
-      (p.obsId, p.stepsList.find(_.id === index + 1))
-        .mapN(
-          (oid, step) =>
-            SeqexecCircuit
-              .dispatchCB(FlipBreakpointStep(oid, step))
-              .when_(step.canSetBreakpoint(index + 1, p.nextStepToRun))
-        )
-        .getOrEmpty
-        .when_(p.canSetBreakpoint)
+      e.preventDefaultCB >>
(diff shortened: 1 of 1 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -462,8 +462,7 @@ object StepsTable extends Columns {
     tableState:      TableState[TableColumn],
     breakpointHover: Option[Int],
     selected:        Option[StepId],
-    scrollCount:     Int,
-    scrollWhileRun:  Boolean
```
cquiroz: Good riddance, this was overcomplicated and not that useful

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -935,15 +930,15 @@ object StepsTable extends Columns {
               ^.cls := className,
               ^.key := s"$key-top",
               SeqexecStyles.expandedTopRow,
-              ^.onClick ==> allowedClick(p, index, onRowClick),
+              ^.onMouseDown ==> allowedClick(p, index, onRowClick),
```
cquiroz: Should we keep both for "legacy" browsers?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
rpiaggio: I probably would have implemented it too. It was one of first things that came to mind when you explained the scroll mechanism.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
cquiroz: Ok, let's keep it this way then

PR #1093, opened 2019-09-18
Update to sbt 1.3 and upgrade ocs libs

cquiroz:
sbt update, This one was on hold until the ocs libs were upgraded to have a matching squants version

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,6 +13,7 @@ name := Settings.Definitions.name
 
 organization in Global := "edu.gemini.ocs"
 
+Global / onChangedBuildSource := ReloadOnSourceChanges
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -13,6 +13,7 @@ name := Settings.Definitions.name
 
 organization in Global := "edu.gemini.ocs"
 
+Global / onChangedBuildSource := ReloadOnSourceChanges
```
tpolecat: I'm curious how this will work out. Worth a try.

cquiroz: I added that to stop sbt bugging me, it seemed to work the couple of times I made a change

PR #1104, opened 2019-09-23
remove unused modules in ocs3 repo

tpolecat:
This removes unused prototype modules, which should speed up the build a little bit. We will probably want to scavenge some code from these modules as part of GSP, which we can do by checking out the `v0.11.19` tag.

Next Steps
- `coreJVM/JS`, `db`, `ephemeris`, and `jsonJVM/JS` will move to a new `gsp-core` repo. The `json` module may go away.
-  It might be nice to pull out `gsp-acm` and `gsp-giapi` (or maybe combined, as `gsp-realtime` or something) since they're really just support libraries.
- There may be some orphaned dependencies. It would be nice to do a sweep through and make sure we're using everything referenced in `project/Settings.scala`.


cquiroz:
Curiously this build took longer, maybe just a one off.

tpolecat:
Weird.

cquiroz:
I'd not split giapi/acm while we are still developing the seqexec. I'd also be a bit wary of splitting core out as we've had multiple PRs spanning core and the seqexec

PR #1107, opened 2019-09-24
SEQNG-1062 Set the nod and shuffle state param

cquiroz:
It seems we are missing setting some N&S params when doing the initial gmos config. This PR should fix it but it needs more testing

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -52,6 +52,11 @@ trait GmosEncoders {
     case Order.TWO  => 2
   }
 
+  implicit val nsStateEncoder: EncodeEpicsValue[NodAndShuffleState, String] = EncodeEpicsValue{
+    case NodAndShuffleState.Classic    => "CLASSIC"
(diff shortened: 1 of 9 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -338,7 +343,9 @@ object GmosControllerEpics extends GmosEncoders {
 
       private def nsParams(state: GmosNSEpicsState, config: NSConfig): List[IO[Unit]] =
         List(
-          applyParam(state.nsPairs, config.nsPairs, (x: Int) => DC.setNsPairs(x))
+          applyParam(state.nsPairs, config.nsPairs, (x: Int) => DC.setNsPairs(x)),
+          applyParam(state.nsRows, config.nsRows, (x: Int) => DC.setNsRows(x)),
```
jluhrs: I would expect for this to not be necessary, because it will be set as part of the first N&S stage. In any case, the Tcl Seqexec does set it, so maybe it is needed.
The value set should correspond to the first stage.

cquiroz: Fixed that and another bug that I just noted on the calculation of `nsPairs`

PR #1108, opened 2019-09-25
Simplify calling React components. `sequence/steps` components.

rpiaggio:
This is a refactor to change the way in which components and their props are defined so that the call site code is simpler.

So far, all components in the `sequence/steps` package are refactored.


tpolecat:
Looks like you need to run `headerCreate`.

rpiaggio:
Done, thank you for the tip @tpolecat.

PR #1110, opened 2019-09-26
Refactoring of engine

cquiroz:
This is a refactoring of `Engine` following the tagless final style. given that we have `Logger` we can remove the need of a `Sync` and like with just `MonadError`

With this refactoring the `engine` module is fully `F` based rather than `IO` based


modules/seqexec/engine/src/test/scala/seqexec/engine/SequenceSpec.scala
```diff
@@ -54,8 +56,10 @@ class SequenceSpec extends AnyFlatSpec {
 
   }
 
+  private implicit def logger: Logger[IO] = Slf4jLogger.unsafeFromName[IO]("seqexec-engine")
```
tpolecat: This is deprecated (here and elsewhere) in favor of `getLoggerFromName` … if you're not getting a deprecation warning you may need to upgrade.

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -89,8 +91,11 @@ package object server {
 
   type ExecutionQueues = Map[QueueId, ExecutionQueue]
 
+  // This is far from ideal but we'll address this in another refactoring
+  private implicit def logger: Logger[IO] = Slf4jLogger.unsafeFromName[IO]("seqexec-engine")
```
tpolecat: I think it's probably fine. It's not actually unsafe.

modules/seqexec/engine/src/main/scala/seqexec/engine/Handle.scala
```diff
@@ -3,40 +3,40 @@
...
-    Handle[D, V, Unit](Applicative[StateT[IO, D, ?]].pure[(Unit, Option[Stream[IO, V]])](((), Some(p))))
+  def fromStream[F[_]: Monad, D, V](p: Stream[F, V]): Handle[F, D, V, Unit] = {
+    Handle[F, D, V, Unit](Applicative[StateT[F, D, ?]].pure[(Unit, Option[Stream[F, V]])](((), Some(p))))
   }
 
-  implicit def handlePInstances[D, V]: Monad[Handle[D, V, ?]] = new Monad[Handle[D, V, ?]] {
-    private def concatOpP[F[_]](op1: Option[Stream[F, V]],
+  implicit def handlePMonad[F[_]: Monad, D, V]: Monad[Handle[F, D, V, ?]] = new Monad[Handle[F, D, V, ?]] {
+    private def concatOpP(op1: Option[Stream[F, V]],
                                 op2: Option[Stream[F, V]]): Option[Stream[F, V]] = (op1, op2) match {
```
jluhrs: Alignment?

cquiroz: :D

PR #1111, opened 2019-09-27
WIP - remove core

tpolecat:
This removes `core`, `db`, `sql`, and `json` and adds dependencies on `gsp-core`. The build is greatly simplified and may be small enough to move back to Travis.


tpolecat:
Let's click the update button and see what happens.

tpolecat:
Todo

- [ ] update readme

tpolecat:
I expect some discussion here of course. The end result is that I would like to see the seqexec app as its own repo with all the ancillary stuff pulled in as libraries. See the updated README here for my thoughts.

cquiroz:
As the PR itself I don't see any issues.

This is probably inevitable as we keep growing but there are real pains. We have had many PRs that touch core (e.g adding keywords) while changing other bits. This becomes a lot more cumbersome as we need two PRs and a publication step in between, this will multiply even more with so many modules. Possibly this could be mitigated by some clever use of submodules but I doubt `sbt` will cooperate.

Besides the obvious issue with the long build times, I'd still prefer a monorepo.

tpolecat:
For local development you can `publishLocal` and depend on the snapshot version. On merge a snapshot will be published and on tag a release version will be published. So yes there is a bit more ceremony but CI times will be much shorter in most cases which I hope will speed things up in general.

swalker2m:
At least on my phone I can’t look at the actual update but I can imagine it is straightforward. Is there anything we can do about the keyword issue so it doesn’t require modifying core? 

Eventually we have to get here but I guess we can discuss timing since the seqexec is supposed to be wrapping up by November?

tpolecat:
If keywords are part of the data model then they need to be in core. Not sure what we can do about this.

cquiroz:
The keywords is just an example but it is a sample of a general theme.

A multi repo makes it easier to do lib development at the cost of more difficult app development.

Of course opinions are different on where to put the line, but given that we don't really expect our libs to be used by third parties (they may but it is certainly not an objective of ours), I'd hope we lean towards making app development easier.

tpolecat:
There are going to be a lot of applications depending on the core model. We can't have them all in the same repo. I think it makes sense to treat the model as a library.

swalker2m:
>  I think it makes sense to treat the model as a library.

I get the argument about the pain of switching to another repo when something you're doing in an app requires you to change the model, but I agree with Rob that ultimately the model is a library that will be widely shared.

I know this isn't the main point of the argument, but I do wonder if the FITS keywords belong in the core model.  It feels seqexec-specific. The issue, as I recall though, is that the machinery for making enums from tables is there and hard to extract?

cquiroz:
The core model is at the moment a library. I don't argue against that but I worry about a core part of every app being in a separate repo and quickly evolving independently

We have terrible stories about this same situation between ocs2 spModel and ITAC. Many of those would have been non issues if ITAC would have been on the same repo.

cquiroz:
I think a minimal mitigation is that at least all libs live in one "libs" repo and they are consistent with each other rather than one repo each, and maybe applications in their own.

swalker2m:
> We have terrible stories about this same situation between ocs2 spModel and ITAC. Many of those would have been non issues if ITAC would have been on the same repo.

The issue with ITAC is that the model is converted to different representations (via XML) as it goes through the system.  Having ITAC in the same repo wouldn't solve that problem, I think.  Nothing would require you to add a new filter to the ITAC model if you add one to the P1 model for example.

cquiroz:
I guess this can be closed

PR #1112, opened 2019-09-27
SEQNG-1068 Disabled GMOS CC configuration for Darks and Biases.

jluhrs:
GMOS Darks and Biases don't need to configure the CC. This change skips the CC configuration for those two observe types.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -350,13 +350,18 @@ object GmosControllerEpics extends GmosEncoders {
           applyParam(state.nsState, encode(config.nsState), DC.setNsState)
         ).flattenOption
 
+      // Don't set CC if Dark or Bias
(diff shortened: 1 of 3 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -350,13 +350,18 @@ object GmosControllerEpics extends GmosEncoders {
           applyParam(state.nsState, encode(config.nsState), DC.setNsState)
         ).flattenOption
 
+      // Don't set CC if Dark or Bias
       private def ccParams(state: GmosCCEpicsState, config: Config[T]#CCConfig): List[IO[Unit]] =
-        (setFilters(state, config.filter) ++
+        if(config.isDarkOrBias) List.empty
```
cquiroz: So in a Dark it doesn't really matter if any mechanism is deployed?

jluhrs: Yes, because the detector's shutter is closed.

PR #1114, opened 2019-09-30
SEQNG-1069A: Populate ns state to send to the client

cquiroz:
This PR adds additional fields to the NSStep to populate the UI. At the moment only static fields are done, we should add other fields in a next PR but this will help @rpiaggio progress

A big reason of the size of the PR is just that classes were moved from server to model to be shared with the UI

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/NSSubexposure.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/NSSubexposure.scala
@@ -7,29 +7,42 @@ import cats.Eq
 import cats.implicits._
 import seqexec.model.enum.NodAndShuffleStage
 import seqexec.model.enum.NodAndShuffleStage._
+import seqexec.model.GmosParameters._
+import shapeless.tag
 
-final case class NSSubexposure(
-  totalCycles: Int,
-  cycle:       Int,
-  id:          Int,
-  stage:       NodAndShuffleStage
+final case class NSSubexposure private (
+  totalCycles: NsCycles, // Total amount of cycles for a N&S step
+  cycle:       NsCycles, // Cycle for this sub exposure
+  stageIndex:  Int,      // Nod or stage index (between 0 and 3)
+  stage:       NodAndShuffleStage // Subexposure stage
 ) {
-  val firstSubexposure = cycle === 0 && id === 0
-  val lastSubexposure  = cycle === totalCycles - 1 && id === NsSequence.length - 1
+  val firstSubexposure = (cycle: Int) === 0 && stageIndex === 0
+  val lastSubexposure  = (cycle: Int) === (totalCycles: Int) - 1 && stageIndex === NsSequence.length - 1
 }
 
 object NSSubexposure {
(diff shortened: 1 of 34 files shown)


modules/seqexec/model/src/main/scala/seqexec/model/NSSubexposure.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import cats.Eq
+import cats.implicits._
+import seqexec.model.enum.NodAndShuffleStage
+import seqexec.model.enum.NodAndShuffleStage._
+
+final case class NSSubexposure(
+  totalCycles: Int,
+  cycle:       Int,
+  id:          Int,
```
tpolecat: Can we call this `index` maybe, since it's the index into `NsSequence`? It would be good to have doc comments on these parameters.

modules/seqexec/model/src/main/scala/seqexec/model/NSSubexposure.scala
```diff
@@ -0,0 +1,35 @@
...
+    Eq.by(x => (x.totalCycles, x.cycle, x.id, x.stage))
+
+  // Calculate the subexposures
+  def subexposures(
+    cycles: Int
+  ): List[NSSubexposure] =
+    (for {
+      i <- 0 until cycles
+      j <- 0 until NsSequence.length
+      stage = NsSequence.toList.lift(j).getOrElse(StageA)
```
tpolecat: This is weird. Can you do `NsSequence.toList.zipWithIndex` or something and get the stage and index at the same time?

modules/seqexec/model/src/main/scala/seqexec/model/NSSubexposure.scala
cquiroz: Yeah, now that you mention I'll also use `NsCycle` for the cycle-related fields

modules/seqexec/web/client/src/main/scala/seqexec/web/client/reusability/package.scala
```diff
@@ -39,10 +39,15 @@ import seqexec.web.client.model.TabSelected
...
 import seqexec.web.client.model.GlobalLog
 import seqexec.web.client.circuit._
+import squants.Time
+import shapeless.tag.@@
 
 package object reusability {
   implicit def enumeratedReuse[A <: AnyRef: Enumerated]: Reusability[A] =
     Reusability.byRef
+  implicit def taggedInt[A]: Reusability[Int @@ A] =
+    Reusability.by(x => x: Int)
```
rpiaggio: I have never used shapeless's tag feature, but I wonder if this can this be generalized (ie: `Reusability[B @@ A]`) ?

cquiroz: It should be assuming B has `Reusability` let me try

cquiroz: This is a good aim, but after several tries I'm going to give up. We'd have to like with less generic versions for `Reusability` and `Pickler`

PR #1115, opened 2019-09-30
SEQNG-1064 Fixed GMOS grating always being configured.

jluhrs:
And also simplified a bit the grating configuration code.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -233,7 +233,7 @@ object GmosControllerEpics extends GmosEncoders {
                                                                            cfg: GmosController.Config[T],
                                                                            d: GmosController.Config[T]#GmosDisperser)
   : List[Option[IO[Unit]]] = {
-    val set: List[Option[IO[Unit]]] = d match {
+    val params: List[Option[IO[Unit]]] = d match {
       case cfg.GmosDisperser.Mirror => List(setDisperser(state, cfg.mirror))
       case cfg.GmosDisperser.Order0(d) =>
         val s0: Option[IO[Unit]] = setDisperser(state, d)
@@ -262,7 +262,7 @@ object GmosControllerEpics extends GmosEncoders {
     }
 
     //If disperser, order or wavelength are set, force mode configuration. If not, check if it needs to be set anyways
-    if(set.exists(_.isDefined)) set :+ CC.setDisperserMode(disperserMode0).some
+    if(params.exists(_.isDefined)) params :+ CC.setDisperserMode(disperserMode0).some
     else List(applyParam(disperserModeDecode(state.disperserMode), disperserMode0, CC.setDisperserMode))
   }
 
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -3,12 +3,13 @@
 
 package seqexec.server.gmos
 
-import cats.effect.{ IO, Timer }
+import cats.effect.{IO, Timer}
 import cats.implicits._
 import edu.gemini.spModel.gemini.gmos.GmosCommonType._
 import edu.gemini.spModel.gemini.gmos.InstGmosCommon.UseElectronicOffsettingRuling
 import fs2.Stream
 import mouse.all._
+
 import scala.concurrent.ExecutionContext
 import org.log4s.getLogger
 import seqexec.model.dhs.ImageFileId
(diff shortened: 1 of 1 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -230,47 +229,41 @@ object GmosControllerEpics extends GmosEncoders {
...
-          } else {
-            List(
-              CC.setDisperserOrder(disperserOrderEncoder.encode(o)).some,
-              s0,
-              setWavelength(state, w)
-            )
-          }
-
-        //If disperser, order or wavelength are set, force mode configuration
-        s.foldMap(_ => s ++ List(CC.setDisperserMode(disperserMode0).some))
```
jluhrs: `s` here was a non empty `List`, not an `Option` ! This caused the Mode to be set always.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -230,47 +229,41 @@ object GmosControllerEpics extends GmosEncoders {
   def setWavelength(state: GmosCCEpicsState, w: Length): Option[IO[Unit]] =
     applyParam(state.disperserWavel, encode(w), CC.setDisperserLambda)
 
-  def setDisperserParams[T <: GmosController.SiteDependentTypes: Encoders](state: GmosCCEpicsState, cfg: GmosController.Config[T], d: GmosController.Config[T]#GmosDisperser): List[Option[IO[Unit]]] = {
+  def setDisperserParams[T <: GmosController.SiteDependentTypes: Encoders](state: GmosCCEpicsState,
+                                                                           cfg: GmosController.Config[T],
+                                                                           d: GmosController.Config[T]#GmosDisperser)
+  : List[Option[IO[Unit]]] = {
     val set: List[Option[IO[Unit]]] = d match {
```
cquiroz: The name `set` maybe a bit confusing, you imagine it is a function rather than a value

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -218,10 +219,8 @@ object GmosControllerEpics extends GmosEncoders {
   def setDisperser[T <: GmosController.SiteDependentTypes: Encoders](state: GmosCCEpicsState, d: T#Disperser): Option[IO[Unit]] = {
     val encodedVal = Encoders[T].disperser.encode(d)
     val s = applyParam(state.disperser.toUpperCase, encodedVal.toUpperCase, (_: String) => CC.setDisperser(encodedVal))
-    // Force setting if disperser is parked
-    s.orElse(
-      applyParam(state.disperserParked, false, (_: Boolean) => CC.setDisperser(encodedVal))
-    )
+    // Force setting if not set and disperser is parked
+    s.orElse(state.disperserParked.option(CC.setDisperser(encodedVal)))
```
cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -218,10 +219,8 @@ object GmosControllerEpics extends GmosEncoders {
   def setDisperser[T <: GmosController.SiteDependentTypes: Encoders](state: GmosCCEpicsState, d: T#Disperser): Option[IO[Unit]] = {
     val encodedVal = Encoders[T].disperser.encode(d)
     val s = applyParam(state.disperser.toUpperCase, encodedVal.toUpperCase, (_: String) => CC.setDisperser(encodedVal))
```
cquiroz: Why don't we need `upperCase` when setting the value?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -230,47 +229,41 @@ object GmosControllerEpics extends GmosEncoders {
...
-        s.foldMap(_ => List(s, CC.setDisperserMode(disperserMode0).some))
+      case cfg.GmosDisperser.Mirror => List(setDisperser(state, cfg.mirror))
       case cfg.GmosDisperser.Order0(d) =>
         val s0: Option[IO[Unit]] = setDisperser(state, d)
         // If disperser is set, force order configuration
-        val s: List[Option[IO[Unit]]] = if(s0.isEmpty) List(setOrder(state, Order.ZERO))
-                else CC.setDisperserOrder(disperserOrderEncoder.encode(Order.ZERO)).some :: List(s0)
-        //If disperser or order are set, force mode configuration
-        s.foldMap(_ => s ++ List(CC.setDisperserMode(disperserMode0).some))
+        if(s0.isEmpty) List(setOrder(state, Order.ZERO))
```
cquiroz: `s0 match ...`?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -230,47 +229,41 @@ object GmosControllerEpics extends GmosEncoders {
...
-              CC.setDisperserOrder(disperserOrderEncoder.encode(o)).some,
-              s0,
-              setWavelength(state, w)
-            )
-          }
-
-        //If disperser, order or wavelength are set, force mode configuration
-        s.foldMap(_ => s ++ List(CC.setDisperserMode(disperserMode0).some))
+        // If disperser is set, the order has to be set regardless of current value.
+        if (s0.isEmpty) {
```
cquiroz: `s0 match { ...`?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: I don't know, this just copy the code in the Tcl Seqexec. My guess is that the instrument changes the name in the status channel.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: I would have used `match` if I were to use the contents of `s0`. But as it is this is more compact.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: Same as above.

PR #1117, opened 2019-10-01
SEQNG-1067 Tcs in-position stabilization time now depends on magnitude of movement.

jluhrs:
The TCS in-position signal takes some time to be stable. This PR makes the time to wait for it to stabilize depend on the subsystems used and the magnitude of the displacement.
The parameters for the calculations are placeholders until they are measured. I don't intend to merge the PR until then.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
@@ -17,7 +17,7 @@ import java.util.concurrent.TimeUnit;
 public class CaWindowStabilizer<T> implements CaAttribute<T> {
 
     private final CaAttribute<T> sa;
-    private final Duration settleTime;
+    private Duration settleTime;
     private final ScheduledExecutorService executor;
     private ScheduledFuture<?> timeoutFuture;
     private final CaAttributeListener<T> valListener;
@@ -147,16 +147,20 @@ public class CaWindowStabilizer<T> implements CaAttribute<T> {
         notifier.removeListener(listener);
     }
 
-    public CaWindowStabilizer<T> reset() {
+    public CaWindowStabilizer<T> restart(Duration settleTime) {
         filteredVal = null;
         //Restart the timer
         if (timeoutFuture != null) {
             timeoutFuture.cancel(true);
         }
+        this.settleTime = settleTime;
         timeoutFuture = executor.schedule(() -> CaWindowStabilizer.this.onTimeout(),
             settleTime.toMillis(), TimeUnit.MILLISECONDS);
 
         return this;
+    }
+    public CaWindowStabilizer<T> restart() {
+        return restart(settleTime);
     }
 
 }
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
@@ -14,6 +14,11 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
+/*
+ * CaWindowStabilizer filters the value of a CaAttribute. A new value is not set unless it does not change during a time
+ * window given by the parameter settleTime.
+ * This filter is more useful for discrete signals. It is used for in-position signals.
+ */
 public class CaWindowStabilizer<T> implements CaAttribute<T> {
 
     private final CaAttribute<T> sa;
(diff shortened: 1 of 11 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsSettleTimeCalculator.scala
```diff
@@ -0,0 +1,67 @@
...
+object TcsSettleTimeCalculator {
+
+  trait SettleTimeCalculator {
+    def calc(displacement: Angle): Time
+  }
+
+  def constantSettleTime(cnst: Time): SettleTimeCalculator = (_: Angle) => cnst
+
+  // Settle time proportional to displacement
+  def linearSettleTime(scale: SettleTimeScale): SettleTimeCalculator = (displacement: Angle) => scale * displacement
```
jluhrs: I may need to implement more formulae, like quadratic or exponential, depending on the results of measuring the stabilization times.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsNorthControllerEpicsAo.scala
```diff
@@ -94,14 +95,17 @@ object TcsNorthControllerEpicsAo {
 
     def sysConfig(current: EpicsTcsAoConfig): IO[EpicsTcsAoConfig] = {
       val params = configParams(current)
+      val stabilizationTime = tcs.tc.offsetA
```
cquiroz: This seems duplicated across north/south, could it be factored out?

modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
```diff
@@ -147,16 +147,20 @@ public void removeListener(CaAttributeListener<T> listener) {
         notifier.removeListener(listener);
     }
 
-    public CaWindowStabilizer<T> reset() {
+    public CaWindowStabilizer<T> restart(Duration settleTime) {
```
cquiroz: Maybe we should document this a bit. What does the `settleTime` do?

jluhrs: Added documentation. `settleTime` is the time a the value of the filtered `CaAttribute` must be unchanged to be set as the output of the filter.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsNorthControllerEpicsAo.scala
jluhrs: Done.

jluhrs:
I put fixed values for the TCS in-position stabilization filter. The values were taken from the fudge factors in the old Seqexec.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
```diff
@@ -380,29 +384,38 @@ final class TcsEpics[F[_]: Async](epicsService: CaService, tops: Map[String, Str
...
-    .flatMap(waitForValueF(_, "TRUE", timeout,"TCS inposition flag"))
+  // Tcs fudge1 (time to wait for in-position to change to false)
+  private val tcsSettleTime = 2.8.seconds
+
+  // This functions returns a F that, when run, first waits tcsSettleTime to absorb in-position transients, then waits
+  // for the in-position to change to true and stay true for stabilizationTime. It will wait up to `timeout`
+  // seconds for that to happen.
+  def waitInPosition(stabilizationTime: Time, timeout: Time)(implicit T: Timer[F]): F[Unit] =
+    T.sleep(FiniteDuration(tcsSettleTime.toMillis, TimeUnit.MILLISECONDS)) *>
+      Sync[F].delay(filteredInPositionAttr.restart(java.time.Duration.ofMillis(stabilizationTime.toMillis)))
```
cquiroz: Will `restart` call into `epics`?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
```diff
@@ -385,12 +389,16 @@ final class TcsEpics[F[_]: Async](epicsService: CaService, tops: Map[String, Str
   private val filteredInPositionAttr: CaWindowStabilizer[String] =
     new CaWindowStabilizer[String](inPositionAttr, java.time.Duration.ofMillis(defaultTcsStabilizeTime.toMillis))
 
+  // Tcs fudge1 (time to wait for in-position to change to false)
+  private val tcsSettleTime = 2.8.seconds
```
cquiroz: This is from tcl-seqexec right?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
jluhrs: No, restart resets the filter

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
jluhrs: Yes

PR #1119, opened 2019-10-02
SEQNG-1069B Send to the client the state of running ns actions

cquiroz:
This is the second part of sending N&S updates to the client. This PR adds a field to `NodAndShuffleStatus` containing the state of the current execution
There are some changes on the way N&S actions are structured to send events before and after each substage is produced

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/NodAndShuffleStatus.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/NodAndShuffleStatus.scala
@@ -10,16 +10,24 @@ import seqexec.model.GmosParameters._
 import monocle.macros.Lenses
 import squants.Time
 
+final case class NSRunningState(action: NSAction, sub: NSSubexposure)
+
+object NSRunningState {
+  implicit val equalNSRunningState: Eq[NSRunningState] =
+    Eq.by(x => (x.action, x.sub))
+}
+
 @Lenses
 final case class NodAndShuffleStatus(
   observing: ActionStatus,
   totalExposureTime: Time,
   nodExposureTime: Time,
-  cycles: NsCycles
+  cycles: NsCycles,
+  state: Option[NSRunningState]
 )
 
 object NodAndShuffleStatus {
 
   implicit val equalNodAndShuffleStatus: Eq[NodAndShuffleStatus] =
-    Eq.by(x => (x.observing, x.totalExposureTime, x.nodExposureTime, x.cycles))
+    Eq.by(x => (x.observing, x.totalExposureTime, x.nodExposureTime, x.cycles, x.state))
 }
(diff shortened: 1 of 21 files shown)


cquiroz:
This PR needs a bit of rework due to #1120

cquiroz:
Updated after #1120 It can be reviewed now

modules/seqexec/model/src/main/scala/seqexec/model/NodAndShuffleStatus.scala
```diff
@@ -10,16 +10,24 @@ import seqexec.model.GmosParameters._
...
+}
+
 @Lenses
 final case class NodAndShuffleStatus(
   observing: ActionStatus,
   totalExposureTime: Time,
   nodExposureTime: Time,
-  cycles: NsCycles
+  cycles: NsCycles,
+  state: Option[NSRunningState]
```
cquiroz: @rpiaggio This will be populated once the sequence starts running indicating the current subexposure and the state

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -882,6 +884,7 @@ object SeqexecEngine extends SeqexecConfiguration {
         // TODO: Sequence completed event not emitted by engine.
         case SystemEvent.Completed(_, _, _, _)                                    => SequenceUpdated(svs)
         case SystemEvent.StopCompleted(id, _, _, _)                               => SequenceStopped(id, svs)
+        case SystemEvent.PartialResult(_, _, _, Partial(_: InternalPartialVal))   => NullEvent
```
cquiroz: This is an extra to avoid sending messages for intermediate steps that don't concern the client

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -48,6 +49,7 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Concurrent: Logge
     r match {
       case ObserveCommandResult.Success =>
         okTail(fileId, dataId, stopped = false, env)
+          .as(Result.Partial(NSFinalObs)) // For normally completed observations send a partial
```
cquiroz: This will let us send messages after the obs completes. It needs to be matched with an `Ok` down below

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -149,15 +158,28 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Concurrent: Logge
...
+            NSSubexposure
+              .subexposures(cycles)
               .map {
                 oneSubExposure(fileId, _, positions, env, post)
               }
               .reduceOption(_ ++ _)
-              .orEmpty
+              .orEmpty ++
+            Stream.emit(Result.Partial(NSComplete(nsLast))) ++
+            Stream.emit(Result.OK(Response.Observed(fileId)))
```
cquiroz: This last `Ok` closes the whole N&S step

PR #1120, opened 2019-10-02
SEQNG-1072 Fixes to execute N&S in Epics

cquiroz:
This PR adds some fixes to successfully run N&S with epics, thanks @jluhrs 
The two main changes are

* Remove the call to set rows in between stages
* Handle the response of the observe pause differently if N&S

There are few other logger related changes

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -11,7 +11,7 @@ import fs2.Stream
 import mouse.all._
 
 import scala.concurrent.ExecutionContext
-import org.log4s.getLogger
+import io.chrisdavenport.log4cats.Logger
 import seqexec.model.dhs.ImageFileId
 import seqexec.model.enum.ObserveCommandResult
 import seqexec.model.GmosParameters._
@@ -60,6 +60,12 @@ trait GmosEncoders {
     case NodAndShuffleState.NodShuffle => "NOD_SHUFFLE"
   }
 
+  implicit val nsStateDecoder: DecodeEpicsValue[String, Option[NodAndShuffleState]] = DecodeEpicsValue{
+    case "CLASSIC"     => NodAndShuffleState.Classic.some
+    case "NOD_SHUFFLE" => NodAndShuffleState.NodShuffle.some
+    case _             => none
+  }
+
   implicit val exposureTimeEncoder: EncodeEpicsValue[ExposureTime, Int] = EncodeEpicsValue(_.toSeconds.toInt)
 
   implicit val disperserLambdaEncoder: EncodeEpicsValue[Length, Double] =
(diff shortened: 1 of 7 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -363,78 +374,81 @@ object GmosControllerEpics extends GmosEncoders {
...
-          sys.post.void
-
       override def observe(fileId: ImageFileId, expTime: Time): IO[ObserveCommandResult] =
         failOnDHSNotConected *>
           sys.observeCmd.setLabel(fileId) *>
           sys.observeCmd.setTimeout[IO](expTime + ReadoutTimeout) *>
-          sys.observeCmd.post[IO]
+          sys.observeCmd.post[IO].flatMap {
+            case ObserveCommandResult.Paused =>
+              sys.nsState.map(GmosEncoders.nsStateDecoder.decode).map {
```
cquiroz: @jluhrs This worked fine in test but there maybe a little risk given that GMOS is automagically changing this to `CLASSIC`. WDYT?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -363,78 +374,81 @@ object GmosControllerEpics extends GmosEncoders {
...
-          sys.configCmd.setTimeout[IO](ConfigTimeout) *>
-          sys.post.void
-
       override def observe(fileId: ImageFileId, expTime: Time): IO[ObserveCommandResult] =
         failOnDHSNotConected *>
           sys.observeCmd.setLabel(fileId) *>
           sys.observeCmd.setTimeout[IO](expTime + ReadoutTimeout) *>
-          sys.observeCmd.post[IO]
+          sys.observeCmd.post[IO].flatMap {
+            case ObserveCommandResult.Paused =>
```
cquiroz: Core change 1. We should respond with `Partial` if it is N&S to let the rest of the engine logic to proceed

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -105,50 +105,37 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Concurrent: Logge
...
     env:       ObserveEnvironment[F],
     post:      (Stream[F, Result[F]], ObserveEnvironment[F]) => Stream[F, Result[F]]
   ): Stream[F, Result[F]] = {
-    val rowsToShuffle = Gmos.rowsToShuffle(sub.stage, rows)
     val nsPositionO   = positions.find(_.stage === sub.stage)
-    // Configure GMOS rows
-    Stream
-      .eval(
-        inst
-          .configureShuffle(rowsToShuffle)
```
cquiroz: Core change 2. We shouldn't set rows to shuffle in between subexposures. Discovered experimentally

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: Yes, but it changes to CLASSIC after the detector readout. At that point the observation state cannot change to PAUSED.

PR #1122, opened 2019-10-03
Use pprint for debug console logging in Seqexec-UI, only in development.

rpiaggio:
**Before**:

image

**After**:

image

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
@@ -203,7 +203,7 @@ object actions {
       PPrinter(defaultHeight = 200, colorApplyPrefix = fansi.Color.Blue)
 
     def apply(x: Any): String =
-      fansi.Str.join(pprinter.tokenize(x, initialOffset = 4).toSeq:_*).toString
+      pprinter(x, initialOffset = 4).toString
   }
 
   private val stepInfo: PartialFunction[Step, Product] = {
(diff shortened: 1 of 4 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
```diff
@@ -195,9 +196,19 @@ object actions {
...
 
-  private val stepInfo: PartialFunction[Step, (StepId, StepState, List[(Resource, ActionStatus)])] = {
+  // This, together with the whole pprint library, is shaken off when linking
+  // for production, since show is not called in that case (see LoggingProcessor).
+  private object Printer {
+    private val pprinter: PPrinter =
+      PPrinter(defaultHeight = 200, colorApplyPrefix = fansi.Color.Blue)
+
+    def apply(x: Any): String =
+      fansi.Str.join(pprinter.tokenize(x, initialOffset = 4).toSeq:_*).toString
```
cquiroz: `fansi`?

tpolecat: It's Haoyi's ANSI coloring library.

cquiroz:
This is great, the use of LinkingInfo avoids any costs on size and in fact we reduce the overall size as the action formatting call is discarded

Do you know if pprint is available on 2.13?

rpiaggio:
> This is great, the use of LinkingInfo avoids any costs on size and in fact we reduce the overall size as the action formatting call is discarded
> 
> Do you know if pprint is available on 2.13?

Yes, latest version 5.5 is released for 2.13.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
rpiaggio: In theory, we could just call `pprinter.apply` here, which does the same as this `.tokenize` call but drops the `initialOffset` parameter (seemed like a bug in pprint).

However, even this way the `initialOffset` parameter is not being honoured, or it does not do what I expect it to do (which is shift the whole block by that number of spaces). It still looks much better than whithout `pprint` though.

I guess that I could simplify back to `pprinter.apply` and either report/submit a PR to `pprint`.

PR #1123, opened 2019-10-03
SEQNG-906 GMOS stop, abort and pause allowed only if acquiring and with enough time left

jluhrs:
GMOS detector controller can crash during an observation if the stop, abort or pause command are issued at the wrong time. This PR assures they are not.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
@@ -18,7 +18,7 @@ import seqexec.model.GmosParameters._
 import seqexec.server.EpicsCodex.EncodeEpicsValue
 import seqexec.server.EpicsUtil._
 import seqexec.server.InstrumentSystem.ElapsedTime
-import seqexec.server.{EpicsUtil, Progress, SeqexecFailure}
+import seqexec.server.{EpicsCommand, EpicsUtil, Progress, SeqexecFailure}
 import seqexec.server.EpicsCodex._
 import seqexec.server.gmos.GmosController.Config._
 import seqexec.server.gmos.GmosController._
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -403,29 +403,50 @@ object GmosControllerEpics extends GmosEncoders {
...
-          sys.stopCmd.mark[IO] *>
-          sys.stopCmd.post[IO]
-
-      override def abortObserve: IO[Unit] =
-        L.info("Abort Gmos exposure") *>
-          sys.abortCmd.setTimeout[IO](DefaultTimeout) *>
-          sys.abortCmd.mark[IO] *>
-          sys.abortCmd.post[IO].void
+      // Remaining time when it is not safe to stop, pause or abort
+      val SafetyCutoff = 3.0 // seconds
```
cquiroz: Could we move this to `GmosParameters` to make ti reachable by the client?
Should it be not a double but a more typed value, e.g. `FiniteDuration`?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -403,29 +403,50 @@ object GmosControllerEpics extends GmosEncoders {
...
+
+      override def stopObserve: IO[Unit] = (sys.dcIsAcquiring, sys.countdown).mapN { case (isAcq, timeLeft) =>
+        if(!isAcq)
```
cquiroz: It seems there isa common logic across observe/stop/abort, should it be factored out?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: Done.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: Done.

PR #1124, opened 2019-10-03
[SEQNG-333] Display of Nod and Shuffle Step

rpiaggio:
Display cycle and subexposure progress for N&S steps in Seqexec UI, shown when a step is selected or running.

Also display graceful and immediate pause and stop buttons when the user is logged in and the step is running.

image

This PR integrates the changes from https://github.com/gemini-hlsw/ocs3/pull/1114 but not yet the ones from https://github.com/gemini-hlsw/ocs3/pull/1119 (no running state yet).

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
@@ -7,123 +7,135 @@ import seqexec.model.enum.Instrument
 import seqexec.model.enum.Instrument._
 
 object operations {
-  // Operations possible at the sequence level
-  sealed trait SequenceOperations
-
-  object SequenceOperations {
-    case object PauseSequence extends SequenceOperations
-    case object StopSequence extends SequenceOperations
-    case object RunSequence extends SequenceOperations
+  sealed trait OperationLevel
+  object OperationLevel {
+    sealed trait Observation extends OperationLevel
+    sealed trait NsCycle extends OperationLevel
+    sealed trait NsNod extends OperationLevel
   }
 
-  // Operations possible at the observation level
-  sealed trait ObservationOperations
+  import OperationLevel._
 
-  object ObservationOperations {
-    case object PauseObservation extends ObservationOperations
-    case object StopObservation extends ObservationOperations
-    case object AbortObservation extends ObservationOperations
-    case object ResumeObservation extends ObservationOperations
-    // Operations for Hamamatsu
-    case object PauseImmediatelyObservation extends ObservationOperations
-    case object StopImmediatelyObservation extends ObservationOperations
-    case object PauseGracefullyObservation extends ObservationOperations
-    case object StopGracefullyObservation extends ObservationOperations
+  sealed trait Operations[A <: OperationLevel]
+  object Operations {
+    // Operations possible at the observation level
+    case object PauseObservation extends Operations[Observation]
+    case object StopObservation extends Operations[Observation]
+    case object AbortObservation extends Operations[Observation]
+    case object ResumeObservation extends Operations[Observation]
+
+    // Operations possible for N&S Cycle
+    case object PauseGracefullyObservation extends Operations[NsCycle]
+    case object StopGracefullyObservation extends Operations[NsCycle]
+
+    // Operations possible for N&S Nod
+    case object PauseImmediatelyObservation extends Operations[NsNod]
+    case object StopImmediatelyObservation extends Operations[NsNod]
   }
+
+  sealed trait OperationLevelType[L <: OperationLevel]
+  implicit object ObservationLevel extends OperationLevelType[Observation]
+  private implicit object NsCycleLevel extends OperationLevelType[NsCycle]
+  private implicit object NsNodLevel extends OperationLevelType[NsNod]
 
   sealed trait SupportedOperations {
-
-    /**
-      * Sorted list of operations supported at the sequence level
-      */
-    def observationOperations(
-      isObservePaused: Boolean): List[ObservationOperations]
-
-    /**
-      * Sorted list of operations supported at the observation (row) level
-      */
-    def sequenceOperations: List[SequenceOperations]
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]]
   }
 
   private val F2SupportedOperations = new SupportedOperations {
-    def observationOperations(
-      isObservePaused: Boolean): List[ObservationOperations] = Nil
-    def sequenceOperations: List[SequenceOperations] = Nil
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
+      Nil
   }
 
   private val GmosSupportedOperations = new SupportedOperations {
-    def observationOperations(
-      isObservePaused: Boolean): List[ObservationOperations] =
-      if (isObservePaused) {
-        List(ObservationOperations.ResumeObservation,
-             ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
-      } else {
-        List(ObservationOperations.PauseObservation,
-             ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
+      level match {
+        case ObservationLevel =>
+          if (isObservePaused) {
+            List(Operations.ResumeObservation,
+                 Operations.StopObservation,
+                 Operations.AbortObservation)
+          } else {
+            List(Operations.PauseObservation,
+                 Operations.StopObservation,
+                 Operations.AbortObservation)
+          }
+        case NsCycleLevel =>
+            List(Operations.PauseGracefullyObservation,
+                 Operations.StopGracefullyObservation)
+        case NsNodLevel =>
+          List(Operations.PauseImmediatelyObservation,
+               Operations.StopImmediatelyObservation)
+        case _                => Nil
       }
-
-    def sequenceOperations: List[SequenceOperations] = Nil
   }
 
   private val GnirsSupportedOperations = new SupportedOperations {
-    def observationOperations(
-      isObservePaused: Boolean): List[ObservationOperations] =
-      if (isObservePaused) {
-        List(ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
-      } else {
-        List(ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
+      level match {
+        case ObservationLevel =>
+          if (isObservePaused) {
+            List(Operations.StopObservation,
+                 Operations.AbortObservation)
+          } else {
+            List(Operations.StopObservation,
+                 Operations.AbortObservation)
+          }
+        case _                => Nil
       }
-
-    def sequenceOperations: List[SequenceOperations] = Nil
   }
 
   private val NiriSupportedOperations = new SupportedOperations {
-    def observationOperations(
-      isObservePaused: Boolean): List[ObservationOperations] =
-      if (isObservePaused) {
-        List(ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
-      } else {
-        List(ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
+      level match {
+        case ObservationLevel =>
+          if (isObservePaused) {
+            List(Operations.StopObservation,
+                 Operations.AbortObservation)
+          } else {
+            List(Operations.StopObservation,
+                 Operations.AbortObservation)
+          }
+        case _                => Nil
       }
-
-    def sequenceOperations: List[SequenceOperations] = Nil
   }
 
   private val NifsSupportedOperations = new SupportedOperations {
-    def observationOperations(
-      isObservePaused: Boolean
-    ): List[ObservationOperations] =
-      if (isObservePaused) {
-        List(ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
-      } else {
-        List(ObservationOperations.StopObservation,
-             ObservationOperations.AbortObservation)
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+                                  (implicit level: OperationLevelType[L]): List[Operations[L]] =
+      level match {
+        case ObservationLevel =>
+          if (isObservePaused) {
+            List(Operations.StopObservation,
+                 Operations.AbortObservation)
+          } else {
+            List(Operations.StopObservation,
+                 Operations.AbortObservation)
+          }
+        case _                => Nil
       }
-
-    def sequenceOperations: List[SequenceOperations] = Nil
   }
 
   private val GsaoiSupportedOperations = new SupportedOperations {
-    def observationOperations(
-      isObservePaused: Boolean
-    ): List[ObservationOperations] =
-      List(ObservationOperations.StopObservation,
-           ObservationOperations.AbortObservation)
-
-    def sequenceOperations: List[SequenceOperations] = Nil
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+                                  (implicit level: OperationLevelType[L]): List[Operations[L]] =
+      level match {
+        case ObservationLevel =>
+          List(Operations.StopObservation,
+               Operations.AbortObservation)
+        case _                => Nil
+      }
   }
 
   private val NilSupportedOperations = new SupportedOperations {
-    def observationOperations(
-      isObservePaused: Boolean): List[ObservationOperations] = Nil
-    def sequenceOperations: List[SequenceOperations] = Nil
+    def apply[L <: OperationLevel](isObservePaused: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
+      Nil
   }
 
   private val instrumentOperations: Map[Instrument, SupportedOperations] = Map(
@@ -138,15 +150,10 @@ object operations {
 
   final implicit class SupportedOperationsOps(val i: Instrument)
       extends AnyVal {
-    def observationOperations(
-      isObservePaused: Boolean): List[ObservationOperations] =
+    def operations[L <: OperationLevel](isObservePaused: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
       instrumentOperations
-        .getOrElse(i, NilSupportedOperations)
-        .observationOperations(isObservePaused)
-    def sequenceOperations: List[SequenceOperations] =
-      instrumentOperations
-        .getOrElse(i, NilSupportedOperations)
-        .sequenceOperations
+        .getOrElse(i, NilSupportedOperations)(isObservePaused)
   }
 
 }
--- a/modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
@@ -35,33 +35,37 @@ object operations {
 
   sealed trait OperationLevelType[L <: OperationLevel]
   implicit object ObservationLevel extends OperationLevelType[Observation]
-  private implicit object NsCycleLevel extends OperationLevelType[NsCycle]
-  private implicit object NsNodLevel extends OperationLevelType[NsNod]
+  implicit object NsCycleLevel extends OperationLevelType[NsCycle]
+  implicit object NsNodLevel extends OperationLevelType[NsNod]
 
   sealed trait SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
       (implicit level: OperationLevelType[L]): List[Operations[L]]
   }
 
   private val F2SupportedOperations = new SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
       (implicit level: OperationLevelType[L]): List[Operations[L]] =
       Nil
   }
 
   private val GmosSupportedOperations = new SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
       (implicit level: OperationLevelType[L]): List[Operations[L]] =
       level match {
         case ObservationLevel =>
-          if (isObservePaused) {
-            List(Operations.ResumeObservation,
-                 Operations.StopObservation,
-                 Operations.AbortObservation)
+          if(isMultiLevel) {
+            List(Operations.AbortObservation)
           } else {
-            List(Operations.PauseObservation,
-                 Operations.StopObservation,
-                 Operations.AbortObservation)
+            if (isObservePaused) {
+              List(Operations.ResumeObservation,
+                   Operations.StopObservation,
+                   Operations.AbortObservation)
+            } else {
+              List(Operations.PauseObservation,
+                   Operations.StopObservation,
+                   Operations.AbortObservation)
+            }
           }
         case NsCycleLevel =>
             List(Operations.PauseGracefullyObservation,
@@ -74,7 +78,7 @@ object operations {
   }
 
   private val GnirsSupportedOperations = new SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
       (implicit level: OperationLevelType[L]): List[Operations[L]] =
       level match {
         case ObservationLevel =>
@@ -90,7 +94,7 @@ object operations {
   }
 
   private val NiriSupportedOperations = new SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
       (implicit level: OperationLevelType[L]): List[Operations[L]] =
       level match {
         case ObservationLevel =>
@@ -106,8 +110,8 @@ object operations {
   }
 
   private val NifsSupportedOperations = new SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
-                                  (implicit level: OperationLevelType[L]): List[Operations[L]] =
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
       level match {
         case ObservationLevel =>
           if (isObservePaused) {
@@ -122,8 +126,8 @@ object operations {
   }
 
   private val GsaoiSupportedOperations = new SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
-                                  (implicit level: OperationLevelType[L]): List[Operations[L]] =
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
+      (implicit level: OperationLevelType[L]): List[Operations[L]] =
       level match {
         case ObservationLevel =>
           List(Operations.StopObservation,
@@ -133,7 +137,7 @@ object operations {
   }
 
   private val NilSupportedOperations = new SupportedOperations {
-    def apply[L <: OperationLevel](isObservePaused: Boolean)
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
       (implicit level: OperationLevelType[L]): List[Operations[L]] =
       Nil
   }
@@ -150,10 +154,10 @@ object operations {
 
   final implicit class SupportedOperationsOps(val i: Instrument)
       extends AnyVal {
-    def operations[L <: OperationLevel](isObservePaused: Boolean)
+    def operations[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean = false)
       (implicit level: OperationLevelType[L]): List[Operations[L]] =
       instrumentOperations
-        .getOrElse(i, NilSupportedOperations)(isObservePaused)
+        .getOrElse(i, NilSupportedOperations)(isObservePaused, isMultiLevel)
   }
 
 }
--- a/modules/seqexec/web/client/src/main/resources/less/style.less
+++ b/modules/seqexec/web/client/src/main/resources/less/style.less
@@ -444,6 +444,15 @@ body {
   margin-left: 20px;
 }
 
+.SeqexecStyles-nodAndShuffleDetailRow {
+  display: flex;
+  justifyContent: spaceBetween;
+}
+
+.SeqexecStyles-nodAndShuffleControls {
+  margin-left: 14px;
+}
+
 .SeqexecStyles-observationProgressBar {
   margin: 0 !important;
   height: 1em !important;
@@ -457,6 +466,11 @@ body {
 .SeqexecStyles-observationLabel {
   text-align: center;
   width: 100%;
+}
+
+.SeqexecStyles-dividedProgress {
+  width: 100%;
+  display: flex;
 }
 
 .SeqexecStyles-dividedProgressSectionLeft {
--- a/modules/seqexec/web/client/src/main/resources/less/style.less
+++ b/modules/seqexec/web/client/src/main/resources/less/style.less
@@ -446,7 +446,6 @@ body {
 
 .SeqexecStyles-nodAndShuffleDetailRow {
   display: flex;
-  justifyContent: spaceBetween;
 }
 
 .SeqexecStyles-nodAndShuffleControls {
--- a/modules/seqexec/web/client/src/main/resources/less/style.less
+++ b/modules/seqexec/web/client/src/main/resources/less/style.less
@@ -538,6 +538,10 @@ body {
   white-space: nowrap;
 }
 
+.SeqexecStyles-progressMessage {
+  font-weight: bold;
+}
+
 .SeqexecStyles-subsystems {
   margin-left: auto;
   padding-right: 7px;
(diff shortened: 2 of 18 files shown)


arturog8m:
This is looking really nice! What's the significance of the two shades of blue?

rpiaggio:
> This is looking really nice! What's the significance of the two shades of blue?

Oh, it's just the bar animation caught midway in the screen capture:

SEQNG-333

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -641,19 +645,19 @@ object StepsTable extends Columns {
       b.props.rowGetter(i),
       b.props.canSetBreakpoint,
       b.state.breakpointHover) match {
-      case (-1, _, _, _) =>
+      case (-1, _, _, _)                                           =>
```
cquiroz: Now that I look at this, the -1 is not that good. Should we give it a name, like `HeaderRow`?

modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
```diff
@@ -7,123 +7,139 @@ import seqexec.model.enum.Instrument
...
-    // Operations for Hamamatsu
-    case object PauseImmediatelyObservation extends ObservationOperations
-    case object StopImmediatelyObservation extends ObservationOperations
-    case object PauseGracefullyObservation extends ObservationOperations
-    case object StopGracefullyObservation extends ObservationOperations
-  }
+  import OperationLevel._
 
-  sealed trait SupportedOperations {
+  sealed trait Operations[A <: OperationLevel]
```
cquiroz: This is a very good idea

modules/seqexec/model/src/main/scala/seqexec/model/operations.scala
```diff
@@ -7,123 +7,139 @@ import seqexec.model.enum.Instrument
...
+    case object StopImmediatelyObservation extends Operations[NsNod]
+  }
+
+  sealed trait OperationLevelType[L <: OperationLevel]
+  implicit object ObservationLevel extends OperationLevelType[Observation]
+  implicit object NsCycleLevel extends OperationLevelType[NsCycle]
+  implicit object NsNodLevel extends OperationLevelType[NsNod]
+
+  sealed trait SupportedOperations {
+    def apply[L <: OperationLevel](isObservePaused: Boolean, isMultiLevel: Boolean)
```
cquiroz: I this can be simplified a bit with `apply[L <: OperationLevel: OperationLevelType]`
Given that `OperationLevelType` already has the L type, perhaps it can be extracted from the typeclass?

modules/seqexec/web/client/src/main/resources/less/style.less
```diff
@@ -429,32 +432,81 @@ body {
   margin-bottom: 1em;
 }
 
-.SeqexecStyles-acProgressRow {
+.SeqexecStyles-tableDetailRow {
```
cquiroz: Will be the same used for ac and ns?

modules/seqexec/web/client/src/main/resources/less/style.less
```diff
@@ -429,32 +432,81 @@ body {
...
   width: 100%;
 }
 
+.SeqexecStyles-dividedProgress {
+  width: 100%;
+  display: flex;
+}
+
+.SeqexecStyles-dividedProgressSectionLeft {
+  border-top-right-radius: 0 !important;
```
cquiroz: Did you try without the `!important`? I can imagine it is required but it is good to avoid it if possible

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
```diff
@@ -0,0 +1,84 @@
...
+  @inline def render: VdomElement = DividedProgress.component(this)
+}
+
+object DividedProgress {
+  type Props = DividedProgress
+
+  type Label = String
+  type Quantity = Int
+
+  implicit val propsReuse: Reusability[Props] = Reusability.byRef
```
cquiroz: I think this should be done with `Reusability.derive[DivideProgress]`, otherwise `byRef` will be essentially always `false`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
```diff
@@ -0,0 +1,84 @@
...
+        if (completeSections === 0)
+          List.empty
+        else
+          SeqexecStyles.dividedProgressBarLeft +:
+            List.fill(completeSections - 1)(SeqexecStyles.dividedProgressBarMiddle) :+
+            SeqexecStyles.dividedProgressBarRight
+
+      <.span(
+        SeqexecStyles.dividedProgress,
+        p.sections.zip(sectionValuesAndColors).zip(sectionProgressStyles).zip(sectionBarStyles.padTo(countSections, Css.Zero)).toTagMod {
```
cquiroz: Should we verify that the all the `zipped` list are of the same length? So we don't end up missing sections?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
```diff
@@ -0,0 +1,84 @@
...
+          case (((label, (sectionValue, sectionColor)), sectionProgressStyle), sectionBarStyle) =>
+            Progress(Progress.Props(
+              label,
+              p.sectionTotal.toLong,
+              sectionValue.toLong,
+              p.indicating,
+              p.progress,
+              sectionColor,
+              p.progressCls :+ sectionProgressStyle,
```
cquiroz: I think you can use a monoid add here? `p.progressCl |+| sectionProgressStyle`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
```diff
@@ -0,0 +1,84 @@
...
+            List.fill(completeSections - 1)(SeqexecStyles.dividedProgressBarMiddle) :+
+            SeqexecStyles.dividedProgressBarRight
+
+      <.span(
+        SeqexecStyles.dividedProgress,
+        p.sections.zip(sectionValuesAndColors).zip(sectionProgressStyles).zip(sectionBarStyles.padTo(countSections, Css.Zero)).toTagMod {
+          case (((label, (sectionValue, sectionColor)), sectionProgressStyle), sectionBarStyle) =>
+            Progress(Progress.Props(
+              label,
+              p.sectionTotal.toLong,
```
cquiroz: Does `Progress` take `Long`? that's a bit odd

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/ControlButtonResolver.scala
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.web.client.components.sequence.steps
+
+import seqexec.model.{SequenceState, Step}
+import seqexec.web.client.model.ClientStatus
+
+trait ControlButtonResolver[A] {
```
cquiroz: Can this be `sealed`?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,165 @@
...
+object NodAndShuffleCycleProgress extends CatsReactExt {
+  type Props = NodAndShuffleCycleProgress
+
+  implicit val propsReuse: Reusability[Props] = Reusability.derive[Props]
+
+  protected val component = ScalaComponent
+    .builder[Props]("NodAndShuffleCycleProgress")
+    .stateless
+    .render_P { p =>
+      val nsStatus = p.state.step.asInstanceOf[NodAndShuffleStep].nsStatus
```
cquiroz: Let's try avoiding `asInstanceOf`. It could be done with a `match` and return an Option

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,165 @@
...
+      val isInError = !p.state.isNSRunning && p.state.isNSInError
+
+      DividedProgress(
+        List.range(1, nsStatus.cycles + 1).map(_.show),
```
cquiroz: Why `show` here.I think it is basically the same as `toString`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,165 @@
...
+        nsStatus.nodExposureTime.toSeconds.toInt * NodAndShuffleStage.NsSequence.length,
+        value = 0,
```
cquiroz: Good, perhaps we should name all params

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,165 @@
...
+  implicit val propsReuse: Reusability[Props] = Reusability.derive[Props]
+
+  private val nodSections: List[DividedProgress.Label] =
+    NodAndShuffleStage.NsSequence.map(_.symbol.name).toList
+
+  protected val component = ScalaComponent
+    .builder[Props]("NodAndShuffleNodProgress")
+    .stateless
+    .render_P { p =>
+      val nsStatus = p.state.step.asInstanceOf[NodAndShuffleStep].nsStatus
```
cquiroz: Same as above

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,165 @@
...
+    }
+    .configure(Reusability.shouldComponentUpdate)
+    .build
+}
+
+final case class NodAndShuffleNodProgress(state: StepStateSnapshot) extends ReactProps {
+  @inline def render: VdomElement = NodAndShuffleNodProgress.component(this)
+}
+
+object NodAndShuffleNodProgress extends CatsReactExt {
```
cquiroz: Why are we using `CatsReactExt`?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/ObservationProgressDisplay.scala
```diff
@@ -89,15 +90,22 @@ object SmoothProgressBar {
...
-        total       = p.total,
-        value       = s.value,
-        color       = "blue".some,
-        progressCls = List(SeqexecStyles.observationProgressBar),
-        barCls      = List(SeqexecStyles.observationBar),
-        labelCls    = List(SeqexecStyles.observationLabel)
-      ))
+      if(p.hideBar)
+        <.div(SeqexecStyles.specialStateLabel,
+
```
cquiroz: extra lines?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/ObservationProgressDisplay.scala
```diff
@@ -89,15 +90,22 @@ object SmoothProgressBar {
...
-      Progress(Progress.Props(
-        label       = label,
-        total       = p.total,
-        value       = s.value,
-        color       = "blue".some,
-        progressCls = List(SeqexecStyles.observationProgressBar),
-        barCls      = List(SeqexecStyles.observationBar),
-        labelCls    = List(SeqexecStyles.observationLabel)
-      ))
+      if(p.hideBar)
```
cquiroz: I guess this is to hide the bar for N&S right?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/ObservationProgressDisplay.scala
```diff
@@ -110,49 +118,60 @@ object SmoothProgressBar {
...
-                color       = "blue".some,
-                progressCls = List(SeqexecStyles.observationProgressBar),
-                barCls      = List(SeqexecStyles.observationBar),
-                labelCls    = List(SeqexecStyles.observationLabel)
-              ))
+              val msg = if (p.paused) s"${p.fileId} - Paused" else p.fileId
+
+              if(p.hideBar)
+                <.div(SeqexecStyles.specialStateLabel,
+
```
cquiroz: remove extra lines

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsControlButtons.scala
```diff
@@ -29,6 +29,127 @@ import seqexec.web.client.semanticui.elements.icon.Icon.IconTrash
...
+) extends ReactProps {
+  @inline def render: VdomElement = ControlButtons.component(this)
+
+  val requestInFlight = tabOperations.stepRequestInFlight
+}
+
+object ControlButtons {
+  type Props = ControlButtons
+
+  implicit val operationsReuse: Reusability[Operations[_]] = Reusability.derive[Operations[_]]
```
cquiroz: I was bit surprised this'd work but then noted that `Operations[T]` doesn't have an internal reference to T. Is that why this compiles?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/package.scala
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2016-2019 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package seqexec.web.client.components.sequence
+
+package object steps {
+  implicit class ControlButtonResolverSyntax[A](val a: A) extends AnyVal {
```
cquiroz: I think we often name these as `ControlButtonResolverOps` but that's a tiny detail

rpiaggio:
> Great work. I'm very happy that you've been able to get this far so quickly. And there are some good creative solutions in the code
> I added a bunch of comments but most of the are not critical
> I'd like to avoid the use of `asInstanceOf` in a few places
> 
> Does it show when you cilck the row? or only when running?

Thank you @cquiroz and thank you for all your comments. I'll be adding commits with new functionality and answering your inline comments.

Yes, the detail rows are shown when a user clicks on the row.

modules/seqexec/web/client/src/main/resources/less/style.less
rpiaggio: Yes, that's the reason for renaming it.

modules/seqexec/web/client/src/main/resources/less/style.less
rpiaggio: Yes, I always try first without the `!important`.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
rpiaggio: I was getting warnings in the console about that and had it pending to look into it. Fixed now, thank you!

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
rpiaggio: That's a good idea. I'll log a warning if the final `zipped` list falls short.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
rpiaggio: Yes, it does. It caught my attention too, given that `Long` is opaque in Scala.js. Shall we change it to `Int`?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
rpiaggio: There's a `Monoid[Css]`, but in this case we are adding a single `Css` to a `List[Css]`, since that's what `Progress` expects. Is there a FP way to do this? Or... shall we change `Progress` to accept a single `Css` instead of `List[Css]`?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
cquiroz: I don't recall the reason, it seem we can do with `Int`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
cquiroz: Ah nevermind then. I got confused

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/ControlButtonResolver.scala
rpiaggio: Yes, good idea.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: Ok. I will first try to parametrize `StepStateSnapshot` on the `Step` type (which will move the problem to the caller, `StepsTable`, but I think we are already doing a `match` there). If it doesn't work, I'll implement as you suggest.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: I assumed I should use `show` whenver possible. When should I use each?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: Definitely. (Had already done it actually ;) )

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: I think I was using `Reusability.byEq` at some point, but now I'm using `.derive[Props]`. Removing it, good catch!

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/ObservationProgressDisplay.scala
rpiaggio: Exactly. The logic to build the message to display is in (previously named) `ObservationProgressBar` and `SmoothProgressBar`, which displayed the message as the progress bar label.

Extracting all the message building logic seemed quite complex, so I opted to rename the classes `*ProgressBar` = `*ProgressDisplay` and make showing the progress bar optional.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/ObservationProgressDisplay.scala
rpiaggio: Good catch, thank you.

(I do that sometimes with styles when I'm trying things out and leave the extra lines to remember to move the style to a class before committing, but this one slipped by.)

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsControlButtons.scala
rpiaggio: TBH, I was a bit surprised too. Probably that's the reason, since a `Reusability[T]` is not required.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
rpiaggio: Ok, I'll look into it.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/package.scala
rpiaggio: Ok, let's unify.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/StepItems.scala
```diff
@@ -215,21 +213,47 @@ object StepItems {
     }
   }
 
-  final case class StepStateSnapshot(step: Step, i: Instrument, t: TabOperations, state: SequenceState) {
+  final case class StepStateSnapshot(step: Step,
```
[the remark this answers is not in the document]
rpiaggio: How about `StepStateSummary` ?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/StepItems.scala
```diff
@@ -215,21 +213,47 @@ object StepItems {
...
+        2
+      else if (isACRunning || isACInError)
+        1
+      else
+        0
+  }
+
+  object StepStateSnapshot {
+    implicit val EqStepStateSnapshot: Eq[StepStateSnapshot] =
+      Eq.fromUniversalEquals
```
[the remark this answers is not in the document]
rpiaggio: Fixed!

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/StepItems.scala
```diff
@@ -215,21 +213,47 @@ object StepItems {
...
       isAC && anyError
 
-    val displayDetails: Boolean =
-      isACRunning || isACInError
+    val isNSInError: Boolean =
+      isNS && anyError
+
+    def detailRows(selected: Option[StepId]): Int =
+      if( (isNS && selected.exists(_ === step.id)) || isNSRunning || isNSInError)
+        2
```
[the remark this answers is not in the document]
rpiaggio: Done: `DetailRows.[NoDetailRows | OneDetailRow | TwoDetailRows]`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
rpiaggio: The -1 is a virtualized thing. The `rowClassName` function where that code appears is passed directly to the virtualized component (which passes -1 to it when querying for the header class name). I'm defining a `HeaderRow` constant anyway for comparison.

cquiroz: Right, it comes from virtualized. I just thought it would be better if we define it as our constant
If you want to do it on the facade please send a PR
It should probably be done in scala style: `HeaderRow`

rpiaggio: Even if it's a constant? (I defined it right below `MIDDLE_CLICK` ;) )

rpiaggio:
PR #1119 integrated.

modules/seqexec/web/client/src/main/resources/less/style.less
```diff
@@ -429,32 +432,81 @@ body {
...
-  padding: @table_row_left_padding;
+  padding: @table_detail_row_top_padding @table_detail_row_right_padding @table_detail_row_bottom_padding @table_row_left_padding;
   z-index: 100;
   position: relative;
   margin-left: 20px;
 }
 
+.SeqexecStyles-nodAndShuffleDetailRow {
+  display: flex;
+  justifyContent: spaceBetween;
```
cquiroz: I think this should be `space-between`
https://css-tricks.com/almanac/properties/j/justify-content/

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/AlignAndCalibProgress.scala
```diff
@@ -95,8 +94,8 @@ object ACProgressBar {
       val msg = if (isInError) "Error" else s.msg
       Progress(Progress.Props(
         s"Align and Calib: $msg",
-        total       = enum.all.length.toLong - 1,
-        value       = max(0, s.counter.toLong),
+        total       = enum.all.length - 1,
```
cquiroz: Where does the `enum` come from. Maybe we should change the name given `enum` will be a reserved word in Dotty

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,266 @@
...
+                                 $.backend.newStateFromProps($.currentProps, $.nextProps))
+    .configure(TimerSupport.install)
+    .configure(Reusability.shouldComponentUpdate)
+    .build
+}
+
+final case class NodAndShuffleProgress(
+  summary: StepStateSummary,
+  sections: NodAndShuffleStatus => List[DividedProgress.Label],
+  quantitiesFromNodMillis: // (nsStatus, nodMillis) => (sectionTotal, currentValue)
```
cquiroz: 👍

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,266 @@
...
+    SeqexecCircuit.connect(SeqexecCircuit.obsProgressReader(summary.obsId, summary.step.id))
+}
+
+object  NodAndShuffleProgress {
+  type Props = NodAndShuffleProgress
+
+  @Lenses
+  final case class State(progressConnect: ReactConnectProxy[Option[ObservationProgress]])
+
+  implicit val propsReuse: Reusability[Props] = Reusability.by(_.summary)
```
cquiroz: This means we don't re-render on changes to `sections` and `quantities`. Is that because these don't change inside a step?
Eventually the user could change the definitions in the OT (e.g. change the amount of cycles) and resync. I believe this deep on the stack it would be refreshed anyway

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,266 @@
...
+    .initialStateFromProps(p => State(p.connect))
+    .render_PS { (p, s) =>
+      s.progressConnect { proxy =>
+        val (totalMillis, remainingMillis) =
+          proxy()
+            .map(p => (p.total.toMilliseconds.toInt, p.remaining.toMilliseconds.toInt))
+            .getOrElse((0, 0))
+        val elapsedMillis = totalMillis - max(0, remainingMillis)
+
+        p.summary.nsStatus.map[VdomElement] { nsStatus =>
```
cquiroz: nice

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,266 @@
...
+  implicit val stateReuse: Reusability[State] = Reusability.always
+
+  protected val component = ScalaComponent
+    .builder[Props]("NodAndShuffleProgress")
+    .initialStateFromProps(p => State(p.connect))
+    .render_PS { (p, s) =>
+      s.progressConnect { proxy =>
+        val (totalMillis, remainingMillis) =
+          proxy()
+            .map(p => (p.total.toMilliseconds.toInt, p.remaining.toMilliseconds.toInt))
```
cquiroz: I think this could be changed to `foldMap` removing the `getOrElse`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -0,0 +1,266 @@
...
+
+trait NodAndShuffleRow[L <: OperationLevel] {
+  type Props <: NodAndShuffleRowProps
+
+  implicit val propsReuse: Reusability[Props]
+
+  implicit val propsControlButtonResolver: ControlButtonResolver[Props] =
+    ControlButtonResolver.build(p => (p.clientStatus, p.stateSummary.state, p.stateSummary.step))
+
+  implicit private val operationLevelType: OperationLevelType[L] = implicitly[OperationLevelType[L]]
```
cquiroz: Who uses this?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/SmoothProgressBar.scala
```diff
@@ -0,0 +1,61 @@
...
+import scala.math.min
+
+trait SmoothProgressBarProps extends ReactProps {
+  val value: Int
+  val maxValue: Int
+  val stopping: Boolean
+  val paused: Boolean
+}
+
+trait SmoothProgressBar[P <: SmoothProgressBarProps] {
```
cquiroz: 👍

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepProgressControl.scala
```diff
@@ -178,10 +181,10 @@ object StepProgressCell {
       ).when(props.step.canRunFrom && props.clientStatus.canOperate),
       <.div(
         SeqexecStyles.specialStateLabel,
-        if (props.stateSnapshot.isAC) {//} && !props.stateSnapshot.anyError) {
-          if (props.stateSnapshot.isACRunning) {
+        if (props.stateSummary.isAC) {//} && !props.stateSnapshot.anyError) {
```
cquiroz: Let's remove the unused code?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -441,6 +445,7 @@ object StepsTable extends Columns {
   type ReceiveProps = ComponentWillReceiveProps[Props, State, Unit]
 
   private val MIDDLE_BUTTON = 1 // As defined by React.js
+  private val HEADER_ROW = -1
```
cquiroz: I guess we want constants to follow the scala style `MiddleButton` and `HeaderRow`

modules/seqexec/web/client/src/main/resources/less/style.less
rpiaggio: Yes! And since it doesn't seem to be affecting the display, we should remove it ;)

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/AlignAndCalibProgress.scala
rpiaggio: That's a `gem.util.Enumerated[AlignAndCalibStep]` defined in the same file. I'm renaming to `acSteps`.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: `sections` and `quantities` are functions, that's why I left them out of reusability.

Which actually means they are being created on each update, and we don't want that. I'm considering using reusable functions (https://github.com/japgolly/scalajs-react/blob/master/doc/PERFORMANCE.md#reusable-functions), but I'm afraid that in that case they may not get updated if the definitions change in the OT.

I'll look into the issue. I will probably try refactoring so that `NodAndShuffleProgress` is a superclass of `NodAndShuffleCycleProgress` and `NodAndShuffleNodProgress` instead of a child component.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: It's passed into `SupportedOperationsOps.operations` in order to be able to `match` on the type parameter `L <: OperationLevel`. I accept suggestions as to other ways of doing this.

BTW, I had to move the implicit resolutions into the subclasses of `NodAndShuffleRow`, that's why the buttons were not being displayed.

rpiaggio: Hmm.... I think I'm going to try to create a 2-parameter typeclass depending on the `[Instrument, OperationLevel]` and simplify the whole logic.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
rpiaggio: I had not fully internalized this convention for Scala constants. I have now ;) Renaming...

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
cquiroz: me too

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
cquiroz: Ok, sounds good, I never could understand the reusable functions thing to be honest

PR #1126, opened 2019-10-04
Sbt...

cquiroz:
Sbt didn't fully like my fix, copy paste to the rescue


cquiroz:
I wish sbt a slow and painful death

PR #1127, opened 2019-10-04
SEQNG-1063 Turn off GCAL lamps for GMOS darks and biases.

jluhrs:
GMOS is picky when it comes to taking darks and biases, so this PR turns the GCAL lamps off for them. A consequence is that GMOS darks and biases will prevent the execution of other calibrations in parallel, because they take control of GCAL.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
@@ -20,20 +20,22 @@ object StepType {
   final case class FlatOrArc(override val instrument: Instrument) extends StepType
   final case class DarkOrBias(override val instrument: Instrument) extends StepType
   final case class DarkOrBiasNS(override val instrument: Instrument) extends StepType
+  final case class ExclusiveDarkOrBias(override val instrument: Instrument) extends StepType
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
```diff
@@ -20,20 +20,22 @@ object StepType {
...
-    case _                                        => false
+    case (CelestialObject(i), CelestialObject(j))         => i === j
+    case (Dark(i), Dark(j))                               => i === j
+    case (NodAndShuffle(i), NodAndShuffle(j))             => i === j
+    case (Gems(i), Gems(j))                               => i === j
+    case (AltairObs(i), AltairObs(j))                     => i === j
+    case (FlatOrArc(i), FlatOrArc(j))                     => i === j
+    case (DarkOrBias(i), DarkOrBias(j))                   => i === j
+    case (DarkOrBiasNS(i), DarkOrBiasNS(j))               => i === j
+    case (ExclusiveDarkOrBias(i), ExclusiveDarkOrBias(j)) => i === j
```
jluhrs: New step type for Darks or Biases that require a dark GCAL.

PR #1132, opened 2019-10-10
[SEQNG-1078] N&S progress messages and total remaining time.

rpiaggio:
I undid the previous changes to the progress bar component and abstracted away the logic to compute the progress message, which is now shared with a specialized component to show it in N&S.

@cquiroz I'd appreciate your input on generalizing the standard and N&S step reusability instances without `config`.

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
@@ -17,7 +17,7 @@ import japgolly.scalajs.react.extra.TimerSupport
 import monocle.macros.Lenses
 import react.common.Css
 import seqexec.model.dhs.ImageFileId
-import seqexec.model.{NSRunningState, NSSubexposure, NodAndShuffleStatus, ObservationProgress, StepId}
+import seqexec.model.{NodAndShuffleStatus, ObservationProgress, StepId}
 import seqexec.model.operations._
 import seqexec.web.client.circuit.SeqexecCircuit
 import seqexec.web.client.model.{ClientStatus, StopOperation}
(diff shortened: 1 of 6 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/reusability/package.scala
```diff
@@ -104,4 +95,20 @@ package object reusability {
     Reusability.derive[M2GuideConfig]
   implicit val configReuse: Reusability[TelescopeGuideConfig] =
     Reusability.derive[TelescopeGuideConfig]
+  val stdStepReuse: Reusability[StandardStep] =
+    Reusability.caseClassExcept('config)
```
cquiroz: There are a few reason for this is, computing `config` could potentially be slow as the map inside could be largish and we do it for each step (at least displayed steps).
Additionally the only reason ever for config to change is if you do a `Sync` and that will refresh the whole table anyway.
Finally in `StepsTable` we don't display the config. we show it on the `StepConfigTable`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -24,6 +26,54 @@ import seqexec.web.client.semanticui._
...
+
+  protected[steps] val connect =
+    SeqexecCircuit.connect(SeqexecCircuit.obsProgressReader(obsId, stepId))
+}
+
+object NodAndShuffleProgressMessage extends ProgressLabel {
+  type Props = NodAndShuffleProgressMessage
+
+  @Lenses
+  protected case class State(progressConnect: ReactConnectProxy[Option[ObservationProgress]])
```
cquiroz: Why is the connect in `State` Wouldn't be enough for it to live in the `Props` and make the component `stateless`?

rpiaggio: I added a comment about it in another use in the same file: https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala#L85

I followed the suggested way at the end of "Wrap and Connect Components" section in https://diode.suzaku.io/UsageWithReact.html.

If we keep in the props, when the props are recreated (which happens when a nod ends and another one starts), the component is unmounted and remounted.

Another option would be to have the `connect` it in a backend, but we would need to store it in a `var` and deal with initialization logic, since we need the `Props` to initialize it.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/reusability/package.scala
rpiaggio: Thank you, it seems then that it's safe to have these global `Reusability` instances that ignore `config`.

rpiaggio:
Although we've already merged, just in case, this PR is about the total remaining observation time message for N&S:

SEQNG-1078

PR #1133, opened 2019-10-10
[SEQNG-1081] Show ellipsis and popup when error messages are too long.

rpiaggio:
**Before:**

image

**After:**

SEQNG-1081-AFTER

The popup also works when the message is touched on mobile devices.


cquiroz:
This is great, thanks. I think `ellipsis` is a bit finicky in terms of browser support. Could you verify at least FF and Chrome. And we can check how it goes on the consoles later on

rpiaggio:
It's working in Chrome and Firefox ESR, therefore I'm merging.

PR #1134, opened 2019-10-10
Improve construction of subsystems

cquiroz:
Constructions of subsystems has been distributed across several places, web server and the seqexec engine. It also uses some unsafe actions. 
This PR tides this removing the `unsafe` calls and it puts the initialization of all systems in one place plus some other internal improvements

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerSim.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerSim.scala
@@ -2,36 +2,33 @@
 // For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
 package seqexec.server.gems
-import cats.effect.Sync
+
+import cats.Applicative
 import cats.implicits._
+import io.chrisdavenport.log4cats.Logger
 import seqexec.server.tcs.Gaos.{PauseConditionSet, PauseResume, ResumeConditionSet}
-import org.log4s.getLogger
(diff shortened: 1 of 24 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerSim.scala
```diff
@@ -2,36 +2,33 @@
...
+
+import cats.Applicative
 import cats.implicits._
+import io.chrisdavenport.log4cats.Logger
 import seqexec.server.tcs.Gaos.{PauseConditionSet, PauseResume, ResumeConditionSet}
-import org.log4s.getLogger
 import seqexec.server.gems.Gems.{GemsWfsState, Cwfs1DetectorState, Cwfs2DetectorState, Cwfs3DetectorState, Odgw1DetectorState, Odgw2DetectorState, Odgw3DetectorState, Odgw4DetectorState}
 import seqexec.server.gems.GemsController.GemsConfig
 
-final class GemsControllerSim[F[_]: Sync] private extends GemsController[F] {
```
jluhrs: But, but, I do like the explicit class. I only like anonymous classes for small definitions.

cquiroz: Ok, I can revert that

jluhrs: No need, it's just a matter of preferences.

PR #1138, opened 2019-10-14
SEQNG-1088 Set electronic offset in between nods

cquiroz:
Tests showed that e-offset wasn't being properly set with N&S.
This PR fixes it, the core reason was a badly read type from the ocs config

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosController.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosController.scala
@@ -23,9 +23,6 @@ import seqexec.server.InstrumentSystem.ElapsedTime
 import seqexec.server._
 import squants.{Length, Time}
 import shapeless.tag
-import shapeless.tag.@@
-
-trait UseElectronicOffsetI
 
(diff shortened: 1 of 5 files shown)


jluhrs:
Wouldn't be better to define a proper type for electronic offsets, instead of using `Boolean`? Something like what is used for GeMS detectors in Gems.scala.

cquiroz:
That makes sense. I'll define an ADT for that

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosController.scala
```diff
@@ -131,6 +130,20 @@ object GmosController {
         Enumerated.of(InBeam, OutOfBeam)
     }
 
+    sealed trait ElectronicOffset extends Product with Serializable
+
+    object ElectronicOffset {
+      case object ElectronicOffsetOn extends ElectronicOffset
+      case object ElectronicOffsetOff extends ElectronicOffset
```
jluhrs: I think `On` and `Off` are enough. In every use of those values it is clear that they refer ro `ElectronicOffset`.

PR #1140, opened 2019-10-15
SEQNG-1083 Refactored paused observation commands

jluhrs:
Paused observations commands were build from two pieces: a function that would send a command to the instrument and wait for a result, and a function that would process that result. The first function could be a continue, a stop or an abort, and it does not depend on the actual observation, only the instrument. The second function does depend on the observation, and was build at the moment the observation was paused.
This scheme covered all instruments, except for GMOS N&S. This PR changes that. Now, when the observation is paused, three functions are built: one to continue the observation, one to stop it, and one to abort it. For normal observations they are built the same way than before. When the user call for resuming, stopping or aborting the paused observation, the right function is retrieved and called. This allows to have specialized functions for N&S without changing the way the commands are called by the web server.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
@@ -166,13 +166,13 @@ trait ObserveActions {
         abortTail(env.systems, env.obsId, fileId)
       case ObserveCommandResult.Paused =>
         env.inst.calcObserveTime(env.config)
-          .flatMap(t => env.inst.observeControl match {
+          .flatMap(totalTime => env.inst.observeControl match {
             case c: CompleteControl[F] => Result.Paused(
               ObserveContext[F](
-                (t: Time) => Stream.eval(c.continue.self(t)).flatMap(observeTail(fileId, dataId, env)),
+                (elapsed: Time) => Stream.eval(c.continue.self(elapsed)).flatMap(observeTail(fileId, dataId, env)),
                 Stream.eval(c.stopPaused.self).flatMap(observeTail(fileId, dataId, env)),
                 Stream.eval(c.abortPaused.self).flatMap(observeTail(fileId, dataId, env)),
-                t
+                totalTime
               )
             ).pure[F].widen[Result[F]]
             case _                     =>
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
@@ -11,9 +11,10 @@ import io.chrisdavenport.log4cats.Logger
 import seqexec.engine._
 import seqexec.model.dhs._
 import seqexec.model.enum.ObserveCommandResult
-import seqexec.server.InstrumentSystem.ElapsedTime
+import seqexec.server.InstrumentSystem.{CompleteControl, ElapsedTime}
 import squants.time.TimeConversions._
 import SeqTranslate.dataIdFromConfig
+import squants.time.Time
 
 /**
   * Methods usedd to generate observation related actions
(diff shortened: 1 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/package.scala
```diff
@@ -71,7 +71,12 @@ package server {
     val default: HeaderExtraData = HeaderExtraData(Conditions.Default, None, None)
   }
 
-  final case class ObserveContext[F[_]](t: ObserveCommandResult => Stream[F, Result[F]], expTime: Time) extends PauseContext[F]
+  final case class ObserveContext[F[_]](
```
cquiroz: I guess this is the core change, right?

modules/seqexec/server/src/test/scala/seqexec/server/SeqTranslateSpec.scala
```diff
@@ -14,7 +14,9 @@ import gem.Observation
 import gem.enum.Site
 import io.chrisdavenport.log4cats.noop.NoOpLogger
 import java.time.LocalDate
+
```
cquiroz: why empty lines?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -226,95 +225,49 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
-      executeEngine.EventType]] = {
-
-      val inst = toInstrumentSys(instrument).toOption
-
-      val observeIndex: Option[(ObserveContext[IO], Option[Time], Int)] =
-        seqState.current.execution.zipWithIndex.find(_._1.kind === ActionType.Observe).flatMap {
+  ): EngineState => Option[Stream[IO, executeEngine.EventType]] = (st: EngineState) => {
+    val observeIndex: Option[(ObserveContext[IO], Option[Time], Int)] =
+      st.sequences.get(seqId)
+        .flatMap(_.seq.current.execution.zipWithIndex.find(_._1.kind === ActionType.Observe).flatMap {
```
cquiroz: One thing here. When I was doing N&S I did it with multiple `Observe` actions and was hit by this `find` which only returns one.
At the end it was not a problem because for other reasons we had to revert to a single `Observe` action but the underlying assumption of a single action remains.
Could/Should we generalize this?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -226,95 +225,49 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
-             tio: Timer[IO]
-  ): EngineState => Option[Stream[IO, executeEngine.EventType]] = {
-    def f(o: ObserveControl[IO]): Option[Time => IO[ObserveCommandResult]] = o match {
-      case CompleteControl(_, _, _, ContinuePausedCmd(a), _, _) => a.some
-      case _                                                    => none
+    (st.sequences.get(seqId).flatMap{x => toInstrumentSys(x.seqGen.instrument).toOption}, observeIndex).mapN{
+      case (ins, (obCtx, t, i)) => Stream.eval(IO(Event.actionResume(seqId, i,
+        ins.observeProgress(obCtx.expTime, ElapsedTime(t.getOrElse(0.0.seconds)))
+          .map(Result.Partial(_))
+          .mergeHaltR(obCtx.resumePaused(obCtx.expTime))
```
cquiroz: Is it `mergeHaltR` because we don't much care about `Progress` halting?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -226,95 +225,49 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
-      case _                                                => none
-    }
-
-    pausedCommand(seqId, f, useCountdown = false)
-  }
+  private def endPaused(seqId: Observation.Id, l: ObserveContext[IO] => Stream[IO, Result[IO]])(st: EngineState)
+  : Option[Stream[IO, executeEngine.EventType]] =
+    st.sequences.get(seqId)
+      .flatMap(
+        _.seq.current.execution.zipWithIndex.find(_._1.kind === ActionType.Observe).flatMap {
```
cquiroz: Same question as above

modules/seqexec/server/src/main/scala/seqexec/server/package.scala
jluhrs: Yes

modules/seqexec/server/src/test/scala/seqexec/server/SeqTranslateSpec.scala
jluhrs: That's probably Idea messing with the imports.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: I am not sure. The premise here is that for each step there is one observe and one image.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: Same answer: one step, one observe, one image.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
cquiroz: I don't think we need to fix it now, but we need to be aware of this, perhaps a comment?

modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -164,13 +165,24 @@ trait ObserveActions {
...
         abortTail(env.systems, env.obsId, fileId)
       case ObserveCommandResult.Paused =>
-        env.inst
-          .calcObserveTime(env.config)
-          .map(e => Result.Paused(ObserveContext(observeTail(fileId, dataId, env), e)))
+        env.inst.calcObserveTime(env.config)
+          .flatMap(t => env.inst.observeControl match {
+            case c: CompleteControl[F] => Result.Paused(
+              ObserveContext[F](
+                (t: Time) => Stream.eval(c.continue.self(t)).flatMap(observeTail(fileId, dataId, env)),
```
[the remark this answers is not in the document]
jluhrs: Ups.

PR #1141, opened 2019-10-15
[SEQNG-1093] & [SEQNG-1095] N&S UI Fixes.

rpiaggio:


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -1071,9 +1077,10 @@ object StepsTable extends Columns {
       TableContainer.Props(
         b.props.hasControls,
         size => {
+          val areaSize = Size(size.height, size.width - 15) // Subtract scroll bar width
```
cquiroz: Where does the 15 come from? It feels it would be browser/OS dependent

rpiaggio: Actually, it is. Let me add some code to detect the browser.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -1023,6 +1032,24 @@ object StepsTable extends Columns {
...
+          case _                     => 0.0
+        }
+      }.getOrElse(0.0)
+    }
+  }
+
+  def didUpdate(b: DidUpdate): Callback = {
+    computeScrollBarWidth >>= {sw => b.modState(State.scrollBarWidth.set(sw)) }
+  }
+
```
rpiaggio: TBH, this is a hack and I don't like it: we are relying on `virtualized`'s internal DOM structure.

However, it was the only way I could find to correctly and (hopefully) universally compensate for the scroll bar width.

It's worth noting that `virtualized` does know the scroll bar width: https://github.com/bvaughn/react-virtualized/blob/bc5ae2fa084e20323b445b31fd72acee90e0ab5f/source/Table/Table.js#L727 but it doesn't expose. Maybe we could submit a PR to them and live with this until it's published?

cquiroz: Right, and I want to remove the jquery dependency. Shouldn't it be possible to call `_setScrollbarWidth` via `js.Dynamic`? I'm not totally sure. and we probably shouldn't spend too much time on this
Somehow I thought that there would be some DOM property that would give you that but I guess there isn't

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -1023,6 +1032,24 @@ object StepsTable extends Columns {
...
+    import org.querki.jquery.$
+
+    ref.get.map(_.getDOMNode.toHtml).asCallback.map{
+      _.flatten.flatMap { tableNode =>
+        // Table has a Grid inside, which is the one actually showing the scroll bar.
+        $(tableNode).find(".ReactVirtualized__Table__Grid").get(0).toOption.map {
+          case gridNode: HTMLElement => gridNode.offsetWidth - gridNode.clientWidth
+          case _                     => 0.0
+        }
+      }.getOrElse(0.0)
```
cquiroz: Maybe we should default to 15?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
```diff
@@ -1023,6 +1032,24 @@ object StepsTable extends Columns {
     }.getOrEmpty
   }
 
+  private val computeScrollBarWidth: CallbackTo[Double] = {
+    import org.querki.jquery.$
+
+    ref.get.map(_.getDOMNode.toHtml).asCallback.map{
+      _.flatten.flatMap { tableNode =>
+        // Table has a Grid inside, which is the one actually showing the scroll bar.
+        $(tableNode).find(".ReactVirtualized__Table__Grid").get(0).toOption.map {
```
cquiroz: Maybe we can use `querySelector` and put it inside the `Option`
`Option(tableNode.querySelctor(".React....A))`?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsTable.scala
rpiaggio: Good idea, done!

PR #1142, opened 2019-10-16
Replace seqexec offsets by gsp offsets

cquiroz:
For a long time the seqexec had an `Offset` class and then `gsp` produced one but they were not unified. This PR does that unification and it should help @rpiaggio with a task.
Thanks to @swalker2m and @tpolecat with their help trying to wrap my mind around the new optics


modules/seqexec/model/src/main/scala/seqexec/model/ModelLenses.scala
```diff
@@ -147,10 +147,14 @@ trait ModelLenses {
     paramValueL(SystemName.Observe.withParam("object")) ^<-? // find the target name
     some                                                     // focus on the option
 
-  private[model] def telescopeOffsetPI: Iso[Double, TelescopeOffset.P] =
-    Iso(TelescopeOffset.P.apply)(_.value)
-  private[model] def telescopeOffsetQI: Iso[Double, TelescopeOffset.Q] =
-    Iso(TelescopeOffset.Q.apply)(_.value)
+  val signedArcsecFormat: Format[String, Angle] =
```
tpolecat: (for posterity … Carlos and I talked about this on Slack) Can we add tests for these? See [AngleSpec](https://github.com/gemini-hlsw/gsp-math/blob/master/modules/tests/shared/src/test/scala/gsp/math/AngleSpec.scala#L35), keeping in mind you'll need multiple strings that map to the same Angle (otherwise it will seem like it's a Prism and our test checks for this) so you the pattern in [ArbAngle](https://github.com/gemini-hlsw/gsp-math/blob/master/modules/testkit/shared/src/main/scala/gsp/math/arb/ArbAngle.scala#L37-L53) may be helpful.

PR #1143, opened 2019-10-16
Update sbt

cquiroz:
sbt update and removal of a deprecated feature (exclusion rule)

I verified that seqexec distributions work


cquiroz:
It only affects giapi and makes the test more verbose, We can probably be ok with that and I don't want the sbt warning

PR #1146, opened 2019-10-17
Remove unused bias time parameter

cquiroz:
The BiasTime parameter is not used and could be removed


cquiroz:
If I delete all the code, do I get a gigantic bonus?

PR #1148, opened 2019-10-18
SEQNG-1069 Bad reported N&S exposure

cquiroz:
The N&S exposure time with the new seqexec and the old differ
Looking at the old code I don't see we dividing the exposure so I'm doing the same here, removing the division
We need to still divide it for the simulator

This needs further on instrument testing


cquiroz:
Tests show this is not the right approach

PR #1149, opened 2019-10-18
SEQNG-328 N&S interrupt commands

jluhrs:
Implementation of the N&S interruption commands pause, pause gracefully, stop, stop gracefully, and abort.
This PR also includes the commands to resume, stop or abort a paused N&S observation.
The commands were tested sending HTTP commands with curl.

### DIFF
--- a/modules/seqexec/model/src/main/scala/seqexec/model/enum/ObserveCommandResult.scala
+++ b/modules/seqexec/model/src/main/scala/seqexec/model/enum/ObserveCommandResult.scala
@@ -12,9 +12,8 @@ object ObserveCommandResult {
   case object Paused  extends ObserveCommandResult
   case object Stopped extends ObserveCommandResult
   case object Aborted extends ObserveCommandResult
-  case object Partial extends ObserveCommandResult // Used for N&S
 
   /** @group Typeclass Instances */
   implicit val ObserveCommandResultEnumerated: Enumerated[ObserveCommandResult] =
-    Enumerated.of(Success, Paused, Stopped, Aborted, Partial)
+    Enumerated.of(Success, Paused, Stopped, Aborted)
 }
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -8,6 +8,7 @@ import cats.data.{EitherT, NonEmptyList, NonEmptySet}
 import cats.effect.{Concurrent, IO, Timer}
 import cats.effect.Sync
 import cats.effect.LiftIO
+import cats.effect.concurrent.Ref
 import cats.implicits._
 import edu.gemini.seqexec.odb.{ExecutedDataset, SeqexecSequence}
 import edu.gemini.spModel.gemini.altair.AltairParams.GuideStarType
@@ -48,10 +49,11 @@ import seqexec.server.altair.AltairKeywordReaderDummy
 import seqexec.server.gems.{Gems, GemsEpics, GemsHeader, GemsKeywordReaderDummy, GemsKeywordReaderEpics}
 import seqexec.server.CleanConfig.extractItem
 import seqexec.server.ConfigUtilOps._
+import seqexec.server.gmos.NSObserveCommand
 import squants.Time
 import squants.time.TimeConversions._
 
-class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings)(implicit L: Logger[IO]) extends ObserveActions {
+class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings, gmosNsCmd: Ref[IO, Option[NSObserveCommand]])(implicit L: Logger[IO]) extends ObserveActions {
   import SeqTranslate._
 
   private def step(obsId: Observation.Id, i: StepId, config: CleanConfig, nextToRun: StepId,
@@ -84,10 +86,7 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
         val env = ObserveEnvironment(systems, config, stepType, obsId, inst, sys.filterNot(inst.equals), headers, ctx)
         // Request the instrument to build the observe actions and merge them with the progress
         // Also catches any errors in the process of runnig an observation
-        ia.observeActions(env, (s, env) =>
-          ia.observationProgressStream(env)
-            .mergeHaltR(s)
-            .handleErrorWith(catchObsErrors[IO]))
+        ia.observeActions(env)
       }
 
       extractStatus(config) match {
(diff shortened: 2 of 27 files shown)


modules/seqexec/model/src/main/scala/seqexec/model/enum/ObserveCommandResult.scala
```diff
@@ -12,9 +12,8 @@ object ObserveCommandResult {
   case object Paused  extends ObserveCommandResult
   case object Stopped extends ObserveCommandResult
   case object Aborted extends ObserveCommandResult
-  case object Partial extends ObserveCommandResult // Used for N&S
 
   /** @group Typeclass Instances */
   implicit val ObserveCommandResultEnumerated: Enumerated[ObserveCommandResult] =
-    Enumerated.of(Success, Paused, Stopped, Aborted, Partial)
+    Enumerated.of(Success, Paused, Stopped, Aborted)
```
cquiroz: Good riddance

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -84,10 +86,7 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
         val env = ObserveEnvironment(systems, config, stepType, obsId, inst, sys.filterNot(inst.equals), headers, ctx)
         // Request the instrument to build the observe actions and merge them with the progress
         // Also catches any errors in the process of runnig an observation
-        ia.observeActions(env, (s, env) =>
-          ia.observationProgressStream(env)
-            .mergeHaltR(s)
-            .handleErrorWith(catchObsErrors[IO]))
+        ia.observeActions(env)
```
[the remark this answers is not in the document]
cquiroz: I see, it is in the action instead

modules/seqexec/server/src/main/scala/seqexec/server/flamingos2/Flamingos2.scala
```diff
@@ -32,7 +31,10 @@ import squants.time.{Seconds, Time}
 
 import scala.reflect.ClassTag
 
-final case class Flamingos2[F[_]: Sync: Timer: Logger](f2Controller: Flamingos2Controller[F], dhsClient: DhsClient[F]) extends DhsInstrument[F] with InstrumentSystem[F] {
+final case class Flamingos2[F[_]: Sync: Timer: Logger: Concurrent](
```
cquiroz: With `Concurrent` you don't need `Sync`

modules/seqexec/server/src/main/scala/seqexec/server/ghost/Ghost.scala
```diff
@@ -30,7 +30,7 @@ import squants.time.Time
 
 import scala.reflect.ClassTag
 
-final case class Ghost[F[_]: Sync: Logger](controller: GhostController[F])
+final case class Ghost[F[_]: Sync: Logger: Concurrent](controller: GhostController[F])
```
cquiroz: Same as above

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -54,17 +57,47 @@ abstract class Gmos[F[_]: MonadError[?[_], Throwable]: Concurrent: Logger, T <:
 
   override val keywordsClient: KeywordsClient[F] = this
 
-  val continueCommand: Time => F[ObserveCommandResult] =
-    controller.resumePaused
+  val nsCmdRef: Ref[F, Option[NSObserveCommand]] = nsCmdR
```
cquiroz: Why are you aliasing?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
```diff
@@ -54,17 +57,47 @@ abstract class Gmos[F[_]: MonadError[?[_], Throwable]: Concurrent: Logger, T <:
...
+        AbortObserveCmd(_ => controller.abortObserve),
+        PauseObserveCmd(_ => controller.pauseObserve),
+        ContinuePausedCmd(controller.resumePaused),
+        StopPausedCmd(controller.stopPaused),
+        AbortPausedCmd(controller.abortPaused)
+      )
+
+  def stopNS(gracefully: Boolean): F[Unit] =
+    if(gracefully)
+      nsCmdRef.set(NSObserveCommand.StopGracefully.some)
```
cquiroz: I'm not totally familiar with `Ref` but I think it is better to use `update` rather than `set` to ensure updates are safe in presence of concurrence

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosControllerSim.scala
```diff
@@ -145,6 +145,7 @@ object GmosControllerSim {
       ): fs2.Stream[F, Progress] =
         sim.observeCountdown(total, elapsed)
 
+      override def nsCount: F[Int] = nsConfig.get.map(_.current.map(_.exposureCount).getOrElse(0))
```
cquiroz: `.foldMap(_.exposureCount)`

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -63,49 +66,97 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Concurrent: Logge
...
 
+  private def continueResult(
+    fileId: ImageFileId,
+    dataId: DataId,
+    env:    ObserveEnvironment[F],
+    nsCfg: NSConfig.NodAndShuffle,
+    subExp: NSSubexposure,
+    nsObsCmd: Option[NSObserveCommand])(obsResult: ObserveCommandResult): F[Result[F]] =
+    (nsObsCmd, obsResult) match {
+      case (Some(PauseImmediately), ObserveCommandResult.Paused) |
```
cquiroz: 👏

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosInstrumentActions.scala
```diff
@@ -138,31 +190,30 @@ class GmosInstrumentActions[F[_]: MonadError[?[_], Throwable]: Concurrent: Logge
...
-           Stream.eval(initialObserve(fileId, env))
-         } else if (sub.lastSubexposure) {
-           Stream.eval(lastObserve(fileId, env))
-         } else {
-           Stream.eval(continueObserve(env))
-         }) ++
-        Stream.emit(Result.Partial(NSSubexposureEnd(sub))),
-      env
-    )
+      observationProgressStream(env)
```
cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosNorth.scala
```diff
@@ -22,7 +23,11 @@ import edu.gemini.spModel.gemini.gmos.InstGmosNorth._
 import squants.Length
 import squants.space.Arcseconds
 
-final case class GmosNorth[F[_]: MonadError[?[_], Throwable]: Concurrent: Logger](c: GmosNorthController[F], dhsClient: DhsClient[F]) extends Gmos[F, NorthTypes](c,
+final case class GmosNorth[F[_]: MonadError[?[_], Throwable]: Concurrent: Logger](
```
cquiroz: Same comment as above, I think you don't need `MonadError` here

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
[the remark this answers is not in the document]
jluhrs: Yes, that way each instrument can build its observation progress in a different way. That was important for N&S, where you have time countdown only inside the subexposures. and need to add nod counts in the progress information.

jluhrs:
There is one details that may need more explanation.The immediate interruptions send commands to the instrument that affect the currently running subexposure. That is reflected in the output result of the subexposure. But if the command is processed just as the subexposure completes, its result may not reflect the attempted command. Because of that, the nsCmd Ref is checked after every subexposure, in case there was an interruption command that failed to affect the subexposure in the instrument.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/Gmos.scala
jluhrs: But the code documentation fro `update` says:
```
   * Satisfies:
   *   `r.update(_ => a) == r.set(a)`
```

cquiroz: You can check Flabio's talk about this. As I understand it, it may happen that updates are lost if multiple threads call set

https://vimeo.com/366191463

TBH I don't know if in practice it makes any difference for us

PR #1150, opened 2019-10-18
Upgrade gsp-math to 0.1.7 and tests to run with scalatest 3.1.0.

rpiaggio:


project/Settings.scala
```diff
@@ -62,7 +62,6 @@ object Settings {
     val flywayVersion           = "6.0.4"
 
     // test libraries
-    val scalaTest               = "3.0.5"
     val scalaCheck              = "1.13.5"
```
cquiroz: I guess we don't need scalaCheck nor discipline anymore

rpiaggio: Ok, removed them too.

PR #1153, opened 2019-10-20
Always pass the epics instances as build params

cquiroz:
This is the continuation of #1152 applied to instruments.

Same set of changes:
* Remove static calls to epics systems
* Convert to a tagless final encoding
* Use Logger where appropriate

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/Systems.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/Systems.scala
@@ -68,7 +68,7 @@ object Systems {
     else                                                          TcsNorthControllerSim[IO].pure[IO]
 
   def altair(settings: Settings)(implicit L: Logger[IO]): IO[AltairController[IO]] =
-    if (settings.altairControl.command && settings.tcsControl.command) AltairControllerEpics.pure[IO]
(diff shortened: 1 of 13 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/Systems.scala
```diff
@@ -68,35 +68,35 @@ object Systems {
...
     else                               GsaoiControllerSim[IO]
 
   def gnirs(settings: Settings)(implicit T: Timer[IO], L: Logger[IO]): IO[GnirsController[IO]] =
-    if (settings.gnirsControl.command) GnirsControllerEpics().pure[IO]
+    if (settings.gnirsControl.command) GnirsControllerEpics(GnirsEpics.instance).pure[IO]
     else                               GnirsControllerSim[IO]
 
   def niri(settings: Settings)(implicit T: Timer[IO], L: Logger[IO]): IO[NiriController[IO]] =
-    if (settings.niriControl.command) NiriControllerEpics().pure[IO]
+    if (settings.niriControl.command) NiriControllerEpics((NiriEpics.instance)).pure[IO]
```
jluhrs: Redundant parentheses.

PR #1156, opened 2019-10-21
[SEQNG-1075] Wire N&S UI buttons to actual actions.

rpiaggio:
This is how a paused N&S step is currently looking.

image

I'm open to suggestions for improvement as to how to place the buttons.

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
@@ -56,14 +56,16 @@ object actions {
       extends Action
 
   // Actions related to executing sequences
-  final case class RequestRun(s:         Observation.Id) extends Action
-  final case class RequestSync(s:        Observation.Id) extends Action
-  final case class RequestPause(s:       Observation.Id) extends Action
-  final case class RequestCancelPause(s: Observation.Id) extends Action
-  final case class RequestStop(id:       Observation.Id, step: Int) extends Action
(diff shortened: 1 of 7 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/actions.scala
```diff
@@ -56,14 +56,16 @@ object actions {
...
-  final case class RequestCancelPause(s: Observation.Id) extends Action
-  final case class RequestStop(id:       Observation.Id, step: Int) extends Action
-  final case class RequestAbort(id:      Observation.Id, step: Int) extends Action
-  final case class RequestObsPause(id:   Observation.Id, step: Int) extends Action
-  final case class RequestObsResume(id:  Observation.Id, step: Int) extends Action
+  final case class RequestRun(s:               Observation.Id) extends Action
+  final case class RequestSync(s:              Observation.Id) extends Action
+  final case class RequestPause(s:             Observation.Id) extends Action
+  final case class RequestCancelPause(s:       Observation.Id) extends Action
+  final case class RequestStop(id:             Observation.Id, step: Int) extends Action
```
cquiroz: We have a StepId alias for the `Int` We may eventually replace it by some richer type

rpiaggio: Fixed

andrewwstephens:
I like that the buttons are different for the cycles ("graceful") and nods.  This will make it easy to describe in the documentation.

PR #1158, opened 2019-10-22
SEQNG-1070A: Model for configuration

cquiroz:
The configuration of the seqexec is not centralized and difficult to follow. It also uses deprecated libs and older practices
This PR is the first part of the process to improve it. I'm putting new classes to store the configuration and move them to the model. Thus the model needs to be converted to a full crossproject with some jvm only classes.
The next PR will use these classes to improve configuration


modules/seqexec/model/jvm/src/main/scala/seqexec/model/config/ControlStrategy.scala
```diff
@@ -0,0 +1,41 @@
...
+  implicit class ControlStrategyOps(v: ControlStrategy) {
+    val connect: Boolean = v match {
+      case ControlStrategy.Simulated => false
+      case _         => true
+    }
+    // If connected, then use real values for keywords
+    val realKeywords: Boolean = connect
+    val command: Boolean = v match {
+      case ControlStrategy.FullControl => true
+      case _           => false
```
jluhrs: Alignment?

tpolecat: How about `Eq` from universal and then `v === ControlStrategy.FullControl`?

modules/seqexec/model/jvm/src/main/scala/seqexec/model/config/SeqexecServerConfiguration.scala
```diff
@@ -0,0 +1,37 @@
...
+                          systemControl:           SystemsControlConfiguration,
+                          odbNotifications:        Boolean,
+                          instForceError:          Boolean,
+                          failAt:                  Int,
+                          odbQueuePollingInterval: FiniteDuration,
+                          gpiUrl:                  Uri @@ GpiSettings,
+                          ghostUrl:                Uri @@ GhostSettings,
+                          gpiGDS:                  Uri @@ GpiSettings,
+                          ghostGDS:                Uri @@ GhostSettings)
+                          // tops:                    String,
```
cquiroz: Hmm, this was updated earlier, I may have missed a commit

modules/seqexec/model/jvm/src/main/scala/seqexec/model/config/SeqexecServerConfiguration.scala
```diff
@@ -0,0 +1,37 @@
...
+import cats.implicits._
+import org.http4s.Uri
+import scala.concurrent.duration.FiniteDuration
+import shapeless.tag.@@
+
+trait GpiSettings
+trait GhostSettings
+
+final case class SeqexecServerConfiguration(
+                          odb:                     String,
```
tpolecat: Can we get some Scaladoc for these parameters? Some are obvious but many aren't. (like this one)

PR #1161, opened 2019-10-23
Prevent calls to epics subsystems before initalization

cquiroz:
While testing the new config I noted we have a bug about the initialization of epics systems
This PR should fix it

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
@@ -3,156 +3,154 @@
 
 package seqexec.server.gems
 
-import cats.ApplicativeError
 import cats.effect.Async
 import cats.implicits._
+import io.chrisdavenport.log4cats.Logger
 import monocle.macros.Lenses
 import mouse.boolean._
 import seqexec.server.gems.GemsController.GemsConfig
 import seqexec.server.gems.Gems._
-import org.log4s.getLogger
 import seqexec.server.gsaoi.GsaoiGuider
 import seqexec.server.tcs.Gaos.{PauseCondition, PauseConditionSet, PauseResume, ResumeCondition, ResumeConditionSet}
 import squants.Time
 import squants.time.TimeConversions._
 
-class GemsControllerEpics[F[_]: Async: ApplicativeError[?[_], Throwable]](epicsSys: GemsEpics[F],
-                                                                          gsaoiGuider: GsaoiGuider[F]
-                                                                         ) extends GemsController[F] {
-  import GemsControllerEpics._
-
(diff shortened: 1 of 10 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
```diff
@@ -3,156 +3,154 @@
...
+        )
+      ).flattenOption
 
-    (
-      p.map(_._1).foldLeft(current){case (a, b) => b(a)},
       p.nonEmpty.option{
-        p.traverse(_._2) *>
-        epicsSys.ApdControl.setTimeout(CmdTimeout) *>
-        epicsSys.ApdControl.post.void
+        p.sequence *>
```
sraaphorst: What is the difference between `p.sequence` and `p.traverse(_._2)`?d

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
```diff
@@ -3,156 +3,154 @@
...
-                                                                          gsaoiGuider: GsaoiGuider[F]
-                                                                         ) extends GemsController[F] {
-  import GemsControllerEpics._
+object GemsControllerEpics {
 
-  override def pauseResume(pauseReasons: PauseConditionSet, resumeReasons: ResumeConditionSet)
-                          (cfg: GemsConfig)
-  : F[PauseResume[F]] = retrieveConfig.map{ c =>
+  private class GemsControllerEpics[F[_]: Async: Logger](
+    epicsSys: => GemsEpics[F],
```
cquiroz: This is one kind of change, pass call by name params

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
```diff
@@ -3,156 +3,154 @@
...
+      val r2 = List(startCwfs(s1, cfg), resume(resumeReasons)).flattenOption
+
+      PauseResume(
+        r1.nonEmpty.option(r1.sequence.void),
+        r2.nonEmpty.option(r2.sequence.void)
+      )
+    }
+
+    import GsaoiGuider.OdgwId._
+    override lazy val stateGetter: GemsWfsState[F] = GemsWfsState(
```
cquiroz: The other kind of change is to make vals that call `epicsSys` lazy

modules/seqexec/server/src/main/scala/seqexec/server/gems/GemsControllerEpics.scala
cquiroz: This wasn't really changed, It looks like a change because indentation changed making for a large diff

cquiroz:
Made obsolete by #1165

PR #1165, opened 2019-10-26
SEQNG-1102 Collected all references to EPICS classes in one place.

jluhrs:
The EPICS initialization code is one of the first pieces put together when we were younger and naiver. This PR makes a big overhaul of it. Now references to EPICS objects are in one place, and initialization was made automatic.
I had to "untaglessy" some functions in `SeqTranslate` to make it compile, but the final goal is to make `SeqTranslate` tagless, and this will make it easier.


modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -365,75 +360,63 @@ class SeqTranslate(site: Site, systems: Systems[IO], settings: TranslateSettings
...
-        GsaoiHeader.header[IO](sys, tcsKReader, gsaoiReader).asRight
+        GsaoiHeader.header[IO](sys, systems.tcsKeywordReader, systems.gsaoiKeywordReader).asRight
       case _                 =>
         TrySeq.fail(Unexpected(s"Instrument ${sys.resource} not supported."))
     }
   }
 
-  private def commonHeaders[F[_]: Sync](epics: => TcsEpics[F], config: CleanConfig, tcsSubsystems: List[TcsController.Subsystem],
-                            inst: InstrumentSystem[F])(ctx: HeaderExtraData): Header[F] =
+  private def commonHeaders(config: CleanConfig, tcsSubsystems: List[TcsController.Subsystem],
```
cquiroz: Why are we going back to `IO`?

modules/seqexec/server/src/main/scala/seqexec/server/Systems.scala
```diff
@@ -43,135 +47,286 @@ final case class Systems[F[_]](
   nifs:       NifsController[F],
   altair:     AltairController[F],
   gems:       GemsController[F],
-  guideDb:    GuideConfigDb[F]
+  guideDb:    GuideConfigDb[F],
```
cquiroz: Mixed alignment?

modules/seqexec/server/src/main/scala/seqexec/server/Systems.scala
```diff
@@ -43,135 +47,286 @@ final case class Systems[F[_]](
...
+    def dhs[F[_] : Effect : Timer : Logger](httpClient: Client[F]): F[DhsClient[F]] =
+      if (settings.systemControl.dhs.command)
+        DhsClientHttp[F](httpClient, settings.dhsServer).pure[F]
+      else
+        DhsClientSim.apply[F]
 
-    (ghostClient, ghostGDS).mapN(GhostController(_, _))
-  }
+    // TODO make instruments controllers generalized on F
+    def gcal(implicit L: Logger[IO]): IO[(GcalController[IO], GcalKeywordReader[IO])] =
```
cquiroz: Given that `instance` is parametrized I'd expect we could make these parametric too. Why don't we?

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairEpics.scala
```diff
@@ -176,9 +175,9 @@ class AltairEpics[F[_]: Async](service: CaService, tops: Map[String, String]) {
 object AltairEpics extends EpicsSystem[AltairEpics[IO]] {
 
   override val className: String = getClass.getName
-  override val Log: Logger = getLogger
   override val CA_CONFIG_FILE: String = "/Altair.xml"
 
-  override def build(service: CaService, tops: Map[String, String]) = new AltairEpics(service, tops)
+  override def build[F[_]: Sync](service: CaService, tops: Map[String, String]):F[AltairEpics[IO]] =
```
cquiroz: Same here, why `AltairEpics[IO]` rather than `AltairEpics[F]`

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: I was forced because the keyword readers now come in `systems`, which is `System[IO]`. But don't worry, we'll get back the `F` in my next PR.

cquiroz: 👍

modules/seqexec/server/src/main/scala/seqexec/server/Systems.scala
jluhrs: The 'F' in `instance` instance is the effect to produce the EPICS class. `EpicsSystem[T].instance[F]` produces a `F[T]`, but in all cases `T` is something like `TcsEpics[IO]`, `GmosEpics[IO]`, etc. We are not yet at the point where we could get rid of that IO, because the different `XXXEpics[IO]` are implemented with objects.

modules/seqexec/server/src/main/scala/seqexec/server/altair/AltairEpics.scala
jluhrs: This is related to the previous question. `AltairEpics[IO]` is the type parameter of the trait implemented by the object where `build` is defined.

jluhrs:
@cquiroz I tested that it started the EPICS connections before creating the PR.

PR #1167, opened 2019-10-28
SEQNG-1106 Progress event for N&S

cquiroz:
This PR adds a new type of progress event specifically for N&S. It is very similar to the current progress event but it includes the current sub exposure
This has been tested in simulation but I'm a bit uncertain about in the instrument. There is some info about the current position of N&S in Gmos epics db but it would be good to verify it


modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
```diff
@@ -8,14 +8,44 @@ import cats.implicits._
...
+  implicit val equalObservationProgress: Eq[ObservationProgress] =
     Eq.by(x => (x.obsId, x.stepId, x.total, x.remaining))
 
 }
+
+final case class NSObservationProgress(obsId:   Observation.Id,
+                                     stepId:    StepId,
+                                     total:     Time,
+                                     remaining: Time,
+                                     sub:       NSSubexposure) extends Progress
```
jluhrs: I would add the total number of subexposures or the total number of cycles to make the information self contained.

cquiroz: The total amount of cycles is in `NSSubexposure` and the amount of subexposures can be calculated out of it

PR #1169, opened 2019-10-29
Tagless auth

cquiroz:
This PR revamps the auth code with the following improvements:
* Update to the latest jwt library
* Replace argonaut for circe
* Use tagless style

The code could have used a bit more improvements (Some bits were use to learn about free monads) but given time constraints I'll leave it as is


modules/seqexec/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -137,7 +137,7 @@ object WebServerLauncher extends IOApp with LogInitialization {
              /_/
 """
     val msg = s"""Start web server for site ${conf.site} on ${conf.mode} mode, version ${OcsBuildInfo.version}"""
-    L.info(banner + msg)
+    Logger[F].info(banner + msg)
```
jluhrs: Why this change? Doesn't `L` still exist in this context?

cquiroz: I thought about that, L is `Logger[IO]` and I wanted `Logger[F]`

PR #1173, opened 2019-10-30
[SEQNG-1107] Use specialized N&S progress messages.

rpiaggio:


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -135,38 +136,39 @@ sealed trait NodAndShuffleProgress {
...
           val nodMillis = nsStatus.nodExposureTime.toMilliseconds.toInt
-          val (sectionTotal, currentValue) = quantitiesFromNodMillis(nsStatus, nodMillis)
+          val (sectionTotal, currentValue) = quantitiesFromNodMillis(cycleIndex, nodIndex, nodMillis)
 
           SmoothDividedProgressBar(
             sections = sections(nsStatus),
             sectionTotal = sectionTotal,
-            value = currentValue.map(_ + elapsedMillis).getOrElse(0),
-            maxValue = currentValue.map(_ + nodMillis).getOrElse(0),
+            value = nsStatus.state.map(_ => currentValue + elapsedMillis).orEmpty, // Only advance smooth bar if actually started
```
cquiroz: `foldMap`

modules/seqexec/web/client/src/main/scala/seqexec/web/client/circuit/package.scala
```diff
@@ -44,6 +46,9 @@ package circuit {
     def zoomL[A: Eq](lens: Lens[M, A]): ModelR[M, A] =
       circuit.zoom[A](lens.get)(fastEq[A])
 
+    def zoomO[A: Eq](lens: Optional[M, A]): ModelR[M, Option[A]] =
```
cquiroz: Cool

modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
```diff
@@ -23,6 +25,12 @@ object Progress {
       case (a: NSObservationProgress, b: NSObservationProgress) => a === b
       case _                                                    => false
     }
+
+  implicit val obsProgressP: Prism[Progress, ObservationProgress] =
```
cquiroz: This seems a bit strange, having an implicit `Prism` but it seems to work fine

rpiaggio: Yup, it's used at https://github.com/gemini-hlsw/ocs3/blob/0f625eb4d91c002956171ae280eb8761a2ff1086/modules/seqexec/web/client/src/main/scala/seqexec/web/client/circuit/SeqexecCircuit.scala#L142 so that we can request a subclass to the lens through type parameters.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: Right, keep forgetting about that! Done.

rpiaggio:
For some reason, these latest changes seem to be making the UI drag a lot in FF during N&S runs. Let's not merge yet until we have a clear picture of what's going on.

PR #1181, opened 2019-11-04
Improve initalization of the ca service

cquiroz:
I was a bit surprised to see the CaService initalized even in full simulation, then I noted it was ok to do so but I want to add a log item just to make it clearer
This PR adds that and moves the init part out of SeqexecEngine

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/CaServiceInit.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/CaServiceInit.scala
@@ -14,15 +14,13 @@ object CaServiceInit {
   // the configuration file or from the environment
   def caInit[F[_]](conf: SeqexecEngineConfiguration)(implicit F: Sync[F], L: Logger[F]): F[CaService] =
     L.info("Init EPICS but all subsystems in simulation").unlessA(conf.systemControl.connectEpics) *>
-    (conf.epicsCaAddrList.map(a => F.delay(CaService.setAddressList(a))).getOrElse {
+    conf.epicsCaAddrList.map(a => F.delay(CaService.setAddressList(a))).getOrElse {
       F.delay(Option(System.getenv("EPICS_CA_ADDR_LIST"))).flatMap {
         case Some(_) => F.unit
         case _       => F.raiseError[Unit](new RuntimeException("Cannot initialize EPICS subsystem"))
       }
     } *>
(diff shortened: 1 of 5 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/CaServiceInit.scala
```diff
@@ -0,0 +1,28 @@
...
+import seqexec.model.config.SeqexecEngineConfiguration
+
+object CaServiceInit {
+  // Ensure there is a valid way to init CaService either from
+  // the configuration file or from the environment
+  def caInit[F[_]](conf: SeqexecEngineConfiguration)(implicit F: Sync[F], L: Logger[F]): F[CaService] =
+    L.info("Init EPICS but all subsystems in simulation").unlessA(conf.systemControl.connectEpics) *>
+    (conf.epicsCaAddrList.map(a => F.delay(CaService.setAddressList(a))).getOrElse {
+      F.delay(Option(System.getenv("EPICS_CA_ADDR_LIST"))).flatMap {
+        case Some(_) => F.unit
```
cquiroz: @jluhrs I think the original intent was to init CA with `EPICS_CA_ADDR_LIST` if available but we don't do that here, Should we remove this piece of code?

modules/seqexec/server/src/main/scala/seqexec/server/CaServiceInit.scala
```diff
@@ -0,0 +1,28 @@
...
+        case _       => F.raiseError[Unit](new RuntimeException("Cannot initialize EPICS subsystem"))
+      }
+    } *>
+      F.delay(CaService.setIOTimeout(java.time.Duration.ofMillis(conf.ioTimeout.toMillis))) *>
+      F.delay(Option(CaService.getInstance())).flatMap {
```
cquiroz: Hmm now that I look at this, it can be better expresed with `ensure`

modules/seqexec/server/src/main/scala/seqexec/server/CaServiceInit.scala
jluhrs: Yes, that was the original intent. Can't you just fix it?

cquiroz: Ok, I will fix it

PR #1182, opened 2019-11-04
[SEQNG-1114] and [SEQNG-1115] Fix Popup tooltip disappearing.

rpiaggio:
Fixing [SEQNG-1114] and [SEQNG-1115] was fairly trivial, but I took the chance to refactor `Popup` and `Button` like we did with other components before, so that they become simpler to declare at call sites.

The tricky part was that both these components accept children, but a mechanism was established for components that accept children from now on.

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/semanticui/elements/popup/Popup.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/semanticui/elements/popup/Popup.scala
@@ -3,17 +3,34 @@
 
 package seqexec.web.client.semanticui.elements.popup
 
-import japgolly.scalajs.react.component.Scala.Unmounted
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.vdom.html_<^._
 import japgolly.scalajs.react.raw.React
+import web.client.ReactPropsWithChildren
 
 /**
   * Produces a popup using javascript
   * This is a barebones component waiting for the proper react component to be made available
   */
+
+final case class Popup(selector: String, content: String) extends ReactPropsWithChildren {
+  @inline def render: Seq[CtorType.ChildArg] => VdomElement = Popup.component(this)
+}
+
 object Popup {
-  final case class Props(selector: String, content: String)
+  type Props = Popup
+
+  private def mountPopup(component: ComponentDom.Mounted, props: Props): Callback = Callback {
+    // Enable menu on Semantic UI
+    import org.querki.jquery.$
+    import web.client.facades.semanticui.SemanticUIPopup._
+
+    component.toElement.foreach { dom =>
+      $(dom).popup(
+        JsPopupOptions.content(props.content)
+        )
+    }
+  }
 
   private val component = ScalaComponent.builder[Props]("Popup")
     .stateless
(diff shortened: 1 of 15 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/semanticui/elements/popup/Popup.scala
```diff
@@ -3,37 +3,45 @@
...
+    }
+  }
 
   private val component = ScalaComponent.builder[Props]("Popup")
     .stateless
-    .renderPC{($, _, _) =>
-      // This is in principle unsafe but we are only allowing Elements on the constructor
-      VdomElement($.propsChildren.only().asInstanceOf[React.Element])
+    .render_C{ children =>
+      if( children.count === 1)
```
cquiroz: Is this the core fix?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/semanticui/elements/popup/Popup.scala
```diff
@@ -3,37 +3,45 @@
...
-        ctx.getDOMNode.toElement.foreach { dom =>
-          $(dom).popup(
-            JsPopupOptions.content(ctx.props.content)
-          )
-        }
-      }
-    ).build
-
-  def apply(p: Props, children: VdomElement): Unmounted[Props, Unit, Unit] = component(p)(children)
+    .componentDidMount(ctx => mountPopup(ctx.getDOMNode, ctx.props))
```
cquiroz: I supose this could be `componentDidMount(mountPopup)` and pass the whole ctx to `mountPopup`. Or is it different for `DidMount` and `DidUpdate`?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/semanticui/elements/popup/Popup.scala
rpiaggio: No, the core fix is adding a `componentDidUpdate` and mount the popup there too, besides doing it in the `componentDidMount`.

This other fix is necessary because of changing the component structure so that its props extend `ReactPropsWithChildren`, which accepts any number of children. Before we were guaranteeing only one child was passed in the `apply` method.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/semanticui/elements/popup/Popup.scala
rpiaggio: I thought of that but `componentDidMount` passes a `ComponentDidMount[P, S, B]` and `componentDidUpdate` passes a `ComponentDidUpdate[P, S, B, SS]` (`SS` seems to be a ShapShot).

They don't have any common ancestor that gives access to `getDomNode` and `props`. I guess we could define a type class but it seems overkill.

cquiroz: 👍

modules/seqexec/web/client/src/main/scala/seqexec/web/client/semanticui/elements/popup/Popup.scala
cquiroz: Ah I see, that's great

PR #1185, opened 2019-11-05
Remove the concept of graceful abort

cquiroz:
It seems to me the concept of graceful abort is not used and could be removed. @jluhrs Why did we have it in the first place?


jluhrs:
Excess of enthusiasm.

PR #1190, opened 2019-11-06
SEQNG-1124: Avoid duplicated abort notifications to the odb

cquiroz:
Some testing shows that on abort we get two abort events on the odb:
image

This is because the abort sends one, and we have a generic one for errors.
A future change will make abort act without using an error but in the meanwhile this PR fixes it

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
@@ -27,17 +27,9 @@ trait ObserveActions {
     * Actions to perform when an observe is aborted
     */
   def abortTail[F[_]: MonadError[?[_], Throwable]](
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -27,17 +27,9 @@ trait ObserveActions {
     * Actions to perform when an observe is aborted
     */
   def abortTail[F[_]: MonadError[?[_], Throwable]](
-    systems:     Systems[F],
-    obsId:       Observation.Id,
-    imageFileId: ImageFileId
+    obsId:       Observation.Id
   ): F[Result[F]] =
-    systems.odb
```
cquiroz: Oddly I'll restore this once `abort` doesn't use an error for signaling

PR #1191, opened 2019-11-07
SEQNG-1038 Custom tolerance function for polarizer angle

cquiroz:
Turns out the original comparison function to decide if we should move GPI's polarizer angle was too naive not taking into account wraparounds on 360
This PR improves it calculating the shortest angle difference and fixes an additional bug on comparisons
This code was tested on instrument and surfaced an extra bug on GPI TLC code: REL-3752

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GpiStatusApply.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gpi/GpiStatusApply.scala
@@ -11,11 +11,13 @@ import gem.enum.GiapiStatus
 import gem.enum.GiapiType
 import gem.enum.Instrument
 import gem.ocs2.Parsers
+import gsp.math.Angle
 import gsp.math.syntax.all._
 import giapi.client.commands.Configuration
 import giapi.client.GiapiStatusDb
 import giapi.client.StatusValue
 import giapi.client.syntax.status._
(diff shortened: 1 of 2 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gpi/GpiStatusApply.scala
```diff
@@ -150,4 +152,33 @@ object GpiStatusApply extends GpiLookupTables {
...
+        (measuredAngle, requestedAngle).mapN {(m, r) =>
+          // This calculates the minimal angle diference so that e.g. 359.9 and 0.1 have 0.2 diff
+          // This should probably be in the Angle class
+          val π = scala.math.Pi
+          val δ: Double = (m - r).toSignedDoubleRadians % 2*π
+          // "normalized" difference
+          val δʹ= abs(if (δ < - π) δ + 2*π else if (δ >= π) δ - 2*π else δ)
+          val ε: Option[Double] = GpiPolarizerAngle.tolerance.map(t => Angle.fromDoubleDegrees(t.toDouble).toSignedDoubleRadians)
+          ε.exists(abs(δʹ) <= _)
+        }.getOrElse(false)
```
jluhrs: Redundant call to `abs`. You already applied it in line 177.

PR #1192, opened 2019-11-10
[SEQENG-1125] & [SEQNG-1116] (partial)

rpiaggio:
[SEQNG-1125]: Disable GMOS N&S "Stop at end of cycle" button while paused

Regarding [SEQNG-1116]: Indicate detector is reading out after Stop
*  I'm changing the message to "Stopping - Reading Out" whenever a stop is requested in any kind of observation. I just need confirmation that a "Reading Out" always takes place whenever a Stop is requested in any kind of step.

* Also, @jluhrs @cquiroz  we seem to be missing an event that should happen whenever a graceful stop starts. Without that, we have no way to display the "Reading Out" message when the end of the cycle is reached. (At least without knowing that a graceful stop has been requested).


cquiroz:
I think @jluhrs is planning to add events for Graceful stop

jluhrs:
Yes, I am working on SEQNG-1104

rpiaggio:
> Yes, I am working on SEQNG-1104

Thank you! It seems that, besides knowing when graceful stop and pause are requested, we also need to know when a graceful stop happens.

PR #1193, opened 2019-11-12
Upgrade fs2

cquiroz:
This PR upgrades fs2 and http4s. It adds `Blocker` which I don't understand that well but I typed random stuff until everything compiled
I'll leave this for review but I want to test it locally before merging


tpolecat:
>  It adds `Blocker` which I don't understand that well

The assumption is that computations are non-blocking, so Cats provides a CPU-bound work-stealing pool (managed by `ContextShift`) for normal operations. It typically has one thread per core for maximum performance, and these threads are never blocked.

When you need to do a blocking operation (like Java file I/O) you use a `Blocker`, which lets you move the computation to another pool where the threads are expected to be blocked all the time. When the blocking result is ready the computation resumes on the main CPU-bound pool. This pool is usually large (or unbounded) because the threads here aren't doing any real work.

The end result is that your compute threads can always make progress and are never blocked by I/O, so you never have an idle core if there's work to be done.

modules/seqexec/server/src/main/scala/seqexec/server/Systems.scala
```diff
@@ -67,7 +67,7 @@ object Systems {
       if (settings.odbNotifications) OdbProxy.OdbCommandsImpl[F](new Peer(settings.odb.renderString, 8442, null))
       else new OdbProxy.DummyOdbCommands[F])
 
-    def dhs[F[_] : Effect : Timer : Logger](httpClient: Client[F]): F[DhsClient[F]] =
+    def dhs[F[_]: Concurrent : Timer : Logger](httpClient: Client[F]): F[DhsClient[F]] =
```
tpolecat: \o/

cquiroz:
I have tested this for a while and it seems fine. @jluhrs Should we merge it?

PR #1194, opened 2019-11-12
SEQNG-1104 Send N&S pause and stop command state to web client.

jluhrs:
The step information send to the client now indicates if there is a stop gracefully or pause gracefully command in progress.

### DIFF
--- a/modules/seqexec/model/shared/src/main/scala/seqexec/model/Step.scala
+++ b/modules/seqexec/model/shared/src/main/scala/seqexec/model/Step.scala
@@ -5,6 +5,7 @@ package seqexec.model
 
 import cats._
 import cats.implicits._
+import gem.util.Enumerated
 import seqexec.model.enum._
 import seqexec.model.dhs._
 import monocle.Prism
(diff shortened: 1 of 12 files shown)


modules/seqexec/model/shared/src/main/scala/seqexec/model/Step.scala
```diff
@@ -212,7 +211,8 @@ final case class NodAndShuffleStep(
   override val skip:       Boolean,
   override val fileId:     Option[ImageFileId],
   configStatus:            List[(Resource, ActionStatus)],
-  nsStatus:                NodAndShuffleStatus
+  nsStatus:                NodAndShuffleStatus,
+  pendingObserveCmd:       Option[NodAndShuffleStep.PendingObserveCmd]
```
jluhrs: Posible pending N&S observation command.

modules/seqexec/server/src/main/scala/seqexec/server/StepsView.scala
```diff
@@ -101,7 +103,7 @@ object StepsView {
 
   private def observeAction[F[_]](executions: List[ParallelActions[F]]): Option[Action[F]] =
     // FIXME This is too naive and doesn't work properly for N&S
-    executions.flatMap(_.toList).filter(_.kind === ActionType.Observe).headOption
+    executions.flatMap(_.toList).find(_.kind === ActionType.Observe)
```
cquiroz: Why are we reverting this? If we do so, let's remove the `FIXME`

jluhrs: This change does not address the `FIXME`. It is just a terser version of the same code.

cquiroz: Your right

PR #1196, opened 2019-11-13
[SEQNG-1105] Show in UI that graceful stop or pause was requested.

rpiaggio:


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsControlButtons.scala
```diff
@@ -143,11 +160,10 @@ object ControlButtons {
            case StopGracefullyObservation =>
              Popup("button", "Stop the current exposure at the end of the cycle")(
                Button(
-                 icon     = Some(IconStop),
                  color    = Some("orange"),
                  onClick  = requestGracefulStop(p.id, p.stepId),
-                 disabled = p.requestInFlight || p.isObservePaused
-               )
+                 disabled = p.requestInFlight || p.isObservePaused || p.nsPendingObserveCmd.isDefined
+               )(stopGracefullyIcon),
```
cquiroz: I bet scalastyle complains about the trailing comma

cquiroz:
Can you add a screenshot?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsControlButtons.scala
rpiaggio: Yup. I changed the logic now anyway so that we can have icon groups in a nice way.

rpiaggio:
> Can you add a screenshot?

Yup, was planning to since I think I need feedback. This is the best my creativity could come up with today, but I'm really open to other suggestions.

**PAUSE**
nsgracefulpause

**STOP**
nsgracefulstop

jluhrs:
The animated screenshot shows that several seconds pass between the time the button is pressed, and when the button gives a feedback. Is that how it look in the real thing?

rpiaggio:
> The animated screenshot shows that several seconds pass between the time the button is pressed, and when the button gives a feedback. Is that how it looks in the real thing?

At least in simulation. I **think** this has to do with the fact that we don't process regular `ServerMessages` all the way into the UI as often as we do Progress messages. But maybe I'm wrong, I'd appreciate your feedback on this @cquiroz.

Other things I noticed and I don't know if they are expected or not, or if they are simulated, or a consequence of simulation:

1. Regular observations take a few (3-4 seconds) to process a regular observation pause:
normalpause

2. N&S immediate pause also takes a long time process (seems similar to graceful pause). It seems longer than with a regular observation:

nspause

3. On graceful pause, the UI seems to get progress event from the next cycle before it gets the `ExposurePaused` event:

image

This results in the UI looking like this when the graceful pause is reached:

image

4. I've actually also seen the opposite happen: a graceful pause is requested, the exposure reaches the end of the cycle and pauses, but the last progress event doesn't get to the UI, leaving the bar in "almost complete" state. I cannot reproduce this now. This happened when my (particular) development environment was thrashing and never happened after I rebooted. But it still worries me this may happen if the server is under stress.

PR #1198, opened 2019-11-14
SEQNG-1126 Added exposure stage status to progress event.

jluhrs:
The current observation stage (preparing, acquiring, reading out) is now read from the instrument and sent to the web client.

### DIFF
--- a/modules/seqexec/server/src/main/resources/Gmos.xml
+++ b/modules/seqexec/server/src/main/resources/Gmos.xml
@@ -469,10 +469,20 @@
             <type>STRING</type>
             <description>GMOS detector controller s/w</description>
         </attribute>
+        <attribute name="obsPrep">
+            <channel>sad:dc:prep.VAL</channel>
+            <type>INTEGER</type>
+            <description>Observation is preparing</description>
+        </attribute>
         <attribute name="obsAcq">
             <channel>sad:dc:acq.VAL</channel>
             <type>INTEGER</type>
-            <description>Observation is aquiring</description>
+            <description>Observation is acquiring</description>
+        </attribute>
+        <attribute name="readingOut">
+            <channel>sad:dc:rdout.VAL</channel>
+            <type>INTEGER</type>
+            <description>Observation is reading out</description>
         </attribute>
     </Status>
     <Status name="gmos::status">
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ProgressUtil.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ProgressUtil.scala
@@ -33,7 +33,7 @@ object ProgressUtil {
       .collect{ case Some(p) => p }
 
   def countdown[F[_]: Applicative: Timer](total: Time, elapsed: Time): Stream[F, Progress] =
-    Stream.emit(ObsProgress(total, RemainingTime(total), ObserveStage.Preparing)).covary[F] ++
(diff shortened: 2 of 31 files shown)


modules/seqexec/server/src/main/resources/Gmos.xml
```diff
@@ -469,10 +469,20 @@
...
+        <attribute name="obsPrep">
+            <channel>sad:dc:prep.VAL</channel>
+            <type>INTEGER</type>
+            <description>Observation is preparing</description>
+        </attribute>
         <attribute name="obsAcq">
             <channel>sad:dc:acq.VAL</channel>
             <type>INTEGER</type>
-            <description>Observation is aquiring</description>
+            <description>Observation is acquiring</description>
```
sraaphorst: Why is it "Acquiring observation" for F2 but "Observation is acquiring" for Gmos? Should these be consistent?

sraaphorst: (Same with others: Reading out observation vs. Observation is reading out and Preparing observation vs. Observation is preparing?)

jluhrs: That's boring :)

modules/seqexec/server/src/main/scala/seqexec/server/ProgressUtil.scala
```diff
@@ -30,12 +33,25 @@ object ProgressUtil {
       .collect{ case Some(p) => p }
 
   def countdown[F[_]: Applicative: Timer](total: Time, elapsed: Time): Stream[F, Progress] =
+    Stream.emit(ObsProgress(total, RemainingTime(total), ObserveStage.Preparing)).covary[F] ++
```
cquiroz: I think you can drop `covary` here

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
```diff
@@ -267,6 +267,7 @@ trait ModelBooPicklers extends GemModelBooPicklers {
   implicit val userNotificationPickler    = generatePickler[UserNotification]
   implicit val guideConfigPickler         = generatePickler[GuideConfigUpdate]
   implicit val queueUpdatedPickler        = generatePickler[QueueUpdated]
+  implicit val observationStagePickler    = enumeratedPickler[ObserveStage]
```
cquiroz: Did you add a test for this?

cquiroz:
What events will be sent to giapi based instruments?

modules/seqexec/web/shared/src/main/scala/seqexec/web/model/boopickle/ModelBooPicklers.scala
jluhrs: Yes.

modules/seqexec/server/src/main/scala/seqexec/server/ProgressUtil.scala
jluhrs: Done

jluhrs:
@cquiroz There are no events "sent" to the instrument here.
GPI uses a simulated countdown. That countdown now goes through the stages Preparing-Acquiring-ReadingOut, so the client will not see a difference.

PR #1201, opened 2019-11-17
[SEQNG-1123] Remember user-edited Observer name

rpiaggio:
The user-edited was being remembered but it seems the current interfaced proved a bit confusing.

This is how it is looking now:

defaultObserver

As always, I accept suggestions regarding UI in general: fonts, colors, behavior, etc.

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/ControlMenu.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/ControlMenu.scala
@@ -3,8 +3,7 @@
 
 package seqexec.web.client.components
 
-import seqexec.web.client.actions.Logout
-import seqexec.web.client.actions.OpenLoginBox
+import seqexec.web.client.actions.{Logout, OpenLoginBox, UpdateDefaultObserver}
 import seqexec.web.client.model.ClientStatus
 import seqexec.web.client.circuit.SeqexecCircuit
 import seqexec.web.client.semanticui.Size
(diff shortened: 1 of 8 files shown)


cquiroz:
Can you elaborate on the relation between the `Observer` field at the top and the one on the status bar. They don't seem to be synchronized. Do we need both?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/ControlMenu.scala
```diff
@@ -44,14 +47,31 @@ object ControlMenu {
            disabled = !enabled,
            inverted = true)(text)
 
+
+  // Ideally we'd do this with css text-overflow but it is not
+  // working properly inside a header item, let's abbreviate in code
+  private def overflowText(text: String): String =
```
cquiroz: We maybe using this in several places, Should we make it common?

cquiroz:
What happens in small screens? Can you still edit the observer name

rpiaggio:
> Can you elaborate on the relation between the `Observer` field at the top and the one on the status bar. They don't seem to be synchronized. Do we need both?

The one in the status bar at the bottom is **exactly** the same as the one that appears on `Observer - Default` at the top. If the top is displaying `Observer - Default` they do synchronize (and the one at the top is disabled). This is the observer that will be used into sequences that are loaded from now on from this terminal.

However, if the top is displaying ` Observer - <INSTRUMENT>` that means we are seeing data from a loaded (and selected) sequence. Then that field is enabled and that's the one we use to change the observer *only for that sequence*.

It seemed to me that with the previous behavior, users were expecting to change the observer for a loaded sequence, and that that observer would be used for sequences loaded from now on. (Which would be a valid behavior too, and we could implement that, but it's not what was happening, and I'm not sure that's what's required).

PR #1202, opened 2019-11-18
SEQNG-908 Handle abort as a regular action

cquiroz:
At the moment an Abort is handled as an error. This feels incorrect as we are using error as a regular control flow mechanism and additionally it makes it hard to tell on the ui side if it is a real error or an abort
I'm sending this for review but I'm missing something to set the state of the step as `Aborted`. Could @jluhrs take a look?


modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -240,6 +241,10 @@ class Engine[F[_]: MonadError[?[_], Throwable]: Logger, S, U](stateL: Engine.Sta
       _.current.execution.forall(Action.completed).option(Handle.fromStream[F, S, EventType](Stream(executed(id))))
     ).getOrElse(unit))
 
+  private def abort[R <: RetVal](id: Observation.Id, i: Int, r: Result.OKAborted[R]): HandleType[Unit] =
+    modifyS(id)(_.mark(i)(r)) *>
```
cquiroz: It seems to me that here we should set the `StepState` to be `Aborted` but I don't know how

modules/seqexec/engine/src/main/scala/seqexec/engine/Step.scala
```diff
@@ -57,7 +57,8 @@ object Step {
         // Return error or continue with the rest of the checks
       }}.map[StepState](StepState.Failed).getOrElse(
         // All actions in this Step were completed successfully, or the Step is empty.
-        if (step.executions.flatMap(_.toList).forall(Action.completed)) StepState.Completed
+        if (step.executions.flatMap(_.toList).exists(Action.aborted)) StepState.Aborted
```
cquiroz: This should work if a step is `Aborted` but I can't set that on the step

modules/seqexec/engine/src/main/scala/seqexec/engine/Execution.scala
```diff
@@ -85,6 +85,7 @@ object Execution {
     s => r match {
       case Result.OK(x)         => s.copy(runState = ActionState.Completed(x))
       case Result.OKStopped(x)  => s.copy(runState = ActionState.Completed(x))
+      case Result.OKAborted(x)  => s.copy(runState = ActionState.Completed(x))
```
jluhrs: Here is your problem (I think). It should go to `ActionState.Aborted`. Then the next time you calculate the step state, you will get an aborted state.

cquiroz: Hmm thanks, that seems to be the problem

jluhrs:
@carlos You should add a unit test to verify that the step state is `Aborted` if an action ends with an `Aborted` value.

cquiroz:
A screenshot
image

cquiroz:
Ok, this is now working. It required a bit of working to get the UI working right as it can be seen on the screenshot

There is one missing "feature". If you load a new sequence or do a sync we loose the "Aborted" state an the step looks pending. It is in principle possible to read the state from the ODB but it involves checking the events so I'll mark it as a nice to have

cquiroz:
Another note. I was using an old text to speech engine which was kind of cumbersome and I tested Amazon Polly. It was a lot easier so I re-recorded all sounds to make them consistent

PR #1206, opened 2019-11-20
Add systemd launcher scripts

cquiroz:
This PR adds scripts to launch the seqexec via systemd. Note that we install the scripts manually rather than using e.g rpms.
The scripts are designed to keep the current location of the seqexec so they should work with new deployments.
I'm not an expert on this an apparently `systemd` would allow much finer grained control but this suffice for our needs
This seems a bit of an arcane art, I expect we'll be iterating about it for a while

### DIFF
--- a/app/seqexec-server/src/main/resources/seqexec-server.env
+++ b/app/seqexec-server/src/main/resources/seqexec-server.env
@@ -1,0 +1,38 @@
+# #####################################
+# ##### Environment Configuration #####
+# #####################################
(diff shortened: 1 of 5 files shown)


cquiroz:
A question is where do these script come from. They are defined on `sbt-native-packager`. I build a debian package and read them from there.
The idea to use the same is that eventually we'd use `sbt-native-packager` and this way it would be consistent

PR #1207, opened 2019-11-20
[SEQNG-1133] N&S configuring too tall for logged out users.

rpiaggio:


modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/StepItems.scala
```diff
@@ -234,8 +237,8 @@ object StepItems {
     val isBias: Boolean =
       step.stepType(instrument).exists(_ === StepType.Bias)
 
-    def detailRows(selected: Option[StepId]): DetailRows =
-      if( (isNS && selected.exists(_ === step.id)) || isNSRunning || isNSInError)
+    def detailRows(selected: Option[StepId], hasControls: Boolean): DetailRows =
+      if( hasControls && ((isNS || isNSInError) && selected.exists(_ === step.id)) || isNSObserving)
```
cquiroz: Maybe it would be clearer to put some of the conditions in a separate variable

rpiaggio: Any suggestions on how to call `isNS || isNSInError` ?

PR #1209, opened 2019-11-22
[SEQNG-1123] Remember user-edited Observer name

rpiaggio:
We go back to the previous way observers were edited, but now we change the default observer for the session whenever the observer is changed for any sequence, thus remembering the change for future sequences previewed or loaded from that terminal.


modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/SequenceExecutionHandler.scala
```diff
@@ -26,7 +26,8 @@ class SequenceExecutionHandler[M](
   def handleUpdateObserver: PartialFunction[Any, ActionResult[M]] = {
     case UpdateObserver(sequenceId, name) =>
       val updateObserverE = Effect(
-        SeqexecWebClient.setObserver(sequenceId, name.value).as(NoAction))
+        SeqexecWebClient.setObserver(sequenceId, name.value).as(NoAction)
+        ) + Effect.action(UpdateDefaultObserver(name))
```
rpiaggio: This is the actual new change. All the rest are reverts.

PR #1210, opened 2019-11-22
[SEQNG-1135] Reading out N&S progress bars misbehave.

rpiaggio:
This is an attempt to interpret correctly the data sent from the instrument channels through the progress events.

This is a fix based on our observations with @cquiroz of the instrument logs, which for the time being differ from our simulation.

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/ObservationsProgressStateHandler.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/ObservationsProgressStateHandler.scala
@@ -27,6 +27,19 @@ class ObservationsProgressStateHandler[M](
     extends ActionHandler(modelRW)
     with Handlers[M, AllObservationsProgressState] {
 
+  // In N&S, whenever a nod finishes (and the observation continues), the instruments show:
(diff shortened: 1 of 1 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/ObservationsProgressStateHandler.scala
```diff
@@ -24,12 +27,44 @@ class ObservationsProgressStateHandler[M](
...
+  // 4) Remaining time resets.
+  // 5) Preparing channel is set to 1.
+  // 6) Preparing channel is set to 0.
+  // 7) Acquiring channel is set to 1.
+  // 8) Remaning time continues.
+  //
+  // Therefore, we should not adjust the current stage index while Acquiring is off (we receive Idle).
+  // Furthermore, when we have a stage === ReadingOut, we force progress variable to complete
+  //   (by setting remaining = 0 and copying subexposure variables from previous progress).
+  private def adjustProgress(newProgress: Progress)(oldProgress: Progress): Progress = newProgress match {
```
cquiroz: This is a very nice workaround. We should test it on the instrument

PR #1214, opened 2019-11-26
SEQNG-1140 Disable OI guiding for GCAL calibrations.

jluhrs:
OIWFS guiding is now forced to freeze for GCAL calibrations inside a night sequence, if there is an OI in use.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/CleanConfig.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/CleanConfig.scala
@@ -7,8 +7,10 @@ import cats.implicits._
 import edu.gemini.spModel.config2.{Config, DefaultConfig, ItemEntry, ItemKey}
 import seqexec.server.ConfigUtilOps._
 import edu.gemini.spModel.gemini.gmos.InstGmosCommon.USE_NS_PROP
-import edu.gemini.spModel.obscomp.InstConstants.{ARC_OBSERVE_TYPE, FLAT_OBSERVE_TYPE, OBSERVE_TYPE_PROP, BIAS_OBSERVE_TYPE}
-import edu.gemini.spModel.seqcomp.SeqConfigNames.INSTRUMENT_KEY
+import edu.gemini.spModel.guide.StandardGuideOptions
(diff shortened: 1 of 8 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/CleanConfig.scala
```diff
@@ -19,11 +21,6 @@ import seqexec.model.enum.SystemName
  * now, but it allows for more to be added in the future.
  */
 final case class CleanConfig(config: Config, overrides: Map[ItemKey, AnyRef]) {
-  //Check that every value in overrides can be assigned instead of the value in config
-  private def checkTypes: Boolean =
-    overrides.forall{ case (k, v) => Option(config.getItemValue(k)).forall(_.getClass.isAssignableFrom(v.getClass))}
-  assert(checkTypes)
```
jluhrs: I had to remove the assertion, because it would fail when overriding Enums.

modules/seqexec/server/src/test/scala/seqexec/server/CleanConfigSpec.scala
```diff
@@ -30,7 +30,7 @@ class CleanConfigSpec extends AnyFlatSpec with Matchers {
     )
   )
 
-  "CleanConfig" must "fail construction if given a set of overrides with conflicting types" in {
+   ignore must "fail construction if given a set of overrides with conflicting types" in {
```
jluhrs: I removed the test after I had to remove the assertion in `CleanConfig`. Some day we will have a properly typed configuration.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -149,19 +149,23 @@ object SeqTranslate {
...
-        case (c, i) => step(obsId, i, c, nextToRun, sequence.datasets).attempt
+        case (c, i) => step(obsId, i, c, nextToRun, sequence.datasets, isNightSeq).attempt
       }.sequence.map(_.separate)
 
       val instName = configs
         .headOption
         .map(extractInstrument)
         .getOrElse(Either.left(SeqexecFailure.UnrecognizedInstrument("UNKNOWN")))
-
+// scalastyle:off console.io
```
cquiroz: Do you need this comment?

PR #1217, opened 2019-11-27
SEQNG-1145 Improve reusability of tables

cquiroz:
We have detected some unnecessary re renderings. I traced at least some of them to the calculations on tables. The original calculation was too naive and we should account for small changes
This PR seems to reduce the warnings of unneeded refreshes


rpiaggio:
LGTM

PR #1218, opened 2019-11-27
Update scalajs-react-clipboard and react-props

cquiroz:
I moved ReactProps to a dependency and this lets me improve the facade of third party libraries. This PR updates `scalajs-react-clipboard` and replaces `ReactProps`


project/Settings.scala
```diff
@@ -28,7 +28,7 @@ object Settings {
     val scalaJQuery             = "1.2"
     val scalaJSReactCommon      = "0.3.0"
```
rpiaggio: I see you have released a 0.3.2 of `scalaJSReactCommon` with `ReactProps`? I think this version bump would need to be in the PR.

cquiroz: Good catch. this gets evicted but we should do better than this. I'll update it

PR #1220, opened 2019-11-27
[SEQNG-1145] Investigate unnecesary React rerenders

rpiaggio:
@cquiroz I wonder if we should add all the confirmed FALSE POSITIVES from http://jira.gemini.edu:8080/browse/SEQNG-1145 into `dev.js`.

BTW: I took the change to convert a few more components to the `ReactProps` mechanism

### DIFF
--- a/modules/seqexec/web/client/src/main/resources/dev.js
+++ b/modules/seqexec/web/client/src/main/resources/dev.js
@@ -6,7 +6,7 @@ import React from "react";
 if (process.env.NODE_ENV !== "production") {
   const { whyDidYouUpdate } = require("why-did-you-update");
   whyDidYouUpdate(React, {
-    exclude: ["Draggable", "DraggableCore", "AutoSizer", "DiodeWrapper"]
+    exclude: ["Draggable", "DraggableCore", "AutoSizer", "DiodeWrapper", "Login"]
   });
 }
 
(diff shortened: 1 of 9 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepSettings.scala
```diff
@@ -59,7 +59,8 @@ final case class ExposureTimeCell(s: Step, i: Instrument) extends ReactProps {
 object ExposureTimeCell {
   type Props = ExposureTimeCell
 
-  implicit val propsReuse: Reusability[Props] = Reusability.derive[Props]
+  implicit val propsReuse: Reusability[Props] =
+    Reusability.by(p => (p.s.config, p.i))
```
cquiroz: Why not `derive`?

modules/seqexec/web/client/src/main/resources/dev.js
```diff
@@ -6,7 +6,7 @@ import React from "react";
 if (process.env.NODE_ENV !== "production") {
   const { whyDidYouUpdate } = require("why-did-you-update");
   whyDidYouUpdate(React, {
-    exclude: ["Draggable", "DraggableCore", "AutoSizer", "DiodeWrapper"]
+    exclude: ["Draggable", "DraggableCore", "AutoSizer", "DiodeWrapper", "Login"]
```
cquiroz: How is `Login` a false positive. It seems it should be possible to make it reusable

rpiaggio: I'm calling false positives the cases where the plugin tells the rerender is not necessary but it is. This is happening with `Login` and other components.

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepSettings.scala
rpiaggio: `derive` will compare all the `Step` properties, and rerender if any of them changes. In this component we only need to rerender if the `config` changes (and even then, only if the exposure time changes).

PR #1224, opened 2019-12-04
SEQNG-1155 Put some order in the real system log messages.

jluhrs:
Added some missing log messages when interacting with external systems, and gave some uniformity to them. For example:
* Now every "start observe" message includes the file id.
* All long commands have a "start" and "completed" message
* All log messages related to interactions with external systems are at the DEBUG level.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerEpics.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerEpics.scala
@@ -153,9 +153,7 @@ object GsaoiControllerEpics {
         checkDhs *>
         epicsSys.observeCmd.setLabel(fileId) *>
         epicsSys.observeCmd.setTimeout[F](calcObserveTimeout(cfg)) *>
-        epicsSys.observeCmd.post[F].flatMap{ r =>
-          L.debug("Completed GSAOI observe").as(r)
-        }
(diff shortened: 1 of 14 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerEpics.scala
```diff
@@ -149,25 +149,27 @@ object GsaoiControllerEpics {
...
 
-      L.info("Start GSAOI observe") *>
+      L.debug(s"Start GSAOI observe, file id $fileId") *>
         checkDhs *>
         epicsSys.observeCmd.setLabel(fileId) *>
         epicsSys.observeCmd.setTimeout[F](calcObserveTimeout(cfg)) *>
-        epicsSys.observeCmd.post[F]
+        epicsSys.observeCmd.post[F].flatMap{ r =>
+          L.debug("Completed GSAOI observe").as(r)
+        }
```
rpiaggio: `flatTap`?

jluhrs: What is this witchery!

jluhrs: I learnt something new today.

PR #1227, opened 2019-12-06
SEQNG-1160 Fixed logic that tests useAo flag.

jluhrs:


cquiroz:
I suppose we should be able to avoid these bugs with a better test suite, Do you think is worth it?

jluhrs:
Yes, but this depends on EPICS channels. I am not sure how to create unit tests for that.

sraaphorst:
Looks like buildkite isn't happy about this:
```
/usr/local/bin/sbt: line 219: 16358 Killed                  "$@"
🚨 Error: The command exited with status 137
```

PR #1234, opened 2019-12-10
Use github actions instead of buildkite

cquiroz:
Buildkite has become very unstable. I propose we switch to github actions

This PR sets it up running tests only for the seqexec part. This is because we are going to split the repo and I don't want to setup postgress yet

On buildkite we measure the asset sizes. I added another action that would do something equivalent using https://app.packtracker.io/organizations/422/projects/321. The latter task would only run on merge thus making the build faster

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,0 +1,29 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Create Release
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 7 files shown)


.github/workflows/create-release.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          body: |
+            Changes in this Release:
+            * First Change
+            * Second Change
```
tpolecat: So the idea is we go in and edit the release notes?

cquiroz: yes, you do a tag and it adds a github release. Maybe is not that useful for us as we track these in jira

PR #1235, opened 2019-12-10
Upgrade http4s and add tests

cquiroz:
I updated http4s and took the opportunity to add tests for most of the http endpoints


cquiroz:
Hmm, buildkite still shows up

PR #1239, opened 2019-12-11
Remove shared module

cquiroz:
I had a shared module when I expected multiple apps to be in the monorepo
This PR removes the shared module and adds the code to seqexec/model
It should make the build a bit faster


cquiroz:
yes we may want to do that. these were just a few utility classes.
Any thoughts about using `tsec`?

cquiroz:
After some effort the action to measure asset sizes is working to my liking

PR #1241, opened 2019-12-12
Reduce the amount of times we build the js file

cquiroz:
This change should build the js only if relevant files have changed. Hopefully it will reduce CI times in some cases


cquiroz:
So this will not build js e.g. if only the engine has changed but the packtracer checks won't be executed either

cquiroz:
I simplifed a bit, it will be a bit slower but clearer

PR #1250, opened 2019-12-19
SEQNG-1164 Implemented engine heartbeat

jluhrs:


modules/seqexec/server/src/main/scala/seqexec/server/gcal/Gcal.scala
```diff
@@ -88,6 +88,6 @@ object Gcal {
 
   // GCAL that always turn off its lamps except for the IR lamp. Used to assure GCAL light does not interfere in a non
   // calibration step
-  def defaultGcal[F[_]: Sync](controller: GcalController[F]): Gcal[F] =
+  def defaultGcal[F[_]: Sync: Logger](controller: GcalController[F]): Gcal[F] =
```
cquiroz: I think we can skip Sync here

PR #1258, opened 2019-12-26
SEQNG-1167: Hearbeart log message is not shown in client

cquiroz:
It was requested that we should hearbeat messages in the client.
I don't like this feature at all but it is implemented here
It pollutes the client log and makes it note very useful
We should first fix the freezing bug or leave the messages server only
image (4)


andrewwstephens:
This is the inverse of what we want to know (like a fire alarm that beeps when there is no fire).  Is there any (easy) way to throw a warning when the seqexec server stops responding to queries?

cquiroz:
If we were able to know that we'd probably fix the bug 😄

jluhrs:
Maybe we could add some processing in the web server that would filter these messages, but generate a different one if it doesn't see a heartbeat in, say, 30 s?

cquiroz:
I updated this PR to use the new heartbeat system, and we only get notified if the heart beat is lost
Seqexec - GS 2020-01-08 20-34-49

PR #1261, opened 2020-01-03
[SEQNG-1163] Display "Reading Out" in BIAS observations, disable control buttons during readout, fix N&S countdown

rpiaggio:
Two of the fixes are related to issues reported in SEQNG-1163:

1. Display "Reading Out" on BIAS observations.
2. Disable observation control buttons during readout (this is implemented now for all kinds of observations).

Also, beyond the reported issues, this PR attempts to fix a problem with a "dancing" (going back and forth) countdown in N&S observations. This has to be tested on the telescope since our N&S simulation is not yet reflecting the same behavior.

### DIFF
--- a/modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
+++ b/modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
@@ -33,6 +33,9 @@ object Progress {
 
   implicit val nsProgressP: Prism[Progress, NSObservationProgress] =
     GenPrism[Progress, NSObservationProgress]
+
+  implicit val progressP: Prism[Progress, Progress] =
+    GenPrism[Progress, Progress]
 }
 
 final case class ObservationProgress(obsId:     Observation.Id,
--- a/modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
+++ b/modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
@@ -7,7 +7,7 @@ import cats.Eq
 import cats.implicits._
 import gem.Observation
 import gem.util.Enumerated
-import monocle.Prism
+import monocle.{Iso, Prism}
 import monocle.macros.GenPrism
 import squants.Time
 
@@ -35,7 +35,7 @@ object Progress {
     GenPrism[Progress, NSObservationProgress]
 
   implicit val progressP: Prism[Progress, Progress] =
-    GenPrism[Progress, Progress]
+    Iso.id[Progress].asPrism
 }
 
 final case class ObservationProgress(obsId:     Observation.Id,
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
@@ -54,20 +54,21 @@ object NodAndShuffleProgressMessage extends ProgressLabel {
       <.div(
         SeqexecStyles.specialStateLabel,
         SeqexecStyles.progressMessage,
-        p.nsStatus.state.map[VdomElement] { nsState =>
+        p.nsStatus.state.map[VdomElement] { _ =>
           s.progressConnect { proxy =>
+            val nsProgress = proxy()
             val nodCount = NodAndShuffleStage.NsSequence.length
             val nodMillis = p.nsStatus.nodExposureTime.toMilliseconds.toInt
             val cycleMillis = nodMillis * nodCount
-            val remainingCycles = p.nsStatus.cycles - nsState.sub.cycle - 1
-            val remainingNods = nodCount - nsState.sub.stageIndex - 1
-            val remainingNodMillis = proxy().foldMap(_.remaining.toMilliseconds.toInt)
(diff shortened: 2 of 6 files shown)


modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
```diff
@@ -33,6 +33,9 @@ object Progress {
 
   implicit val nsProgressP: Prism[Progress, NSObservationProgress] =
     GenPrism[Progress, NSObservationProgress]
+
+  implicit val progressP: Prism[Progress, Progress] =
```
cquiroz: This seems odd? prism form Progress to Progress?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
```diff
@@ -54,20 +54,21 @@ object NodAndShuffleProgressMessage extends ProgressLabel {
       <.div(
         SeqexecStyles.specialStateLabel,
         SeqexecStyles.progressMessage,
-        p.nsStatus.state.map[VdomElement] { nsState =>
+        p.nsStatus.state.map[VdomElement] { _ =>
           s.progressConnect { proxy =>
+            val nsProgress = proxy()
```
cquiroz: The use of proxy has been problematic in the past. I prefer to extract the content of the proxy when the component is created and stor it in the props

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/StepsControlButtons.scala
```diff
@@ -88,89 +92,92 @@ object ControlButtons {
...
-         .map {
-           case PauseObservation =>
-             Popup("button", "Pause the current exposure")(
-               Button(
-                 icon     = Some(IconPause),
-                 color    = Some("teal"),
-                 onClick  = requestObsPause(p.id, p.stepId),
-                 disabled = p.requestInFlight || p.isObservePaused
+      p.connect{ proxy =>
+        val isReadingOut = proxy().exists(_.stage === ObserveStage.ReadingOut)
```
cquiroz: Same comment as above about proxies

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/sequence/steps/NodAndShuffleProgress.scala
rpiaggio: The proxy is provided by the `connect`. The `connect` is defined in the `Props`, but we cannot extract the value outside the `connect` block.

modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
rpiaggio: I agree, I had my doubts about this.

This comes from the requirement to have a `connect` that provides a generic `Progress` and get the `ObserveStage` to check out if we are if we are `ReadingOut`. It is the first time we need such generic `Progress`, up until now we always requested the `connect` knowing whether we are in a regular observation or a N&S one (which have different `Progress` subclasses).

We obtain the `A <: Progress` `connect` from https://github.com/gemini-hlsw/ocs3/blob/develop/modules/seqexec/web/client/src/main/scala/seqexec/web/client/circuit/SeqexecCircuit.scala#L140, which requires an implicit `Prism[Progress, A]`, so this was the way I found to satisfy  such implicit.

Maybe there's a better way to do it?

rpiaggio: I see now that an `Iso` seems to be a valid `Prism`. Maybe it would be cleaner to define an `Iso` in this case?

rpiaggio:
> I added a few comments inline. Could you add a screenshot?

Yes, definitely:

Screenshot of BIAS observation while reading out now:
image

Screenshot of regular observation while reading out:
image

modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala
rpiaggio: I tried: 

```
implicit val progressP: Prism[Progress, Progress] =
    Iso.id[Progress]
```

but it fails to compile with:
```
[error] /mnt/c/dev/gemini/ocs3/modules/seqexec/model/shared/src/main/scala/seqexec/model/ObservationProgress.scala:38:11: type mismatch;
[error]  found   : monocle.Iso[seqexec.model.Progress,seqexec.model.Progress]
[error]     (which expands to)  monocle.PIso[seqexec.model.Progress,seqexec.model.Progress,seqexec.model.Progress,seqexec.model.Progress]
[error]  required: monocle.Prism[seqexec.model.Progress,seqexec.model.Progress]
[error]     (which expands to)  monocle.PPrism[seqexec.model.Progress,seqexec.model.Progress,seqexec.model.Progress,seqexec.model.Progress]
[error]     Iso.id[Progress]
[error]           ^
[info] monocle.Iso[seqexec.model.Progress,seqexec.model.Progress] <: monocle.Prism[seqexec.model.Progress,seqexec.model.Progress]?
[info] false
```

rpiaggio: Ah, this seems to work:

```
  implicit val progressP: Prism[Progress, Progress] =
    Iso.id[Progress].asPrism
```

PR #1275, opened 2020-01-15
SEQNG-1181 Removed redundant observe timeout, adjusted timeouts.

jluhrs:
The `timeoutTo` became redundant after the observe timeout in the acm code was fixed, and was causing problems because it was triggered before the ACM one in some cases.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
@@ -167,28 +167,23 @@ trait ObserveActions {
       case ObserveCommandResult.Aborted =>
         abortTail(env.systems, env.obsId, fileId)
       case ObserveCommandResult.Paused =>
-        val timeout = env.inst.observeTimeout
         env.inst.calcObserveTime(env.config)
           .flatMap(totalTime => env.inst.observeControl(env.config) match {
             case c: CompleteControl[F] =>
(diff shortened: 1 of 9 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ObserveActions.scala
```diff
@@ -127,8 +127,7 @@ trait ObserveActions {
       _ <- notifyObserveStart(env).timeoutTo(env.inst.observeTimeout, Concurrent[F].raiseError(SeqexecFailure.ObsSystemTimeout(fileId)))
       _ <- env.headers(env.ctx).traverse(_.sendBefore(env.obsId, fileId))
       _ <- info(s"Start ${env.inst.resource.show} observation ${env.obsId.format} with label $fileId")
-      t <- env.inst.calcObserveTime(env.config).map(t => t + env.inst.observeTimeout)
-      r <- env.inst.observe(env.config)(fileId).timeoutTo(new FiniteDuration(t.millis, MILLISECONDS), Concurrent[F].raiseError(SeqexecFailure.ObsTimeout(fileId)))
+      r <- env.inst.observe(env.config)(fileId)
```
cquiroz: Should we  remove the timeouts in the other actions, like pause/abort/etc?

jluhrs: Yes, you are right.

cquiroz:
There is a compilation error

PR #1277, opened 2020-01-16
SEQNG-1183 Guard against null CLID value in CaObserveSenderImpl.

jluhrs:
This should fix the NPE problem.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
@@ -728,7 +728,7 @@ public class CaObserveSenderImpl<C extends Enum<C> & CarStateGeneric> implements
 
         @Override
         public CaObserveSenderImpl.ApplyState onCarValChange(final CarStateGeneric val) {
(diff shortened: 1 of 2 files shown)


modules/acm/src/main/java/edu/gemini/epics/acm/CaObserveSenderImpl.java
```diff
@@ -728,7 +728,7 @@ public String signature() {
 
         @Override
         public CaObserveSenderImpl.ApplyState onCarValChange(final CarStateGeneric val) {
-            if ((carState == null || carState.isIdle()) && val.isBusy() && carClid > 0) {
+            if ((carState == null || carState.isIdle()) && val.isBusy() && carClid != null && carClid > 0) {
```
jluhrs: The state event handlers are not called with null values, there is no need to check `val`.

jluhrs:
@cquiroz `carClid` in the `WaitApplyPreset` is `null` until `WaitApplyPreset.onCarClidChange` is called at least once. It could be that in some cases, `WaitApplyPreset.onCarValChange` is called  first.

cquiroz:
I guess we could reproduce that in a test?

PR #1279, opened 2020-01-17
SEQNG-1178 Always write CRPA even for calibrations

cquiroz:
We should always write CRPA


sraaphorst:
I originally didn't read that as CRPA.

PR #1288, opened 2020-01-22
SEQNG-1184 Fixed timeouts for all EPICS commands.

jluhrs:
There has been several problems with EPICS commands executed without a timeout. I put the timeout as parameter of the `post` command so we never forget to set it.

Additionally, I found the test that was causing the build to take forever. It was testing the observe timeout that was removed in a previous PR, so I removed the test.


modules/seqexec/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -34,15 +34,14 @@ import scala.math.abs
 import scala.collection.JavaConverters._
 
 trait EpicsCommand[F[_]] {
-  def post: F[ApplyCommandResult]
+  def post(timeout: Time): F[ApplyCommandResult]
```
cquiroz: This is very good, a great improvement. Should we use `FiniteDuration` rather than `Time`?

jluhrs: Why can't you let me be lazy ?

PR #1289, opened 2020-01-22
remove gsp-core sources, add gsp-core dependency

tpolecat:
This deletes a lot of stuff.

image

All of the deleted code has moved to the [gsp-core](https://github.com/gemini-hlsw/gsp-core) project, which publishes three artifacts that are now pulled in as a dependency.

`gsp-core` version `0.1.3` is up to date with ocs3 commit 0f67f874f237be247afc742bfe07ae43da0b26d3 which is the current `develop` head right now. If you want to confirm that they're consistent you can check out gsp-core at tag `0.1.3` and check out the ocs3 `develop` branch and run `./update` in gsp-core, which will re-copy all the sources over (gsp-core and ocs3 must be sibling directories). Following up with `git status` should show that there are no differences.

There is probably some remaining cruft (dependencies declared in `Settings.sbt` but never used, for instance).


tpolecat:
It will be interesting to see how much this speeds up CI.

build/buildkite-build.sh
```diff
@@ -10,7 +10,6 @@ cd `dirname $0`/..
 echo "--- :scala: Compiling main codebase"
 /usr/local/bin/sbt                  \
   -jvm-opts build/buildkite-jvmopts \
-  -Docs3.skipDependencyUpdates      \
```
cquiroz: I think we can completely delete this file

tpolecat: Yeah there's probably a lot of cleanup that can be done now but I don't want to hold this up anymore because I have to start over if any of the deleted code changes in `develop`.

PR #1307, opened 2020-02-06
SEQNG-1162 Monitor WebSocket pings and kill unresponsive clients, plu…

cquiroz:
We have been observing a strange behaviour when the seqexec would hang for several minutes with the web side active but the engine blocked

There is evidence that sometimes clients go "zombie", possibly due to the re connection logic on the client or some odd behavior at the port mapping iptables. But we have seen multiples clients pointing to a given ip while we can see on the UI just one.
The hypothesis is that the server keeps trying to push WS messages to those clients but they are not getting them blocking the server upstream.

Unfortunately this is hard to reproduce but a possible remediation is offered in this PR:

WebSocket supports the concept of Ping/Pong where the server pushes ping control frames and the client should answer. http4s doesn't do this by default but the seqexec sends pings as a way to ensure the connection remains alive
However, the pongs are discarded and nobody is checking if they arrive timely

This PR adds a check that if Pongs don't arrive after n intervals the client connection is terminated. It is hard to prove this will solve the problem, the suggestion is to install it and field test it

Finally a bit more debugging is including, including printing every certain amount of time a summary of the clients currently connected

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/WebSocketHandler.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/WebSocketHandler.scala
@@ -80,7 +80,7 @@ class WebSocketHandler[M](modelRW: ModelRW[M, WebSocketConnection])
     def onClose(): Unit =
       // Increase the delay to get exponential backoff with a minimum of 200ms and a max of 1m
       if (value.autoReconnect) {
-        SeqexecCircuit.dispatch(ConnectionRetry(math.min(60000, math.max(200, value.nextAttempt * 2))))
+        SeqexecCircuit.dispatch(ConnectionRetry(math.min(60000, math.max(1000, value.nextAttempt * 2))))
       }
 
     ws.binaryType = "arraybuffer"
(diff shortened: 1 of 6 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/handlers/WebSocketHandler.scala
```diff
@@ -80,7 +80,7 @@ class WebSocketHandler[M](modelRW: ModelRW[M, WebSocketConnection])
     def onClose(): Unit =
       // Increase the delay to get exponential backoff with a minimum of 200ms and a max of 1m
```
rpiaggio: The comment still says 200ms.

tpolecat: Comments are bad!

cquiroz: I think @tpolecat is right. I should remove the comment 😄

PR #1321, opened 2020-02-18
Upgrade to scalajs-react+ecosystem versions.

rpiaggio:


rpiaggio:
I brought `io.github.cquiroz.react.{common, react-virtualized, react-draggable, react-sortable-hoc, react-clipboard}` to the latest versions.

There's a behavior in `common` which I don't know if you want to keep: if we `import react.common._` and `import react.common.implicits._`, then there are amiguous implicits since both packages extend `AllSyntax`.

Also, this PR is not ready to merge yet. I detected a problem with the router, which now prevents ambiguous routes (it seemes that previous to `scalajs-react` `1.5.0` it used the first match).

cquiroz:
You should only need `react.common._` as it now includes `syntax`

PR #1322, opened 2020-02-19
SEQNG-1189 Removed WCS for NIFS.

jluhrs:
Seqexec does not need to write the WCS FITS keywords for NIFS anymore, because the writing of the keywords was moved to the instrument (as it should be).
This PR must not be merged until the new NIFS DC software is officially deployed.


cquiroz:
Did we very MDJ_OBS is included?

PR #1334, opened 2020-02-25
SEQNG-1191 Update ocs ghost model

cquiroz:
This PR brings the latest ocs model for GHOST and reads the `FiberAgitator` parameter

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ghost/Ghost.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ghost/Ghost.scala
@@ -106,27 +106,29 @@ object Ghost {
     EitherT {
       Sync[F].delay {
         (for {
-          baseRAHMS     <- raExtractor(SPGhost.BaseRAHMS)
-          baseDecDMS    <- decExtractor(SPGhost.BaseDecDMS)
+          baseRAHMS     <- raExtractor(SPGhost.BASE_RA_HMS)
+          baseDecDMS    <- decExtractor(SPGhost.BASE_DEC_DMS)
 
-          srifu1Name    = extractor[String](SPGhost.SRIFU1Name)
-          srifu1RAHMS   <- raExtractor(SPGhost.SRIFU1RAHMS)
-          srifu1DecHDMS <- decExtractor(SPGhost.SRIFU1DecDMS)
+          fiberAgitator = extractor[Boolean](SPGhost.FIBER_AGITATOR)
+          srifu1Name    = extractor[String](SPGhost.SRIFU1_NAME)
+          srifu1RAHMS   <- raExtractor(SPGhost.SRIFU1_RA_HMS)
+          srifu1DecHDMS <- decExtractor(SPGhost.SRIFU1_DEC_DMS)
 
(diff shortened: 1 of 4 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ghost/Ghost.scala
```diff
@@ -106,27 +106,29 @@ object Ghost {
...
+          hrifu2RAHMS   <- raExtractor(SPGhost.HRIFU2_RA_HMS)
+          hrifu2DecHDMS <- decExtractor(SPGhost.HRIFU2_DEC_DMS)
 
           config <- GhostConfig(
             (baseRAHMS, baseDecDMS).mapN(Coordinates.apply),
             1.minute,
+            FiberAgitator.fromBoolean(fiberAgitator.getOrElse(false)),
             srifu1Name, (srifu1RAHMS, srifu1DecHDMS).mapN(Coordinates.apply),
             srifu2Name, (srifu2RAHMS, srifu2DecHDMS).mapN(Coordinates.apply),
             hrifu1Name, (hrifu1RAHMS, hrifu1DecHDMS).mapN(Coordinates.apply),
```
sraaphorst: Do we need something here like:
`(hrifu1RAHMS, hrifu2DecHDMS).mapN(Coordinates.apply)`
?
(Maybe this will come later... it just seems strange that the SRIFU1, SRIFU2, and HRIFU1 coordinates all appear here, and the HRIFU2 coordinates don't, even though they will always be named `"Sky"` if they do exist.)

modules/seqexec/server/src/main/scala/seqexec/server/ghost/GhostController.scala
```diff
@@ -81,6 +103,80 @@ sealed trait GhostConfig {
...
+
+  private[ghost] def ifuPark(ifuNum: IFUNum): Configuration = {
+    def cfg[P: GiapiConfig](paramName: String, paramVal: P) =
+      Configuration.single(s"${ifuNum.ifuStr}.$paramName", paramVal)
+      cfg("target", IFUTargetType.NoTarget.targetType) |+|
+      cfg("type", DemandType.DemandPark.demandType)
+  }
+
+  def apply(baseCoords: Option[Coordinates],
+            expTime: Duration,
```
sraaphorst: This is the part I'm questioning: we no longer have a single `expTime`, but two `expTimes`: one for the red camera and one for the blue camera, along with exposure counts. We also have spectral / spatial binning for each of the two cameras, and read / gain.

sraaphorst: Do we need to pass these through the seqexec?

sraaphorst: (Definitely the `expTime` seems like it should be removed, I would think.)

cquiroz: I expect that we send on the sequence the blue and red times and a total exp time (or maybe have a way to calculate it in the seqexec). We need a total exp time to set the timeouts

PR #1363, opened 2020-03-17
Support newer jdks

cquiroz:
This PR lets us run the seqexec on newer jdks. It requires some minor code updates and the explicit addition of jaxb as a dependency
I tested running with jdk8


cquiroz:
@jluhrs I deployed this to seqexec-test and it starts fine running on jdk 8

cquiroz:
I did further testing deploying to production with our current 1.8 JRE and also with JDK14

PR #1389, opened 2020-04-17
Update cats-effect and sbt-gsp

cquiroz:
Updates some libs and headers.


cquiroz:
Yes those are the changes, the sbt-gsp
Update required and update on the headers

PR #1424, opened 2020-06-04
Port to scala.js 1.0 and scalajs-react 1.7.0

cquiroz:
This is a work in progress
Pending:
* Use public versions of diode and boopickle
* Eliminate deprecated calls
* Testing

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/ConnectionState.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/ConnectionState.scala
@@ -14,9 +14,8 @@ import seqexec.web.client.icons._
 import seqexec.web.client.model.WebSocketConnection
 import seqexec.web.client.reusability._
 
-final case class ConnectionState(u: WebSocketConnection) extends ReactProps {
-  @inline def render: VdomElement = ConnectionState.component(this)
-}
+final case class ConnectionState(u: WebSocketConnection)
+    extends ReactProps[ConnectionState](ConnectionState.component)
 
 /**
   * Alert message when the connection disappears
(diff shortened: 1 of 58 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/ConnectionState.scala
```diff
@@ -34,7 +33,7 @@ object ConnectionState {
       f"${delay / 1000}%d"
     }
 
-  private val component = ScalaComponent
+  val component = ScalaComponent
```
rpiaggio, line 36: Just for future reference, we can keep these `protected`.

project/Settings.scala
```diff
@@ -120,6 +121,12 @@ object Settings {
       "net.logstash.logback" % "logstash-logback-encoder" % LibraryVersions.logstash
     )
     val Log4s = Def.setting("org.log4s" %%% "log4s" % LibraryVersions.log4s)
+    val Log4CatsLogLevel = Def.setting(
+      Seq(
+        "io.chrisdavenport" %%% "log4cats-core"     % LibraryVersions.log4cats,
+        "com.rpiaggio"      %%% "log4cats-loglevel" % LibraryVersions.log4catsLevel
```
rpiaggio, line 127: I think we are not using this one.

project/plugins.sbt
```diff
@@ -4,7 +4,7 @@ resolvers  ++= Seq(
 )
 // Gives support for Scala.js compilation
 val scalaJSVersion =
-  Option(System.getenv("SCALAJS_VERSION")).getOrElse("0.6.33")
+  Option(System.getenv("SCALAJS_VERSION")).getOrElse("1.1.1")
```
rpiaggio, line 7: I think we are not using this.

cquiroz:
This has been tested enough and we need to move forward to work on other fixes. I'll merge it and keep testing on master

PR #1434, opened 2020-06-16
SEQNG-1197 FIx for TCS in-position timeout problem.

jluhrs:
The main purpose of this PR is to fix the TCS in-position timeout problem. The fix was to avoid waiting for the in-position to stabilize when there was no offset applied. I found other minor problems that I fixed.
Other changes:

- Consolidated the construction of thread pools in the ACM module.
- Callback calls are now done in their own thread.

### DIFF
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/AttributeNotifier.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/AttributeNotifier.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+ * For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+ */
+
 package edu.gemini.epics.acm;
 
 import java.util.LinkedList;
--- a/modules/acm/src/main/java/edu/gemini/epics/acm/AttributeNotifier.java
+++ b/modules/acm/src/main/java/edu/gemini/epics/acm/AttributeNotifier.java
@@ -24,7 +24,15 @@ public class AttributeNotifier<T> {
     }
 
     public void notifyValueChangeSingle(CaAttributeListener<T> listener, List<T> vals) {
-        executor.execute(() -> listener.onValueChange(vals));
+        executor.execute(() -> {
+            synchronized (AttributeNotifier.this) {
+                // Check that the listener has not been removed between the call to notifyValueChangeSingle and
+                // the deferred execution of the callback.
+                if(AttributeNotifier.this.listeners.contains(listener)) {
+                    listener.onValueChange(vals);
+                }
+            }
+        });
     }
 
     synchronized public void notifyValueChange(List<T> newVals) {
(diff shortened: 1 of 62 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
```diff
@@ -786,14 +810,19 @@ final class TcsEpicsImpl[F[_]: Async](epicsService: CaService, tops: Map[String,
   // for the in-position to change to true and stay true for stabilizationTime. It will wait up to `timeout`
   // seconds for that to happen.
   override def waitInPosition(stabilizationTime: Duration, timeout: FiniteDuration)(implicit T: Timer[F]): F[Unit] =
-    T.sleep(FiniteDuration(tcsSettleTime.toMillis, TimeUnit.MILLISECONDS)) *>
-      Sync[F].delay(filteredInPositionAttr.restart(stabilizationTime))
-        .flatMap(waitForValueF(_, "TRUE", timeout,"TCS inposition flag"))
+    T.sleep(FiniteDuration(tcsSettleTime.toMillis, TimeUnit.MILLISECONDS)) *> (
```
cquiroz, line 813: I guess this is the core fix

modules/seqexec/server/src/test/scala/seqexec/server/EpicsUtilSpec.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.scalatest.flatspec.AnyFlatSpec
+import java.util.concurrent.TimeUnit.MILLISECONDS
+
+import scala.concurrent.duration.FiniteDuration
+
+class EpicsUtilSpec extends AnyFlatSpec {
+
+  val executor: ScheduledExecutorService = new ScheduledThreadPoolExecutor(2)
+
+  "EpicsUtil waitForValue" should "work with CaWindowStabilizer" in {
```
cquiroz, line 21: 👍

modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
```diff
@@ -28,13 +28,15 @@
     private final CaAttributeListener<T> valListener;
     private T filteredVal;
     private T lastVal;
+    private final AttributeNotifier<T> notifier;
 
-    public CaWindowStabilizer(CaAttribute<T> sa, Duration settleTime) {
+    public CaWindowStabilizer(CaAttribute<T> sa, Duration settleTime, ScheduledExecutorService executor) {
```
cquiroz, line 33: I don't recall this class from before 😧

modules/acm/src/main/java/edu/gemini/epics/acm/AttributeNotifier.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    synchronized public void removeListener(CaAttributeListener<T> listener) {
+        listeners.remove(listener);
+    }
+
+    public void notifyValueChangeSingle(CaAttributeListener<T> listener, List<T> vals) {
+        executor.execute(() -> {
+            synchronized (AttributeNotifier.this) {
```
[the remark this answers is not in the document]
jluhrs: Yes. This is one consequence of having the callback calls in their own threads. It open the possibility that the listener is removed after the callback call is scheduled, but before it is executed. That is why I added a check in the actual call.

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
jluhrs: Yes it is.

modules/acm/src/main/java/edu/gemini/epics/acm/CaWindowStabilizer.java
jluhrs: Which one, `CaWindowStabilizer`?

modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsEpics.scala
jluhrs: Yes and no. It is an improvement that would have fixed the problem, but the only case when this was called with a zero stabilization time was when there was no offset applied, and that case does not apply anymore.

PR #1450, opened 2020-07-03
SEQNG-512 Add the concept of unknown conditions

cquiroz:
The old seqexec included the concept of unknown for conditions but it wasn't included in the new seqexec. This PR adds that value and sets it as the default
A special case for unknown is added to write keywords with that value

### DIFF
--- a/modules/seqexec/model/shared/src/main/scala/seqexec/model/Conditions.scala
+++ b/modules/seqexec/model/shared/src/main/scala/seqexec/model/Conditions.scala
@@ -18,6 +18,14 @@ final case class Conditions(
 
 object Conditions {
 
+  val Unknown: Conditions =
+    Conditions(
+      CloudCover.Unknown,
+      ImageQuality.Unknown,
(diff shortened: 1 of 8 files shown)


modules/seqexec/model/shared/src/main/scala/seqexec/model/Conditions.scala
```diff
@@ -45,12 +53,10 @@ object Conditions {
     )
 
   val Default: Conditions =
-    Worst // Taken from ODB
+    Unknown // Taken from ODB
 
   implicit val equalConditions: Eq[Conditions] =
     Eq.by(x => (x.cc, x.iq, x.sb, x.wv))
 
-  implicit val showConditions: Show[Conditions] =
```
rpiaggio, line 53: This wasn't being used I presume?

cquiroz: nope, we were using `Show` for the conditions components though

PR #1457, opened 2020-07-10
Port to scala 2.13.3

cquiroz:
Updates the code to scala 2.13.3, it should be fairly mechanical


modules/seqexec/server/src/main/scala/seqexec/server/tcs/TcsController.scala
```diff
@@ -215,7 +215,7 @@ object TcsController {
 
   final case class InstrumentOffset(p: Angle@@OffsetP, q: Angle@@OffsetQ) {
     def toFocalPlaneOffset(iaa: Angle): FocalPlaneOffset = FocalPlaneOffset(
-      tag[OffsetX]((-p * iaa.cos - q * iaa.sin) / FOCAL_PLANE_SCALE),
+      tag[OffsetX](((p * -1 * iaa.cos) - q * iaa.sin) / FOCAL_PLANE_SCALE),
```
cquiroz, line 218: This is fairly ugly. Scala 2.13.3 forbids calls to methods that take parens without them. In this case we are calling squant's `unary_-()` method and it won't allow it
We maybe able to get a PR into squants to solve it

PR #1461, opened 2020-07-14
SEQNG-1205 Fixed RELEASE keyword.

jluhrs:
Use proprietaryMonths to calculate release date regardless of the value of headerVisibility, as the old Seqexec does.


modules/seqexec/server/src/main/scala/seqexec/server/keywords/ObsKeywordReader.scala
```diff
@@ -295,10 +295,7 @@ object ObsKeywordReader extends ObsKeywordsReaderConstants {
...
     override def headerPrivacy: F[Boolean] = headerPrivacyF
 
-    private val noProprietaryMonths: F[String] =
-      LocalDate.now(ZoneId.of("GMT")).format(DateTimeFormatter.ISO_LOCAL_DATE).pure[F]
-
-    private val calcProprietaryMonths: F[String] =
+    private val calcReleaseDate: F[String] =
       EitherT(
         config.extractAs[Integer](PROPRIETARY_MONTHS_KEY)
           .recoverWith {
```
cquiroz, line 301: There is a call to LocalDate.now inside.We should put that inside a `Sync.delay`

PR #1466, opened 2020-07-30
SEQNG-1206 Send event sequenceStart to ODB

jluhrs:
The event was sent only when executing the first step, which meant there was not event when starting a partially executed sequence.
I also found several cases where the server was not notifying the web clients when a sequence started, particularly when run from the calibration queue. That was also fixed.

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Handle.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Handle.scala
@@ -72,12 +72,13 @@ object Handle {
   // are concatenated in the reverse order.
   implicit class HandleReverseMap[F[_]: Monad, D, V, A](self: Handle[F, D, V, A]) {
     private def reverseConcatOpP(op1: Option[Stream[F, V]],
-                              op2: Option[Stream[F, V]]): Option[Stream[F, V]] = (op1, op2) match {
-          case (None, None) => None
-          case (Some(p1), None) => Some(p1)
-          case (None, Some(p2)) => Some(p2)
-          case (Some(p1), Some(p2)) => Some(p2 ++ p1)
-        }
+                                 op2: Option[Stream[F, V]]): Option[Stream[F, V]] = (op1, op2) match {
+      case (None, None) => None
+      case (Some(p1), None) => Some(p1)
+      case (None, Some(p2)) => Some(p2)
+      case (Some(p1), Some(p2)) => Some(p2 ++ p1)
+    }
+
     def reversedStreamFlatMap[B](f: A => Handle[F, D, V, B]):Handle[F, D, V, B] = Handle[F, D, V, B](
       self.run.flatMap {
         case (a, op1) => f(a).run.map {
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Handle.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Handle.scala
@@ -68,6 +68,25 @@ object Handle {
     }
   }
 
+  // This class adds a method to Hadle similar to flatMap, but the Streams resulting from both Handle instances
+  // are concatenated in the reverse order.
+  implicit class HandleReverseMap[F[_]: Monad, D, V, A](self: Handle[F, D, V, A]) {
+    private def reverseConcatOpP(op1: Option[Stream[F, V]],
+                              op2: Option[Stream[F, V]]): Option[Stream[F, V]] = (op1, op2) match {
+          case (None, None) => None
+          case (Some(p1), None) => Some(p1)
+          case (None, Some(p2)) => Some(p2)
+          case (Some(p1), Some(p2)) => Some(p2 ++ p1)
+        }
+    def reversedStreamFlatMap[B](f: A => Handle[F, D, V, B]):Handle[F, D, V, B] = Handle[F, D, V, B](
+      self.run.flatMap {
+        case (a, op1) => f(a).run.map {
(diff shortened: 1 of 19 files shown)


modules/seqexec/engine/src/main/scala/seqexec/engine/UserEvent.scala
```diff
@@ -19,7 +19,7 @@ sealed trait UserEvent[+F[_], S, U] extends Product with Serializable {
 
 object UserEvent {
 
-  final case class Start[S, U](id: Observation.Id, user: Option[UserDetails], clientId: ClientId, userCheck: S => Boolean) extends UserEvent[Nothing, S, U]
+  final case class Start[S, U](id: Observation.Id, user: Option[UserDetails], clientId: ClientId) extends UserEvent[Nothing, S, U]
```
jluhrs, line 22: The `userCheck` was removed because the resources check is not done inside `Engine` anymore.

modules/seqexec/engine/src/main/scala/seqexec/engine/Handle.scala
```diff
@@ -68,6 +68,25 @@ object Handle {
...
+  // are concatenated in the reverse order.
+  implicit class HandleReverseMap[F[_]: Monad, D, V, A](self: Handle[F, D, V, A]) {
+    private def reverseConcatOpP(op1: Option[Stream[F, V]],
+                              op2: Option[Stream[F, V]]): Option[Stream[F, V]] = (op1, op2) match {
+          case (None, None) => None
+          case (Some(p1), None) => Some(p1)
+          case (None, Some(p2)) => Some(p2)
+          case (Some(p1), Some(p2)) => Some(p2 ++ p1)
+        }
+    def reversedStreamFlatMap[B](f: A => Handle[F, D, V, B]):Handle[F, D, V, B] = Handle[F, D, V, B](
```
cquiroz, line 81: line in between?

modules/seqexec/engine/src/main/scala/seqexec/engine/Handle.scala
```diff
@@ -68,6 +68,25 @@ object Handle {
     }
   }
 
+  // This class adds a method to Hadle similar to flatMap, but the Streams resulting from both Handle instances
+  // are concatenated in the reverse order.
+  implicit class HandleReverseMap[F[_]: Monad, D, V, A](self: Handle[F, D, V, A]) {
```
cquiroz, line 73: Nice concept

modules/seqexec/engine/src/main/scala/seqexec/engine/UserEvent.scala
cquiroz: Where is it done then?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -810,7 +849,7 @@ object SeqexecEngine {
     SequenceView(seq.id, SequenceMetadata(instrument, obsSeq.observer, obsSeq.seqGen.title), st.status, engineSteps(seq), None)
   }
 
-  private def toSeqexecEvent[F[_]](ev: EventResult[SeqEvent], qState: EngineState[F]): SeqexecEvent = {
+  private def toSeqexecEvent[F[_]](ev: EventResult[SeqEvent], qState: EngineState[F]): Stream[F, SeqexecEvent] = {
```
jluhrs, line 852: The change to `Stream` for the result allows to produce more than one `SeqexecEvent` in response to a `SeqEvent`.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -115,17 +110,20 @@ object SeqTranslate {
         extractStatus(config) match {
           case StepState.Pending if i >= nextToRun => SequenceGen.PendingStepGen(
             i,
+            dataId,
             config,
             calcResources(sys),
-            StepActionsGen(initialStepExecutions, configs, rest)
+            StepActionsGen(List.empty, configs, rest)
```
cquiroz, line 116: Do we need the first param anymore?

modules/seqexec/server/src/main/scala/seqexec/server/SeqexecEngine.scala
cquiroz: Thanks, I was wondering about that

jluhrs:
One major change is that I don't use the `Start` event anymore. Instead, I follow the example of `SeqexecEngine.startFrom`, and implement the start inside a `ModifyState` event.

jluhrs:
The other major change is that I moved the extraction of the `dataId` from the configuration, from the time when the step is run to the time the step is built. That simplified a bit the code, and allowed me to have the `dataId` value available for the `sequenceStart` event.

modules/seqexec/engine/src/main/scala/seqexec/engine/UserEvent.scala
jluhrs: In `SeqexecEngine`. I already had to do something similar in the implementation of `SeqexecEngine.startFrom`.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: Well, it is not used anymore.

PR #1472, opened 2020-08-03
SEQNG-1209 Fail fast if sending keywords fail

cquiroz:
We are being too lenient about DHS errors and thus we could end up e.g. that in some cases we get an error but continue observing or mark it as done even though some keyword acquisitions failed

Note that with this change we could have a long exposure and a keyword collection failse, we'd fail the whole step.


modules/seqexec/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -270,32 +270,31 @@ package object keywords {
...
-  def sendKeywords[F[_]: MonadError[?[_], Throwable]](
+  def sendKeywords[F[_]: MonadError[*[_], Throwable]](
       id: ImageFileId,
       inst: InstrumentSystem[F],
       b: List[KeywordBag => F[KeywordBag]]): F[Unit] =
     (for {
       bag <- inst.keywordsClient.keywordsBundler.bundleKeywords(b)
       _   <- inst.keywordsClient.setKeywords(id, bag, finalFlag = false)
     } yield ())
-    .handleError(_ => ()) // In case there is an error ignore them to let other keywords be added
```
cquiroz, line 298: This is the relevant bugfix

modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
```diff
@@ -70,7 +70,7 @@ class DhsClientHttp[F[_]: Concurrent](base: Client[F], baseURI: Uri)(implicit ti
           "keywords" := keywords.keywords
         )
       ),
-      baseURI / id / "keywords"
+      if (finalFlag) baseURI / id / "eywords" else baseURI / id / "keywords"
```
jluhrs, line 73: Wait, what??

PR #1485, opened 2020-08-31
SEQNG-1218 Properly read the Observing wavelength from the OT

cquiroz:
This fixes a fault reported where the Wavelength wasn't properly set on each step.
It was tracked to the value from the OT using a different type than expected and 
the error was silently discarded.
This makes the values of the Wavelength on the headers incorrect but worse the TCS
wasn't getting the correct values. It may invalidate some observations.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,6 +52,7 @@ val dateFormatter = java.time.format.DateTimeFormatter.BASIC_ISO_DATE
 
 inThisBuild(
   List(
+    scalaVersion in ThisBuild := "2.13.3",
     version := dateFormatter.format(
       dynverCurrentDate.value.toInstant.atZone(java.time.ZoneId.of("UTC")).toLocalDate
     ) + dynverGitDescribeOutput.value.mkVersion(versionFmt,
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -52,6 +52,7 @@ val dateFormatter = java.time.format.DateTimeFormatter.BASIC_ISO_DATE
 
 inThisBuild(
   List(
+    scalaVersion in ThisBuild := "2.13.3",
```
rpiaggio, line 55: Isn't this set by `sbt-gsp`?

modules/seqexec/server/src/main/scala/seqexec/server/SequenceConfiguration.scala
```diff
@@ -51,7 +51,10 @@ trait SequenceConfiguration {
     }.getOrElse(StepState.Failed("Logical error reading step status"))
 
   def extractWavelength(config: CleanConfig): Option[Wavelength] =
-    config.extractAs[Wavelength](OBSERVING_WAVELENGTH_KEY).toOption
+    config
+      .extractAs[String](OBSERVING_WAVELENGTH_KEY)
+      .flatMap(v => Either.catchNonFatal(v.toDouble))
+      .map(Wavelength.fromMicrons[Double](_)).toOption
```
rpiaggio, line 57: Just to know what's happening here: what happens if parsing fails?

cquiroz: The response gets to `None` which is what you want. but now that you mention it maybe we should log something

build.sbt
cquiroz: Yes for the most part but not the root one and `metals` was complaining. We have an older `sbt-gsp` here waiting for the scala.js upgrade

PR #1486, opened 2020-09-02
SEQNG-1218 Further refine the search for different Wavelengths sent b…

cquiroz:
The OT sends different types in the `OBSERVING_WAVELENGTH` field. This was detected while testing and it affects GNIRS in particular.

This PR includes support for the `GNIRS` wavelength type and adds some extra safeguards for other possible types
The logic has been changed so that if we can't decode we'll in fact forbid loading a sequence. This is undesirable but better than allowing bad sequences to be executed.
Hopefully after this has been ironed out in testing we can go back to the simpler version of the code

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -11,6 +11,7 @@ import cats.implicits._
 import edu.gemini.seqexec.odb.{ExecutedDataset, SeqexecSequence}
 import edu.gemini.spModel.gemini.altair.AltairParams.GuideStarType
 import edu.gemini.spModel.obscomp.InstConstants.{DATA_LABEL_PROP, OBSERVE_TYPE_PROP, SCIENCE_OBSERVE_TYPE}
+import edu.gemini.spModel.core.Wavelength
 import fs2.Stream
 import gem.Observation
 import gem.enum.Site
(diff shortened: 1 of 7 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/SequenceConfiguration.scala
```diff
@@ -50,13 +51,33 @@ trait SequenceConfiguration {
...
+          .extractAs[Wavelength](OBSERVING_WAVELENGTH_KEY)
+      }.orElse{
+        // Just in case
+        config
+          .extractAs[java.lang.Double](OBSERVING_WAVELENGTH_KEY)
+          .map(v => Wavelength.fromMicrons[Double](v.doubleValue))
+      }.leftMap {
+        case _ =>
+          Unexpected(s"Error reading wavelength ${config.itemValue(OBSERVING_WAVELENGTH_KEY)}: ${config.itemValue(OBSERVING_WAVELENGTH_KEY).getClass()}")
+      }
```
swalker2m, line 77: I was worried that we might store wavelength in different units for different instruments but I checked and that doesn't seem to be the case.  It's safe to assume that they are always microns.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -323,29 +324,37 @@ object SeqTranslate {
     private def flatOrArcTcsSubsystems(inst: Instrument): NonEmptySet[TcsController.Subsystem] =
       NonEmptySet.of(AGUnit, (if (inst.hasOI) List(OIWFS) else List.empty): _*)
 
+    private def tryWavelength(inst: Instrument, config: CleanConfig): F[Option[Wavelength]] =
+      extractWavelength(config) match {
+        case Left(x) => Logger[F].error(s"Cannot decode the wavelength for ${inst.label}") *> MonadError[F, Throwable].raiseError(new SeqexecFailure.Execution(s"Cannot decode the wavelength from the sequence $x"))
+        case Right(w) => w.some.pure[F]
+      }
```
swalker2m, line 331: As discussed we'll need to check whether `Wavelength` is present at all.  If that's not an error then we'd return `None` in that case.

cquiroz: I added code to handle the case of no wavelength defined in the sequence. Local tests with this case went ok.

cquiroz:
Added a fix for the case of no wavelength and improved the log output using pprint

cquiroz:
Thanks @swalker, the no wavelength case was quite important

PR #1489, opened 2020-09-04
Cats 2.2.0

cquiroz:
scala-steward made some fixes for cats 2.2.0 but missed others. This PR completes the upgrade

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
@@ -4,7 +4,7 @@
 package seqexec.web.client.components
 
 import cats.implicits._
-import scala.scalajs.js.JSConverters._
+import scala.scalajs.js.JSConverters._
 import japgolly.scalajs.react._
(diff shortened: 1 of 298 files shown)


tpolecat (approved):

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
```diff
@@ -3,8 +3,8 @@
 
 package seqexec.web.client.components
 
-import scala.scalajs.js.JSConverters._
 import cats.implicits._
+import scala.scalajs.js.JSConverters._
```
rpiaggio, line 7: There seems to be an extra newline.

rpiaggio (approved):
LGTM

modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/DividedProgress.scala
cquiroz: Oh that's odd

cquiroz:
I'll run some experiments with packtracker along the next few PRs.

PR #1497, opened 2020-09-07
Use Zipper from gem-ui

cquiroz:
Though we'll soon move to lucuma-ui we may as well remove this implementation of `Zipper` to ease the transition


rpiaggio (approved):
LGTM

PR #1498, opened 2020-09-07
SEQNG-1219 Skip logging when an unsupported instrument's sequence is loaded

cquiroz:
We put a log entry when an unsupported instrument shows on the queue.
This log is repeated every one minute polluting the logs

This feature had more sense when we had partial instrument support, but now it just adds noise.

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
@@ -8,7 +8,7 @@ import cats.Endo
 import cats.MonadError
 import cats.effect.Concurrent
 import cats.effect.Timer
-import cats.syntax.all._
+import cats.implicits._
 import edu.gemini.spModel.core.SPProgramID
 import edu.gemini.spModel.obscomp.InstConstants
 import edu.gemini.spModel.seqcomp.SeqConfigNames.OCS_KEY
@@ -19,6 +19,7 @@ import seqexec.engine.Sequence
 import seqexec.server.ConfigUtilOps._
 import seqexec.server.SeqEvent._
 import seqexec.server.SeqexecFailure.SeqexecException
+import seqexec.server.SeqexecFailure.UnrecognizedInstrument
 
 final class ODBSequencesLoader[F[_]: ApplicativeError[?[_], Throwable]: Logger](
   odbProxy: OdbProxy[F],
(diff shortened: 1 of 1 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
```diff
@@ -8,7 +8,7 @@ import cats.Endo
 import cats.MonadError
 import cats.effect.Concurrent
 import cats.effect.Timer
-import cats.syntax.all._
+import cats.implicits._
```
rpiaggio, line 11: Why isn't `cats.syntax.all_` enough here?

modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
```diff
@@ -53,35 +54,41 @@ final class ODBSequencesLoader[F[_]: ApplicativeError[?[_], Throwable]: Logger](
...
+          .fold(ODBSequencesLoader.loadSequenceEndo(seqId, seqg, execEngine))(_ =>
+            ODBSequencesLoader.reloadSequenceEndo(seqId, seqg, execEngine)
           )(st)
       }.withEvent(LoadSequence(seqId)).toHandle)
 
     t.map {
-      case (err :: _, None) =>
-        List(Event.logDebugMsgF[F, EngineState[F], SeqEvent](explain(err)))
-      case (errs, Some(seq)) =>
+      case (UnrecognizedInstrument(_) :: _, None) =>
```
rpiaggio, line 72: I don't know whether we need this, but is an unrecognized instrument logged at least once somewhere else or not at all anymore?

rpiaggio (approved):
LGTM

modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
cquiroz: Actually I took another route and it required the change and I forgot to revert it.
At any rate I'm tending to use the `cats.implicits._` import

modules/seqexec/server/src/main/scala/seqexec/server/ODBSequencesLoader.scala
cquiroz: That would be ideal but it forces to keep an state and I want to avoid that.
In reality this log was useful during development but not anymore.
This function is called every 30s thus polluting the log

PR #1500, opened 2020-09-08
Remove TrySeq alias

cquiroz:
We've been trying  to reduce the use pf type aliases in the seqexec.
This OR removes `TrySeq`

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
@@ -49,11 +49,9 @@ object ConfigUtilOps {
   def explainExtractError(e: ExtractFailure): SeqexecFailure =
     SeqexecFailure.Unexpected(ConfigUtilOps.explain(e))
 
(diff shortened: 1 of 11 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
```diff
@@ -50,10 +50,10 @@ object ConfigUtilOps {
     SeqexecFailure.Unexpected(ConfigUtilOps.explain(e))
 
   implicit class TrySeqed[A] private[server] (r: Either[ExtractFailure, A]) {
```
rpiaggio, line 52: Nitpicking about the name (it is not actually used).

rpiaggio (approved):
LGTM

modules/seqexec/server/src/main/scala/seqexec/server/ConfigUtilOps.scala
cquiroz: Oh good point

PR #1501, opened 2020-09-08
Support copying the seqexec version

cquiroz:
It is not possible now to copy the version of the seqexec from the UI.
We want to make it easier for users to report the version number.
This PR will copy the version when the area is clicked

I also used this as an opportunity to test semantic-ui-toasts for notifications.

ezgif com-video-to-gif(3)

### DIFF
--- a/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/Footer.scala
+++ b/modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/Footer.scala
@@ -52,7 +52,7 @@ object Footer {
     )
 
   private val component = ScalaComponent
-    .builder[Props]("Footer")
+    .builder[Props]
     .stateless
     .render_P(p =>
       Menu(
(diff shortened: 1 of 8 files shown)


modules/seqexec/web/client/src/main/scala/seqexec/web/client/components/Footer.scala
```diff
@@ -32,31 +41,40 @@ object Footer {
...
+                   icon = Icon("clipboard"),
+                   size = Small,
+                   tpe = ToastType.Success,
+                   time = Dismissal.On(500.millisecond)
+      )
+    )
+
   private val component = ScalaComponent
-    .builder[Props]("SeqexecAppBar")
+    .builder[Props]("Footer")
```
rpiaggio, line 55: We can actually omit the names now.

rpiaggio (approved):
LGTM

PR #1509, opened 2020-09-11
SEQNG-1220 Don't use a custom mask if FPU is NONE

cquiroz:
A fault was reported for a GMOS sequence containing a custom mask for some steps and FPU none for others.
The expected behavior is to set custom or no on the respective steps but the seqexec
is trying to set the custom mask at each step

The root cause is that the OT indicates that each step uses a custom mask mode, and that indication
is used to select the mask.
Ideally this would be fixed in the OT but otherwise we can workaround checking the particular case that FPU is none for custom mask mode.

The change is required both in the backend and the frontend as we don't have real types at that level on the frontend, relying instead on the String representation provided by the OT.

The UI shows the difference
Before:
image (9)

After:
image (7)
 
It require testing against a few sequences

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosSouth.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosSouth.scala
@@ -3,7 +3,6 @@
 
 package seqexec.server.gmos
 
-import cats.Eq
 import cats.effect.Concurrent
 import cats.effect.Timer
 import cats.effect.concurrent.Ref
@@ -34,7 +33,6 @@ final case class GmosSouth[F[_]: Concurrent: Timer: Logger] private (
 ) extends Gmos[F, SouthTypes](
   c,
   new SiteSpecifics[SouthTypes] {
-    val fpuEq: Eq[GmosSouthType.FPUnitSouth] = Eq.fromUniversalEquals
     val fpuDefault: GmosSouthType.FPUnitSouth = FPU_NONE
     def extractFilter(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, SouthTypes#Filter] =
       config.extractInstAs[SouthTypes#Filter](FILTER_PROP)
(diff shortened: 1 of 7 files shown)


cquiroz:
This may need a bit of rework to ignore the fpuMode property altogether

swalker2m:
Looking at it this morning, it seems like the `fpuMode` configuration is not reliable coming from the WDBA/ODB.  It also appears to be redundant (if it were correct) since the value can be inferred (`Builtin` vs `CustomMask`) from the FPU itself. I think it should maybe be removed from the configuration and ignored.

rpiaggio (approved):
LGTM

cquiroz:
I changed the PR to ignore the fpu mode which is bogus in some cases and instead
rely on the newly added `isCustom` flag to the FPU enums
I'll merge this to do field testing today

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosSouth.scala
```diff
@@ -44,6 +42,7 @@ final case class GmosSouth[F[_]: Concurrent: Timer: Logger] private (
       config.extractInstAs[SouthTypes#FPU](FPU_PROP_NAME)
     def extractStageMode(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosSouthType.StageModeSouth] =
       config.extractInstAs[SouthTypes#GmosStageMode](STAGE_MODE_PROP)
+    def isCustomFPU(config: CleanConfig): Boolean = extractFPU(config).map(_.isCustom()).getOrElse(false)
```
swalker2m, line 45: Will require an ODB update to deploy in production?  (I know you updated the code in a separate PR, just wondering if we need to deploy an updated ODB to use it.)

cquiroz: Not really, I exported the new code as a lib but since we are converting xml to ocs code in the seqexec it should work

My local tests worked fine with an older odb

swalker2m: Oh, that's right.  :+1:

PR #1511, opened 2020-09-11
SEQNG-1217: Target check UI

cquiroz:
This PR adds support to do target checks mostly from the UI point of view though the hooks are ready to setup the backend. 

A target check is done when a run is started (either from the top level or from a given step) and
will compare the target in the TCS and the sequence. If they don't match a prompt will ask the user to either Stop (default) or just continue.

The prompt looks like:

Seqexec - GN-2020B-Q-205-55 2020-09-11 15-57-12

Note that in practice this is now not active as the target check always passes.
We still need to implement the following to close this.

* Do the actual target check comparing the TCS and sequence.
* Include the target names on the user notification.


modules/seqexec/model/shared/src/main/scala/seqexec/model/UserPrompt.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+  implicit class UserPromptOps(val p: UserPrompt) extends AnyVal {
+    def question: List[String]         = UserPrompt.question(p)
+    def title: String                  = UserPrompt.title(p)
+    def okButton: String               = UserPrompt.okButton(p)
+    def cancelButton: String           = UserPrompt.cancelButton(p)
+    def okColor: PromptButtonColor     = UserPrompt.okColor(p)
+    def cancelColor: PromptButtonColor = UserPrompt.cancelColor(p)
+  }
```
rpiaggio, line 69: Could these be in the `UserPrompt` trait?

cquiroz: Yes, I just followed the encoding we used in `Notification`, it probably doesn't matter much.

modules/seqexec/model/shared/src/main/scala/seqexec/model/UserPrompt.scala
```diff
@@ -0,0 +1,79 @@
...
+
+package seqexec.model
+
+import cats.Eq
+import cats.implicits._
+import gem.Observation
+
+sealed trait UserPrompt extends Product with Serializable
+
+object UserPrompt {
```
rpiaggio, line 12: Upon second look, maybe we could keep the UI details on the client side? As per our Slack conversation, I understand that adhering to this would mean changing `Notification` too.

jluhrs (approved):
LGTM

PR #1514, opened 2020-09-14
Fix showing custom masks

cquiroz:
This is a fix to show the custom mask properly on the steps table for GMOS:

image

### DIFF
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosNorth.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosNorth.scala
@@ -38,8 +38,6 @@ final case class GmosNorth[F[_]: Concurrent: Timer: Logger] private (
       config.extractInstAs[NorthTypes#Disperser](DISPERSER_PROP)
     def extractFPU(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosNorthType.FPUnitNorth] =
       config.extractInstAs[NorthTypes#FPU](FPU_PROP_NAME)
-    def extractCustomFPU(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, String] =
-      config.extractInstAs[String]("fpuCustomMask")
     def extractStageMode(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosNorthType.StageModeNorth] =
       config.extractInstAs[NorthTypes#GmosStageMode](STAGE_MODE_PROP)
     val fpuDefault: GmosNorthType.FPUnitNorth = FPU_NONE
@@ -49,7 +47,6 @@ final case class GmosNorth[F[_]: Concurrent: Timer: Logger] private (
         case (_, Right(_))                             => true
         case _                                         => false
       }
-
   },
   nsCmdR
 )(northConfigTypes) {
--- a/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosNorth.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosNorth.scala
@@ -30,7 +30,7 @@ final case class GmosNorth[F[_]: Concurrent: Timer: Logger] private (
   c: GmosNorthController[F],
   dhsClient: DhsClient[F],
   nsCmdR: Ref[F, Option[NSObserveCommand]]
-) extends Gmos[F, NorthTypes](c,
+  ) extends Gmos[F, NorthTypes](c,
   new SiteSpecifics[NorthTypes] {
     def extractFilter(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, NorthTypes#Filter] =
       config.extractInstAs[NorthTypes#Filter](FILTER_PROP)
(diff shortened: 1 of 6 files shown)


modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosNorth.scala
```diff
@@ -30,18 +30,26 @@ final case class GmosNorth[F[_]: Concurrent: Timer: Logger] private (
...
+  ) extends Gmos[F, NorthTypes](c,
   new SiteSpecifics[NorthTypes] {
     def extractFilter(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, NorthTypes#Filter] =
       config.extractInstAs[NorthTypes#Filter](FILTER_PROP)
     def extractDisperser(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosNorthType.DisperserNorth] =
       config.extractInstAs[NorthTypes#Disperser](DISPERSER_PROP)
     def extractFPU(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosNorthType.FPUnitNorth] =
       config.extractInstAs[NorthTypes#FPU](FPU_PROP_NAME)
+    def extractCustomFPU(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, String] =
+      config.extractInstAs[String]("fpuCustomMask")
```
rpiaggio, line 42: Should we put the property name in a constant for consistency with the other ones?

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosSouth.scala
```diff
@@ -38,11 +38,18 @@ final case class GmosSouth[F[_]: Concurrent: Timer: Logger] private (
       config.extractInstAs[SouthTypes#Filter](FILTER_PROP)
     def extractDisperser(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosSouthType.DisperserSouth] =
       config.extractInstAs[SouthTypes#Disperser](DISPERSER_PROP)
+    def extractCustomFPU(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, String] =
+      config.extractInstAs[String]("fpuCustomMask")
```
rpiaggio, line 42: Same here, we could reference the same constant.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosSouth.scala
```diff
@@ -38,11 +38,18 @@ final case class GmosSouth[F[_]: Concurrent: Timer: Logger] private (
...
     def extractFPU(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosSouthType.FPUnitSouth] =
       config.extractInstAs[SouthTypes#FPU](FPU_PROP_NAME)
     def extractStageMode(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosSouthType.StageModeSouth] =
       config.extractInstAs[SouthTypes#GmosStageMode](STAGE_MODE_PROP)
-    def isCustomFPU(config: CleanConfig): Boolean = extractFPU(config).map(_.isCustom()).getOrElse(false)
+    def isCustomFPU(config: CleanConfig): Boolean =
```
rpiaggio, line 47: Seems to be the same logic at both sites, can it be moved to `SiteSpecifics`?

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/StepItems.scala
```diff
@@ -38,45 +37,40 @@ object StepItems {
...
-        fpuL = if (isNoneFPU(i) || mode === FPUMode.BuiltIn) instrumentFPUO
-        else instrumentFPUCustomMaskO
-        fpu <- fpuL.getOption(s)
-      } yield fpuNameMapper(i)(fpu).getOrElse(fpu)
+      (i, instrumentFPUO.getOption(s), instrumentFPUCustomMaskO.getOption(s)) match {
+        case (Instrument.GmosS | Instrument.GmosN | Instrument.F2, Some("CUSTOM_MASK"), c) => c
+        case (Instrument.GmosS | Instrument.GmosN | Instrument.F2, None, c @ Some(_))      => c
+        case (_, Some(b), _)                                                               => fpuNameMapper(i)(b)
+        case _                                                                             => none
+      }
```
rpiaggio, line 73: If I understand correctly, we are not defaulting to `BuiltIn` anymore, is that the fix?

rpiaggio (approved):
LGTM

modules/seqexec/web/client/src/main/scala/seqexec/web/client/model/StepItems.scala
cquiroz: There are several options how the OT presents the sequence depending on whether there are iterators and what not.
This basically tries to find if there is a custom mask present first and if not it goes the build in
There is a flag to indicate whether we use custom or built in and it was used previously
but it turns out it is unreliable and there are cases where the flag indicates we have a built in mask
while in reality it is custom

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosSouth.scala
cquiroz: The logic looks the same but they are different as they act on different types representing FPU North/South
`extractCustomFPU` can be made common though.

modules/seqexec/server/src/main/scala/seqexec/server/gmos/GmosNorth.scala
```diff
@@ -41,7 +41,12 @@ final case class GmosNorth[F[_]: Concurrent: Timer: Logger] private (
     def extractStageMode(config: CleanConfig): Either[ConfigUtilOps.ExtractFailure, GmosNorthType.StageModeNorth] =
       config.extractInstAs[NorthTypes#GmosStageMode](STAGE_MODE_PROP)
     val fpuDefault: GmosNorthType.FPUnitNorth = FPU_NONE
-    def isCustomFPU(config: CleanConfig): Boolean = extractFPU(config).map(_.isCustom()).getOrElse(false)
+    def isCustomFPU(config: CleanConfig): Boolean =
+      (extractFPU(config), extractCustomFPU(config)) match {
+        case (Right(builtIn), _) if builtIn.isCustom() => true
+        case (_, Right(_))                             => true
```
swalker2m, line 47: I would have expected the previous version to work too.  How are the sequence steps generated?  There is an option that gives you the full configuration at each step and the FPU should be correct there.  If you want to send me a sequence where it is wrong, I'll take a look.

swalker2m: Thanks, I looked at the sequence you sent.  It's clearly a bug in the `ocs` code but this looks like it should work around it.  :-/

jluhrs (approved):
LGTM

PR #1515, opened 2020-09-16
SEQNG-1222 Compare current GMOS exposure time using same channel as Tcl Seqexec

jluhrs:
The channel used was not suitable, because it was not updated until an observe was executed. Now Seqexec uses the same channel used by the old Seqexec, which is updated as soon as the new exposure time is configured.


cquiroz:
Do we need to update the keywords' values?

jluhrs:
> Do we need to update the keywords' values?

No. The values for the GMOS keywords are read after the observation, so even though it uses the SAD channel, it will have the right value.

cquiroz (approved):
LGTM

PR #1516, opened 2020-09-17
Fix to read FPU in some composite cases

cquiroz:
This is another attempt to properly read from the OT when to use the built in FPU mask or a custom mask
All provided sample sequences work as expected


swalker2m (approved):
It's ridiculous that you have to do this, but the sequence is a mess.  Looks good. :+1:

PR #1524, opened 2020-10-06
SEQNG-1223 Accept that other commands are executed in TCS while configuring it.

jluhrs:
I removed the restriction that the apply.VAL cannot change while executing an EPICS command.


cquiroz (approved):
LGTM

PR #1525, opened 2020-10-09
SEQNG-1224 Use order 0 when GMOS is missing the disperser order parameter

jluhrs:
It happens that a missing Disperser Order in a GMOS sequence was not an error, but a "feature".


cquiroz (approved):
LGTM

andrewwstephens:
I'm a bit worried that order=0 is rarely the desired value...

andrewwstephens:
I chatted with Kristin and she agrees that, if we are going to make an assumption about the grating order, it should be that order = 1.

PR #1528, opened 2020-10-14
SEQNG-1226 Don't activate system confíg or run from when doing alt cl…

cquiroz:
This fixes a bug where if you alt click on a row it should flip the breakpoint but it also can configure a subsystem if clicked on top of a button.


jluhrs (approved):
LGTM

PR #1553, opened 2020-11-17
Update github actions

cquiroz:
This started as an effort to fix the build but it now also includes a dependabot action and updates to some libs


jluhrs (approved):
LGTM

PR #1578, opened 2020-11-30
System override implemented on server side.

jluhrs:
This was way more complicated than I expected. The control code of all the systems follows a similar model. For each one exists a "Controller" class that encapsulates the interaction with the real system. I implemented the system override my switching each Controller class with one that does nothing, except logging some messages. I had to delay the creation of class instances that used these Controller classes, so I could reapply them each time I switch between the real Controller and the dumb one.
This was easy to do for TCS and GCAL, but it was a pain for the instruments. The main issue was that the interface `InstrumentSystem` had grown to encompass every instrument specific aspect. I had to split it to separate the parts that depended on a Controller.

### DIFF
--- a/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
+++ b/modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
@@ -83,7 +83,7 @@ class Engine[F[_]: MonadError[?[_], Throwable]: Logger, S, U](stateL: Engine.Sta
     val x = for {
       seq <- stateL.sequenceStateIndex(c.sid).getOption(st)
       if (seq.status.isIdle || seq.status.isError) && !seq.getSingleState(c.actCoords).active
-      act <- seq.getSingleAction(c.actCoords)
+      act <- seq.rollback.getSingleAction(c.actCoords)
     } yield act.gen
 
     x.map(p =>
--- a/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
@@ -7,18 +7,15 @@ import scala.concurrent.duration._
 
 import cats.data.Kleisli
 import fs2.Stream
-import lucuma.core.enum.LightSinkName
 import seqexec.model.dhs.ImageFileId
 import seqexec.model.enum.Instrument
 import seqexec.model.enum.ObserveCommandResult
 import seqexec.server.keywords.KeywordsClient
-import squants.Length
 import squants.Time
 
-trait InstrumentSystem[F[_]] extends System[F] with InstrumentGuide {
+trait InstrumentSystem[F[_]] extends System[F] {
   override val resource: Instrument
-  // The name used for this instrument in the science fold configuration
-  def sfName(config: CleanConfig): LightSinkName
+
   val contributorName: String
 
   def observeControl(config: CleanConfig): InstrumentSystem.ObserveControl[F]
@@ -35,10 +32,6 @@ trait InstrumentSystem[F[_]] extends System[F] with InstrumentGuide {
   def observeProgress(total: Time, elapsed: InstrumentSystem.ElapsedTime): Stream[F, Progress]
 
   def instrumentActions(config: CleanConfig): InstrumentActions[F]
-
-  override val oiOffsetGuideThreshold: Option[Length] = None
-
-  override def instrument: Instrument = resource
 
 }
 
--- a/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
@@ -36,9 +36,6 @@ trait InstrumentSystem[F[_]] extends System[F] with InstrumentGuide {
 
   def instrumentActions(config: CleanConfig): InstrumentActions[F]
 
-  def calcStepType(config: CleanConfig, isNightSeq: Boolean): Either[SeqexecFailure, StepType] =
-    SequenceConfiguration.calcStepType(config, isNightSeq)
-
   override val oiOffsetGuideThreshold: Option[Length] = None
 
   override def instrument: Instrument = resource
--- a/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
+++ b/modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
@@ -30,7 +30,6 @@ import seqexec.model.dhs._
 import seqexec.model.enum.Instrument
 import seqexec.model.enum.Resource
 import seqexec.model.{Progress => _, _}
-import seqexec.server
 import seqexec.server.CleanConfig.extractItem
 import seqexec.server.ConfigUtilOps._
 import seqexec.server.InstrumentSystem._
@@ -148,7 +147,7 @@ object SeqTranslate {
         insSpecs = instrumentSpecs(inst)
         stepType <- insSpecs.calcStepType(config, isNightSeq).fold(_.raiseError[F, StepType], _.pure[F])
         dataId   <- dataIdFromConfig[F](config)
-        is       <- toInstrumentSys(inst)
+        is       = toInstrumentSys(inst)
         systems  <- calcSystems(config, stepType, insSpecs)
         headers  <- calcHeaders(config, stepType, inst)
       } yield buildStep(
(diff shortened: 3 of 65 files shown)


cquiroz:
Build is broken, some headers are missing

modules/seqexec/server/src/main/scala/seqexec/server/InstrumentSystem.scala
```diff
@@ -7,18 +7,15 @@ import scala.concurrent.duration._
 
 import cats.data.Kleisli
 import fs2.Stream
-import lucuma.core.enum.LightSinkName
 import seqexec.model.dhs.ImageFileId
 import seqexec.model.enum.Instrument
 import seqexec.model.enum.ObserveCommandResult
 import seqexec.server.keywords.KeywordsClient
-import squants.Length
 import squants.Time
```
cquiroz, line 14: Unrelated to this PR, but we should consider transitioning to `coulomb`

cquiroz:
One general question. It seems that the selection of disabled subsystems is applied when the steps are generated, thus we know to use the "disabled" controller. Is it so?
What happens when you click disabling subsystem x, does the sequence gets re generated? 
Somehow I'd expect a dynamic condition like disabling a subsytem to be applied dynamically as well, i.e. at execution time rather than by regenerating the steps.
Would we be able e.g. to change the subsystems disabled while running?

jluhrs:
@cquiroz Yes, the sequence is regenerated with the new value. This works the same as for the observer name or observing conditions. And yes, it can be changed while running. Only the pending parts of the sequence will be updated.

modules/seqexec/server/src/main/scala/seqexec/server/SequenceGen.scala
```diff
@@ -52,28 +52,27 @@ object SequenceGen {
...
 
-  final case class StepActionsGen[F[_]](pre: List[ParallelActions[F]],
-                                  configs: Map[Resource, Action[F]],
-                                  post: HeaderExtraData => List[ParallelActions[F]]) {
-    def generate(ctx: HeaderExtraData): List[ParallelActions[F]] =
-      pre ++
-        NonEmptyList.fromList(configs.values.toList).toList ++
-        post(ctx)
+  final case class StepActionsGen[F[_]](configs: Map[Resource, SystemOverrides => Action[F]],
+                                        post: (HeaderExtraData, SystemOverrides) => List[ParallelActions[F]]) {
```
jluhrs, line 64: The changes to `StepActionsGen` are the core of the systems override feature. Making the `Action`s dependent on a `SystemOverrides` allows to reapply it when it changes.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -320,20 +317,47 @@ object SeqTranslate {
...
+      case Instrument.Nifs  => { ov:SystemOverrides =>
+        Nifs(overriddenSystems.nifs(ov), overriddenSystems.dhs(ov)):InstrumentSystem[F]
+      }.pure[F]
+      case Instrument.Gsaoi => { ov:SystemOverrides =>
+        Gsaoi(overriddenSystems.gsaoi(ov), overriddenSystems.dhs(ov)):InstrumentSystem[F]
+      }.pure[F]
     }
 
-    private def calcResources(sys: List[System[F]]): Set[Resource] = sys.map(_.resource).toSet
+    def instrumentSpecs(instrument: Instrument): InstrumentSpecifics = instrument match {
```
cquiroz, line 350: Could we have an `Instrument{ def specifics: InstrumentSpecifics}` method?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -342,219 +366,253 @@ object SeqTranslate {
...
+        tryWavelength(inst.instrument, config).flatMap{w => site match {
           case Site.GS => if(useGaos)
-            Gems.fromConfig[F](systems.gems, systems.guideDb)(config).map(a =>
-              TcsSouth.fromConfig[F](systems.tcsSouth, subs, a.some, inst, systems.guideDb)(
+            Gems.fromConfig[F](systemss.guideDb, config).map(a => (ov:SystemOverrides) =>
+              TcsSouth.fromConfig[F](overriddenSystems.tcsSouth(ov), subs, a(overriddenSystems.gems(ov)).some, inst,
+                systemss.guideDb)(
                 config, LightPath(lsource, inst.sfName(config)), w
-              )
+              ):System[F]
```
cquiroz, line 382: `.widen[System[F]]`?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -342,219 +366,253 @@ object SeqTranslate {
...
           useGaos = true,
-          sys,
+          instSpec,
           TcsController.LightSource.AO,
           config
-        ).map{ List(sys, _, Gcal.defaultGcal(systems.gcal)) }
+        ).map{ x =>
+          Map(
+            Resource.TCS -> x,
+            Resource.Gcal -> {ov: SystemOverrides => Gcal.defaultGcal(overriddenSystems.gcal(ov))}
```
cquiroz, line 464: It seems this is repeated a few times, could it be moved to its own method?

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -342,219 +366,253 @@ object SeqTranslate {
...
+        List(
+          commonHeaders(config, allButGaos.toList, kwClient)(ctx),
+          gwsHeaders(kwClient),
+          calcInstHeader(config, instrument, kwClient)
+        )
+      }.pure[F]
+
+      case StepType.AltairObs(_)    => Altair.guideStarType[F](config).map { gst =>
+        kwClient: KeywordsClient[F] => ctx: HeaderExtraData =>
+          // Order is important
```
cquiroz, line 561: If order is important should we use a structure that preserves order? e.g. `SortedMap`?

modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
```diff
@@ -13,7 +13,6 @@ sealed trait StepType {
 
 object StepType {
   final case class CelestialObject(override val instrument: Instrument) extends StepType
-  final case class Dark(override val instrument: Instrument) extends StepType
```
cquiroz, line 16: No darks?

modules/seqexec/server/src/main/scala/seqexec/server/SystemOverrides.scala
```diff
@@ -0,0 +1,55 @@
...
+  def disableInstrument: SystemOverrides
+  def enableInstrument: SystemOverrides
+  def disableGcal: SystemOverrides
+  def enableGcal: SystemOverrides
+  def disableDhs: SystemOverrides
+  def enableDhs: SystemOverrides
+}
+
+object SystemOverrides {
+  val allEnabled: SystemOverrides = SystemOverridesImpl(
```
cquiroz, line 24: Often constants are capitalized like `AllEnabled`

modules/seqexec/server/src/main/scala/seqexec/server/keywords/DhsClientDisabled.scala
```diff
@@ -0,0 +1,27 @@
...
+import seqexec.server.SystemOverrides.overrideLogMessage
+
+class DhsClientDisabled[F[_]: Sync: Logger] extends DhsClient[F] {
+
+  val format: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMdd")
+
+  override def createImage(p: DhsClient.ImageParameters): F[ImageFileId] = for {
+    _ <- overrideLogMessage("DHS", "setKeywords")
+    date <- Sync[F].delay(LocalDate.now)
+    time <- Sync[F].delay(System.currentTimeMillis % 1000)
```
cquiroz, line 22: Maybe should be consistent and use e.g. `Instant` here?

cquiroz (approved):
Looks good though it is a bit hard to follow the logic as it is spread across the code.
I added some minor comments

modules/seqexec/engine/src/main/scala/seqexec/engine/Engine.scala
```diff
@@ -83,7 +83,7 @@ class Engine[F[_]: MonadError[?[_], Throwable]: Logger, S, U](stateL: Engine.Sta
     val x = for {
       seq <- stateL.sequenceStateIndex(c.sid).getOption(st)
       if (seq.status.isIdle || seq.status.isError) && !seq.getSingleState(c.actCoords).active
-      act <- seq.getSingleAction(c.actCoords)
+      act <- seq.rollback.getSingleAction(c.actCoords)
```
jluhrs, line 86: This was a bit of a bug. Before, it could have happened that `startSingle` was called on a partially run step. It happened to worked OK. But it would not work properly with the implementation of the system overrides. That is because the update method in Step only updates the pending `Execution`s, not the run ones or the `focus`. But it properly updates the backup copy of the actions, so calling `rollback` gets the updated actions.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
```diff
@@ -92,45 +81,52 @@ trait SeqTranslate[F[_]] extends ObserveActions {
...
+                     ): F[StepGen[F]] = {
       def buildStep(
         dataId: DataId,
-        inst: InstrumentSystem[F],
-        sys: List[System[F]],
-        headers: HeaderExtraData => List[Header[F]],
+        instRes: Resource,
+        insSpec: InstrumentSpecifics,
+        instf: SystemOverrides => InstrumentSystem[F],
+        otherSysf: Map[Resource, SystemOverrides => System[F]],
```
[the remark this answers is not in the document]
jluhrs: It could be, but that would introduce some problems. The first one is that all the systems in `otherSysf` are treated generically, but we need to access the special behavior of `InstrumentSystem`. Also, the `InstrumentSystem` is mandatory, and that will not be reflected if we pass it in `otherSysf`.

modules/seqexec/server/src/main/scala/seqexec/server/SeqTranslate.scala
jluhrs: It would be handy, but Instrument and InstrumentSpecifics live in separate packages with separate functions. It also would invert the current dependency between those packages.

modules/seqexec/server/src/main/scala/seqexec/server/StepType.scala
jluhrs: The type was not used anywhere. `DarkOrBias` was used instead.

PR #1583, opened 2020-12-04
SEQNG-1203 Implemented observing conditions check.

jluhrs:
The observing conditions check was combined with the already existing target check, using the same user prompt box and the override mechanism.


cquiroz (approved):
LGTM

PR #1592, opened 2020-12-09
SEQNG-1231 Optional limit to precision of floating point keywords.

jluhrs:
Tcl Seqexec can limit the number of decimals for the value of floating point keywords, as a consequence of everything been strings in Tcl.  I implemented a similar feature in the new Seqexec.

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/gmos/GmosHeader.scala
+++ b/modules/server/src/main/scala/seqexec/server/gmos/GmosHeader.scala
@@ -117,7 +117,7 @@ object GmosHeader {
             buildDouble(gmosReader.gratingWavelength, KeywordName.GRWLEN),
             buildDouble(gmosReader.gratingAdjustedWavelength, KeywordName.CENTWAVE),
             buildInt32(gmosReader.gratingOrder, KeywordName.GRORDER),
-            buildDouble(gmosReader.gratingTilt, KeywordName.GRTILT),
+            buildDoublePrecision(gmosReader.gratingTilt, 4, KeywordName.GRTILT),
             buildDouble(gmosReader.gratingStep, KeywordName.GRSTEP),
             buildDouble(gmosReader.dtaX, KeywordName.DTAX),
             buildDouble(gmosReader.dtaY, KeywordName.DTAY),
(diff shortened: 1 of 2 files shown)


modules/server/src/main/scala/seqexec/server/keywords/package.scala
```diff
@@ -133,6 +138,15 @@ package keywords {
       extends Keyword[Boolean](name, TypeBoolean, value)
   final case class StringKeyword(name: KeywordName, value: String)
       extends Keyword[String](name, TypeString, value)
+  final case class FloatPrecisionKeyword(name: KeywordName, precision: Int, value: Float)
+      extends Keyword[Float](name, TypeFloat, value)   {
+    override val stringValue: String =
+      s"%.${precision}f".formatLocal(new Locale("en", "US"), value.toDouble)
```
cquiroz, line 144: I think we are launching with a set `Locale` to en_US, perhaps we can use the default?

cquiroz: At any rate you may want to put the enUS locale in a separate variable. Instantiating a locale has a certain cost.

modules/server/src/main/scala/seqexec/server/gmos/GmosHeader.scala
```diff
@@ -117,7 +117,7 @@ object GmosHeader {
             buildDouble(gmosReader.gratingWavelength, KeywordName.GRWLEN),
             buildDouble(gmosReader.gratingAdjustedWavelength, KeywordName.CENTWAVE),
             buildInt32(gmosReader.gratingOrder, KeywordName.GRORDER),
-            buildDouble(gmosReader.gratingTilt, KeywordName.GRTILT),
+            buildDoublePrecision(gmosReader.gratingTilt, 4, KeywordName.GRTILT),
```
cquiroz, line 120: Is this the only keyword ever need a specific precission?

modules/server/src/main/scala/seqexec/server/keywords/package.scala
jluhrs: Ok, I'll do.

modules/server/src/main/scala/seqexec/server/gmos/GmosHeader.scala
jluhrs: So far, but I already told Andy we can do this, he surely will come with a list of other keywords.

cquiroz (approved):
LGTM

PR #1600, opened 2020-12-14
Rework of the main grid

cquiroz:
The seqexec used semantic ui grids to organize the main layout but that was more complicated than needed, especially when making the UI responsive

This PR changes all that making it use flexbox instead

### DIFF
--- a/modules/web/client/src/main/scala/seqexec/web/client/components/tabs/TabsArea.scala
+++ b/modules/web/client/src/main/scala/seqexec/web/client/components/tabs/TabsArea.scala
@@ -10,10 +10,7 @@ import japgolly.scalajs.react.extra.router.RouterCtl
 import japgolly.scalajs.react.vdom.html_<^._
 import lucuma.core.enum.Site
 import react.common._
-import react.semanticui.collections.grid.GridColumn
-import react.semanticui.widths._
 import seqexec.web.client.circuit._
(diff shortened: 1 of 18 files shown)


jluhrs (approved):
LGTM

modules/web/client/src/main/scala/seqexec/web/client/components/tabs/TabsArea.scala
```diff
@@ -34,18 +31,16 @@ object TabsArea {
     .builder[Props]("TabsArea")
     .stateless
     .render_P(p =>
-      GridColumn(width = Sixteen, clazz = SeqexecStyles.sequencesArea)(
+      React.Fragment(
         SeqexecTabs(p.router),
         tabsConnect(x =>
           React
             .Fragment(
               x().toList.collect {
```
rpiaggio, line 39: I think we can `collect[VdomNode]` instead of the type ascriptions in each case.

rpiaggio (approved):
LGTM

modules/web/client/src/main/scala/seqexec/web/client/components/tabs/TabsArea.scala
cquiroz: Cool, works nicely. I'll add it in the next PR

PR #1603, opened 2020-12-17
SEQNG-1193 Remember that Gems was paused because of a Sky between steps.

jluhrs:
Seqexec interacts with GeMS by notifying it of events that may affect GeMS AO and guiding. There are three types defined:
1. Sky: Natural guide stars will not be available for an observation because of a big offset.
2. Dither: The telescope will apply an offset.
3. Filter: It only applies if using GSAOI ODGW. ODGW will be unavailable.

Seqexec sends events when the conditions appears (pause command) and when it ends (resume command). Dither is the easiest one, because both events occur inside the setup of a single step (pause Dither -> TCS offset -> resume Dither -> observe).

The issue this PR fixes is for the Sky events. The pause Sky event must be sent before the first unguided exposure, and the resume Sky must be sent before the next guided one. I added a mechanism to carry the state of sending the sky event between steps.

Something similar must be done for the "pause Filter" event when implementing support for ODGW.

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/tcs/Gaos.scala
+++ b/modules/server/src/main/scala/seqexec/server/tcs/Gaos.scala
@@ -74,6 +74,11 @@ object Gaos {
       case b: FixedPauseCondition => PauseConditionSet(offsetO, fixed + b)
     }
 
+    def -(v: PauseCondition): PauseConditionSet = v match {
+      case _: OffsetMove          => PauseConditionSet(none, fixed)
+      case b: FixedPauseCondition => PauseConditionSet(offsetO, fixed - b)
+    }
+
     def contains(v: FixedPauseCondition): Boolean = fixed.contains(v)
 
   }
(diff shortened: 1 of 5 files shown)


modules/server/src/main/scala/seqexec/server/tcs/Gaos.scala
```diff
@@ -74,6 +74,11 @@ object Gaos {
       case b: FixedPauseCondition => PauseConditionSet(offsetO, fixed + b)
     }
 
+    def -(v: PauseCondition): PauseConditionSet = v match {
```
cquiroz, line 77: I don't often see symbolics operators outside libraries but 👍

cquiroz (approved):
LGTM

modules/server/src/main/scala/seqexec/server/tcs/GuideConfigDb.scala
```diff
@@ -186,6 +194,6 @@ object GuideConfigDb {
     )(TelescopeGuideConfig(_, _, _))(mountGuideDecoder, m1GuideDecoder, m2GuideDecoder)
 
   implicit val guideConfigDecoder: Decoder[GuideConfig] =
-    Decoder.forProduct2("tcsGuide", "gaosGuide")(GuideConfig)
+    Decoder.forProduct2("tcsGuide", "gaosGuide")(GuideConfig(_, _, false))
```
jluhrs, line 197: Every time TCC sets a new guide configuration on GeMS, the pause/resume logic is reset. That is why the flag is set to `false`.

PR #1608, opened 2020-12-20
SEQNG-1232 Fixed useAo flag being checked for night time calibrations.

jluhrs:
Night time calibrations inside an Altair sequence were failing with the error "Found useAo set for non AO step".
The issue was that class `TcsControllerEpicsCommon` is used to setup the telescope for calibrations and non AO observations, but assumed that the useAo flags must be always off. That is not the case for calibrations inside an Altair sequence.
Strictly speaking, having the useAo flag set is only an error if PWFS2 is used. I used this fact to make the check more restricted.


cquiroz (approved):
LGTM

PR #1627, opened 2021-01-19
SEQNG-1235 CRFOLLOW value is now calculated from TCS rotator tracking frame

jluhrs:
The change affects the keyword CRFOLLOW, set for GPI, GeMS and Altair.


cquiroz (approved):
Looks great, thanks

PR #1628, opened 2021-01-19
SEQNG-1238 Fixed problem with trying to run a sequence from a step other than the first

jluhrs:
Running from a step other than the first is accomplished by marking all the previous steps as skipped, and then starting the sequence normally. Of course, the process of marking the previous steps as skipped left completed or already skipped steps untouched. The bug was that partially run steps were also left untouched, so when calling start on the sequence, it founded this unrun step not skipped, and would start from it.
After the changes some code became redundant. I remove it.


cquiroz (approved):
LGTM

PR #1642, opened 2021-02-05
SEQNG-1242 Workaround for FR-41232: GMOS Numrois broken.

jluhrs:
Always set the number of ROIs parameter. To be reverted when the GMOS status channel is fixed.


cquiroz (approved):
LGTM

PR #1645, opened 2021-02-10
SEQNG-1243 Expanded and standardized log messages for interaction with external systems.

jluhrs:
I put a "Starting" and "Completed" log message around every interaction with external systems. That way if something breaks while trying to command an external system, we will know exactly where it broke.


cquiroz (approved):

rpiaggio (approved):
LGTM

PR #1647, opened 2021-02-11
SEQNG-1234 Added the system name to errors from EPICS commands.

jluhrs:
Added the system name to EPICS errors to make clear that the error was coming from an external system and not from Seqexec.


cquiroz (approved):
LGTM

PR #1649, opened 2021-02-12
SEQNG-1244 Used sequence value for keyword DETNROI.

jluhrs:
This is a workaround for the problem with the GMOS status channel that gives the number of ROIs used. Instead of relying on that channel, the number of ROIs given in the sequence is used. That number is also used to determine how many ROI parameter sets must be written in the header.


cquiroz (approved):
LGTM

PR #1677, opened 2021-02-25
SEQNG-1245 Updated library dependencies to use CA reads with retries.

jluhrs:


modules/server/src/main/scala/seqexec/server/CaServiceInit.scala
```diff
@@ -26,6 +26,7 @@ object CaServiceInit {
           case _       => F.raiseError[Unit](new RuntimeException("Cannot initialize EPICS subsystem"))
         }
       } *>
+      F.delay(CaService.setReadRetries(conf.readRetries)) *>
```
cquiroz, line 29: I wonder if we should include all these in one `F.delay`. Meaning configuring the retries and IOTimeout in one go.

cquiroz (approved):
LGTM with one little comment

PR #1684, opened 2021-03-02
SEQNG-1248 Mark GMOS ROI CAD when changing unconnected parameters.

jluhrs:
When configuring GMOS, whenever the number of roy or binning is changed, the ROI CAD is marked. This is necessary because the ROI configuration depends on those parameters, but they are not connected to the ROI CAD.
I also went back in the workaround for the DETNROI keyword, because at the point the value is read from GMOS, it already got the updated value.


cquiroz (approved):
LGTM

PR #1689, opened 2021-03-05
SEQNG-1251 Fixed status acceptor used for Altair Slow Focus status.

jluhrs:
Code was looking for the two Altair attributes in the wrong status acceptor.


cquiroz (approved):
LGTM

PR #1694, opened 2021-03-08
SEQNG-1252 Increased estimated readout time for observe timeout.

jluhrs:
The logs show instances of timeout errors where the actual time taken by GMOS to finish the observation was close to the timeout time used by Seqexec, probably causing a false timeout error. I increased the timeout for GMOS.
GMOS is not the only instrument that has frequent observation timeouts. Because of that, I also increased the timeout for other instruments. That would either fix the timeout problem, or allows us to discard that they are caused by a too short waiting time.


cquiroz (approved):
LGTM

PR #1745, opened 2021-04-09
SEQNG-1258 Altair: execute mirror flattening before reenabling loops.

jluhrs:
Seqexec was trying to flatten Altair DM at the same time it was enabling Altair guiding, which caused a problem. That was an error when translating the old Tcl code. With this PR the two commands are sequenced. Also, I noticed that the Tcl code also enables the BTO loop on resuming, something I forgot to put in the Web Seqexec. That was also fixed.


cquiroz (approved):
LGTM

PR #1750, opened 2021-04-13
SEQNG-1262 Restored retries for DHS keyword messages.

jluhrs:
Actually, it was disabled only for the last package (the "final" one).
I also adjusted the DHS timeout after analyzing the record of request service time. 24s is the worst case.

### DIFF
--- a/app/seqexec-server-gn/src/main/resources/app.conf
+++ b/app/seqexec-server-gn/src/main/resources/app.conf
@@ -64,5 +64,5 @@ seqexec-engine {
   epicsCaAddrList = "10.2.2.255 10.2.10.21 10.2.126.101"
   readRetries = 4
   ioTimeout = 4 seconds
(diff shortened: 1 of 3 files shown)


cquiroz:
I think there was a reason for disabling at final, to avoid starting the write more than once if I recall, would the new dhs proxy solve that?

jluhrs:
Yes, that is the change. The DHS Proxy now keeps track of the requests, and ignores duplicates. That applies only to setKeywords messages.
One problem I noticed is that the proxy remembers the completed requests. That means that if Seqexec retries a request before the first attempt completes, the proxy will execute it. That is the reason I searched for the longest service time and used it as the new timeout, to avoid that situation. It is not ideal, because I also found that only 1/100000 of the requests take more than 20s. 99% take less than 1s.

cquiroz (approved):
LGTM

PR #1759, opened 2021-04-22
SEQNG-1258 Fixed the instance of the flatten command I missed before.

jluhrs:


modules/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
```diff
@@ -305,10 +309,8 @@ object AltairControllerEpics {
         L.debug("Resuming Altair guiding") *>
           epicsAltair.btoLoopControl.setActive("ON") *>
           epicsAltair.btoLoopControl.post(DefaultTimeout) *>
-          epicsTcs.aoFlatten.mark *>
```
cquiroz, line 308: Why was there twice before?

cquiroz (approved):
LGTM

PR #1794, opened 2021-05-24
Do Not Merge: Changes the GDS client from XMLRPC to HTTP

toddburnside:
This PR is in preparation for the rewrite of the GDS portion of GMP. The GDS server will switch from XMLRPC to using standard REST-like HTTP. Eventually, these changes will need to be ported to `Observe` as well.

I don't know if there are additional configuration changes that will need to be made when it is deployed.

There are unit tests, but it has not yet been tested against the modified GDS.


app/seqexec-server-gs/src/main/resources/app.conf
```diff
@@ -63,7 +63,7 @@ seqexec-engine {
   ioTimeout = 4 seconds
   dhsTimeout = 24 seconds
   gpiUrl = "failover:(tcp://172.17.107.50:61616)?timeout=4000"
-  gpiGDS = "http://172.17.107.50:8888/xmlrpc"
+  gpiGDS = "http://172.17.107.50:8888"
```
cquiroz, line 66: We may want to keep the endpoint in a route as the GMP could be exposing other endpoints over http. I guess we could use a different for for different endpoints though

cquiroz (commented):
Thanks for getting this started. I'm afraid we need to keep the old xmlrpc client along the new json one. At the moment GHOST is still planned to use the previous version of the GMP. 
If we could manage to get GHOST into the new version we could plan to delete the old client

modules/server/src/main/scala/seqexec/server/keywords/GdsClient.scala
```diff
@@ -56,142 +57,60 @@ object GdsClient {
...
-          }</string></value>
-        }
-      }
-            </data>
-          </array>
-        </value>
-      </param>
+  case class KeywordRequest(id: String, ks: KeywordBag)
+  case class OpenObservationRequest(obsId: Observation.Id, id: String, ks: KeywordBag)
+  case class CloseObservationRequest(id: String)
```
cquiroz, line 90: Maybe we should add an abort message too, When the seqexec aborts we tell GDS to release the keywords it has in memory. We'd still need to do garbage collection on GDS but that would make it a bit cleaner

The DHS side doesn't have that so we may need to ignore it there.

toddburnside:
> Thanks for getting this started. I'm afraid we need to keep the old xmlrpc client along the new json one. At the moment GHOST is still planned to use the previous version of the GMP.

I can add the xmlrpc one back in and see what I need to do to have GHOST use the old one.

cquiroz:
We can remove the xmlrpc endpoint on the GMP, it is just that GHOST will use the older version of the GMP

But in that scenario the seqexec must support both styles

cquiroz:
I'll close this PR, the changes have been included in #2388

PR #1816, opened 2021-06-27
SEQNG-1268 Log reasons to change TCS configuration.

jluhrs:
Log every EPICS parameter changed in TCS. The log message has the parameter name, old value, and new value.
The new messages are output only if seqexec.server.tcs.trace is defined.

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/tcs/TcsControllerEpicsCommon.scala
+++ b/modules/server/src/main/scala/seqexec/server/tcs/TcsControllerEpicsCommon.scala
@@ -5,9 +5,7 @@ package seqexec.server.tcs
 
 import java.time.Duration
 import java.util.concurrent.TimeUnit.SECONDS
-
 import scala.concurrent.duration.FiniteDuration
-
 import cats._
 import cats.data._
 import cats.effect.Async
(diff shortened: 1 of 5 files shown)


modules/server/src/main/scala/seqexec/server/tcs/TcsControllerEpicsCommon.scala
```diff
@@ -239,64 +245,94 @@ object TcsControllerEpicsCommon {
     demand:  T,
     act:     T => F[Unit],
     lens:    Lens[C, T]
-  ): Option[C => F[C]] =
-    (used && current =!= demand).option(c => act(demand) *> lens.set(demand)(c).pure[F])
+  )(name:    String): Option[WithDebug[C => F[C]]] =
```
cquiroz, line 248: Would it make sense to make a `WithDebug` typeclass and attach it to `F` like `Logger` or natchez
https://github.com/tpolecat/natchez

jluhrs: Do you mean, like to be able to use an `F: Applicative` or `F: Monad`, and define a way to combine the debug messages at the same time you do `map` or `flatMap` on the `F` ?

cquiroz: Yeah use it like `F: WithDebug` but it is just an algebra with methods like `_.withDebug`

jluhrs: I don't think that applies that way. `withDebug` is really a constructor that takes an `A` and produces a `WithDebug[A]`.  There is no point on having and `x: F` (with `F: WithDebug`) just to be able to call `x.withDebug`

cquiroz:
How are these changes going to be included in `observe`?

cquiroz (approved):
LGTM

PR #1830, opened 2021-07-15
FR-72094: increased timeout when waiting for science fold in-position.

jluhrs:
This is a small PR to adjust the timeout used when waiting for the science fold to get in position. It happens that sometimes it takes longer than 60 seconds.


cquiroz (approved):
LGTM

PR #1831, opened 2021-07-18
SEQNG-1268 Fixed comparison of current and demanded offset.

jluhrs:
Fixed comparison of current and demanded offset to decide if it needs to be applied. Also, added a tolerance when comparing wavelength, and added unit tests for setting offset and wavelength.


modules/server/src/main/scala/seqexec/server/tcs/TcsControllerEpicsCommon.scala
```diff
@@ -28,6 +28,8 @@ import seqexec.server.EpicsCodex.encode
 import seqexec.server.EpicsCommand
 import seqexec.server.SeqexecFailure
 import seqexec.server.tcs.TcsController._
+import squants.Length
```
cquiroz, line 31: Unrelated but we are moving towards using coulomb rather than squants. Maybe it should be changed at least on observe

cquiroz (approved):

PR #1853, opened 2021-08-04
SEQNG-1270: Check selected OI before trying to park it.

jluhrs:
Seqexec was trying to park the OI probe when none was selected in the A&G Sequencer. The result was a no operation that wasted a couple of seconds on each step. This PR makes sure that Seqexec parks the OI only if it selected in the A&G Sequencer.


cquiroz (approved):
LGTM

PR #1860, opened 2021-08-10
FR-72196 Removed name translation of major bodies for target check

jluhrs:
I removed the translation once used to fix the target check for major bodies, because it is not needed anymore and is causing problems when in the OT, the Horizon code is used instead of the body name.


cquiroz (approved):
Does this mean we don't need to do translation anymore? Who does it then?

jluhrs:
In TCS you can define planets as targets just by its name, and TCS does all the calculations internally. That mode is the one that required a translation, but it is not used anymore. Now planets, and all other non-sidereal targets, always use ephemerides.

PR #1868, opened 2021-08-19
FR-72238 Fixed bug where Seqexec was not setting the fpu if it was None.

jluhrs:


modules/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
```diff
@@ -354,23 +354,26 @@ object GmosControllerEpics extends GmosEncoders {
...
         cc:    GmosFPU
       ): List[Option[F[Unit]]] = {
-        def builtInFPU(fpu: T#FPU): (Option[String], Option[String]) = Encoders[T].fpu.encode(fpu)
+        def builtInFPU(fpu: T#FPU): Option[(String, String)] = Encoders[T].fpu.encode(fpu)
 
-        def customFPU(name: String): (Option[String], Option[String]) = name match {
-          case "None" => (none, beamEncoder.encode(Beam.OutOfBeam).some)
-          case _      => (name.some, beamEncoder.encode(Beam.InBeam).some)
+        def customFPU(name: String): (String, String) = name match {
+          case "None" => ("None", beamEncoder.encode(Beam.OutOfBeam))
```
cquiroz, line 360: I guess this is the name for the instrument? It is unfortunate that matches the scala `None` type

cquiroz (approved):
LGTM

modules/server/src/main/scala/seqexec/server/gmos/GmosControllerEpics.scala
jluhrs: Yes. I think my confusion came from this Tcl code:
```
switch $fpu {
      "None" {
         set inbeam "OUT-OF-BEAM"
         }
      "Longslit 0.25 arcsec" {
         set inbeam "IN-BEAM"
         set fpu 0.25arcsec
         }
```
I think I saw that and thought "OK, so then if the ODB value for fpu is None, then there is none"

PR #1880, opened 2021-09-01
SEQNG-1273 Fixed keyword DATE-OBS for GSAOI and F2.

jluhrs:
The problem is that local time was used instead of UT.


cquiroz (approved):
LGTM

PR #1898, opened 2021-09-22
SEQNG-1274 Run GSAOI dome flats even if GeMS was not configured.

jluhrs:
GSAOI dome flats are special, because they are calibration observations taken on OBJECT targets (the dome position for the flat screen). Seqexec was complaining because OBJECT steps with a GeMS component should be run with a GeMS configuration coming from TCC, which is not the case for dome flats (GeMS is ignored in TCC).
This PT makes Seqexec ignore the problem if the step does not have any GeMS guider active.


modules/server/src/main/scala/seqexec/server/gems/Gems.scala
```diff
@@ -140,7 +144,7 @@ object Gems {
   def fromConfig[F[_]: MonadError[*[_], Throwable]](
```
cquiroz, line 144: Maybe we put `Logger` in the `F` constraints?

cquiroz (approved):
LGTM

modules/server/src/main/scala/seqexec/server/gems/Gems.scala
rpiaggio: Also, there's the `MonadThrow` alias now.

PR #1914, opened 2021-10-07
SEQNG-1275 Increased settle time for PWFS2.

jluhrs:
I increased the time to wait for PWFS2 to settle after an offset, after night time tests showed star elongations


modules/server/src/main/scala/seqexec/server/tcs/TcsSettleTimeCalculator.scala
```diff
@@ -41,7 +41,7 @@ object TcsSettleTimeCalculator {
   val settleTimeCalculators: Map[Subsystem, SettleTimeCalculator] = Map(
     Subsystem.Mount -> constantSettleTime(1.seconds),
     Subsystem.PWFS1 -> constantSettleTime(1.seconds),
-    Subsystem.PWFS2 -> constantSettleTime(1.seconds)
+    Subsystem.PWFS2 -> constantSettleTime(2.seconds)
```
cquiroz, line 44: - [ ] Will this apply to GN and GS? Wasn't the smearing mostly in the south?

jluhrs: It was detected at GS when using F2 + PWFS2. Yes, the change will affect both sites.

PR #1925, opened 2021-10-12
Fixed bug in warning message for Altair control matrix

jluhrs:


modules/server/src/main/scala/seqexec/server/tcs/TcsSettleTimeCalculator.scala
```diff
@@ -41,7 +41,7 @@ object TcsSettleTimeCalculator {
   val settleTimeCalculators: Map[Subsystem, SettleTimeCalculator] = Map(
     Subsystem.Mount -> constantSettleTime(1.seconds),
     Subsystem.PWFS1 -> constantSettleTime(1.seconds),
-    Subsystem.PWFS2 -> constantSettleTime(2.seconds)
+    Subsystem.PWFS2 -> constantSettleTime(1.seconds)
```
cquiroz, line 44: Didn't you chabge this to 2 secs a couple of weeks ago?

jluhrs: Yes, but it was decided to do more tests to better asses the problem before decided to increase this time.

cquiroz: ah ok

PR #1948, opened 2021-11-03
Fix querying display name for noirlab users

cquiroz:
To get the display name of new users we needed to change the LDAP query used.
Also a bit more logging was added


jluhrs (approved):
LGTM

cquiroz:
I got confirmation this fix is working

PR #1956, opened 2021-11-12
Observer owns sequences

cquiroz:
This is a non trivial change where sequences (and the queue) are "owned" by the observer. An observer can use her/his name from LDAP or change it and when any action is triggered it will become the owner of that sequence

The idea is to simplify the way the observer name is set. A bit of design discussion goes here:
https://gist.github.com/cquiroz/86266d14bd8e4bd448ec40e05874c20b

And a screenshot:
image

This is a first iteration, it may need some further tweaking after user testing

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -171,11 +171,21 @@ trait SeqexecEngine[F[_]] {
 
   def requestRefresh(q: EventQueue[F], clientId: ClientId): F[Unit]
 
-  def stopObserve(q: EventQueue[F], seqId: Observation.Id, observer: Observer, graceful: Boolean): F[Unit]
+  def stopObserve(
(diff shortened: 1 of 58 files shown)


modules/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -72,28 +72,41 @@ trait SeqexecEngine[F[_]] {
     q:           EventQueue[F],
     id:          Observation.Id,
     user:        UserDetails,
+    observer:    Observer,
```
jluhrs, line 75: Wouldn't have made sense to include the name used for Observer in `UserDetails` ? Something like `final case class UserDetails(username: String, displayName: String, observerName: String)` And then `observerName` would be the same as `displayName`, until the user writes something different.

cquiroz: In hindsight that's probably correct, I'm not sure if there is enough time to fix it

PR #1958, opened 2021-11-15
SEQNG-1240 Support refreshing after sync

cquiroz:
Sync should update the table column's values if e.g. the exposure time changed in the backend. The current caching mechanism was too aggressive and that didn't happen.


jluhrs (approved):
LGTM

PR #1986, opened 2021-12-16
SEQNG-1279 Restore M2 guide after offset for Altair LGS mode.

jluhrs:
Fixed a bug: Seqexec was not restoring M2 guiding after an offset for Altair LGS mode.


cquiroz (approved):
LGTM

PR #2017, opened 2022-01-13
SEQNG-1281 Fixed problem with system configuration indicator.

jluhrs:
The problem was that, in order to see the change of the observer name when commanding a system configuration, the system was forcing a sequence refresh in the client. That made the client load the new observer's name, but also reset the subsystems state.
I change the way the observer name is set when triggering other actions. Now it will generate an event to the client to refresh just the observer's name, removing the need to force a full refresh.

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/SeqexecEngine.scala
+++ b/modules/server/src/main/scala/seqexec/server/SeqexecEngine.scala
@@ -185,19 +185,31 @@ trait SeqexecEngine[F[_]] {
     q:        EventQueue[F],
     seqId:    Observation.Id,
     observer: Observer,
+    user:     UserDetails,
     graceful: Boolean
   ): F[Unit]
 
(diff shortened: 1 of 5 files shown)


modules/server/src/main/scala/seqexec/server/SeqexecEngine.scala
```diff
@@ -1265,17 +1282,17 @@ object SeqexecEngine {
       q:        EventQueue[F],
       sid:      Observation.Id,
       observer: Observer,
+      user:     UserDetails,
       stepId:   StepId,
       sys:      Resource,
       clientID: ClientId
     ): F[Unit] =
-      q.enqueue1(Event.modifyState[F, EngineState[F], SeqEvent](setObserver(sid, observer))) *>
+      setObserver(q, sid, user, observer) *>
```
cquiroz, line 1290: Nice

cquiroz (approved):
Thanks a lot, LGTM

PR #2018, opened 2022-01-14
REL-4054 Implemented special beahvior for F2 filters in lyot wheel.

jluhrs:


modules/server/src/main/scala/seqexec/server/flamingos2/Flamingos2ControllerEpics.scala
```diff
@@ -122,16 +123,28 @@ object Flamingos2ControllerEpics extends Flamingos2Encoders {
       _ <- sys.dcConfigCmd.setBiasMode(encode(dc.b))
     } yield ()
 
+    private def filterAndLyot(cc: CCConfig): (Option[String], String) =
```
jluhrs, line 126: Filter encoder has to return an `Option` because of obsolete values in `Filter` (i.e. filters that don't exist anymore in the instrument)

cquiroz (approved):
LGTM

PR #2060, opened 2022-03-01
SEQNG-1282 Fixed handling of Altair's SFO control.

jluhrs:
Control of Altair's SFO is weird, in that it does not use a CAD record. All that is needed is to write a value to an EPICS channel. But Seqexec was handling it as it was another CAD, which caused problems.


cquiroz (approved):
LGTM

PR #2061, opened 2022-03-02
Added debug info to NIFS mask configuration

jluhrs:


cquiroz (approved):
LGTM

modules/server/src/main/scala/seqexec/server/nifs/NifsControllerEpics.scala
```diff
@@ -271,25 +271,27 @@ object NifsControllerEpics extends NifsEncoders {
...
+              L.debug(
+                s"Check NIFS mask ($check): !(current ($current) === ${'"'}INVALID${'"'} && lsm ($lsm) === mask ($mask) && mo ($mo) =!= 0.0)"
+              )
+          } yield check.option(setMaskIO)
 
         // We need an even smarter set param
         for {
           instMask     <- epicsSys.mask
           setIfInvalid <- checkInvalid(instMask)
-        } yield if (instMask =!= mask) setMaskIO.some else setIfInvalid
```
jluhrs, line 283: If `instMask =!= mask`, it does not necessarily mean that we need to set the mask. It could be that the right mask is selected, but a mask offset was applied, which is exactly the condition we check inside `checkInvalid()`

PR #2069, opened 2022-03-14
REL-4082 Workaround for GNIRS filter 2 problem.

jluhrs:
Workaround to reduce the need to move GNIRS filter wheel 2. Now we try to keep it always at position XD.


cquiroz (approved):
LGTM

PR #2104, opened 2022-04-29
Added VAL field to definition of TCS channels.

jluhrs:
Added explicitly the VAL field to TCS channels that did not have a field. VAL is the default field, so the change has no major effect.
The change was required because of a problem with the new TCS Simulator written in Scala.


cquiroz (approved):
LGTM

PR #2125, opened 2022-06-01
Fixed GeMS NGS status, added tests.

jluhrs:
There are two changes, The first is to the GeMS NGS status, that changed with NGS2. The second is that I added unit tests to the logic that pauses and resumes guiding in GeMS.


cquiroz (approved):
Looks good, I'm looking forward to see Gems using the web seqexec

PR #2187, opened 2022-07-02
FR-72718 Inverted logic for target filter, assume it is always on.

jluhrs:
Changed the logic to control the target filter to make it as in the Tcl Seqexec


cquiroz (approved):
LGTM

PR #2217, opened 2022-07-20
Disable Altair in LGS mode when using offsets.

jluhrs:
Seqexec does not stop AO guiding in LGS mode, as the laser star is always visible. This PR reverts that behavior to keep it in line with the Tcl Seqexec. This is only for a test.


cquiroz (approved):
LGTM

PR #2220, opened 2022-07-22
Flatten mirror in LGS mode.

jluhrs:


modules/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
```diff
@@ -454,7 +455,18 @@ object AltairControllerEpics {
...
+    )(cfg:           AltairConfig): F[AltairPauseResume[F]] = {
+      val unguidedStep                                = pauseReasons.contains(GaosGuideOff)
+      def guideOff(turnOff: Boolean): Option[F[Unit]] =
+        (turnOff || unguidedStep).option(
+          L.debug(
+            s"Pausing Altair guiding"
+          ) *>
+            setCorrectionsOff *>
+            L.debug("Altair guiding paused") *>
+            (L.debug("Flatting Altair DM") *> dmFlattenAction).whenA(unguidedStep).void
```
cquiroz, line 467: I guess you don't need the outer parens here

cquiroz (approved):
LGTM

modules/server/src/main/scala/seqexec/server/altair/AltairControllerEpics.scala
jluhrs: I think I need. The debug message and the action to flatten de mirror must go together.

PR #2247, opened 2022-09-02
Reverted workaround for F2 window problem.

jluhrs:
F2 window mechanism was fixed, so I reverted the workaround I put in place.


cquiroz (approved):
LGTM

PR #2273, opened 2022-10-13
Removed guide use from decision to park probes.

jluhrs:
Don't check if guider is used in current guider configuration to decide if it must be parked. There are use cases where it is valid to have a guider probe following a target and not use it to guide.


cquiroz (approved):
LGTM

PR #2278, opened 2022-10-19
Moved to use Java 17.

jluhrs:
I configured to run on Java 17 to be able to use newer versions of logback (which is incompatible with Java 1.8 since version 1.4.0)


.github/workflows/packtracker.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Scala
         uses: olafurpg/setup-scala@v12
         with:
-          java-version: 1.8
+          java-version: 17
```
cquiroz, line 26: We should probably delete `packtracker` completely. It is deprecated now

cquiroz (approved):
LGTM

PR #2280, opened 2022-10-27
FR-42381: Removed check for GeMS and Altair configuration in GeMS and Altair mode.

jluhrs:
There are many cases where Seqexec complained about not having received a GeMS or Altair guide configuration from TCC, which were actually valid (i.e. the user actually wanted to run a GeMS or Altair observation without setting up GeMS or Altair in TCC). One example is running a pupil alignment sequence with GSAOI.
I removed the checks for a GeMS or Altair guide configuration, and used a default instead.


cquiroz (approved):
Not sure if I follow the logic but the code looks good to me

PR #2287, opened 2022-11-09
Use jdk http client

jluhrs:
The goal was to have Seqexec use the JDK HTTP client for DHS communication. I brought that change from Observe, together with all the changes done in observe to use cats effects 3, monocle 3, and the latest versions of lucuma.


build.sbt
```diff
@@ -228,7 +229,8 @@ lazy val seqexec_web_client = project
       "terser-webpack-plugin"         -> "3.0.6",
       "html-webpack-plugin"           -> "4.3.0",
       "css-minimizer-webpack-plugin"  -> "1.1.5",
-      "favicons-webpack-plugin"       -> "4.2.0"
+      "favicons-webpack-plugin"       -> "4.2.0",
+      "@packtracker/webpack-plugin"   -> "2.3.0"
```
cquiroz, line 233: I'm guessing we don't need this

modules/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -250,10 +242,8 @@ object WebServerLauncher extends IOApp with LogInitialization {
   def seqexec: IO[ExitCode] = {
 
     // Override the default client config
-    def clientConfig(timeout: FiniteDuration): AsyncHttpClientConfig =
-      new DefaultAsyncHttpClientConfig.Builder(AsyncHttpClient.defaultConfig)
-        .setRequestTimeout(timeout.toMillis.toInt) // Change the timeout
-        .build()
+    def mkClient(timeout: FiniteDuration): Resource[IO, Client[IO]] =
+      JdkHttpClient.simple[IO].map(c => Client(r => c.run(r).timeout(timeout)))
```
cquiroz, line 246: Ok. here is where we change the client

cquiroz (approved):
LGTM

cquiroz:
Considering the size of this PR and the changes done switching multiple libraries we should extensively test this against hardware before making an official release. WDYT?

modules/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -273,44 +263,40 @@ object WebServerLauncher extends IOApp with LogInitialization {
...
     val seqexec: Resource[IO, ExitCode] =
       for {
-        b      <- Blocker[IO]
         _      <- Resource.eval(configLog[IO]) // Initialize log before the engine is setup
-        conf   <- Resource.eval(config[IO].flatMap(loadConfiguration[IO](_, b)))
+        conf   <- Resource.eval(config[IO].flatMap(loadConfiguration[IO]))
         _      <- Resource.eval(printBanner(conf))
-        cli    <- AsyncHttpClient
-                    .resource[IO](clientConfig(conf.seqexecEngine.dhsTimeout))
+        cli    <- mkClient(conf.seqexecEngine.dhsTimeout)
```
jluhrs, line 283: mkClient creates a Client using JdkHttpClient, instead of using AsyncHttpClient

jluhrs:
@cquiroz I agree, we'll need to do an extensive regression test before promoting this to production.

PR #2292, opened 2022-11-11
Split keyword messages in smaller chunks.

jluhrs:
Limit the size of keyword messages for the DHS. Long list of keywords are now split into smaller chunks before sending them to the DHS.

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
+++ b/modules/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
@@ -5,9 +5,7 @@ package seqexec.server.keywords
 
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
-
 import scala.concurrent.duration._
-
 import cats.FlatMap
 import cats.data.EitherT
 import cats.effect.Sync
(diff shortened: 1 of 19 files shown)


modules/server/src/main/scala/seqexec/server/keywords/DhsClientHttp.scala
```diff
@@ -68,27 +67,64 @@ class DhsClientHttp[F[_]](base: Client[F], baseURI: Uri)(implicit timer: Tempora
...
+
+  override def setKeywords(
+    id:        ImageFileId,
+    keywords:  KeywordBag,
+    finalFlag: Boolean
+  ): F[Unit] = {
+
+    val minKeywords: Int = 10
+
+    val limit = if (minKeywords > maxKeywords) minKeywords else maxKeywords
```
cquiroz, line 90: `minKeywords.max(maxKeywords)`

modules/server/src/test/scala/seqexec/server/TestCommon.scala
```diff
@@ -160,7 +164,8 @@ object TestCommon {
     Some("127.0.0.1"),
     0,
     3.seconds,
-    10.seconds
+    10.seconds,
+    32
```
cquiroz, line 168: Is the amount of keywords coming from some logic or just a heuristic?

cquiroz:
I guess this won't affect GDS instruments, right?

cquiroz (approved):
LGTM

modules/server/src/test/scala/seqexec/server/TestCommon.scala
jluhrs: Heuristics with a bit of real data. I tabulated the size of the keywords messages that failed with a timeout, and then chose 32 because I like powers of 2.

jluhrs:
@cquiroz Yes, it does not affect GDS instruments.

PR #2309, opened 2022-11-21
FR-42504 Fixed configuring GSAOI read method parameter.

jluhrs:
The GSAOI read method must be written as an integer, not a String. (I was confused because the status gives a String)

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerEpics.scala
+++ b/modules/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerEpics.scala
@@ -89,10 +89,10 @@ object GsaoiControllerEpics {
   }
 
   // This looks a bit silly, but I prefer to keep it in case the definition is changed in the future.
-  private def readModeFromMode(rm: ReadMode): String = rm match {
-    case ReadMode.BRIGHT     => "FOWLER"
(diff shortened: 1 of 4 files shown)


modules/server/src/main/scala/seqexec/server/gsaoi/GsaoiControllerEpics.scala
```diff
@@ -89,10 +89,10 @@ object GsaoiControllerEpics {
...
 
   // This looks a bit silly, but I prefer to keep it in case the definition is changed in the future.
-  private def readModeFromMode(rm: ReadMode): String = rm match {
-    case ReadMode.BRIGHT     => "FOWLER"
-    case ReadMode.FAINT      => "FOWLER"
-    case ReadMode.VERY_FAINT => "FOWLER"
+  private def readModeFromMode(rm: ReadMode): ReadMethod = rm match {
+    case ReadMode.BRIGHT     => ReadMethod.Fowler
+    case ReadMode.FAINT      => ReadMethod.Fowler
+    case ReadMode.VERY_FAINT => ReadMethod.Fowler
```
cquiroz, line 95: coul it be `case _ => ReadMethod.Fowler`?

cquiroz (approved):
LGTM

PR #2314, opened 2022-11-24
Fixes an edge case exposed on scala.js 1.12.0

cquiroz:


jluhrs (approved):
It could be better...

PR #2319, opened 2022-11-28
Pass the conditions to Ghost

cquiroz:
Ghost needs to know the selected conditions to specify exposure times for the guide camera and the slit viewer.

This PR adds an extra `Ref` to hold the current configurations and it lets `Ghost` read it at the time of configuration.

I'm not totally happy having conditions both in the core model and on the parallel Ref as they need to be on sync but this works as it is. Alternatives are welcome


jluhrs:
An alternative would be to take the conditions out of the core model and model them as an external service that just happens to be implemented inside Seqexec Server. I mean, with its own API for setting/retrieving values.

jluhrs (approved):
LGTM

PR #2320, opened 2022-11-28
FR-42534: Fixed some bugs in the guide control for GeMS.

jluhrs:
I found and fixed two major bugs when running GeMS sequences:
- Seqexec was disabling guide on every step without an offset.
- The logic to find if GeMS was active was wrong.


cquiroz (approved):
LGTM. Nice we got a unit test for this

PR #2367, opened 2023-03-02
FR-42705 Don't enable M1 guiding for GeMS on skies.

jluhrs:
GeMS opens internally opens the LGS loop on skies. Accordingly, Seqexec should not enable M1 corrections on skies, because GeMS will produce none.


cquiroz (approved):
LGTM

PR #2370, opened 2023-03-06
FR-42704 Fixed bug in GeMS control, where Seqexec disabled guiding.

jluhrs:
The main issue was that some logic depended on reading the statuses of c1, c2 and c3, but in RTC2 they don't work.
Also, fixed some unit tests.


cquiroz (approved):
LGTM

PR #2372, opened 2023-03-10
FR-42703 Park GeMS guiders when not in use.

jluhrs:
Seqexec had troubles with GeMS sequences where the old Tcl Seqexec had not. The problem was that TCS declared that the ODGW, and sometimes the Canopus guiders, where not in position, causing timeouts when applying offsets. The main difference between both Seqexec is that the Tcl Seqexec sends the park command for each Canopus guider and ODGW that is not in use.
I implemented the parks in this Seqexec hoping that that fixes the problem.

### DIFF
--- a/modules/server/src/main/java/edu/gemini/seqexec/server/tcs/ParkState.java
+++ b/modules/server/src/main/java/edu/gemini/seqexec/server/tcs/ParkState.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+ * For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+package edu.gemini.seqexec.server.tcs;
+
+public enum ParkState {
+
+    NOT_PARKED("NOT PARKED"),
+    PARKED("PARKED");
+
+    final String name;
+
+    ParkState(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public String toString() {
+        return this.name;
+    }
+
+}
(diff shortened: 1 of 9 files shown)


modules/server/src/main/java/edu/gemini/seqexec/server/tcs/ParkState.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+ * For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+package edu.gemini.seqexec.server.tcs;
+
+public enum ParkState {
```
cquiroz, line 8: Why are these in Java? Due to the XML parser?

cquiroz (approved):
LGTM

modules/server/src/main/scala/seqexec/server/tcs/TcsSouthControllerEpicsAo.scala
```diff
@@ -213,27 +207,34 @@ object TcsSouthControllerEpicsAo {
...
-        .flatMap(setCwfs1Guide(_, subsystems, current.cwfs1.tracking, demand.cwfs1.tracking)),
-      current.mapping
-        .get(Cwfs2)
-        .flatMap(setCwfs2Guide(_, subsystems, current.cwfs2.tracking, demand.cwfs2.tracking)),
-      current.mapping
-        .get(Cwfs3)
-        .flatMap(setCwfs3Guide(_, subsystems, current.cwfs3.tracking, demand.cwfs3.tracking)),
-      current.mapping
-        .get(Odgw1)
-        .flatMap(setOdgw1Probe(_)(subsystems, current.odgw1.tracking, demand.odgw1.tracking)),
```
jluhrs, line 227: The parking was implemented for ODGW, but this logic implied that the ODWG would be configured only if it was included in the Virtual Telescope Mapping, which never happened if the telescope configuration required it to be parked!

modules/server/src/main/java/edu/gemini/seqexec/server/tcs/ParkState.java
jluhrs: Due to the giapi epics library.

PR #2375, opened 2023-03-13
Fixed retrieval of GeMS probes state when parked.

jluhrs:
This problem was like the parallel of the one where Seqexec did not parked GeMS probes if they were not defined in the TCS mapping, but now it was that Seqexec did not recognize they were parked if not defined in the TCS mapping. Which they never were, because they were parked. This caused Seqexec to send the park command on each step. Not a critical problem, but it made it waste ~5 seconds per step.


cquiroz (approved):
LGTM

PR #2436, opened 2023-07-04
Downgraded cats-effect.

jluhrs:
Downgraded cats-effect to version 3.4.11 because 3.5.1 was suspected of causing problems.


cquiroz (approved):
LGTM

PR #2444, opened 2023-07-13
REL-4355 Updated GMOS-S filter table to move OIII y OIIIC.

jluhrs:
Filters OIII y OIIIC were moved from filter wheel 2 to filter wheel 2. I changed the internal table to reflect that change.


cquiroz (approved):
LGTM, this makes me wonder if this should be in configuration

PR #2453, opened 2023-07-24
Update to Cats Effect 3.5 and friends

armanbilge:
Because of the cancelation changes in Cats Effect 3.5.x, we must upgrade it together with FS2, http4s, Blaze, and the JDK HTTP client which include critical bug fixes. Updating Cats Effect by itself without upgrading the other libraries is known to cause issues.

### DIFF
--- a/modules/web/server/src/main/scala/scala/web/server/common/StaticRoutes.scala
+++ b/modules/web/server/src/main/scala/scala/web/server/common/StaticRoutes.scala
@@ -9,6 +9,7 @@ import cats.data.OptionT
 import cats.effect.Sync
 import cats.instances.string._
 import cats.syntax.eq._
+import fs2.compression.Compression
 import org.http4s.CacheDirective._
 import org.http4s.HttpRoutes
 import org.http4s.Request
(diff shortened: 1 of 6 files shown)


modules/web/server/src/main/scala/scala/web/server/common/StaticRoutes.scala
```diff
@@ -17,7 +18,7 @@ import org.http4s.StaticFile
 import org.http4s.headers.`Cache-Control`
 import org.http4s.server.middleware.GZip
 
-class StaticRoutes[F[_]: Sync](
+class StaticRoutes[F[_]: Sync: Compression](
```
cquiroz, line 21: What is this? `Compression`?

armanbilge: Yes, this is needed for Gzip middleware. The constraint must now be declared explicitly so you know that it is using it.

modules/web/server/src/main/scala/seqexec/web/server/http4s/WebServerLauncher.scala
```diff
@@ -112,7 +114,7 @@ object WebServerLauncher extends IOApp with LogInitialization {
   }
 
   /** Resource that yields the running web server */
-  def webServer[F[_]: Logger: Async](
+  def webServer[F[_]: Logger: Async: Dns: Compression](
```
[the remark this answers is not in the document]
armanbilge: Same here, this capability was already being used without being properly declared.

PR #2455, opened 2023-07-31
Giapi enums

cquiroz:
Some giapi related enums are in lucuma-core but it makes more sense to internalize them in the seqexec giapi module


jluhrs (approved):
LGTM

PR #2489, opened 2023-10-03
CE 3.5 updates

armanbilge:
Backports changes from https://github.com/gemini-hlsw/observe/pull/644.

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/EpicsUtil.scala
+++ b/modules/server/src/main/scala/seqexec/server/EpicsUtil.scala
@@ -44,20 +44,21 @@ abstract class EpicsCommandBase[F[_]: Async](sysName: String) extends EpicsComma
   override def post(timeout: FiniteDuration): F[ApplyCommandResult] = setTimeout(timeout) *>
     Async[F]
       .async[ApplyCommandResult] { (f: Either[Throwable, ApplyCommandResult] => Unit) =>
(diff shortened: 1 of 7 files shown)


modules/server/src/main/scala/seqexec/server/EpicsUtil.scala
```diff
@@ -43,18 +43,21 @@ abstract class EpicsCommandBase[F[_]: Async](sysName: String) extends EpicsComma
...
+              }
             }
-          }
-          // It should call f on all execution paths, thanks @tpolecat
-        }.void
-          .getOrElse(f(SeqexecFailure.Unexpected("Unable to trigger command.").asLeft))
+            // It should call f on all execution paths, thanks @tpolecat
+          }.void
+            .getOrElse(f(SeqexecFailure.Unexpected("Unable to trigger command.").asLeft))
+          Some(Async[F].unit)
```
jluhrs, line 59: Wouldn't this simple discard all the code above? I think the correct expression would be `Async[F].delay{...}.as(Some(Async[F].unit))`

jluhrs: Oh wait, my bad. I got confused by all the brackets.
But I still think that my suggestion is  prettier :)

jluhrs (approved):
LGTM

PR #2492, opened 2023-10-11
Fix to for authentication

cquiroz:
This fix will let the seqexec use a different LDAP server

There is also a new utility that can be used to test logins in the command line


jluhrs (approved):
LGTM

PR #2571, opened 2024-03-25
Add endObserve event

cquiroz:
For IGRINS2 we need a way to inform the instrument when the sequence has been completed

This PR adds a new method for `InstrumentSystem`, `sequenceComplete` which will be called when the last  step gets completed.

In this PR none of  the instrument will use this action but in the near future IGRINS2 will add an implementation

### DIFF
--- a/modules/server/src/main/scala/seqexec/server/SequenceGen.scala
+++ b/modules/server/src/main/scala/seqexec/server/SequenceGen.scala
@@ -75,12 +75,22 @@ object SequenceGen {
   }
 
   final case class StepActionsGen[F[_]](
-    configs: Map[Resource, SystemOverrides => Action[F]],
-    post:    (HeaderExtraData, SystemOverrides) => List[ParallelActions[F]]
+    configs:          Map[Resource, SystemOverrides => Action[F]],
+    post:             (HeaderExtraData, SystemOverrides) => List[ParallelActions[F]],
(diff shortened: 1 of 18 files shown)


modules/server/src/main/scala/seqexec/server/SequenceGen.scala
```diff
@@ -75,12 +75,22 @@ object SequenceGen {
...
-    def generate(ctx: HeaderExtraData, overrides: SystemOverrides): List[ParallelActions[F]] =
+
+    def generate(
+      ctx:       HeaderExtraData,
+      overrides: SystemOverrides
+    ): List[ParallelActions[F]] = {
+      val postActions =
+        this.endObserveAction.fold(post(ctx, overrides))(a =>
+          post(ctx, overrides) :+ NonEmptyList.one(a(overrides))
+        )
```
jluhrs, line 90: Isn't a bit risky to put `endObserveAction` together with `post` ? They will run in parallel, but what if something in `post` fails?
I think it should be run after `post`.

jluhrs, line 90: OK, I was wrong: they will be executed sequentially. But then there is another problem, that the execution can be stopped between `post` and `endObserveAction`, leaving the Step unfinished. If that happens, the step will be run again when resuming execution.

jluhrs (approved):
LGTM

PR #2582, opened 2024-04-15
REL-4344:  Fix OBJECT keyword for GHOST

cquiroz:
GHOST OBJECT keyword needs to be customized for Dual targets. This PR adds code that will write the keywords correctly as specified by the user.

The UI is also updated to show the same target names on the screen


jluhrs (approved):
LGTM

PR #2597, opened 2024-05-06
[GNFR-73603] Show offsets column whenever there's an Object step in the sequence

rpiaggio:
Addresses [GNFR-73603](https://noirlab.atlassian.net/browse/GNFR-73603) / [REL-4430](https://noirlab.atlassian.net/browse/REL-4430).

The offsets column was hidden unless the first step was of type `Object`. This PR changes the logic so that the column is now hidden if there are no `Object` steps at all in the sequence.


cquiroz (approved):
LGTM

PR #2599, opened 2024-05-08
Added GN readme, docker files, and REL-4415 mod

asylvester-noirlab:
Adding a new readme and docker files to enable building the seqexec at Gemini North. This also includes a small modification to complete REL-4415 pertaining to GNIRS.

### DIFF
--- a/GN-Utils/Dockerfile.base
+++ b/GN-Utils/Dockerfile.base
@@ -1,0 +1,45 @@
+#
+#  Run Seqexec locally in docker
+#
+#  -- on host --
+#  docker build -t registry.gitlab.com/nsf-noirlab/gemini/rtsw/user-tools/seqexec-tester/seqexec-base -f Dockerfile.base .
+
+FROM ubuntu:22.04
+
+# Environment variables
+ENV SCALA_VERSION=2.13.8 \
+    SBT_VERSION=1.6.2 \
+    JAVA_VERSION=17 \
+    JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64/ \
+    PATH=$JAVA_HOME/bin:$PATH
+
+# Install system packages
+RUN apt-get update && apt-get install -y \
+    wget \
+    tree \
+    htop \
+    zip \
+    unzip \
+    lsb-core \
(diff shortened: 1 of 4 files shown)


GN-Utils/Dockerfile.base
```diff
@@ -0,0 +1,45 @@
+#
+#  Run Seqexec locally in docker
+#
+#  -- on host --
+#  docker build -t registry.gitlab.com/nsf-noirlab/gemini/rtsw/user-tools/seqexec-tester/seqexec-base -f Dockerfile.base .
+
+FROM ubuntu:22.04
+
+# Environment variables
+ENV SCALA_VERSION=2.13.8 \
```
cquiroz, line 10: This may need to be 2.3.10 though it is not a big deal

GN-Utils/Dockerfile.base
```diff
@@ -0,0 +1,45 @@
...
+    SBT_VERSION=1.6.2 \
```
cquiroz, line 11: We are using 1.7.3 bu again not a big deal

modules/server/src/main/scala/seqexec/server/gnirs/GnirsControllerEpics.scala
```diff
@@ -201,7 +201,7 @@ object GnirsControllerEpics extends GnirsEncoders {
         c:    Camera,
         slit: Option[SlitWidth]
       ): F[Option[F[Unit]]] = {
-        lazy val specialConversion: String = s"${cameraStr(c)}+MIR${disperserStr(s.disperser)}"
+        lazy val specialConversion: String = s"${cameraStr(c)}${disperserStr(s.disperser)}+MIR"
```
cquiroz, line 204: This seems fine according to the task. I wonder if this has any effect on the keywords
@jluhrs may know

jluhrs: The keyword value is read back from the instrument. This may change it, but I don't think that would be a problem.

jluhrs (approved):
LGTM

PR #2601, opened 2024-05-21
Update ocs libs

cquiroz:
The ocs was not sending target coordinates, this is solved with the upgraded libs on this PR


swalker2m (approved):
:+1:

PR #2624, opened 2024-07-17
Changed the wheel for GMOS filter Y

jluhrs:
This brings Seqexec up to date with the last change in the instrument.


cquiroz (approved):
LGTM are we going to deploy to GS only?

PR #2664, opened 2024-12-07
REL-4609 Use a different connection name for different instances

cquiroz:
We need to make the giapi connection name unique.
If we have multiple seqexec instances (production/test) connected at the same time there will be a conflict.

This PR solves it by attaching the seqexec host to the connection name


jluhrs (approved):
LGTM

PR #2686, opened 2025-02-26
REL-4535 Added new grating R400-_G5310 for GMOS-N

jluhrs:


modules/server/src/main/scala/seqexec/server/gmos/GmosNorthControllerEpics.scala
```diff
@@ -97,6 +97,7 @@ object GmosNorthEncoders extends GmosControllerEpics.Encoders[NorthTypes] {
       case Disperser.B600_G5307  => "B600+_G5307"
       case Disperser.R600_G5304  => "R600+_G5304"
       case Disperser.R400_G5305  => "R400+_G5305"
+      case Disperser.R400_G5310  => "R400-_G5310"
```
cquiroz, line 100: Unlike the other dispersers, why is it a `-` in this case?

jluhrs: That comes from the REL task.

cquiroz: Got this from science:
`The value for the new R400 corresponding to your example is R400+_G5310.`

cquiroz:
I tested this with a local db and it is reading the sequences with both gratings correctly

cquiroz:
I installed a test version with this changes for the users to test

cquiroz (approved):
LGTM

PR #2720, opened 2025-05-27
REL-4610: Add support for GMOS-N new R400 grating in seqexec

cquiroz:
Supersedes #2686


modules/model/shared/src/main/scala/seqexec/model/enumerations.scala
```diff
@@ -194,13 +194,14 @@ object enumerations {
...
     val GmosNDisperser: Map[String, String] = Map(
       "MIRROR"      -> "Mirror",
-      "B1200_G5301" -> "B1200",
-      "R831_G5302"  -> "R831",
-      "B600_G5307"  -> "B600",
-      "B480_G5309"  -> "B480",
-      "R600_G5304"  -> "R600",
-      "R400_G5305"  -> "R400",
-      "R150_G5308"  -> "R150"
+      "B1200_G5301" -> "B1200_G5301",
```
toddburnside, line 197: Isn't this just mapping each string to itself?

cquiroz: sort of. it is from old ocs values to what we want to show on the screen

PR #2721, opened 2025-05-27
REL-4660 Allow GMOS biases & darks to run in parallel with other GCAL…

cquiroz:
… calibrations


modules/server/src/main/scala/seqexec/server/SequenceConfiguration.scala
```diff
@@ -119,9 +119,9 @@ trait SequenceConfiguration {
         case SCIENCE_OBSERVE_TYPE                                    => extractGaos(inst)
         case BIAS_OBSERVE_TYPE | DARK_OBSERVE_TYPE                   =>
           inst match {
-            case Instrument.GmosN | Instrument.GmosS | Instrument.Ghost =>
+            case Instrument.Ghost =>
               StepType.ExclusiveDarkOrBias(inst).asRight
-            case _                                                      =>
+            case _                =>
```
toddburnside, line 124: So, GMOS is part of this case, now?

toddburnside (approved):
LGTM

PR #2817, opened 2026-04-01
REL-4855 Don't move the ifus on calibrations

cquiroz:


modules/server/src/main/scala/seqexec/server/ghost/GhostConfig.scala
```diff
@@ -256,8 +256,7 @@ sealed trait GhostConfig extends GhostLUT {
     baseConfiguration |+| slitMaskConfiguration |+| (
       if (!isScience(obsType)) {
         // as per REL-4855 we move ifus to focus for calibrations
```
jluhrs, line 258: So, it moves the IFUs to focus or it doesn't?

cquiroz: Thanks for spotting this. This worked anyway because calibrations are always unguided but I'll restore the call to be consistent

cquiroz:
Superseded by #2850

PR #2850, opened 2026-06-26
REL-4855 Don't move IFUs when guiding

cquiroz:
Second run to get the fix for REL-4855 working
This has been tested on the telescope on calibrations and at ENT


jluhrs (approved):
LGTM

